PR #13, opened 2021-06-21
refactor: move global Q config to document links

adamdbradley:


src/core/import/qImport.ts
```diff
@@ -117,33 +95,40 @@ export function toImportPath(url: URL): string {
...
+    let value = getConfig(doc, `protocol.` + protocol);
+    if (value && !value.endsWith('/')) {
       value = value + '/';
     }
-    return value;
+    return value || all;
   });
 }
 
+const configCache = new Map<string, string | undefined | null>();
```
mhevery, line 120: As it stands now, I think this will not behave correctly. On a server, one can serve different apps. These apps will have different configurations but may have the same protocol. The above will leak data between the applications. Let's remove this cache altogether.

src/core/import/qImport.ts
```diff
@@ -20,19 +18,14 @@ declare const __mockImport: (path: string) => Promise<any>;
...
  * @param url - A relative URL (as `string` or `QRL`) or fully qualified `URL`
  * @returns A cached value synchronously or promise of imported value.
  * @public
  */
-export function qImport<T>(
-  base: Element | Document | string | QConfig,
-  url: string | QRL<T> | URL
-): T | Promise<T> {
+export function qImport<T>(base: Element | Document, url: string | QRL<T> | URL): T | Promise<T> {
   if (!importCache) importCache = new Map<string, unknown | Promise<unknown>>();
```
mhevery, line 26: As it stands now, I think this will not behave correctly. On a server, one can serve different apps. These apps will have different configurations but may have the same protocol. The above will leak data between the applications. Let's remove this cache altogether.

src/core/render/jsx/mark_dirty.ts
```diff
@@ -126,34 +129,39 @@ export function toAttrQuery(key: string): any {
  * @returns a `Promise` of all of the `HostElements` which were re-rendered.
  * @internal
  */
-export function scheduleRender(document: QDocument): Promise<HostElements> {
-  const promise = document.$qScheduledRender;
+export function scheduleRender(doc: QDocument): Promise<HostElements> {
+  const requestAnimationFrame = doc.defaultView!.requestAnimationFrame!;
```
mhevery, line 133: Is there a reason why you pulled this line up? Now it is read even if `promise` already exists.

src/core/render/jsx/render.unit.tsx
```diff
@@ -232,7 +235,7 @@ describe('render', () => {
 
   describe('<Host>', () => {
     it('should merge component host with <Host>', async () => {
-      const fixture = new ElementFixture();
+      const fixture = new ElementFixture(TEST_CONFIG);
```
mhevery, line 238: Not related to this line, but

Line 269 has an unused component. Can we delete it?

mhevery (approved):
Conditionally approve on fixing the caches per our discussion.

src/core/render/jsx/mark_dirty.ts
adamdbradley: just to group and reduce the `const`s so there's only one, but will put it back

PR #15, opened 2021-06-22
refactor: render document from host.ownerDocument

adamdbradley:


integration/hello_server/server_index.tsx
```diff
@@ -29,5 +29,5 @@ export async function serverMain(document: Document) {
       </body>
     </html>
   );
-  return jsxRender(document, doc, document);
+  return jsxRender(document, doc);
```
mhevery, line 32: ok, this is super confusing. `(document, doc)`. I know it was like this before but could we change it to something better? `(document, jsx)`? maybe?

src/core.api.gold.md
```diff
@@ -227,7 +227,7 @@ export interface JSXNode<T extends string | null | JSXFactory | unknown> {
 }
 
 // @public
-export function jsxRender(host: Element | Document, jsxNode: JSXNode<unknown>, overrideDocument?: Document): Promise<HostElements>;
+export function jsxRender(host: Element | Document, jsxNode: JSXNode<unknown>): Promise<HostElements>;
```
mhevery, line 230: Absolutely ❤️  this change!

mhevery (approved):
LGTM,

- [ ] If you could take care of my comment that would be nice, but it is not strictly necessary.
- [ ] Also there is a lint warning which it would be nice to fix.

PR #21, opened 2021-07-06
fix(core): allow invoking base injector when using cjs

manekinekko:
When the user's app is built using CommonJS, Qwik cannot invoke the base injector function because it's exported as `default`. This change fixes that issue.

### DIFF
--- a/src/core/injector/base_injector.ts
+++ b/src/core/injector/base_injector.ts
@@ -33,6 +33,10 @@ export abstract class BaseInjector implements Injector {
     self?: SELF | null,
     ...rest: REST
   ): Promise<RET> {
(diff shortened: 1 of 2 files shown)


adamdbradley:
In a branch that hasn't landed in `main` yet, we're building both an `.mjs` and `.cjs` modules, and the cjs version will be able to avoid this issue. So far the branch has been able to build on node's esm and classic require. Once it lands in main we'll double check this again. Thanks for the PR!

adamdbradley:
Thanks @manekinekko! You solved the issue perfectly, thanks for the PR!

manekinekko:
Thank you @adamdbradley 😃

PR #22, opened 2021-07-13
build system refactoring

mhevery:


.gitignore
```diff
@@ -10,4 +11,7 @@ external
 etc
 temp
 tsdoc-metadata.json
-**/.DS_Store
\ No newline at end of file
+**/.DS_Store
+integration/**/*.js
```
mhevery, line 15: why do these files ever make it to the integration folder?

.prettierignore
```diff
@@ -4,7 +4,6 @@
 **/.DS_Store
 *.
 .history
-.vscode
```
mhevery, line 7: Why remove?

integration/devserver.ts
```diff
@@ -0,0 +1,255 @@
...
+import type { RenderToStringResult } from '@builder.io/qwik/server';
+import srcMap from 'source-map-support';
+srcMap.install();
+
+/**
+ * Verbose dev-server tooling built specifically for inspecting/debugging this local
+ * project's source files. Standard qwik development would not have most of this.
+ */
+async function main(integrationSrcDir: string) {
+  const qwikDir = join(integrationSrcDir, '..', 'dist-dev', '@builder.io-qwik');
```
mhevery, line 32: Does this still work with bazel?

src/core/render/jsx/factory.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
mhevery, line 1: why disable lint on the whole file? I would rather if we disable specific rules on specific line numbers.

PR #37, opened 2021-07-21
fix(todo): make blur event always working in editing

otodockal:
**Existing flow**
- dblclick an item
- **input.edit element is not focused without a user click**
- blur element (it's not working on a non-focused element)

**Changed flow in the PR**
- dblclick an item
- focus input.edit element
- properly set cursor position once the item is focused
- blur element


mhevery:
Thank you for the PR. Can you share reproduction steps so that we can add tests for this.

otodockal:
@mhevery updated PR description with reproduction steps + added e2e test.

mhevery (approved):
Thank you for your contribution! Very nice fix and tests! ❤️

PR #42, opened 2021-07-26
chore: fix type imports

adamdbradley:


adamdbradley:
Came across a few more imports that esbuild didn't like to see that were not using `type`

PR #46, opened 2021-07-30
feat(prefetch): add prefetch script to populate cache with web-worker.

mhevery:


src/server/types.ts
```diff
@@ -26,6 +26,7 @@ export interface QwikDocument extends Document {}
  */
 export interface DocumentOptions {
   url?: string;
+  params?: Record<string, string>;
```
adamdbradley, line 29: Couldn't we get the params from the `url` here? I was assuming the `url` would be a full `URL`, to include the origin and params. So internally we could do `new URL(opts.url)`.

adamdbradley (approved):

src/server/types.ts
mhevery: good, point!  let me look into it.

mhevery: removed

PR #50, opened 2021-08-09
Experiment to use marker functions to tell the tooling how to generate QRLs automatically.

mhevery:


steve8708:
love the direction, think is huge improvement to be clean and intuitive

random things that jumped out when reading through, all pretty minor:
- where does `getState()` come from? `import { getState } from '@builder.io/qwik/core'`?
- is there another word we can use for `markDirty` than `dirty`? have had it come up before that the concept of `dirty` isn't always intuitive to ppl, as opposed to maybe like `markUpdated` or something?
- love that `state` is a simple mutable object. would be great to wrap it in a proxy and call markDirty on props automatically any time a top level property is changed
- my intuition says the components should be PascalCase but not the handlers or state factories (feels to me like those should be camelCase)

mhevery:
> love the direction, think is huge improvement to be clean and intuitive

thanks
> 
> random things that jumped out when reading through, all pretty minor:
> 
> * where does `getState()` come from? `import { getState } from '@builder.io/qwik/core'`?
Yes
> * is there another word we can use for `markDirty` than `dirty`? have had it come up before that the concept of `dirty` isn't always intuitive to ppl, as opposed to maybe like `markUpdated` or `markChanged` or even `markNeedsUpdate` or something?
Open to suggestions

> * love that `state` is a simple mutable object. would be great to wrap it in a proxy and call markDirty on props automatically any time a top level property is changed
Yes, that could be done. Such improvement would be independent of the current API, and can be added later.

> * my intuition says the components should be PascalCase but not the handlers or state factories (feels to me like those should be camelCase)
Yes, we need to have a discussion about conventions.

otodockal:
@mhevery Could we go even further and remove the need to define qComponent e.g.:```export const MyCounter = qComponent<MyCounter>({state, view})``` ?
In this case, Qwik component is a Single File Component as a constrain.

So Qwik component would be composed of qState, qView and qHandler(s) without qComponent in a single file.

mhevery:
@otodockal thanks for the feedback.
> @mhevery Could we go even further and remove the need to define qComponent e.g.:`export const MyCounter = qComponent<MyCounter>({state, view})` ?
> In this case, Qwik component is a Single File Component as a constrain.

My current understanding is that `qComponent` is needed because the parent component needs a facade for the child component. How would you imagine this working without `qComponent`? What would the parent component import and put in there?

file: `child.tsx`
```typescript
export MyChild = qComponent(....)
```

file: `parent.tsx`
```
// What would parent component import of the child would not have qComponent?
import {MyChild} from './child';

export MyParentView = qView(() => {
  // What symbol would you place at this location if child qComponent would not exist?
  returne <MyChild/>
})
```

otodockal:
@mhevery 

Yeah, the symbol...When I saw your proposal basically **Svelte components** came to my mind, so I was thinking whether Qwik could be even simpler that's why I called it Single File Component(**not a superset of HTML like Svelte, but valid TS program using different extension - .qwik**). Probably not worth the trouble, but just ideas:

file: ```MyChild.qwik```
```ts
export const MyChildState = qState();
export const MyChildView = qView();
export const MyChildHandler = qHandler();

/* ...generated by a Qwik compiler */
export const MyChild = qComponent({
  state: MyChildState,
  view: MyChildView,
});
```

file: ```MyParent.qwik```
```ts
// qComponent is not declared manually, but generated inside. 
// Qwik component is not .ts but .qwik.
import {MyChild} from './MyChild.qwik';

export const MyParentView = qView(() => {
  return <MyChild/>
})
```

And yeah, this would require some Qwik Language server(if even doable), probably a lot of work for very little gain. And definitely can be done later on...

mhevery:
@otodockal Ohh! I get what you are saying. Few issues with it.

1. `MyParentView` needs to import `MyChild`. Because it is generated (not is the source) all of the tooling would be confused. TS would give error saying that there is no such exported symbol. This would be a lot of work to overcome. IDEs would complain etc.
2. Not shown in the above example but the developer could add additional information. See below.

```typescript
export const MyChild = qComponent({
  element: 'my-child', // generates <my-child>
  style: {...} // generates <my-child style="...">
  someProp: 123, // generates <my-child some-prop="123">
  state: MyChildState,
  view: MyChildView,
});
```

`MyChild` acts as a host element and it allows the developer to decorate it with additional information. Having the tooling generate it would mean that the developer would not have a place to put such additional info.

otodockal:
> `MyChild` acts as a host element and it allows the developer to decorate it with additional information. Having the tooling generate it would mean that the developer would not have a place to put such additional info.

@mhevery Yeah, I see now that the SFC solution would be very limited and complicated without qComponent. Thanks for the example.

PR #62, opened 2021-10-11
feat(eslint): add eslint-plugin-qwik

otodockal:


luki215:
> @mhevery My plan is to use RuleTester provided by eslint together with already used Jest. I will look at it today afternoon CET. https://eslint.org/docs/developer-guide/working-with-rules#rule-unit-tests

@otodockal  Might be worth to check out this repo https://github.com/erikbrinkman/eslint-plugin-no-closure/blob/main/index.test.ts where the setup is with Jest is already done.

PR #69, opened 2021-10-26
feat: rust based optimizer

manucorporat:
## MVP

- [x] Avoid name collision on symbols
- [x] Plan for code move without bundler
- [x] Rust API for multiple files
- [x] Rust class that keeps state across builds
- [x] Implement wasm rust project that maps core APIs
- [x] Implement napi rust build 
- [x] Implement JS API, that calls wasm, napi + exposes FS
- [x] Todo app works
- [x] Implement fixture/snapshot testing
- [x] Implement rust CLI => reading / writing files from FS
- [x] Implement grouping API
- [x] File filtering strategies
    -  [x] Glob
    -  [x] Find qwik import
- [x] Import of h function
- [x] Error handling
- [ ] Make prerelease 
- [ ] Test rollup plugin in real environments
- [x] Rollup working
- [x] CI based cross building for NAPI
- [x] Automated release process in the CI
- [ ] Optimiser output works out of the box without bundling (client and server)
- [x] Bundler output works out of the box without bundling (client and server)
- [ ] Standardise and generate client build manifest.json
- [ ] Adapt SSR render to use the client build manifest.json in order to generate correct qURLs
- [ ] Avoid name collisions when declaring custom symbol names (qHook's second argument)

## Nice to have
- [x] Implement simple playground using the Wasm / JS API
- [x] Enable more linting
- [x] Refactor code to avoid even more clone(), use more lifecycle borrowing
- [x] Support for custom symbol name
- [x] Cache rust builds to speed up CI
- [ ] Test sourcemaps
- [ ] Emit custom error diagnostics, when a qHook references an unknown identifier
- [ ] Properly collect used identifiers 
- [ ] Inject array of global identifiers
- [ ] Test native bindings in google cloud lambda
- [ ] Apple code-sign for macOS binaries
- [ ] Webpack plugin
- [ ] Vite integration
- [ ] Parcel integration

## Performance
- [x] Add PURE comments to qHooks and qComponents
- [x] Perf: Parallel transform
- [ ] Perf: use visit_mut instead
- [ ] Perf: use serde_bytes to pass data between worlds
- [x] Optimize string clones. Look into atoms
- [x] Track identifiers with SyntaxContext


integration/todo/components.qwik.tsx
```diff
@@ -19,7 +19,7 @@ import {
   Todos,
   toggleItem,
   updateFilter,
-} from './state';
+} from './state.qwik';
```
mhevery, line 22: What is the thinking behind renaming `state` to `state.qwik`?

manucorporat: It should not be required, i just renamed all files. But since it does not contain any qwik it should not be necesary

PR #81, opened 2021-11-11
Feat: styles

mhevery:


src/core/component/q-component-ctx.ts
```diff
@@ -101,3 +108,20 @@ export function getHostElement(element: Element): HTMLElement | null {
...
+  if (style) {
+    const styleId = styleKey(style as any)!;
+    const document = ctx.hostElement.ownerDocument;
+    const head = document.querySelector('head')!;
+    if (!head.querySelector(`style[q\\:style="${styleId}"]`)) {
+      const styleImport = Promise.resolve(qImport<string>(document, style));
+      styleImport.then((styles: string) => {
+        const style = document.createElement('style');
+        style.setAttribute('q:style', styleId);
+        style.textContent = styles.replace(/�/g, styleId);
```
adamdbradley, line 122: What is `/�/g`?

src/core/render/q-render.unit.tsx
```diff
@@ -304,14 +304,18 @@ describe('q-render', () => {
         expectRendered(
           <hello-world
             on:q-render={HelloWorld.onRender}
-            q:style={HelloWorld.styles as any}
+            q:sstyle={HelloWorld_styles}
```
adamdbradley, line 307: is `q:sstyle` correct?

src/core/util/markers.ts
```diff
@@ -40,7 +40,12 @@ export const enum AttributeMarker {
   /**
    * Component Styles.
    */
-  ComponentStyles = 'q:style',
+  ComponentStyles = 'q:sstyle',
```
adamdbradley, line 43: Ok guess it's correct, but why the change from `q:style` to `q:sstyle`?

adamdbradley (approved):

src/core/render/q-render.unit.tsx
mhevery: Scoped Style. Not loving it, but for now until we figure out something better.

src/core/util/markers.ts
mhevery: So we can have both scoped and unscoped styles.

PR #90, opened 2021-11-23
feat(eslint): add eslint-plugin-qwik

otodockal:
## MVP
- [x] unit tests integration with Jest
- [x] eslint 8.x
- [x] TS
- [ ] publish eslint-plugin-qwik to NPM

the implementation follows these spec rules https://gist.github.com/mhevery/f56cd107bd358625a0a2753eba63147a#file-spec-tsx


mhevery:
@otodockal I am going to close this as we now have it implemented here: https://github.com/BuilderIO/qwik/tree/main/packages/eslint-plugin-qwik

otodockal:
Yeah, I see @mhevery, it's OK, I could learn something new. Either way, the new API feels way more ergonomic. This rule is very strict, so it's better not to need it, definitely.

shairez:
Thanks for the feedback @otodockal ! (and for your contribution) 🙏

PR #93, opened 2021-11-30
feat: add support for `q:base`

mhevery:


manucorporat:
LGTM but, I had a good next proposal for multiple bases in a single app. IE, multiple qwik apps. Like the idea of
`base:entry#symbol`

I could even propose to make it:
`:entry#symbol`
for the default base

src/bootloader-shared.ts
```diff
@@ -26,24 +26,26 @@
  */
 export const qrlResolver = (
   doc: Document,
-  eventUrl: string | null | undefined,
-  linkElm?: HTMLLinkElement,
-  href?: string,
-  url?: URL
+  element: Element | null,
+  eventUrl?: string | null,
+  _url?: string,
```
adamdbradley, line 31: Not important, but why the `_` here?

src/bootloader-shared.ts
```diff
@@ -26,24 +26,26 @@
...
-      eventUrl.replace(/^(\w+):(\/)?/, (str, protocol, slash) => {
-        linkElm = doc.querySelector(`[rel="q.protocol.${protocol}"]`) as HTMLLinkElement;
-        href = linkElm && linkElm.href;
-        if (!href) error(protocol + ' not defined');
-        return href + (href!.endsWith('/') ? '' : slash || '');
-      }),
-      doc.baseURI
-    );
-    url.pathname += '.js';
+  if (eventUrl === undefined) {
```
adamdbradley, line 34: `if (eventUrl)` won't do, can we use `if (eventUrl == null)`?

src/core/import/qImport.ts
```diff
@@ -35,9 +33,9 @@ export function qImport<T>(node: Node | Document, url: string | QRL<T> | URL): T
       return Promise.resolve<T>(testSymbol);
     }
   }
-  const doc: QDocument = node.ownerDocument || (node as Document);
+  const doc: QDocument = element.ownerDocument!;
```
adamdbradley, line 36: Not possible for `document` to be an element?

adamdbradley (approved):

src/bootloader-shared.ts
mhevery: because it is cut and paste from a bootloader where they are parameters to save bytes and I wanted to make it clear that they are not to be used.

src/bootloader-shared.ts
mhevery: NO, it specifically is looking for `undefined`

src/core/import/qImport.ts
mhevery: not any more.

PR #100, opened 2021-12-10
chore(docs): fix start pre-requirements in developer.md

valorkin:
While following default dev env setup instructions found a missing part about rust install


mhevery (approved):

mhevery:
merged as af60f98

PR #125, opened 2021-12-29
feat: implicit qhooks

manucorporat:
Allows to write qwik components without explicitly creating qHook:

```tsx
const Header = qComponent({
  onMount: () => { console.log("mount") },
  onRender: () => {
    return (
      <div
        on:click={(ctx) => console.log(ctx)}
       />
    );
  }
});
```


src/optimizer/core/src/collector.rs
```diff
@@ -434,9 +434,34 @@ impl Visit for HookCollect {
         self.expr_ctxt.pop();
     }
 
+    fn visit_jsx_opening_element(&mut self, node: &ast::JSXOpeningElement) {
+        let mut stacked = false;
+        if let ast::JSXElementName::Ident(ref ident) = node.name {
+            let ident_name = ident.sym.as_ref().chars().next().unwrap();
+            if let 'A'..='Z' = ident_name {
```
mandx, line 441: Maybe you can avoid the `unwrap` by also pattern matching on the `Option`
```suggestion
            let ident_name = ident.sym.as_ref().chars().next();
            if let Some('A'..='Z') = ident_name {
```

mandx (approved):

mandx (approved):

PR #132, opened 2022-01-11
Vendor yarn executable

Dkendal:
## Description

Adds a vendored copy of the yarn executable to version control as well as a 
config file (.yarnrc.yml) to instruct clients which version to use.

Purpose of this is to ensure that everyone uses the same version, leading to
reproducible builds - and ensures that the project builds properly when mounted
as a submodule of a directory that also has a yarn lock file. In effect this
ensures that this project is built with yarn 1 and treated as it's own root.

## Changelog
- chore: vendor yarn version
- chore: change git ignore for yarn files
- chore: "add-target" make target

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,0 +1,1 @@
+yarnPath: .yarn/releases/yarn-1.23.0-20211220.1904.cjs
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 *.json  linguist-language=JSON-with-Comments
+.yarn/releases/** binary
+.yarn/plugins/** binary
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ target
 *.node
 todo-express/
 qwik-app/
+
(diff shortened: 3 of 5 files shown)


.yarnrc.yml
```diff
@@ -0,0 +1 @@
+yarnPath: .yarn/releases/yarn-1.23.0-20211220.1904.cjs
```
mandx, line 1: Apologies for the dumb question, but anyways: This means one needs to have `yarn` installed globally, then that `yarn` executable will look at that this file, read the `yarnPath` key, and just execute (or shell out to) this binary instead, and nothing more?

If this is correct, then devs (and CI I guess) would just install _any_ Yarn version (assuming all versions have the same behavior of reading `.yarnrc.yml` and reading `yarnPath` from it)?

mandx (approved):

.yarnrc.yml
Dkendal: yes I believe that is accurate

PR #156, opened 2022-01-28
docs: add `QRL` docs.

mhevery:


jimt:
I've made some typo and grammatical suggestions in mhevery/qwik#1

I also made a couple of inline comments above where it was either a question of style or I didn't understand your intent.

mhevery:
@jimt Thank you for your improvments. I have incorporated them into the PR.

PR #165, opened 2022-02-01
feat(core): implement `onWatch`

### LINKED ISSUE #63: Design `qEffect` API
Components have a need for composability. `qEffect` should provide such a solution.

## Example 

Code example is worth a thousand words!

Imagine you have a piece of code that reports mouse position. A pseudo-code may look like this:

```typescript
interface MousePosition {
  x: number;
  y: number;
}

const useMousePosition = qEffect<{}, MousePosition>({
  onMount: () => ({ x: 0, y: 0 }),
  listen: {
    ['on:document:mousemove']: qHook((_, state: MousePosition) => {
      const mEvent = useEvent<MouseEvent>();
      state.x = mEvent.x;
      state.y = mEvent.y;
    }),
  },
});

```

Such code can then be used like so:

```typescript
export const Header = qComponent<{}, { mouse: MousePosition }>({
  onMount: qHook(() => ({ mouse: useMousePosition({}) })),
  onRender: qHook((_, { mouse }) => (
    <span>
      {mouse.x}, {mouse.y}
    </span>
  )),
});
```
## Goals
- The effect should be initialized on server and survive serialization to the client.
- The effect should be hydrated independently from the component using it.
- The effect should be composable (used by other effects)

mhevery:
Closes #63


src/core/api.md
```diff
@@ -129,6 +129,11 @@ export interface JSXNode<T = any> {
 // @public
 export function notifyRender(hostElement: Element): Promise<void>;
 
+// @public
+export interface Observer {
+    <T extends {}>(obj: T): T;
```
manucorporat, line 134: it would be interesting to make this more strongly typed, can we observe any kind of objects? I have an idea of how to achieve this, we can develop on a call.

anyway, good to merge for now!

src/core/api.md
```diff
@@ -165,6 +170,12 @@ export const onUnmount$: (first: () => void) => void;
 // @public
 export function onUnmount(unmountFn: QRL<() => void>): void;
 
+// @public
+export const onWatch$: (first: (obs: Observer) => unknown | (() => void)) => void;
```
manucorporat, line 174: if we can return a promise, then we should use ValueOrPromise
```suggestion
export const onWatch$: (first: (obs: Observer) => ValueOrPromise<unknown | (() => void)) => void>;

```

manucorporat (approved):
LGTM!

src/core/api.md
mhevery: Yes, let's get a discussion going.

src/core/api.md
manucorporat: gotcha! makes sense

PR #167, opened 2022-02-02
fix(qwikloader): Add support for multiples QRLs int single `on:event`

mhevery:


manucorporat (approved):
love it! but when merged I would call it a feat(), since it changes the types and developers can do a new thing that previously was not possible.

PR #183, opened 2022-02-07
fix: charset prop, update jsx types

adamdbradley:
Improve converting JSX props to HTML attribute and element properties.


manucorporat (approved):
merge!

Issue #231, opened 2022-02-16
aria-hidden="true" creates aria-hidden="'true'"


adamdbradley:
JSX

```jsx
<svg aria-hidden="true">
```

HTML
```jsx
<svg aria-hidden="'true'">
```

manucorporat:
Fixed https://github.com/BuilderIO/qwik/commit/7fadd186e2358c95cd5902a800c24b448e6d19c5

PR #232, opened 2022-02-17
fix(runtime): fix svg rendering

### LINKED ISSUE #134: Support for <svg>
Currently all elements are created using `createElement()` the result is that `<svg>` is not correctly created. Instead we should detect `<SVG>` and user `createElementNS()` to support SVG in the JSX.

manucorporat:
fixes #134 #231


retepibarra (approved):

manucorporat:
Part of the new runtime PR now

PR #241, opened 2022-02-26
Include Vite .gitignore for better DX

codercatdev:


manucorporat:
Can you update this file instead? https://github.com/BuilderIO/qwik/blob/main/starters/apps/base/.gitignore

codercatdev:
Updated to base @manucorporat

PR #401, opened 2022-04-14
Bug style undefined

armgjoka:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

This JSX: <div style={{display: undefined}}">
Results in: <div style="display: undefined">
Should be: <div>

Link: (https://github.com/BuilderIO/qwik/issues/340)


src/core/render/cursor.ts
```diff
@@ -546,7 +546,11 @@ const getSlots = (componentCtx: ComponentCtx | undefined, hostElm: Element): Slo
 };
 
 const handleStyle: PropHandler = (ctx, elm, _, newValue) => {
-  setAttribute(ctx, elm, 'style', stringifyClassOrStyle(newValue, false));
+  const style = stringifyClassOrStyle(newValue, false);
+  if (Object.keys(style).length > 0) {
```
manucorporat, line 550: I think it should be:
```
  if (style.length > 0) {
```
since. its already a string.

Actually, why do we need this if statement at all, I can see a problem that styles never get removed.

armgjoka: @manucorporat, yes you are right, for some reason saw that as object. My idea was to handle the case where style would be an empty string: 

<div style={{display: undefined}}></div> to <div style=''></div>.

PR #411, opened 2022-04-21
Added Node version file to Cloudflare Pages template.

isaac-mcfadyen:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

- By default, the Cloudflare Pages builder uses a Node version of 12 (because it's an older build image). This is not compatible with Qwik.
- The workaround is to add a NODE_VERSION environment variable or a `.node-version` file with a version number (like `16`).
- However, this is not done by default and the user should not have to do this (the template should "just work"). This PR adds a `.node-version` file to the Cloudflare Pages template so that the template works out-of-the-box.

# Use cases and why

- Make builds work without the user needing to add their own NODE_VERSION environment variable (not added by default on Pages).

# Checklist:

- [ ] I have made corresponding changes to the documentation **(Not applicable)**
- [ ] Added new tests to cover the fix / functionality **(Not applicable)**


adamdbradley:
Awesome, great catch, thanks!

PR #485, opened 2022-05-16
docs: Add clock example

mhevery:


adamdbradley:
Added in https://github.com/BuilderIO/qwik/pull/512

PR #587, opened 2022-06-09
docs: getting started

mhevery:


manucorporat:
I want to review where all this content fits inside the docs as a whole, wanna make sure there is not duplicated content and makes sense in terms of learning curve! Let’s talk

PR #655, opened 2022-06-21
feat: add @builder.io/qwik-react

manucorporat:
Implemented `qwikify` allowing usage of React components in a partial hydration fashion inside the no-hydration / resumability model of qwik components.

```tsx
/** @jsxImportSource react */

import { qwikify$ } from '../react/qwikify';
import { Button } from '@mui/material';

export const App = qwikify$(() => {
  return (
    <>
      <Button variant="contained">Hola</Button>
    </>
  );
});
```

Allowing the usage of the qwik component inside the existing qwik app:
```tsx
import { App } from './examples/app';
export const Root = () => {
  return (
    <html>
      <body>
        <App client:visible />
      </body>
    </html>
  );
};
```

or no hydration at all:
```tsx
import { App } from './examples/app';
export const Root = () => {
  return (
    <html>
      <body>
        <App />
      </body>
    </html>
  );
};
```


manucorporat:
oh! good catch! i meant no client:PROP at all. it will ssr the react component in the server, and never ever download or execute anything!

PR #666, opened 2022-06-25
Fix syntax errors for React example code

Lacoz:
# What is it?

- [x] Docs / tests


manucorporat:
Thanks so much! It would be nice to add even more examples too!

PR #691, opened 2022-06-28
fix: window and document events

manucorporat:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/qwikloader.ts
+++ b/packages/qwik/src/qwikloader.ts
@@ -9,13 +9,12 @@
  */
 export const qwikLoader = (doc: Document, hasInitialized?: number, prefetchWorker?: Worker) => {
   const Q_CONTEXT = '__q_context__';
-  const ON_PREFIXES = ['on:', 'on-window:', 'on-document:'];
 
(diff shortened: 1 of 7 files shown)


packages/qwik/src/qwikloader.ts
```diff
@@ -139,8 +132,9 @@ export const qwikLoader = (doc: Document, hasInitialized?: number, prefetchWorke
     }
   };
 
-  const addDocEventListener = (eventName: string) =>
-    doc.addEventListener(eventName, processEvent, { capture: true });
+  const addDocEventListener = (eventName: string) => {
+    window.addEventListener(eventName, processWindowEvent, { capture: true });
```
mhevery, line 136: This line breaks listening on `mousenter`. 

My theory: `mouseenter` is a non-bubbling event. This is the reason why we have to do `capture: true`. However, because it is non-bubbling it never gets to the `window` it only stays on the document. So I think this breaks the code.

PR #709, opened 2022-07-01
added docs about useMount$

tidiview:
according to see Miško#9885
https://discord.com/channels/842438759945601056/992449727994011759/992470886554812429

### What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

### DIFF
--- a/packages/docs/src/pages/docs/components/hooks.mdx
+++ b/packages/docs/src/pages/docs/components/hooks.mdx
@@ -169,7 +169,7 @@ const Cmp = component$(() => {
     users: []
   });
   useMount$(async () => {
-    // This code will run on the server, then on the client
+    // This code will run on component creation to fetch the data.
     store.users = await db.requestUsers();
   });
   return (
--- a/packages/docs/src/pages/docs/components/hooks.mdx
+++ b/packages/docs/src/pages/docs/components/hooks.mdx
@@ -155,7 +155,11 @@ const Timer = component$(() => {
 
 ## useMount$()
 
-Registers a mount hook, that will run on the server, then on the client.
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/docs/components/hooks.mdx
```diff
@@ -152,6 +152,43 @@ const Timer = component$(() => {
 });
 ```
 
+
+## useMount$()
+
+Registers a mount hook, that will run on the server, then on the client.
```
mhevery, line 158:
```suggestion
Register a hook that executes on component creation.  `useMount$()` will block the rendering of the component until after `useMount$()` callback resolves. (This is useful for fetching asynchronous data and delaying rendering until data is received, ensuring that the rendered component contains the data.).

While `useMount$()` executes both on the server and on the client, it runs exactly once. (Either on the server or on the client depending on where the component got first rendered.)

See also: `useClientMount$()` and `useServerMount$()` that have same semantics but only run on client or server respectively.  
```

packages/docs/src/pages/docs/components/hooks.mdx
```diff
@@ -152,6 +152,43 @@ const Timer = component$(() => {
...
+
+### Example
+
+```tsx
+const Cmp = component$(() => {
+  const store = useStore({
+    users: []
+  });
+  useMount$(async () => {
+    // This code will run on the server, then on the client
```
mhevery, line 168:
```suggestion
    // This code will run on component creation to fetch the data.
```

mhevery (changes requested):

mhevery (approved):

PR #764, opened 2022-07-13
docs: Light Component - edit pass

craigshoemaker:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Edit pass.

# Use cases and why

n/a

# Checklist:

### DIFF
--- a/packages/docs/src/pages/tutorial/component/inline/index.mdx
+++ b/packages/docs/src/pages/tutorial/component/inline/index.mdx
@@ -3,8 +3,12 @@ title: Light Component
 layout: tutorial
 ---
 
-Qwik components are lazy-loadable. This is achieved through `component$()` method. The presence of `$` crates a lazy-loaded boundary. The `$` will be discussed later. Qwik also supports inline components. Inline components are not by themselves lazy loadable and instead load as part of the parent component where they are used. (Inline components are equivalent to regular components in most other frameworks.)
+Qwik components are lazy-loadable, which lies at the heart of what makes Qwik different from most other frameworks.
 
-In this example we have pre-declared an `<App>` and a `<Greeter>` component. The `<Greeter/>` component is currently a Qwik Component because it is declared with `component$()`. Remove the `component$()` to convert `<Greeter>` to inline component. Inline components do not have host elements, (described later.)
+You can mark a component as lazy-loadable by using the `component$()` method. When you see a `$` in Qwik code, you're encountering a lazy-loading boundary.
 
-Open the `Symbols` tab and notice that the `<Greeter/>` component is no longer an independent export, and instead it is bundled as part of the `<App>` component.
+If you want to ensure a component loads with another component, you create an inline component. Inline components load as a part of the parent component and are equivalent to how most other frameworks deal with components.
+
+In this example, the `<App>` and a `<Greeter>` components are prepared for you. The `<Greeter/>` component is declared using the `component$()` method and is a Qwik component.  Remove the `component$()` to convert `<Greeter>` to an inline component.
+
+Open the `Symbols` tab and notice that the `<Greeter/>` component is no longer an independent export, but instead is bundled as part of the `<App>` component.


packages/docs/src/pages/tutorial/component/inline/index.mdx
```diff
@@ -3,8 +3,12 @@ title: Light Component
 layout: tutorial
 ---
 
-Qwik components are lazy-loadable. This is achieved through `component$()` method. The presence of `$` crates a lazy-loaded boundary. The `$` will be discussed later. Qwik also supports inline components. Inline components are not by themselves lazy loadable and instead load as part of the parent component where they are used. (Inline components are equivalent to regular components in most other frameworks.)
+Qwik components are lazy-loadable, which lies at the heart of what makes Qwik different from most other frameworks.
 
-In this example we have pre-declared an `<App>` and a `<Greeter>` component. The `<Greeter/>` component is currently a Qwik Component because it is declared with `component$()`. Remove the `component$()` to convert `<Greeter>` to inline component. Inline components do not have host elements, (described later.)
+You can mark a component as lazy-loadable by using the `component$()` method. When you see a `$` in Qwik code, you're encountering a lazy-loading boundary.
```
shairez, line 8: @craigshoemaker I would change the word "encountering" to "entering" like so -  "when you see a '$' in Qwik code, you are entering a lazy loaded boundry. "

@mhevery WDYT?

shairez:
Thanks @craigshoemaker 
Added some comments

packages/docs/src/pages/tutorial/component/inline/index.mdx
```diff
@@ -3,8 +3,12 @@ title: Light Component
 layout: tutorial
 ---
 
-Qwik components are lazy-loadable. This is achieved through `component$()` method. The presence of `$` crates a lazy-loaded boundary. The `$` will be discussed later. Qwik also supports inline components. Inline components are not by themselves lazy loadable and instead load as part of the parent component where they are used. (Inline components are equivalent to regular components in most other frameworks.)
+Qwik components are lazy-loadable, which lies at the heart of what makes Qwik different from most other frameworks.
```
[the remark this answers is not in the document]
mhevery: Yes totally agree. 

Qwik components are:
- Lazy-loadable=> But all frameworks claim that => Qwik can lazy-load even the components that are in the current DOM (other frameworks can only lazy-load components that are currently not in the DOM)
- Can be lazy-loaded out of order => Meaning a middle component can be resumed without waking up the parent or a child component. (Other frameworks always lazy from root towards children, and they can never skip over things)
- Loaded on interaction (Other frameworks always have to eagerly load code before interaction)
- But it it is not just components. Qwik really lazy loads any closure in this way. Event listeners, effects, behavior etc...

mhevery (approved):
Rest of the changes look great!

packages/docs/src/pages/tutorial/component/inline/index.mdx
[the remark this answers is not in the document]
shairez: Thanks @mhevery !

@craigshoemaker what are your thoughts? you're the docs master here :)
Can you think of a good way to incoporate this feedback in this doc? 

thanks 🙏

packages/docs/src/pages/tutorial/component/inline/index.mdx
shairez: @mhevery thanks!
Maybe it's a good point to mention the limitations of what "crossing the boundary" means? (to give more weight to it, so developers will know "this is an important point, I should make a mental note here")

WDYT?

mhevery: Yes! Every time you see `$`, you (as a developer) should be thinking special rules apply here. Not every valid JS is valid Qwik code.

packages/docs/src/pages/tutorial/component/inline/index.mdx
craigshoemaker, line 6:
```suggestion
One of Qwik's super powers lies in its lazy-loading features. 

Qwik lazy-loads:

- **Independent of DOM presence**: Components can load before and after they are added to the DOM.
- **Independent of hierarchy**: Components can be loaded out of order. For example, the script for a child component can load before its parent script.
- **Based on interaction**: Script loading is deferred until a user interacts with a component.
- **More than just components**: Qwik lazy-loads any closure including components, event listeners, effects, and behaviors.
```

craigshoemaker (commented):
@mhevery / @shairez - here's another pass that integrates the feedback into the introduction

packages/docs/src/pages/tutorial/component/inline/index.mdx
shairez: @mhevery do you think we should add these rules in this section as well?

packages/docs/src/pages/tutorial/component/inline/index.mdx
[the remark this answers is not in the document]
shairez: The new changes based on this thread LGTM
Thanks @craigshoemaker !

Let's see what Misko is thinking on the other thread, I think we should add the rules, WDYT Craig?

packages/docs/src/pages/tutorial/component/inline/index.mdx
mhevery: Hmm, they should probably be both here and in the https://qwik.builder.io/docs/advanced/optimizer!
```
- all captured variables must be declared `const`
- all captured variables must be either:
  - serializable
  - importable (either `import` or `export` in this file)
```

packages/docs/src/pages/tutorial/component/inline/index.mdx
```diff
@@ -3,8 +3,19 @@ title: Light Component
...
 ---
 
-Qwik components are lazy-loadable. This is achieved through `component$()` method. The presence of `$` crates a lazy-loaded boundary. The `$` will be discussed later. Qwik also supports inline components. Inline components are not by themselves lazy loadable and instead load as part of the parent component where they are used. (Inline components are equivalent to regular components in most other frameworks.)
+One of Qwik's super powers lies in its lazy-loading features. 
 
-In this example we have pre-declared an `<App>` and a `<Greeter>` component. The `<Greeter/>` component is currently a Qwik Component because it is declared with `component$()`. Remove the `component$()` to convert `<Greeter>` to inline component. Inline components do not have host elements, (described later.)
+Qwik lazy-loads:
 
-Open the `Symbols` tab and notice that the `<Greeter/>` component is no longer an independent export, and instead it is bundled as part of the `<App>` component.
+- **Independent of DOM presence**: Components can load before and after they are added to the DOM.
```
mhevery, line 10:
```suggestion
```
Not sure what you were going for, but the above does not make sense without the context of Qwik. So I would just remove it.

packages/docs/src/pages/tutorial/component/inline/index.mdx
```diff
@@ -3,8 +3,19 @@ title: Light Component
...
+- **Independent of hierarchy**: Components can be loaded out of order. For example, the script for a child component can load before its parent script.
```
mhevery, line 11:
```suggestion
- **Independent of hierarchy**: Components can be loaded out of order. For example, the code for a child component can load before its parent code.
```

packages/docs/src/pages/tutorial/component/inline/index.mdx
```diff
@@ -3,8 +3,19 @@ title: Light Component
...
+- **Based on interaction**: Script loading is deferred until a user interacts with a component.
```
mhevery, line 12:
```suggestion
- **Based on interaction**: Code loading is deferred until a user interacts with a component.
```

packages/docs/src/pages/tutorial/component/inline/index.mdx
```diff
@@ -3,8 +3,19 @@ title: Light Component
...
+- **More than just components**: Qwik lazy-loads any closure including components, event listeners, effects, and behaviors.
+
+You can mark a component as lazy-loadable by using the `component$()` method. When you see a `$` in Qwik code, you're crossing a lazy-loading boundary.
```
mhevery, line 15:
```suggestion
`$` marks a closure as lazy-loadable. For example `component$()` method makes the component lazy-loadable. When you see a `$` in Qwik code, you're crossing a lazy-loading boundary and have to be aware of special rules:
- any lexically scoped variables must be declared as `const`
- a captured variable/symbol must be either:
  - serializable 
  - importable (Either from a different file as `import` or from this file using `export`)
```

packages/docs/src/pages/tutorial/component/inline/index.mdx
shairez: Awesome!
Thanks @mhevery

packages/docs/src/pages/tutorial/component/inline/index.mdx
craigshoemaker: It was an attempt to work with this comment:

> Lazy-loadable=> But all frameworks claim that => Qwik can lazy-load even the components that are in the current DOM (other frameworks can only lazy-load components that are currently not in the DOM)

perhaps there's another way to word it?

packages/docs/src/pages/tutorial/component/inline/index.mdx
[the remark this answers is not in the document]
craigshoemaker: @shairez - they are now in the text 🎉
@mhevery - feel free to merge if you like the current version

packages/docs/src/pages/tutorial/component/inline/index.mdx
craigshoemaker: Here's #777 to add these rules to the Optimizer doc 😎

mhevery (approved):

PR #841, opened 2022-07-26
DX: Refine useEndpoint generic

nnelgxorz:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Currently it is hard to type your useEndpoint call to your endpoint function.

Consider we have an `onGet` defined as an `EndpointHandler<string>`

Current behavior:
```ts
const onGet: EndpointHandler<string> = () => {
  return "Oh Qwik..."
}
const resource = useEndpoint(); // ResourceReturn<unknown>
const resource = useEndpoint<typeof onGet>(); // ResourceReturn<EndpointHandler<string>
const resource = useEndpoint<ReturnType<typeof onGet>>(); // ResouceReturn<string|void|null|undefined|Promise<string|void|null|undefined>>
```

Of course, you could type it yourself, ie `useEndpoint<string>`, but any changes to the return type of `onGet` do not trigger a type error for `useEndpoint`, making it a brittle solution. Also, you have the option of incorrectly type your `useEndpoint` call and not receiving an error until runtime. :grimacing: 

By adding a little helper type to the useEndpoint<T> definition, we can check the generic, and if T extends EndpointHandler, extract the EndpointHandler type, otherwise use the original T type.

This allows developers to pass the relatively terse `useEndpoint<typeof onGet>()` and receive proper typing.

This gives the following behavior:

```ts
const onGet: EndpointHandler<string> = () => {
  return "Yay Qwik!"
}
const resource = useEndpoint(); // ResourceReturn<unknown>
const resource = useEndpoint<boolean>(); // ResourceReturn<boolean>
const resource = useEndpoint<null>(); // ResourceReturn<null>
//...etc
const resource = useEndpoint<typeof onGet>(); // ResourceReturn<string>
``` 

# Checklist:

### DIFF
--- a/packages/docs/src/pages/qwikcity/endpoint/data.mdx
+++ b/packages/docs/src/pages/qwikcity/endpoint/data.mdx
@@ -67,7 +67,7 @@ interface ProductData { ... }
 export const onGet: EndpointHandler<EndpointData> = async ({ params }) => { ... };
 
 export default component$(() => {
-  const resource = useEndpoint<EndpointData>();
+  const resource = useEndpoint<typeof onGet>(); // equivalent to useEndpoint<EndpointData>
   return (
     <Resource 
       resource={resource}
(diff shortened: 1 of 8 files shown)


manucorporat:
Thanks a lot for the PR!! love it!

PR #907, opened 2022-08-03
feat: qwik-city non-blocking endpoint response

adamdbradley:
By default, an endpoint handler waits on the returned promise, to include waiting on any data loading. In the example below we set the headers immediately, but have to wait on the data to finish loading before we can continue, which means the streaming is held up.

```ts
const onGet: EndpointHandler = ({ response }) => {
    response.headers.set('name', 'value');
    const data = await longDatabaseQuery();
    return data;
};
```

This PR allows for the endpoint handlers to also have the ability to return a callback instead, which doesn't hold up the http response and streaming.

```ts
const onGet: EndpointHandler = ({ response }) => {
    response.headers.set('name', 'value');
    return () => {
        const data = await longDatabaseQuery();
        return data;
    };
};
```

Now the endpoint can set the headers/status immediately, and let HTTP streaming can continue. And by the time the renderer gets to where the data is actually needed, it would have already streamed up until that point. Next the renderer/stream will wait on the promise.


mhevery:
Really love how this turned out.

PR #912, opened 2022-08-03
refactor: ship latest API stabilizations

manucorporat:
# Breaking changes

- `useScopedStyles()` -> `useStylesScoped()`
- `useWatch$((track) => {})` -> `useWatch$(({track}) => {})`
- `useClientEffect$((track) => {})` -> `useClientEffect$(({track}) => {})`
- `useServerMount$(): Resource<>` -> `useServerMount$(): void`
- `useMount$(): Resource<>` -> `useMount$(): void`

### DIFF
--- a/.vscode/qwik.code-snippets
+++ b/.vscode/qwik.code-snippets
@@ -66,7 +66,7 @@
 		"prefix": "useClientEffect",
 		"description": "useClientEffect$() function hook",
 		"body": [
-			"useClientEffect$((track) => {",
+			"useClientEffect$(({track}) => {",
 			"  $0",
 			"});",
 			""
@@ -77,7 +77,7 @@
 		"prefix": "useWatch",
 		"description": "useWatch$() function hook",
 		"body": [
-			"useWatch$((track) => {",
+			"useWatch$(({track}) => {",
 			"  track($1);",
 			"  $0",
 			"});",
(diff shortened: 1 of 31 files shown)


shairez:
Thanks Manu!

One issue that I forgot to mention yesterday - what's the difference between `useWatch` and `useClientEffect` ?

I'm guessing `useCLientEffect$` is running only on the client... but why should we use `useWatch` then?

manucorporat:
useWatch runs before the render, and can run in both places server and client. In the client it never runs eagerly.

We might need more docs about this

shairez:
Thanks! 

If that's the case, I think it might be less confusing to call it something similar to useClientEffect - like... `useEffect` ... or to change `useClientEffect` to `useClientWatch` to match them 

Does it make sense?

manucorporat:
This API has went through a lot of renamed in fact;
useWatchEffect,
useEffect, and finally useClientEffect.

We want to tag as client or server APIs that only run on those platforms, for now only `useClientEffect` and `useServerMount`.

useEffect would conflict with useEffect semantics in React, and that was also a reason to avoid that name.

manucorporat:
Also, watch and effect has different semantics:
- effect: after render
- watch: on data change before render (giving a better time for updating data before render)

PR #921, opened 2022-08-04
Change default prefetch to Worker-Fetch

nnelgxorz:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Link prefetch has been shown to only work on Chrome, other browsers ignore it. Worker-fetch is a better default as it works across all browsers.

# Checklist:

### DIFF
--- a/packages/docs/src/entry.cloudflare.tsx
+++ b/packages/docs/src/entry.cloudflare.tsx
@@ -3,7 +3,7 @@ import { qwikCity } from '@builder.io/qwik-city/middleware/cloudflare-pages';
 
 const qwikCityMiddleware = qwikCity(render, {
   prefetchStrategy: {
-    implementation: 'link-prefetch',
+    implementation: 'worker-fetch',
   },
 });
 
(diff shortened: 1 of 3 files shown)


manucorporat:
While i agree there is a problem with the current default, this mode has problem too, and shows worse performance than the native link... I wonder if we can create something better that uses the best of chrome while supported in other browsers..

I dont like the idea of choosing the lowest common denominator approach. If Safari/firefox is bad, it should not affect chrome users.

nnelgxorz:
I have no doubt there's a better solution here.

For the time being, I think using the worker fetch implementation is the lesser of two evils. I'm just thinking of the Firefox/Safari user who deploys a starter and gets no prefetching at all.

adamdbradley:
Closed with https://github.com/BuilderIO/qwik/pull/935

kylecordes:
I wish there was a polyfill-like solution that helps other browsers actually prefetch when give the standard, declarative hints.

PR #933, opened 2022-08-04
feat: change default host element

manucorporat:
By default, qwik component will not create a `<div>`, instead it will render a `q:host`, which by default will get `display: contents` and be transparent in terms of layout.

But the component still exists, so it not transparent for css selectors, like sibling, last, end, or direct children


mhevery:
Wait! We are duplicating work: https://github.com/BuilderIO/qwik/pull/936

mhevery:
i believe this can be closed.

PR #934, opened 2022-08-04
Redirect docs and onGet usage clarification

nnelgxorz:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Some users may be confused by the current docs implying the onGet function is only for data fetching.

From on Bjorn on discord:

> My understanding (which isn't a lot at this stage as I'm quite new to QwikCity) is that the idea behind endpoints is to fetch data for a component. For my use case it's solely to be used as a redirect. 
Is there another way to handle this use case, which I might be missing? 

# Checklist:

### DIFF
--- a/packages/docs/src/routes/qwikcity/endpoint/non-200.mdx
+++ b/packages/docs/src/routes/qwikcity/endpoint/non-200.mdx
@@ -70,7 +70,9 @@ export default component$(() => {
 
 ## 3xx - Redirect
 
-Let's say that a user does a request to an invalid `skuId` such as `https://example.com/product/999/details`. In this case, we would like to return a 404 HTTP status code and render a 404 page. The place where we determine if the request is valid or not is the data endpoint by looking into the database. Even if the response is non-200, the component still gets a chance to render a page (Except in the redirect case.)
+Sometimes you want to redirect a user from the current page to another page.
+
+Let's say a user tries to go to a dashboard page, but has not logged in yet. We want them to be redirect to a log-in page so they can be authenticated.
 
 ```typescript
 // File: src/routes/dashboard.js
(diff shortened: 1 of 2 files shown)


packages/docs/src/routes/qwikcity/endpoint/non-200.mdx
```diff
@@ -65,3 +67,39 @@ export default component$(() => {
...
+
+The `response.redirect` function takes a url and optionally a status code as the second argument. 
+
+```tsx
+return response.redirect('/login', 301) // Permanent redirect
+```
+
+Redirect status codes are between 300-399. 
+
+If you do not provide a status code, Qwik City will default to a 307 temporty redirect status.
```
mhevery, line 103:
```suggestion
If you do not provide a status code, Qwik City will default to a 307 temporary redirect status.
```

packages/docs/src/routes/qwikcity/endpoint/non-200.mdx
```diff
@@ -65,3 +67,39 @@ export default component$(() => {
   );
 });
 ```
+
+## 3xx - Redirect
+
+Let's say that a user does a request to an invalid `skuId` such as `https://example.com/product/999/details`. In this case, we would like to return a 404 HTTP status code and render a 404 page. The place where we determine if the request is valid or not is the data endpoint by looking into the database. Even if the response is non-200, the component still gets a chance to render a page (Except in the redirect case.)
```
mhevery, line 73: I think the text sholud talk about 303 not 404. Looks like copy-pasta.

mhevery (changes requested):

packages/docs/src/routes/qwikcity/endpoint/non-200.mdx
nnelgxorz: Fixed!

packages/docs/src/routes/qwikcity/endpoint/non-200.mdx
nnelgxorz: Fixed!

adamdbradley (approved):

PR #935, opened 2022-08-04
feat: update prefetch implementation options

adamdbradley:
- Add the ability to render both html link, and run web worker fetch
- Change the prefetch implementation option to an object rather than string

Co-Authored-By: Glenn Becker <nnelgxorz@gmail.com>


adamdbradley:
Closes #921

PR #936, opened 2022-08-04
fix(host): Set host to `<q:host>` and `display: contents`

### LINKED ISSUE #930: Remove <Host/> component

If it's "possible to remove `<Host/>`, let's remove it.

Idea from a discussion @mhevery and I just had:
- Internally, there always needs to be an actual actual <div> (or whatever) host component
- However, what if the <Host/> component user's write never existed at all, it's not something they import and use
- And the "host" component qwik generates, ALWAYS is [display: contents](https://developer.mozilla.org/en-US/docs/Web/CSS/display#box)??
- So "qwik's" host is kinda never really there to begin with.
- It's a host element for qwik to interact with, but it's not something to the user has to worry about since it's basically silent.
- The "user's" host is whatever the heck they want it to be, or no host at all
- Exactly like react in how it can return a top level element, or a fragment with an array of children

So user could return

```tsx
return (
  <footer>my footer</footer>
)
```

And qwik would generate

```tsx
<div q:host>
  <footer>myfooter</footer>
</div>
```

Or the user could return

```tsx
return (
  <Fragment>
    <span>a</span>
    <span>b</span>
  </Fragment>
)
```

And qwik would generate

```tsx
<div q:host>
    <span>a</span>
    <span>b</span>
</div>
```

And now the component's css would flow how they'd expect, and their component could continue act like display inline if that's what the top level components were. Just like how react devs would expect their components to render and flow. 

This allows us to completely remove `<Host/>`, and now users wouldn't have to worry about adding `{ tagName: 'footer' }`, which are both strange to traditional JSX frameworks.

```tsx
    return (
      <Host>
        <ul>
          ...
        </ul>
      </Host>
    );
  },
  {
    tagName: 'footer',
  }
);
```

becomes

```tsx
    return (
      <footer>
        <ul>...</ul>
      </footer>
    );
  } 
);
```

I also think this makes the internal code much easier now, because it doesn't have to worry about merging attributes.

For example, let's say that below is a React app.

```tsx
export default function App() {
  return (
    <div>
      <CmpA style={ { color: 'red' } }/>
    </div>
  );
}

function CmpA() {
  return (
    <>
      <h1>Hello!</h1>
    </>
  );
}
```

The red color `styles` wouldn't actually apply to the `h1`, cuz what would the "style" be getting assigned to? It's a component prop, not an actual attribute. The example above wouldn't work in React (which is expected).

Thoughts

Keep `<Host/>` :

### Additional context

_No response_

mhevery:
Fixes #930

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
     "release.prepare": "yarn lint && yarn test.unit && tsm scripts/index.ts --tsc --build --api --eslint --platform-binding --wasm --prepare-release",
     "test": "yarn build.full && yarn test.unit && yarn test.e2e",
     "test.unit": "tsm node_modules/uvu/bin.js packages unit.ts --tsmconfig tsm.cjs",
+    "test.unit.styles": "tsm node_modules/uvu/bin.js packages scoped-stylesheet.unit.ts --tsmconfig tsm.cjs",
+    "test.unit.styles.debug": "tsm --inspect-brk node_modules/uvu/bin.js packages scoped-stylesheet.unit.ts --tsmconfig tsm.cjs",
     "test.unit.debug": "tsm --inspect-brk node_modules/uvu/bin.js packages unit.ts --tsmconfig tsm.cjs",
     "test.watch": "watchlist packages unit.ts -- yarn test.unit",
     "test.rust": "make test",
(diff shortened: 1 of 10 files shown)


manucorporat:
We cant merge since there is a e2e test, something with this change affects slot projection

PR #984, opened 2022-08-09
Make Content-Type json validation less rigid so it works with things like application/manifest+json (Fixes #983)

### LINKED ISSUE #983: (Endpoints) Content-Type validation is too rigid and gives an error
### Qwik Version

Qwik-City: 0.0.26

Node

_No response_

Qwik City

### Expected Behaviour

I have `routes/manifest.webmanifest.ts` so I dinamically have /manifest.webmanifest URL, but I can't set the headers properly.

```js
response.headers.set('Content-Type', 'application/manifest+json; charset=utf-8')
```

This gives: `Unsupport response body type` when trying to access the URL.

Using `application/json` works fine, but it's not correct.

### Actual Behaviour

See above

### Additional Information

_No response_

felixsanz:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

There are Content-Types that are variations of `application/json`, like `application/manifest+json`, which is needed to output a route like `/manifest.webmanifest.ts`. This PR make it less rigid so it works with those variations.

# Use cases and why

- 1. One use case

Using `routes/manifest.webmanifest.ts` needs to print  `application/manifest+json`.

- 2. Another use case

`application/ld+json` and other variations.

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/request-handler/endpoint-handler.ts
+++ b/packages/qwik-city/middleware/request-handler/endpoint-handler.ts
@@ -18,7 +18,7 @@ export function endpointHandler<T = any>(
   }
 
   // check so we can know later on if we should JSON.stringify the body
-  const isJson = headers.get('Content-Type')!.includes('application/json');
+  const isJson = headers.get('Content-Type')!.includes('json');
 
   return response(status, headers, async ({ write }) => {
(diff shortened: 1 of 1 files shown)


felixsanz:
This doesn't fix it, my bad

adamdbradley:
great idea, thanks!

PR #1019, opened 2022-08-10
feat: remove component$ host element

manucorporat:
This is a big change, but 100% worthy. component$() is not longer generating a host element, removing the need for <Host> and tagName API of component$().

WIP


nnelgxorz:
This is awesome work, Manu!

PR #1041, opened 2022-08-11
fix: upgrade docs-site qwik and city version

youngboy:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

fixed https://github.com/BuilderIO/qwik/issues/1037

and I think docs site no longer need to pin specific version of qwik. so add `^` version prefix to match workspace package if possible.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -19,8 +19,8 @@
   },
   "devDependencies": {
     "@builder.io/partytown": "^0.6.4",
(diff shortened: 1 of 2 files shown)


manucorporat:
Let release 0.0.100 first, we merged the version bump, but still we need to trigger the release process

PR #1070, opened 2022-08-16
docs: fix and improve React examples

rajington:
# What is it?

- [x] Docs / tests

# Description

Some React examples had unnecessary `$` or missed things like `key` (which would be good to call out if qwik behaves differently).

# Use cases and why

- 1. Better comparison to status quo

# Checklist:

- [NA] Added new tests to cover the fix / functionality

### DIFF
--- a/packages/docs/src/routes/docs/cheat-sheet/index.mdx
+++ b/packages/docs/src/routes/docs/cheat-sheet/index.mdx
@@ -76,7 +76,7 @@ export function Component() {
   const [value, setValue] = useState(0);
   return (
     <div>
-      Value is: ${value}
+      Value is: {value}
     </div>
   )
 }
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/cheat-sheet/index.mdx
```diff
@@ -299,7 +299,7 @@ export function DynamicBackground() {
   const [blue, setBlue] = useState(0);
   return (
     <div style={
-      'background': `rgb(${red}, ${green}, ${blue})`
+      'background': `rgb({red}, {green}, {blue})`
```
farukEncoded, line 302: Actually, you will need those `$` signs here. and also an extra pair of curly braces. like this 
```jsx
style={{
      'background': `rgb(${red}, ${green}, ${blue})`
    }}
```

farukEncoded (changes requested):
Actually, you will need those `$` signs here. and also an extra pair of curly braces. like this
```jsx
style={{
      'background': `rgb(${red}, ${green}, ${blue})`
    }}
```

packages/docs/src/routes/docs/cheat-sheet/index.mdx
rajington: great call

manucorporat:
Thanks a lot for the PR! merged!

PR #1073, opened 2022-08-17
fix: runtimeQRL

youngboy:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

fixes https://github.com/BuilderIO/qwik/issues/860

Not fully understand the context thus not very confident on my changes currently, just tested fine against the issue example, feel free to close it 😂. But I think there's two problems here. 

1) not able to get `runtimeQRL` on event dispatch ;

original way `window object & import` just won't works,  tried look into changes history, failed to find why `window` involved here(maybe sort of custom prefetch ?). Instead, I just directly dig into element's internal attributes to get runtime handler.

2) each time re-render will add new event handler copy, cause bind events handler function duplicately.

from the source code, every time in pure client side (no optimizer processing) calling `runtimeUrl` or `$` will generate new `qrl` object resulting failed to compare with existed $listeners and remove old on. Thinking in client side logic need to use inlineQRL instead.

WIP: have a closer look on unit testing

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/props/props-on.ts
+++ b/packages/qwik/src/core/props/props-on.ts
@@ -1,8 +1,7 @@
-import { parseQRL } from '../import/qrl';
+import { inlinedQrl, parseQRL } from '../import/qrl';
 import { isQrl, isSameQRL, QRLInternal } from '../import/qrl-class';
 import type { QContext } from './props';
 import { isArray } from '../util/types';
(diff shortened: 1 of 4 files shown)


youngboy:
https://github.com/BuilderIO/qwik/issues/860 seems fixes already

PR #1075, opened 2022-08-17
feat/fix: Export mdx frontmatter variables to the developer

### LINKED ISSUE #1020: MDX frontmatter is kinda broken
### Qwik Version

0.0.41

Node

_No response_

Qwik City

### Expected Behaviour

Qwik city imports `remark-frontmatter` but seems like is not used since `packages/qwik-city/buildtime/markdown/frontmatter.ts` file contains this code:

```ts
export function parseFrontmatterAttrs(attrs: FrontmatterAttrs, yaml: string) {
  if (typeof yaml === 'string') {
    yaml = yaml.trim();
    if (yaml !== '') {
      const lines = yaml.split(/\r?\n|\r|\n/g);
      for (const line of lines) {
        const parts = line.split(':');
        if (parts.length > 1) {
          const attrName = parts[0].trim();
          const attrValue = parts.slice(1).join(':').trim();
          attrs[attrName] = attrValue;
        }
      }
    }
  }
  return null;
}
```

But this function does not work well, it doesn't accept yaml arrays or objects, etc. Can we remove this and make use of the remark plugin instead since it's already mentioned in package.json?

Also, the function below it (`frontmatterAttrsToDocumentHead`) removes all frontmatter variables except title and description, so you can't have a decent frontmatter with dates, tags, author, etc!

### Actual Behaviour

.

### Additional Information

_No response_

felixsanz:
# What is it?

- [x] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Currently the frontmatter is not exported to the user, instead it just sets `head: { title, description }` if they are found.

This PR exports all the variables the user has defined in the frontmatter.

This PR also replaces the custom parser implementation with the `yaml` module, because the custom impl. only parsed string variables and this one accepts lists, dictionaries, multi-line strings, etc (full yaml syntax).

Also added some tests and modified some of the existing ones because they were testing for invalid yaml syntax, which should never pass. AKA: It may be a breaking change if people is using yaml incorrectly. (Specifically using semicolon `:` symbol without wrapping the string in quotes).

Fixes #1020

# Use cases and why

- 1. One use case

Ability to create variables in the mdx file and make use of them.

- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/buildtime/markdown/frontmatter.ts
+++ b/packages/qwik-city/buildtime/markdown/frontmatter.ts
@@ -2,6 +2,7 @@ import type { Transformer } from 'unified';
 import type { BuildContext, FrontmatterAttrs } from '../types';
 import { normalizePath } from '../utils/fs';
 import { visit } from 'unist-util-visit';
+import { parse as parseYaml } from 'yaml';
 import type { ResolvedDocumentHead } from '../../runtime/src';
 
 export function parseFrontmatter(ctx: BuildContext): Transformer {
@@ -9,7 +10,10 @@ export function parseFrontmatter(ctx: BuildContext): Transformer {
     const attrs: FrontmatterAttrs = {};
 
     visit(mdast, 'yaml', (node: any) => {
-      parseFrontmatterAttrs(attrs, node.value);
+      const parsedAttrs = parseFrontmatterAttrs(node.value);
+      for (const k in parsedAttrs) {
+        attrs[k] = parsedAttrs[k];
+      }
     });
 
     if (Object.keys(attrs).length > 0) {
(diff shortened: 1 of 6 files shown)


packages/qwik-city/buildtime/markdown/frontmatter.ts
```diff
@@ -2,14 +2,18 @@ import type { Transformer } from 'unified';
...
 import type { ResolvedDocumentHead } from '../../runtime/src';
 
 export function parseFrontmatter(ctx: BuildContext): Transformer {
   return (mdast, vfile) => {
     const attrs: FrontmatterAttrs = {};
 
     visit(mdast, 'yaml', (node: any) => {
-      parseFrontmatterAttrs(attrs, node.value);
+      const parsedAttrs = parseFrontmatterAttrs(node.value);
+      for (const k in parsedAttrs) {
```
mhevery, line 14: I am a bit confused. Getting `parsedAttrs` and then coping it over to `attrs` seems like an unnecessary step. Can we just assign directly to `attrs`?

packages/qwik-city/buildtime/types.ts
```diff
@@ -14,7 +14,7 @@ export interface BuildContext {
 }
 
 export interface FrontmatterAttrs {
-  [attrName: string]: string;
+  [attrName: string]: any;
```
mhevery, line 17: why `any` Why is `string` not good enough? Can we do better then`any`?

packages/qwik-city/buildtime/markdown/frontmatter.ts
felixsanz: It is indeed unnecessary but that's how it was before and I maintained that. In theory it let you specify multiple frontmatters in the file, so all variables will just append (but who wants that anyway? haha)

packages/qwik-city/buildtime/types.ts
felixsanz: Because yaml accepts arrays (lists), objects (dictionaries), etc

mhevery (approved):

PR #1102, opened 2022-08-21
Update index.mdx

tidiview:
if you're not telling that a noSerialize import exist, how is it possible to guess it ?!?

# What is it?
- [x] Docs / tests


manucorporat:
Thanks a lot!

PR #1105, opened 2022-08-22
Use req.url as the base path to construct the url in the express middleware

steverandy:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

When using the request object from an endpoint, currently there is a slight difference between dev and serve (express middleware). The request.url in dev includes search string but when using express middleware the search query is missing.

# Use cases and why

Actual:

request.url => `/path`

Expected: 

request.url => `/path?param=1`

# Checklist:


manucorporat:
Thanks a lot for the PR!!

PR #1122, opened 2022-08-23
fix: serializing strings starting with newline/tab

adamdbradley:
If a string starts with `\n`, the serializer thought it was an `Error` object.


manucorporat:
this is not a good fix

PR #1133, opened 2022-08-23
fix: improve dev ssr performance

adamdbradley:
Inject Qwik City and Qwik dev middleware before Vite internal middleware to avoid unnecessary slow response times. Manually check request urls for known vite urls as a fast path.


packages/qwik-city/buildtime/vite/dev-server.ts
```diff
@@ -94,7 +99,20 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
         }
       }
 
-      // static file does not exist, 404
```
youngboy, line 97: current 404 path throw error. guest here's next goes to `qwik/optimizer/vite`'s plugin render. then dev404Middleware never trigger. IMHO that 404 handler put into previous code block is ok

packages/qwik-city/buildtime/vite/dev-server.ts
```diff
@@ -28,6 +28,11 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
       const url = new URL(req.originalUrl!, `http://${req.headers.host}`);
       const pathname = url.pathname;
 
+      if (isViteInternalRequest(pathname)) {
```
youngboy, line 31: I think safely check if `req.headers.accept.includes('text/html')` then continue otherwise go next should be fine?

packages/qwik-city/buildtime/vite/dev-server.ts
youngboy: oh, 404 might need to be after staticDistMiddleware. (if not checking dist ouput filepath prefix)

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -420,137 +420,128 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
...
             return;
           }
 
-          try {
-            if (req.headers.accept && req.headers.accept.includes('text/html')) {
-              const envData: Record<string, any> = {
-                ...(res as QwikViteDevResponse)._qwikEnvData,
-                url: url.href,
-              };
+          if (req.headers.accept && req.headers.accept.includes('text/html')) {
```
youngboy, line 434: can we skip this if city plugin enable ?

packages/qwik-city/buildtime/vite/dev-server.ts
youngboy: nvm. I was wrong. city can handling other type requests too.

packages/qwik/src/optimizer/src/plugins/vite.ts
youngboy: nvm, the answer is not 😅 page type endpoint rendering is delegated to qwik

PR #1148, opened 2022-08-26
TUTORIAL: make useClientEffect use track

jmmon:
# What is it?

- [ x ] Docs / tests

# Description

The instructions say to use `track` to make the `useClientEffect` update. The previous solution did not do this, it simply set up a `setInterval` on `useClientEffect`. The new solution `tracks` the store, and sets up a `setTimeout` to update the store. This results in the `useClientEffect` running when the store changes, and the store changes every second because of `setTimeout`. (Inefficient, but uses `track` like the instructions are trying to demonstrate.)

# Use cases and why

Expected the solution to use `track` to cause the `useClientEffect` to update the clock, because that's what the instructions said to do.
Actual: The solution did not use `track` but instead simply set up a `setInterval` once on useClientEffect; so it did not follow the instructions.

- 1. Demonstrate how `track` can be used with `useClientEffect`

# Checklist:

- [ x ] My code follows the [developer guidelines of this project](../CONTRIBUTING.md)
- [ n/a ] I have made corresponding changes to the documentation
- [ n/a ] Added new tests to cover the fix / functionality


jmmon:
I see the failed check because I didn't use prettier - I simply edited in the tutorials playground on the website. It should still be readable, and is styled the same as the previous solution.

mhevery (approved):

mhevery:
LGTM, can you please fix the lint error.

PR #1151, opened 2022-08-26
TUTORIAL: useResource now has a task

jmmon:
# What is it?

- [ x ] Docs / tests

# Description

I expected there to be a task for this tutorial but the given code was the same as the given solution, leaving nothing for users to do on this task. Now, users have two tasks:

1. create the `useResource$()` function (there is an exact answer already provided in the text on the left to make it easy)
2. create the <Resource> component (I left the `onResolved` callback function in a comment, so all they need to do is lay out the component and props, like the example in the text on the left)

# Use cases and why

- 1. Provide a task for users for this tutorial
- 2. Provide an opportunity for people to use the described function/component

# Checklist:

- [ x ] My code follows the [developer guidelines of this project](../CONTRIBUTING.md)
- [ n/a ] I have made corresponding changes to the documentation
- [ n/a ] Added new tests to cover the fix / functionality


mhevery (approved):

mhevery:
LGTM, please fix lint errors.

PR #1162, opened 2022-08-28
Update index.mdx

tidiview:
1- for some reason `useEndpoint` import line does not appear in Firefox.
2- Also, 2 import lines are redundant.
For that, I change it to one single import line, hoping to solve both problems at one.

# What is it?

- [ x] Docs
- Below capture of missing line of useEndpoint import
capture-useEndpoint

### DIFF
--- a/packages/docs/src/routes/qwikcity/routing/data/index.mdx
+++ b/packages/docs/src/routes/qwikcity/routing/data/index.mdx
@@ -50,8 +50,7 @@ An endpoint `onGet` function retrieves data and makes it available to the compon
 
 ```typescript
 import { Resource, component$, useStore } from '@builder.io/qwik';
-import type { RequestHandler } from '@builder.io/qwik-city';
-import { useEndpoint } from "@builder.io/qwik-city";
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/qwikcity/routing/data/index.mdx
```diff
@@ -50,8 +50,7 @@ An endpoint `onGet` function retrieves data and makes it available to the compon
 
 ```typescript
 import { Resource, component$, useStore } from '@builder.io/qwik';
-import type { RequestHandler } from '@builder.io/qwik-city';
-import { useEndpoint } from "@builder.io/qwik-city";
+import type { RequestHandler, useEndpoint } from '@builder.io/qwik-city';
```
manucorporat, line 53: This is a type import, byt useEndpoint() is a value. Unfortunately this is not correct and it will not compile.

manucorporat (changes requested):

manucorporat:
Unfortunately this change is not good, as it's today is correct

PR #1188, opened 2022-09-01
[docs] Remove useless $ sign

leader22:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

In Qwik JSX, `{state.count}` is OK, instead of `${state.count}`?


adamdbradley:
thanks

PR #1197, opened 2022-09-02
fix(pathname.ts): ssg dynamic routes

### LINKED ISSUE #1196: SSG: dynamic routes with optional parameter & nested pages not working
### Qwik Version

v0.0.107

Windows/Ubuntu

16

Qwik City

### Expected Behaviour

When I run ssg with the following routes folder:

```
routes
    [...id]
        page
            index.tsx
        index.tsx
```

I expect in dist folder (using _myId_ and an empty value for `id`:

```
myId
    page
        index.html
    index.html

page
    index.html
index.html
```

### Actual Behaviour

```
myId
    page
        index.html
    index.html

index.html
```

When the `id `parameter is empty, nested page is removed.

Reproduction (stop the dev server and run `npm run build`): https://stackblitz.com/edit/vitejs-vite-5bdbbf

### Additional Information

After investigating, I found that the issue comes from here: https://github.com/BuilderIO/qwik/blob/main/packages/qwik-city/utils/pathname.ts#L59

Using replace in this way, when the parameter is missing, results in a double `//` that is no longer normalized. 

I will propose a pr that solves the problem in `normalizePathname` function.

robisim74:
fix #1196

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

The issue comes from here: https://github.com/BuilderIO/qwik/blob/main/packages/qwik-city/utils/pathname.ts#L59

Using replace in this way, when the parameter is missing, results in a double `//` that is no longer normalized. 

I preferred to fix the `normalizePathname` function rather than the `getPathnameForDynamicRoute`, so as to avoid other similar situations.

# Use cases and why

When I run ssg with the following routes folder:

```
routes
    [...id]
        page
            index.tsx
        index.tsx
```

I expect in dist folder (using _myId_ and an empty value for `id`:

```
myId
    page
        index.html
    index.html

page
    index.html
index.html
```

Instead, when the `id `parameter is empty, nested page is removed.

# Checklist:

### DIFF
--- a/packages/qwik-city/utils/pathname.ts
+++ b/packages/qwik-city/utils/pathname.ts
@@ -10,6 +10,9 @@ export function normalizePathname(
 
     if (pathname !== '') {
       try {
+        // remove duplicate forward slashes
+        pathname = pathname.replace(/\/+/g, '/');
+
         if (pathname.startsWith('/')) {
           pathname = pathname.slice(1);
         }
--- a/packages/qwik-city/utils/pathname.unit.ts
+++ b/packages/qwik-city/utils/pathname.unit.ts
@@ -141,6 +141,26 @@ test('dynamic rest pathname', () => {
   equal(p, '/blog/what-is-resumability');
 });
 
+test('dynamic, empty rest pathname in root', () => {
+  const p = getPathname({
+    originalPathname: '/[...id]',
+    params: {
+      id: '',
+    },
+  });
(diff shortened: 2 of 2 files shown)


adamdbradley:
This is great, and thanks for adding tests!

robisim74:
(Thanks you for your work.... you are solving a lot of problems of web development)

PR #1212, opened 2022-09-06
feat: support custom events

manucorporat:


juanpmarin:
@manucorporat how are these types of events registered?  I'm using a component library made with lit I can't get the click event to work for example

PR #1216, opened 2022-09-08
docs: add mutable to the Greeter component

stewones:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

add mutable to the Greeter component. otherwise it's going to throw an exception 

```
Code(19): Props are immutable by default. If you need to change a value of a passed in prop, please wrap the prop with "mutable()" <:virtual name={mutable(...)}> 
 - Component: :virtual 
 - Prop: name 
 - Old value: World_2 
 - New value: World_1 Error: Code(19): Props are immutable by default.
```

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/components/rendering/index.mdx
+++ b/packages/docs/src/routes/docs/components/rendering/index.mdx
@@ -49,7 +49,7 @@ const Parent = component$(() => {
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


manucorporat:
mutable() is not longer needed! original example should work :)

Thanks for the PR! really made us think ways to remove mutable()

PR #1222, opened 2022-09-11
feat(eslint): add new no-prevent-default eslint rule

### LINKED ISSUE #1205: Feature Request: Warnings for `preventDefault` calls
In order to run the `preventDefault` of an event we need to add a `preventdefault` attribute (as described in the [docs](https://qwik.builder.io/docs/components/events/#prevent-default)) to the element, but possible `preventDefault` calls inside the callback can still be present even if they do nothing.

The different way to use `preventDefault` is something that developers could forget or not even be aware of (if they missed it in the docs) and currently they wouldn't get a clear indication of what is wrong with their code, so I was wondering if it would be possible to detect the use of `preventDefault` calls inside callbacks and warn the user that those will not have any effect (and pointing them to the docs).

For example, If I were to create a component like this:
```ts
const MyBtn = component$(() => {
  return (
    <button
      onClick$={(event) => {
        event.preventDefault();
        console.log('clicked');
      }}
    >
      click me
    </button>
  );
});
```

I would like to get somewhere a warning that the `event.preventDefault()` won't have any effect and that `preventdefault:click` should be used instead.

___

Some things I have in mind:
- It would be great to get the warning on the client side but I think that any sort of check would happen on the server, so the warning could potentially just be in the terminal, which could anyways still be helpful
- I get the feeling that the callbacks are generic so I am not sure if they have enough context attached to them to tell that they are callbacks, so that may be a challenge
- naturally the argument doesn't need to be called `event` it could be called `evt` instead, or destructured, etc..., so a possible solution would need to be quite robust (also based on the previous point, I could also have a function which takes an argument and calls it's `preventDefault`, being this some custom logic or even an event, so the sort of check would need to be applied specifically only on callbacks, as it would be undesirable/incorrect in other scenarios)

___

Anyways this type of warnings is not fundamental or anything, I just feel like it would help with DX and potentially avoid confusing situations, but if implementing something like this is too expensive/complex it may not be a worthwhile addition.

dario-piotrowicz:
add a new rule to the eslint-plugin-qwik package which warns developers not to use preventDefault, since they should be using the preventdefault attribute instead

resolves #1205

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Addition of a new eslint rule to help developers not to use `preventDefault`

The rule applies to any function expression, I figured this is the only reliable way to have the rule applied for all cases since there are cases in which I don't think we can even discern what the context is (for example if a function is in a standalone file and it is imported and used by a component).

I think this should be rather ok since I'd imagine that cases of a function not being a callback and just happening to have its first argument being an object with a function called `preventDefault` seems quite unlikely (and it would create a false positive which may just be slightly annoying).

If the above is not desirable I've added an option called `inlineOnly` which implements the check only if the function is inlined as the jsx attribute's value (which seems to me like a good enough compromise in case the normal version is too generic). 

I hope this is ok, please let me know what you think.

Also I was considering adding fixes/suggestions, but they should both remove the `preventDefault` call and add the attribute to the jsx, which based on what I said above could be only done in the inline case (since in a generic case you may not know who is going to use your function) also I'd have to do some string manipulating for the attribute making this slightly brittle, so I figured it could be better to keep things simple and just present the error/warning.

I added the rule to the plugin README list, without much details, I didn't add them since I can for now see if this change gets at least accepted and also I can see that the other rules don't have much documentation, so I wasn't sure if you'd like me to anyways (and ask here in case you'd like me to add some documentation)

# Result

Screenshot at 2022-09-11 21-36-02

# Checklist:

### DIFF
--- a/packages/eslint-plugin-qwik/src/noPreventDefault.ts
+++ b/packages/eslint-plugin-qwik/src/noPreventDefault.ts
@@ -1,0 +1,155 @@
+import { AST_NODE_TYPES, ESLintUtils } from '@typescript-eslint/utils';
+import type {
+  ArrowFunctionExpression,
+  BlockStatement,
+  Expression,
+  FunctionExpression,
+  ObjectPattern,
+} from '@typescript-eslint/types/dist/generated/ast-spec';
+
+const createRule = ESLintUtils.RuleCreator(
+  () => 'https://github.com/BuilderIO/qwik/tree/main/packages/eslint-plugin-qwik'
+);
+
+export const noPreventDefault = createRule({
+  name: 'no-prevent-default',
+  defaultOptions: [
+    {
+      inlineOnly: false,
+    },
+  ],
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'no preventDefault calls',
+      recommended: false,
+    },
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          inlineOnly: {
+            type: 'boolean',
+          },
+        },
+        default: {
+          inlineOnly: false,
+        },
+      },
+    ],
+    messages: {
+      errorMessage:
+        "Traditional `preventDefault()` calls don't work in Qwik, use the `preventdefault:{eventName}` attribute instead.",
+    },
+  },
+
+  create(context) {
+    const inlineOnly = context.options[0]?.inlineOnly ?? false;
+
+    function handleFunctionExpression(
+      expressionNode: ArrowFunctionExpression | FunctionExpression
+    ) {
+      if (inlineOnly) {
+        const isInlineCall =
+          expressionNode.parent?.type === AST_NODE_TYPES.JSXExpressionContainer &&
+          expressionNode.parent.parent?.type === AST_NODE_TYPES.JSXAttribute &&
+          (expressionNode.parent.parent.name.name as string)?.endsWith('$');
+
+        if (!isInlineCall) {
+          return;
+        }
+      }
+
+      const firstParam = expressionNode.params?.[0];
+      const eventParamName =
+        firstParam?.type === AST_NODE_TYPES.Identifier ? firstParam.name : null;
+      const preventDefaultParamName =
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin-qwik/src/noPreventDefault.ts
```diff
@@ -0,0 +1,122 @@
+import { AST_NODE_TYPES, ESLintUtils } from '@typescript-eslint/utils';
+import type { ArrowFunctionExpression, BlockStatement, Expression, FunctionExpression, ObjectPattern } from '@typescript-eslint/types/dist/generated/ast-spec';
+
+const createRule = ESLintUtils.RuleCreator(() => 'https://github.com/BuilderIO/qwik/tree/main/packages/eslint-plugin-qwik');
```
dario-piotrowicz, line 4: The url here points to the eslint plugin README, I could change this to the [Prevent Default Section](https://qwik.builder.io/docs/components/events/#prevent-default) of the docs instead, I think it could make more sense... although I would imagine the best would actually be to have more info in the plugin README(s) and eventually redirect from there to the docs.

manucorporat:
I think we dont need this rule anymore since we are relying on custom event types, closing this PR! but thanks so much for your amazing work!

Feel free to reopen if you think it still applies, and we should discuss further

dario-piotrowicz:
Hi @manucorporat I understand :slightly_smiling_face::+1: 

From my point of view ESlint rules provide a better DX even if there are clear runtime errors which developers can see (they could for example not be looking at the console and miss the errors), the rules also don't effect at all the output js so it's not like having extra rules slows things down or anything like that (so I think that there isn't any real downside in having them).

So from my point of view I would say that the rule could still be valuable (better DX with no extra costs), but anyways I understand if you want to keep the ESlint plugin as small as possible and only include what's strictly necessary.

If you'll change your mind the PR's still here :smiley:

PR #1232, opened 2022-09-14
Fix/store obj

samijaber:
# What is it?

- [x] Bug

# Description

Honestly not sure what I'm looking at, but looks like @manucorporat 's changes in https://github.com/BuilderIO/qwik/pull/1217/files#diff-c80d393c4d0acda1ba573e20ff62b3bc96a4169298f7d1af7898ccd8fb60638a stopped checking that the mutated `obj` is still a valid `object` before calling `getProxyTarget`. So now we sometime provide `null` to `getProxyTarget`, causing crashes.

### DIFF
--- a/packages/qwik/src/core/object/store.ts
+++ b/packages/qwik/src/core/object/store.ts
@@ -354,7 +354,7 @@ export const _pauseFromContexts = async (
           suffix += '_';
         }
       }
-      const target = getProxyTarget(obj);
+      const target = isObject(obj) && getProxyTarget(obj);
       if (target) {
         suffix += '!';
         obj = target;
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/object/store.ts
```diff
@@ -340,20 +340,21 @@ export const _pauseFromContexts = async (
...
+      const { value, resolved } = getPromiseValue(obj);
+      obj = value;
+      if (resolved) {
+        suffix += '~';
+      } else {
+        suffix += '_';
       }
+    }
+
+    if (isObject(obj)) {
```
samijaber, line 357: we need to check the `ojb` _after_ `obj = obj.mut` to make sure `the value given to `getProxyTarget` is actually a valid object, and non-null

PR #1257, opened 2022-09-15
fix: dev server redirect to trailing slash

adamdbradley:


adamdbradley:
https://github.com/BuilderIO/qwik/issues/1241

PR #1329, opened 2022-09-20
don't understand

tidiview:
- [x] Docs / tests


tidiview:
I don't understand this sentence and its relation to previous sorry

manucorporat:
misspelled... i want to say that this is not a problem with Qwik, Qwik app regardless of SPA / MPA only ship with 1KB of JS. MPA and SPA is a matter of using <a href> or <Link href>

PR #1331, opened 2022-09-20
suggestion

tidiview:
- [x] Docs / tests


tidiview:
suggestion

manucorporat:
We are preparing a while showcase page, we should link to that when it's ready!

PR #1363, opened 2022-09-21
feat: useSignal()

manucorporat:
<a href="https://www.loom.com/share/49ff1a08cf63474e839924188179711a">
    
  </a>

<a href="https://www.loom.com/share/2675c75cac1743fd8fb07f7feb6456ef">
    
  </a>


shairez:
Wow this might matche Qwik "after load" performance with solid's performance right? 

Amazing! 👏

manucorporat:
it will improve it a lot in some cases, but solid goes a long way extra, but generating custom code, heavy transforming the JSX

NachoVazquez:
I'm very excited about this @manucorporat . It seems a step in the right direction. If these experiments go right, would Qwik consider ditching the Virtual Dom before V1 is released, or do you think the beta announcement is a strong enough commitment to force you to preserve the existing API?

 Do you believe that other technical barriers would make it too hard to make such a change?
 
 Thanks in advance for your thoughts.

manucorporat:
Which API you thinking that could be removed? `useStore()`? VDOM is an implementation detail, from my point of view it could replaced anytime without warning :)

NachoVazquez:
You are right; in that case, yes, `useStore()` would be the API that would need to change or be deleted. 

Although it is convenient, I would prefer not to have two different mechanisms of UI-State reconciliation. Not only because every KB counts but also because the overhead and temptation of using the two models simultaneously could create errors that are hard to reason about.

NachoVazquez:
By changing the `useStore,` I mean to keep the API as it is, but instead of triggering a rerender, the behavior would be to have a fine-grained reactive object like what you would get from Solid's `createStore.`

That's a breaking change because people may expect that component to rerender and have some logic.

That's why I asked if you consider it possible to drop that behavior entirely for version 1.

NachoVazquez:
Awesome @manucorporat !!!!

mfp22:
Excited to try this out. What release will this be part of?

PR #1375, opened 2022-09-21
fix(jsx): tabindex should be lowercase

ulic75:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

A type error.  And in at least chrome and Firefox, using `tabIndex` works until the component is re-rendered, then it's removed.  Using `tabindex` works as expected.


ulic75:
Obviously that broke other places that are using `tabIndex`.  Not sure if the right solution is to update all those places too or if the issue is really something else.

PR #1405, opened 2022-09-22
feat: include CSS Property types for inline styles

KenAKAFrosty:
# What is it?

- [X] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Currently when applying in-line styles, there is no type suggestion or auto complete, nor in-source documentation on those properties.  This PR offers that feature. 

# Use cases and why

Having that assistance for a developer using inline styling can be very helpful and is a nice boost to DX in those cases.  It can save an entire google search to remind the exact property name, or the valid string literal values for those properties that expect them. 

image
image

NOTE: To support fallbacks for irregular/custom CSS properties (per @manucorporat in the [#qwik-contribution Discord channel](https://discord.com/channels/842438759945601056/996551595359740145)), there is a tradeoff.  This is a "looser" version that allows for any string index as a catch-all, including things like css variables  `style={{'--thing':  any}}`  without errors. But at the cost that incorrect properties will not be flagged:

Notice the improperly lowercase 'c' in color that does not show an error:
image

We could get slightly more specific and allow for custom css variables with a template literal type, while removing the catch-all which would result in:
image
If we think that's a valuable middle-ground I can add that to this PR, just let me know!

Thanks 🙂

# Checklist:


manucorporat:
you might need to run:
npm run api.update
npm run lint

PR #1406, opened 2022-09-22
fix(jsx): tabindex should be lowercase

### LINKED ISSUE #1379: tabindex issue due to incorrect typeing?
### Qwik Version

0.9.0

Chrome

_No response_

Qwik Runtime

### Expected Behaviour

I expect to be able to set a tabindex on an element.

### Actual Behaviour

When component is re-rendered the tabindex is removed from the dom.

### Additional Information

I believe this is due to an incorrect typing, but I'm not sure if it's another issue.  In the jsx-generated.ts file it's typed as `tabIndex` in HTMLAttributes but `tabindex` in SVGAttributes.  Notice the capitalization of the `I`.  When `tabIndex` is used it's find on the first render, but then fails after an update.  However, if I change it to `tabindex` ignoring type errors it works as expected.

I changed the type in #1375, but found then it breaks places that type is uses, which I suspect are themselves broken but I haven't had the time to test yet.

ulic75:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

fixes #1379

### DIFF
--- a/packages/docs/src/components/docsearch/doc-search-modal.tsx
+++ b/packages/docs/src/components/docsearch/doc-search-modal.tsx
@@ -236,7 +236,7 @@ export const DocSearchModal = component$(
           .filter(Boolean)
           .join(' ')}
         role="button"
-        tabIndex={0}
+        tabindex={0}
         onMouseDown$={(event) => {
           if (event.target === containerRef.current) {
             onClose$.apply(undefined, []);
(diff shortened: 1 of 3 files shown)


manucorporat:
We fixed this issue, but correctly handling tabIndex! thanks for the PR though :)

PR #1474, opened 2022-09-26
docs: dark mode

Varixo:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Dark mode for the docs. Added many new CSS variables for future color changes.

https://www.loom.com/share/715ce5d7f94745dd91a314f428b2f91e

TODO: Startpage - I can't change it, because of importing `Resource` from Builder 
https://github.com/BuilderIO/qwik/blob/a8ed4e406bc48a653d46053e2cd0cd209d993afa/packages/docs/src/routes/index.tsx#L25

# Use cases and why

- it no longer hurts the eyes 😄 

# Checklist:


packages/docs/src/components/code-block/code-block.tsx
```diff
@@ -25,9 +24,11 @@ export const CodeBlock = component$((props: CodeBlockProps) => {
         : undefined;
   }
 
+  const theme = window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
```
manucorporat, line 27: This will not work in SSR... wondering if we remove all this JS, and do the trick with CSS!

manucorporat (commented):
this is truly amazing work!! let's get it finish and merged

manucorporat:
Might need to `npm run fmt` at the root of the repo!

packages/docs/src/components/code-block/code-block.tsx
Varixo: Right, I'll create two separate css files - one for light mode and one for dark mode, then import them conditionally:
```css
@import "./code-block.light.css" screen;
@import "./code-block.dark.css" screen and (prefers-color-scheme: dark);
```
This should work

Varixo:
What about the start page? Looks bad in dark mode right now @manucorporat ?

Varixo:
Maybe there should be a toggle for switching between light and dark mode?

wtho:
If you want a toggle, JS will be required, and to avoid a flash in the wrong color, a script has to be added to `<head>`, e. g.:
```tsx
  const darkThemeScript = `document.documentElement.classList.add(\`theme-\${localStorage.getItem('theme') ?? (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : undefined) ?? 'light'}\`)`
  // ...
  <script dangerouslySetInnerHTML={themeScript} />
```
(here using localstorage to read the persisted value if the toggle was used).

On top, this would have to be synced with the monaco repl theme as well.

Varixo:
Yes, I was thinking about something like: https://web.dev/building-a-theme-switch-component/. I will try to add it later.

wtho:
Maybe in a follow-up PR to get this work into qwik first

Varixo:
So only the start page is left, but I think I can't do that:
https://github.com/BuilderIO/qwik/blob/a8ed4e406bc48a653d46053e2cd0cd209d993afa/packages/docs/src/routes/index.tsx#L25

packages/docs/src/components/theme-toggle/sun-and-moon.css
```diff
@@ -0,0 +1,82 @@
+.sun-and-moon {
+  --ease-elastic-3: cubic-bezier(0.5, 1.25, 0.75, 1.25);
+  --ease-elastic-4: cubic-bezier(0.5, 1.5, 0.75, 1.25);
+  --ease-out-5: cubic-bezier(0, 0, 0, 1);
+  --ease-3: cubic-bezier(0.25, 0, 0.3, 1);
+}
+
+.sun-and-moon > :is(.moon, .sun, .sun-beams) {
```
Varixo, line 8: it is ok to use the `:is`?

packages/docs/src/components/theme-toggle/README.md
```diff
@@ -0,0 +1,3 @@
+# Theme switch
+
+Based on: [GUI challenges - Theme Toggle](https://github.com/argyleink/gui-challenges/tree/main/theme-switch)
```
Varixo, line 3: I hope we can use this? (I rewritten this to qwik, but logic, animation, css are almost the same)

packages/docs/src/components/theme-toggle/theme-toggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+  useClientEffect$(() => {
+    state.theme = getColorPreference();
+  });
+
+  useClientEffect$(() => {
+    return colorSchemeChangeListener((isDark) => {
+      state.theme = isDark ? 'dark' : 'light';
+      setPreference(state.theme);
+    });
+  });
```
Varixo, line 51: there is one case when this not work, because of component is not rendered yet. After openning the page on a mobile, changing preference to light/dark mode will not have no effect. It is ok? I think to fix this I can move this code to another "hidden" component always rendered for both - desktop and mobile (or there is any better solution?)

packages/docs/src/components/theme-toggle/README.md
Varixo: video before a few fixes, but still shows the toggle
https://www.loom.com/share/8d88f7f3b6fb4c9a9e95e9e8fc85f31c

jrson83:
@Varixo Thank you for the PR, it was a great example help to implement dark mode. If the PR is merged, it would be awesome if you could write a small Qwik doc on how to implement dark mode in Qwik?

Varixo:
Sure, I will try :)

manucorporat:
lets get this merged soon! is anything stopping this?

Varixo:
Unfortunatelly yes, there is no important classes in builder resource (start page). If div has a `dark-mode` class, then will work, but some divs dont have that class. Adam was supposed to add this, but I suspect there is no time for that. (It was done partially or wrong). Sometimes div only has `dark-`, sometimes it doesn't, sometimes the class applies to the parent. 
So, only divs with these classes should have an additional `dark-mode` class:
1. c8g8p4n (now it is ok)
obraz

2. cc2mqar (now it is ok)
obraz

3. c8wpiyw (missing here)
obraz

4. cvid0z5 (missing here, instead it was applied incorrectly to the parent and looks like below)
obraz

And here are unnecessary:
1. cvk60o0 (unnecessary `dark-` class)
2. cdxhzhc (unnecessary `dark-` class)
3. c4e7x5r (as I wrote above -  it was applied incorrectly here instead of child)

Varixo:
And one more - about licence - I am not expert. We can use this? 
https://github.com/argyleink/gui-challenges/tree/main/theme-switch
All rewritten to qwik, but animations, svgs are the same

manucorporat:
ok! i will add the classes,

Varixo:
Maybe we can manually search for this node and apply the inverted style? The disadvantage of this solution is a incorrectly applied style if the structure changes, but maybe it does not bother us for now?

Varixo:
So like this:
```css
:root[data-theme='dark'] .csoetdv {
  filter: invert(90%) hue-rotate(180deg);
}

:root[data-theme='dark'] .cvha93q {
  filter: invert(90%) hue-rotate(180deg);
}
```

It works well with these styles

manucorporat:
Thank you so much for this amazing piece of work! and sorry so much for the delay!

PR #1527, opened 2022-09-30
fix: fulfill platform field for netlify edge

youngboy:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Add platform context for netlify edge middleware

@nnelgxorz 

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


nnelgxorz:
Nice @youngboy! 

Would it be easier to just set the platform field directly to the context so people can use platform.whatever instead of platform.context.whatever?

Also, we should export an Platform interface/type so people can cast the platform to it. I know we're still waiting on the Netlify team for that though.

adamdbradley:
Thanks!

PR #1543, opened 2022-10-01
Removed double counter/counter in the component import

socmov:
This doesn't make sense and isn't followed by the other components that are imported in the examples.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ x] Docs / tests

# Description
The directory structure only has one counter level. I could be wrong but it seems strange. If it is correct an explanation would be great. 

# Use cases and why

More accurate documentation

# Checklist:

- [ x] I have performed a self-review of my own code
- [ x] I have made corresponding changes to the documentation

### DIFF
--- a/packages/docs/src/routes/qwikcity/content/mdx/index.mdx
+++ b/packages/docs/src/routes/qwikcity/content/mdx/index.mdx
@@ -47,7 +47,7 @@ src/
 # File: src/routes/some/path/index.mdx
 title: Hello World Title
 ---
-import { Counter } from "../../../components/counter/counter";
+import { Counter } from "../../../components/counter";
 
 This is a simple hello world component.
 


ChibiBlasphem:
As you see in the `Counter` example, the path is correct
https://github.com/BuilderIO/qwik/blob/bc96788b12d6ba0e4f33f526e15dd1c7b2600340/packages/docs/src/routes/qwikcity/content/mdx/index.mdx?plain=1#L59

shairez:
Thanks @socmov !

As you see by @ChibiBlasphem comment (thanks as well btw) - this appears to be the correct behavior (as the counter has its own library, so there is a duplication there but on purpose) 

So closing for now (if you want to continue the discussion about it, you can re-open it again)

Issue #652, opened 2022-06-21
Eslint: detect bad usage of use-methods


manucorporat:
### Is your feature request related to a problem?

use- methods can only be called in specific situations, and right now only the runtime will error after execution. It should be easy to detect at linting time using eslint:

## Rules

- They can only be called at the root level of a function / arrow function context, not inside branches or other conditional blocks
- They can only be called in component$ 
- They can only be called from another use- method. (allowing composition)

### Describe alternatives you've considered

Improve errors in the runtime for this particular issue

PR #1545, opened 2022-10-01
docs: Hook rules

nnelgxorz:
… in the Resource docs.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Adds the hook rules documented in #652

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/components/lifecycle/index.mdx
+++ b/packages/docs/src/routes/docs/components/lifecycle/index.mdx
@@ -16,6 +16,21 @@ When using lifecycle hooks, you must adhere to the following rules:
   - They can only be called at the root level of a function / arrow function context, not inside of branches or conditional blocks
   - They can only be called from another `use*$` method, allowing for composition
 
+```tsx
+useHook() // <-- ❌ does not work
+export default component$(() => {
+  useHook() // <-- ✅ does work
+  if (condition) {
+    useHook() // <-- ❌ does not work
+  }
+  const myQrl = $(() => useHook()) // <-- ✅ does work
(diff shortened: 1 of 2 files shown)


ulic75:
I have a question about this.  Is it simply not recommended or is it really not supposed to work?  I do the below in a component and it DOES appear to work as I expect.

```js
if (handleChange$) {
  useWatch$(({ track }) => {
    const value = track(store, "open");
    handleChange$(value);
  });
}
```

If it's not a good idea I could of course just do something like:
```js
useWatch$(({ track }) => {
  const value = track(store, "open");
  handleChange$ && handleChange$(value);
});
```

wtho:
@nnelgxorz Here the updated example snippet after some super helpful clarifications by ChibiKookie:
```tsx
useHook() // <-- ❌ does not work
const useCustomHook = () => {
  useHook() // <-- ✅ does work
}
  useHook() // <-- ✅ does work
  if (condition) {
    useHook() // <-- ❌ problems on rerender
  }
  const myQrl = $(() => useHook()) // <-- ❌ does not work, wrapped in QRL
  return <button onClick$={() => useHook() }>B</button> // <-- ❌ does not work, QRL
})
```

wtho:
@ulic75 What you do there is totally fine. `handleChange$` is not a `use`-Hook, so you can call it anywhere.
Only thing is that it's a QRL and therefore a promise and you are not awaiting it and also not `.catch`/`.then`/`.finally`, so you do not handle at all what happens inside. But if the QRL can get resolved, it will get called, and that is totally valid Qwik code.

mhevery (approved):

PR #1552, opened 2022-10-02
fix(ssrdevmiddleware): skip request in Vite Dev Server if detecting a ping.

meenie:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

After loading your app locally and then stopping the Vite dev server, Vite starts up a ping to the WebSocket connection by running a fetch on it over and over again until you start back up the server.  When you do, that ping will trigger the `onGet` handler for the `/` route which can cause side effects and confusion for developers.  You can see the behavior here: https://github.com/vitejs/vite/blob/7a6d4bc0d7fa614d3ac469ca35352a23aaef8232/packages/vite/src/client/client.ts#L300-L322

This might be a naive approach because there is very little to use to detect this particular issue. I'm pretty sure that no one will want to use this combination of headers anyway. If you think it's too brittle, I'm fine just living with the issue.

# Use cases and why

Use [this Stackblitz](https://stackblitz.com/edit/qwik-starter-ccb5nb?file=src%2Froutes%2Findex.tsx) to repro the issue by loading up the route `/a/b/c` in the built-in browser, stop the server and then restart it. You will notice that the `console.log()` within the `onGet` of the `/` route gets called.

# Checklist:

### DIFF
--- a/packages/qwik-city/buildtime/vite/dev-server.ts
+++ b/packages/qwik-city/buildtime/vite/dev-server.ts
@@ -55,7 +55,7 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
     try {
       const url = new URL(req.originalUrl!, `http://${req.headers.host}`);
 
-      if (skipRequest(url.pathname)) {
+      if (skipRequest(url.pathname) || isVitePing(url.pathname, req.headers)) {
         next();
         return;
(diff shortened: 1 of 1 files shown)


adamdbradley:
Think this is a great fix, thanks!

PR #1572, opened 2022-10-03
Return the StyleID for delegation

Exahilosys:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Simply reveals the `StyleID` when using either `useStylesScoped$` or `useStyles$`. 

There is no reactivity for styles at the moment aside from adding or removing classes. Being able to render an additional `<style scoped>` before each element would help override some styles (or css variables) defined with `useStylesScoped$` depending on props or state, making styles reactive.

As of now, while rendering an additional `<style scoped>` is possible (although it weirdly gets its own `class=...`), the styles in there do not override the ones from `useStylesScoped$` as the latter are prefixed with a unique `StyleID`.

Returning this `StyleID` allows the user to inject it in the content of the additional aforementioned `<style scoped>`, allowing its styles to be applied to the element and properly override the default ones.

# Use cases and why

This future enables delegation of the `styleID` to per-element styles for overriding the per-component styles.

```jsx
const Room = component$(function Room ({very}) {

    // Cannot override here
    // useStylesScoped$ does not accept local identifiers 
    const styleID = useStylesScoped$(
        `.chilly {
            background-color: lightblue;
        }`
    )

    return (
        <>
            <style scoped> 
                {
                    // Can override here
                    // "very" is a local identifier
                    `.chilly.⭐️${styleID} {
                        ${very ? 'background-color: lightcyan;'  : ''}
                    }`
                }
            </style>
            <div class='chilly'>
                BRRR!
            </div>
        </>
    )
})
```

This example demonstrates swapping a discreet style, which *could* be done by simply adding or removing classes. But the same cannot be said for discreet styles such as numbers, percentages, etc... 

Take trying to set `width` or a color's alpha to a number passed through props for instance.

# Checklist:


packages/qwik/src/core/use/use-styles.ts
```diff
@@ -34,7 +34,7 @@ import { assertDefined } from '../assert/assert';
  */
 // </docs>
 export const useStylesQrl = (styles: QRL<string>): void => {
```
ChibiBlasphem, line 36:
```suggestion
export const useStylesQrl = (styles: QRL<string>): string => {
```

packages/qwik/src/core/use/use-styles.ts
```diff
@@ -88,7 +88,7 @@ export const useStyles$ = /*#__PURE__*/ implicit$FirstArg(useStylesQrl);
  */
 // </docs>
 export const useStylesScopedQrl = (styles: QRL<string>): void => {
```
ChibiBlasphem, line 90:
```suggestion
export const useStylesScopedQrl = (styles: QRL<string>): string => {
```

packages/qwik/src/core/use/use-styles.ts
ChibiBlasphem, line 36:
```suggestion
export const useStylesQrl = (styles: QRL<string>): string => {
```

ChibiBlasphem (commented):
It seems to like proper return type now that `_useStyles` result is returned.

manucorporat:
not sure i like the idea of returning a scopeID as string, i think we should return an object, like `{scopeId}` so we can extend the API in the future in case we need to return more things without breaking changes

PR #1577, opened 2022-10-03
docs: fix links

adamdbradley:


packages/docs/src/routes/qwikcity/menu.md
```diff
@@ -58,7 +57,7 @@
 - [useDocumentHead()](api/use-document-head/index.mdx)
 - [useLocation()](api/use-location/index.mdx)
 
-## Middleware
+## Adaptors
```
meenie, line 60: If you are changing this here, you probably want to update `middleware/overview/index.mdx` as well since the title is still using `Middleware`. Same thing with the pages under the `middleware` route.  I'm not sure how you would redirect all of the `/middleware/**` pages to `/adapters/**`, though. Possibly in the server side?

PR #1594, opened 2022-10-05
feat: add support for working with locale

mhevery:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/request-handler/page-handler.ts
+++ b/packages/qwik-city/middleware/request-handler/page-handler.ts
@@ -6,12 +6,17 @@ import type {
   RenderResult,
   RenderToStringResult,
 } from '@builder.io/qwik/server';
-import type { ClientPageData, QwikCityEnvData } from '../../runtime/src/library/types';
+import type {
+  ClientPageData,
+  QwikCityEnvData,
+  QwikCityMode,
+} from '../../runtime/src/library/types';
 import { getErrorHtml } from './error-handler';
 import { HttpStatus } from './http-status-codes';
 import type { QwikCityRequestContext, UserResponseContext } from './types';
 
 export function pageHandler<T = any>(
+  mode: QwikCityMode,
   requestCtx: QwikCityRequestContext,
   userResponse: UserResponseContext,
   render: Render,
@@ -21,6 +26,8 @@ export function pageHandler<T = any>(
   const { status, headers } = userResponse;
   const { response } = requestCtx;
   const isPageData = userResponse.type === 'pagedata';
+  const requestHeaders: Record<string, string> = {};
+  requestCtx.request.headers.forEach((value, key) => (requestHeaders[key] = value));
 
   if (isPageData) {
     // page data should always be json
@@ -35,7 +42,7 @@ export function pageHandler<T = any>(
     try {
       const result = await render({
         stream: isPageData ? noopStream : stream,
-        envData: getQwikCityEnvData(userResponse),
+        envData: getQwikCityEnvData(requestHeaders, userResponse, requestCtx.locale, mode),
         ...opts,
       });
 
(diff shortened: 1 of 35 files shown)


manucorporat:
This PR introduces a new API, which implementation i am not sure it's entirely correct since use methods can only be used in  rendering context, what's this for?

packages/qwik/src/core/render/container.ts
```diff
@@ -79,6 +79,17 @@ export const createContainerState = (containerEl: Element) => {
   return containerState;
 };
 
+const envDataFromContainer = (containerEl: Element): Record<string, any> => {
```
manucorporat, line 82: envData is not container attributes, we are mixing semantics, not sure i like this, we are telling developers that envData is server only (for now), but it comes from `envData`, this breaks that semantic

packages/qwik/src/core/api.md
```diff
@@ -268,6 +268,9 @@ export const inlinedQrl: <T>(symbol: T, symbolName: string, lexicalScopeCapture?
 // @internal (undocumented)
 export const inlinedQrlDEV: <T = any>(symbol: T, symbolName: string, opts: QRLDev, lexicalScopeCapture?: any[]) => QRL<T>;
 
+// @alpha
+export const isInUseContext: () => boolean;
```
manucorporat, line 272: i would make it `@internal` is we really need it, the fact that `_context` is defined does not mean that all use methods will work. Only useLexicalScope() works in all contexts

manucorporat (changes requested):

packages/qwik-city/middleware/request-handler/page-handler.ts
```diff
@@ -130,14 +137,21 @@ function getPrefetchBundleNames(result: RenderResult, routeBundleNames: string[]
   return bundleNames;
 }
 
-export function getQwikCityEnvData(userResponse: UserResponseContext): {
+export function getQwikCityEnvData(
+  requestHeaders: Record<string, string>,
+  userResponse: UserResponseContext,
+  mode: QwikCityMode
+): {
```
wmertens, line 144: Would it not be better to switch this to an object argument to allow adding parameters in the future?

packages/qwik-city/middleware/request-handler/page-handler.ts
```diff
@@ -21,6 +26,8 @@ export function pageHandler<T = any>(
   const { status, headers } = userResponse;
   const { response } = requestCtx;
   const isPageData = userResponse.type === 'pagedata';
+  const requestHeaders: Record<string, string> = {};
+  requestCtx.request.headers.forEach((value, key) => (requestHeaders[key] = value));
```
wmertens, line 30: We have something similar in our React codebase, but we allow registering functions that extra environment data from a request object.

This way, more than headers can be used to set the environment data, it avoids having to copy all headers, and it allows running the render without a request object, like when you're rendering a PDF.

wmertens (commented):
- how can one set the first part of the URL to the locale?
- how can one change the locale in the browser without losing state, e.g. while filling in a form?

packages/qwik/src/core/render/container.ts
wmertens: in our React codebase, we have plugins that can extract server data during render (via `<Provider>` mosty), and then populate a global state object that gets serialized to the client, which gets read by the Providers on the client.
This has proven very useful, and I think this `envData` is similar to that.

DustinJSilk:
This is great, it also solves a [PR](https://github.com/BuilderIO/qwik/pull/1605) I sent to get access to request headers.

PR #1636, opened 2022-10-07
Fix: Let developers make their own error pages design

MasterMousa:
# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

Make simple changes to let developers make their own error pages and design as they want
**the developer should make designs for each error page**
if this PR merged and accepted, I'll update the rest of files and make more error pages to be ready for design

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/request-handler/error-handler.ts
+++ b/packages/qwik-city/middleware/request-handler/error-handler.ts
@@ -69,37 +69,8 @@ export function getErrorHtml(status: number, e: any) {
 }
 
 function minimalHtmlResponse(status: number, message?: string, stack?: string) {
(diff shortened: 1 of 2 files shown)


RafaelZasas:
I really like this implementation, although it does shoehorn users into naming their pages with `err-<errStatus>-page` and limited options for the url of error routes.

Would it be possible to allow the user to define the dir of their choice for errors as well as the component name? maybe to set  window.location.href to a `${errPageLocation}` or even having an optional JSX Element prop which would return the actual component itself instead of navigating to a pre defied route?

adamdbradley:
Thanks for opening the PR, but we had to fix this with a different approach to allow for a custom 404 to work and various directories. https://github.com/BuilderIO/qwik/pull/2175

PR #1645, opened 2022-10-08
Signal: add valueOf and toJSON for ease of use

wmertens:
# What is it?

- [x] Feature / enhancement

# Description

Adds valueOf, toString and toJSON so that a signal can be coerced to its value. This makes it easier to pass signals to components expecting simple values, in many cases it will just work.

# Use cases and why

This makes it easier to write components that accept both values and signals of values.

Given a signal `const count = useSignal(0)`, the following examples will work:
- `+count`: `0`
- `${count}`: `"0"`
- `console.log(count)`: outputs `0`
- `JSON.stringify(count)`: `0`

The following examples won't work:
- `count++`: use `count.value++` instead
- `typeof count`: `"object"`

For the rest it is still a signal and the subscribing etc works as before + it also tracks the examples above now.

# Checklist:


shairez:
Thanks @wmertens !

Can you run "prettier" locally to fix the issues?

Plus it seems that the E2E tests are failing

Thanks!

shairez:
@wmertens mentioned on discord: "It only works when the value is coerced by js, so if you use it with an operator. If you do typeof, you'll get "object"

I wonder - if there is a potential inconsistency here, could it potentially damage the DX in someway? (unexpected surprises / bugs etc)

wmertens:
Right, the failing e2e is due to the JSON changing. I think it makes sense that it serializes as the value and not the object, which means updating the tests. Checking.

wmertens:
Right, the failing e2e is due to the JSON changing. I think it makes sense that it serializes as the value and not the object, which means updating the tests. Checking.

manucorporat:
The problem is that now the test, is actually not testing it's getting a Signal... i am not sure, `signal !== signal.value`

manucorporat:
This is unfotunately a bad idea that will affect performance, right now the optimizer can optimize out all usages of:
```prop={signal.value}```
allowing to not rerender, encouraging an API that allows `{+count}` will be encouraging an API that leads to worse perf. I am all for a toJSON() or toString() that helps debugging, but not as a development pattern.

**Comment by wmertens** [comment]
> The problem is that now the test, is actually not testing it's getting a Signal... i am not sure, signal !== signal.value

The test can check if `signal typeof Signal`, which is a better test, no?

> This is unfotunately a bad idea that will affect performance, right now the optimizer can optimize out all usages of: `prop={signal.value}`

I'm really surprised now. When I tested rendering, a child that simply embedded the value would re-render when I passed `signal.value` but not when I passed `signal`. 

Example: Paste this into the playground
```jsx
import { component$, useSignal } from '@builder.io/qwik';

export const C = ({who, count}) => (console.log('render', who),<div>Count {who} is {count}</div>)

export const App = component$(() => {
  const signal=useSignal(0)
  // signal.valueOf = function() { return this.value; }
  // signal.toString = function() { return `${this.value}`; }
  // signal.toJSON = function() { return this.value; }

  return <div style={{display: 'flex', flexDirection:'column'}}>
    <button onClick$={()=>signal.value++}>Click</button>

    <C who={1} count={signal.value}/>
    <C who={2} count={signal}/>
  </div>
});
```

When you click, both C re-render. When you remove the first C, C2 doesn't re-render. When you remove the second C, C1 still re-renders. Note that this is without this PR applied.
If you uncomment the added functions, the rendering will still happen the same way.

So I thought that the only way to make the render avoidance work, is to pass the signal as a signal. This is also what I find most intuitive; if the optimizer would do some special thing with signal.value that would be a little bit too magic maybe.

manucorporat:
check this out:
https://qwik.builder.io/playground#version=0.10.0&buildMode=development&entryStrategy=hook&files=JTVCJTdCJTIycGF0aCUyMiUzQSUyMiUyRmFwcC50c3glMjIlMkMlMjJjb2RlJTIyJTNBJTIyJTVDbmltcG9ydCUyMCU3QiUyMGNvbXBvbmVudCUyNCUyQyUyMHVzZVNpZ25hbCUyMCU3RCUyMGZyb20lMjAnJTQwYnVpbGRlci5pbyUyRnF3aWsnJTNCJTVDbiU1Q25leHBvcnQlMjBjb25zdCUyMEMlMjAlM0QlMjAocHJvcHMpJTIwJTNEJTNFJTIwJTdCJTVDbiUyMCUyMHJldHVybiUyMCUzQ2RpdiUzRUNvdW50JTIwJTdCcHJvcHMud2hvJTdEJTIwaXMlMjAlN0Jwcm9wcy5jb3VudCU3RCUzQyUyRmRpdiUzRSU1Q24lN0QlNUNuJTVDbmV4cG9ydCUyMGNvbnN0JTIwQXBwJTIwJTNEJTIwY29tcG9uZW50JTI0KCgpJTIwJTNEJTNFJTIwJTdCJTVDbiUyMCUyMGNvbnN0JTIwc2lnbmFsJTNEdXNlU2lnbmFsKDApJTVDbiUyMCUyMCUyRiUyRiUyMHNpZ25hbC52YWx1ZU9mJTIwJTNEJTIwZnVuY3Rpb24oKSUyMCU3QiUyMHJldHVybiUyMHRoaXMudmFsdWUlM0IlMjAlN0QlNUNuJTIwJTIwJTJGJTJGJTIwc2lnbmFsLnRvU3RyaW5nJTIwJTNEJTIwZnVuY3Rpb24oKSUyMCU3QiUyMHJldHVybiUyMCU2MCUyNCU3QnRoaXMudmFsdWUlN0QlNjAlM0IlMjAlN0QlNUNuJTIwJTIwJTJGJTJGJTIwc2lnbmFsLnRvSlNPTiUyMCUzRCUyMGZ1bmN0aW9uKCklMjAlN0IlMjByZXR1cm4lMjB0aGlzLnZhbHVlJTNCJTIwJTdEJTVDbiU1Q24lMjAlMjByZXR1cm4lMjAlM0NkaXYlMjBzdHlsZSUzRCU3QiU3QmRpc3BsYXklM0ElMjAnZmxleCclMkMlMjBmbGV4RGlyZWN0aW9uJTNBJ2NvbHVtbiclN0QlN0QlM0UlNUNuJTIwJTIwJTIwJTIwJTNDYnV0dG9uJTIwb25DbGljayUyNCUzRCU3QigpJTNEJTNFc2lnbmFsLnZhbHVlJTJCJTJCJTdEJTNFQ2xpY2slM0MlMkZidXR0b24lM0UlNUNuJTVDbiUyMCUyMCUyMCUyMCUzQ0MlMjB3aG8lM0QlN0IxJTdEJTIwY291bnQlM0QlN0JzaWduYWwudmFsdWUlN0QlMkYlM0UlNUNuJTIwJTIwJTNDJTJGZGl2JTNFJTVDbiU3RCklM0IlMjIlN0QlMkMlN0IlMjJwYXRoJTIyJTNBJTIyJTJGZW50cnkuc2VydmVyLnRzeCUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjAlN0IlMjByZW5kZXJUb1N0cmluZyUyQyUyMFJlbmRlck9wdGlvbnMlMjAlN0QlMjBmcm9tJTIwJyU0MGJ1aWxkZXIuaW8lMkZxd2lrJTJGc2VydmVyJyUzQiU1Q25pbXBvcnQlMjAlN0IlMjBSb290JTIwJTdEJTIwZnJvbSUyMCcuJTJGcm9vdCclM0IlNUNuJTVDbmV4cG9ydCUyMGRlZmF1bHQlMjBmdW5jdGlvbiUyMChvcHRzJTNBJTIwUmVuZGVyT3B0aW9ucyklMjAlN0IlNUNuJTIwJTIwcmV0dXJuJTIwcmVuZGVyVG9TdHJpbmcoJTNDUm9vdCUyMCUyRiUzRSUyQyUyMG9wdHMpJTNCJTVDbiU3RCU1Q24lMjIlN0QlMkMlN0IlMjJwYXRoJTIyJTNBJTIyJTJGcm9vdC50c3glMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJTdCJTIwQXBwJTIwJTdEJTIwZnJvbSUyMCcuJTJGYXBwJyUzQiU1Q24lNUNuZXhwb3J0JTIwY29uc3QlMjBSb290JTIwJTNEJTIwKCklMjAlM0QlM0UlMjAlN0IlNUNuJTIwJTIwcmV0dXJuJTIwKCU1Q24lMjAlMjAlMjAlMjAlM0NodG1sJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQ2hlYWQlM0UlNUNuJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTNDdGl0bGUlM0VIZWxsbyUyMFF3aWslM0MlMkZ0aXRsZSUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0MlMkZoZWFkJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQ2JvZHklM0UlNUNuJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTNDQXBwJTIwJTJGJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQyUyRmJvZHklM0UlNUNuJTIwJTIwJTIwJTIwJTNDJTJGaHRtbCUzRSU1Q24lMjAlMjApJTNCJTVDbiU3RCUzQiU1Q24lMjIlN0QlNUQ%3D

wmertens:
@manucorporat That is not making a difference for me. When I put `console.log('render')` in component C, it still shows that it re-renders unless I pass `signal` instead of `signal.value`.

manucorporat:
Works in 0.11.0!
https://qwik.builder.io/playground#version=0.11.0&buildMode=development&entryStrategy=hook&files=JTVCJTdCJTIycGF0aCUyMiUzQSUyMiUyRmFwcC50c3glMjIlMkMlMjJjb2RlJTIyJTNBJTIyJTVDbmltcG9ydCUyMCU3QiUyMGNvbXBvbmVudCUyNCUyQyUyMHVzZVNpZ25hbCUyMCU3RCUyMGZyb20lMjAnJTQwYnVpbGRlci5pbyUyRnF3aWsnJTNCJTVDbiU1Q25leHBvcnQlMjBjb25zdCUyMEMlMjAlM0QlMjAocHJvcHMpJTIwJTNEJTNFJTIwJTdCJTVDbiUyMCUyMGNvbnNvbGUubG9nKCdyZW5kZXInKSUzQiU1Q24lMjAlMjByZXR1cm4lMjAlM0NkaXYlM0VDb3VudCUyMCU3QnByb3BzLndobyU3RCUyMGlzJTIwJTdCcHJvcHMuY291bnQlN0QlM0MlMkZkaXYlM0UlNUNuJTdEJTVDbiU1Q25leHBvcnQlMjBjb25zdCUyMEFwcCUyMCUzRCUyMGNvbXBvbmVudCUyNCgoKSUyMCUzRCUzRSUyMCU3QiU1Q24lMjAlMjBjb25zdCUyMHNpZ25hbCUzRHVzZVNpZ25hbCgwKSU1Q24lMjAlMjAlMkYlMkYlMjBzaWduYWwudmFsdWVPZiUyMCUzRCUyMGZ1bmN0aW9uKCklMjAlN0IlMjByZXR1cm4lMjB0aGlzLnZhbHVlJTNCJTIwJTdEJTVDbiUyMCUyMCUyRiUyRiUyMHNpZ25hbC50b1N0cmluZyUyMCUzRCUyMGZ1bmN0aW9uKCklMjAlN0IlMjByZXR1cm4lMjAlNjAlMjQlN0J0aGlzLnZhbHVlJTdEJTYwJTNCJTIwJTdEJTVDbiUyMCUyMCUyRiUyRiUyMHNpZ25hbC50b0pTT04lMjAlM0QlMjBmdW5jdGlvbigpJTIwJTdCJTIwcmV0dXJuJTIwdGhpcy52YWx1ZSUzQiUyMCU3RCU1Q24lNUNuJTIwJTIwcmV0dXJuJTIwJTNDZGl2JTIwc3R5bGUlM0QlN0IlN0JkaXNwbGF5JTNBJTIwJ2ZsZXgnJTJDJTIwZmxleERpcmVjdGlvbiUzQSdjb2x1bW4nJTdEJTdEJTNFJTVDbiUyMCUyMCUyMCUyMCUzQ2J1dHRvbiUyMG9uQ2xpY2slMjQlM0QlN0IoKSUzRCUzRXNpZ25hbC52YWx1ZSUyQiUyQiU3RCUzRUNsaWNrJTNDJTJGYnV0dG9uJTNFJTVDbiU1Q24lMjAlMjAlMjAlMjAlM0NDJTIwd2hvJTNEJTdCMSU3RCUyMGNvdW50JTNEJTdCc2lnbmFsLnZhbHVlJTdEJTJGJTNFJTVDbiUyMCUyMCUzQyUyRmRpdiUzRSU1Q24lN0QpJTNCJTIyJTdEJTJDJTdCJTIycGF0aCUyMiUzQSUyMiUyRmVudHJ5LnNlcnZlci50c3glMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJTdCJTIwcmVuZGVyVG9TdHJpbmclMkMlMjBSZW5kZXJPcHRpb25zJTIwJTdEJTIwZnJvbSUyMCclNDBidWlsZGVyLmlvJTJGcXdpayUyRnNlcnZlciclM0IlNUNuaW1wb3J0JTIwJTdCJTIwUm9vdCUyMCU3RCUyMGZyb20lMjAnLiUyRnJvb3QnJTNCJTVDbiU1Q25leHBvcnQlMjBkZWZhdWx0JTIwZnVuY3Rpb24lMjAob3B0cyUzQSUyMFJlbmRlck9wdGlvbnMpJTIwJTdCJTVDbiUyMCUyMHJldHVybiUyMHJlbmRlclRvU3RyaW5nKCUzQ1Jvb3QlMjAlMkYlM0UlMkMlMjBvcHRzKSUzQiU1Q24lN0QlNUNuJTIyJTdEJTJDJTdCJTIycGF0aCUyMiUzQSUyMiUyRnJvb3QudHN4JTIyJTJDJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMCU3QiUyMEFwcCUyMCU3RCUyMGZyb20lMjAnLiUyRmFwcCclM0IlNUNuJTVDbmV4cG9ydCUyMGNvbnN0JTIwUm9vdCUyMCUzRCUyMCgpJTIwJTNEJTNFJTIwJTdCJTVDbiUyMCUyMHJldHVybiUyMCglNUNuJTIwJTIwJTIwJTIwJTNDaHRtbCUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0NoZWFkJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUyMCUyMCUzQ3RpdGxlJTNFSGVsbG8lMjBRd2lrJTNDJTJGdGl0bGUlM0UlNUNuJTIwJTIwJTIwJTIwJTIwJTIwJTNDJTJGaGVhZCUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0Nib2R5JTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUyMCUyMCUzQ0FwcCUyMCUyRiUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0MlMkZib2R5JTNFJTVDbiUyMCUyMCUyMCUyMCUzQyUyRmh0bWwlM0UlNUNuJTIwJTIwKSUzQiU1Q24lN0QlM0IlNUNuJTIyJTdEJTVE

wmertens:
Here's a comparison. Most React devs will use destructuring, and then passing a signal works without issues whereas passing `signal.value` only works if you don't destructure.

And as you can see, adding the helper functions does not change anything about rendering.

So would it not be better to have the helpers, so that most components work with both values and signals, and destructuring still works?

```jsx

import { component$, useSignal } from '@builder.io/qwik';

export const Props = (props) => {
  console.log(`render Props ${props.who}`);
  return <div>Props count {props.who} is {props.count}</div>
}

export const Destructure = ({who, count}) => {
  console.log(`render Destructure ${who}`);
  return <div>Destructure count {who} is {count}</div>
}

export const App = component$(() => {
  const signal=useSignal(0)
  signal.valueOf = function() { return this.value; }
  signal.toString = function() { return `${this.value}`; }
  signal.toJSON = function() { return this.value; }

  return <div style={{display: 'flex', flexDirection:'column'}}>
    <button onClick$={()=>signal.value++}>Click</button>

    <Props who="val" count={signal.value}/>
    <Props who="sig" count={signal}/>
    {/* If you uncomment the below, it will force rendering all children * /}
    <Destructure who="val" count={signal.value}/>
    {/ * */}
    <Destructure who="sig" count={signal}/>
  </div>
});
```

manucorporat:
This is an interesting discussion, but accepting a `Signal<number>` or `number` is not a decision i should make as a author of a component, i dont know how developers will actually consume my component. 

If you want, you can still pass signal, but it's not the pattern I think it's recommended, developers want to pass data to the props, not signals. 

The second aspect is that props destructing is an anti-pattern in Qwik, because of our render we can still make it work (Solid can not, destrucuting in solid is not allowed), but this pattern hurts fine grained updates and efficienty rerenders, because it forces a creation of subscriptions at the component level.

The third aspect is that props destructuring does not work in some cases, ie, when trying to watch for prop changes in a useWatch();

```
component$(({stuff}) => {
  useWatch(({track}) => {
    track(() => stuff);
  });
});
```

I think our efforts should focus on:
- Better docs for this
- Eslint that discourages props destructing
- Optimizer pass that remove the destructuing

wmertens:
@manucorporat So if I understand correctly, it's best to avoid passing signals, so that subscriptions are kept to a minimum?

And the optimizer uses some behind-the-scenes magic to know what to update when passing the signal's value to a child, but that's just an optimization and devs should not worry about it?

What is still confusing to me is that when you take my example above, updating the signal updates only the text nodes, but as soon as you uncomment one of the use cases, the entire subtree needs to re-render. Is that lack of some optimization? I would think only the child with the special case needs rendering?

wmertens:
:facepalm: mystery solved, I wasn't using `component$`. With it, everything works as expected, but interestingly my using Calc from App results in a new signal whereas Calc from sub gets the original signal? Could you check if this is expected?

https://qwik.builder.io/playground#version=0.11.0&buildMode=development&entryStrategy=hook&files=JTVCJTdCJTIycGF0aCUyMiUzQSUyMiUyRmFwcC50c3glMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJTdCJTIwY29tcG9uZW50JTI0JTJDJTIwdXNlU2lnbmFsJTIwJTdEJTIwZnJvbSUyMCclNDBidWlsZGVyLmlvJTJGcXdpayclM0IlNUNuJTVDbmV4cG9ydCUyMGNvbnN0JTIwUHJvcHMlMjAlM0QlMjBjb21wb25lbnQlMjQoKHByb3BzKSUyMCUzRCUzRSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZyglNjByZW5kZXIlMjBQcm9wcyUyMCUyNCU3QnByb3BzLndobyU3RCU2MCklM0IlNUNuJTIwJTIwcmV0dXJuJTIwJTNDZGl2JTNFUHJvcHMlMjBjb3VudCUyMCU3QnByb3BzLndobyU3RCUyMGlzJTIwJTdCcHJvcHMuY291bnQlN0QlM0MlMkZkaXYlM0UlNUNuJTdEKSU1Q24lNUNuZXhwb3J0JTIwY29uc3QlMjBEZXN0cnVjdHVyZSUyMCUzRCUyMGNvbXBvbmVudCUyNCgoJTdCd2hvJTJDJTIwY291bnQlN0QpJTIwJTNEJTNFJTIwJTdCJTVDbiUyMCUyMGNvbnNvbGUubG9nKCU2MHJlbmRlciUyMERlc3RydWN0dXJlJTIwJTI0JTdCd2hvJTdEJTYwKSUzQiU1Q24lMjAlMjByZXR1cm4lMjAlM0NkaXYlM0VEZXN0cnVjdHVyZSUyMGNvdW50JTIwJTdCd2hvJTdEJTIwaXMlMjAlN0Jjb3VudCU3RCUzQyUyRmRpdiUzRSU1Q24lN0QpJTVDbiU1Q25leHBvcnQlMjBjb25zdCUyMENhbGMlMjAlM0QlMjBjb21wb25lbnQlMjQoKHByb3BzKSUyMCUzRCUzRSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZyglNjByZW5kZXIlMjBDYWxjJTIwJTI0JTdCcHJvcHMud2hvJTdEJTYwKSUzQiU1Q24lMjAlMjByZXR1cm4lMjAlM0NkaXYlM0VDYWxjJTIwZG91YmxlJTIwJTdCcHJvcHMud2hvJTdEJTIwaXMlMjAlN0Jwcm9wcy5jb3VudCoyJTdEJTNDJTJGZGl2JTNFJTVDbiU3RCklNUNuJTVDbmV4cG9ydCUyMGNvbnN0JTIwU3ViJTIwJTNEJTIwKHByb3BzKSUyMCUzRCUzRSUyMCU3QiU1Q24lMjAlMjBjb25zb2xlLmxvZyglNjByZW5kZXIlMjBTdWIlNjApJTVDbiUyMCUyMHJldHVybiUyMCglM0MlM0UlNUNuJTIwJTIwJTIwJTIwJTIwJTIwJTNDUHJvcHMlMjB3aG8lM0QlNUMlMjJ2YWwlNUMlMjIlMjBjb3VudCUzRCU3QnByb3BzLnNpZ25hbC52YWx1ZSU3RCUyRiUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0NQcm9wcyUyMHdobyUzRCU1QyUyMnNpZyU1QyUyMiUyMGNvdW50JTNEJTdCcHJvcHMuc2lnbmFsJTdEJTJGJTNFJTVDbiU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0NEZXN0cnVjdHVyZSUyMHdobyUzRCU1QyUyMnZhbCU1QyUyMiUyMGNvdW50JTNEJTdCcHJvcHMuc2lnbmFsLnZhbHVlJTdEJTJGJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQ0Rlc3RydWN0dXJlJTIwd2hvJTNEJTVDJTIyc2lnJTVDJTIyJTIwY291bnQlM0QlN0Jwcm9wcy5zaWduYWwlN0QlMkYlM0UlNUNuJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQ0NhbGMlMjB3aG8lM0QlNUMlMjJ2YWwlNUMlMjIlMjBjb3VudCUzRCU3QnByb3BzLnNpZ25hbC52YWx1ZSU3RCUyRiUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0NDYWxjJTIwd2hvJTNEJTVDJTIyc2lnJTVDJTIyJTIwY291bnQlM0QlN0Jwcm9wcy5zaWduYWwlN0QlMkYlM0UlNUNuJTIwJTIwJTNDJTJGJTNFKSU1Q24lN0QlNUNuJTVDbmV4cG9ydCUyMGNvbnN0JTIwQXBwJTIwJTNEJTIwY29tcG9uZW50JTI0KCgpJTIwJTNEJTNFJTIwJTdCJTVDbiUyMCUyMGNvbnNvbGUubG9nKCU2MHJlbmRlciUyMEFwcCU2MCklNUNuJTIwJTIwY29uc3QlMjBzaWduYWwlM0R1c2VTaWduYWwoMCklNUNuJTVDbiUyMCUyMCUyRiUyRiUyMFVuY29tbWVudCUyMHRvJTIwbWFrZSUyMENhbGMlMjBzaWduYWwlMjB3b3JrJTVDbiUyMCUyMHNpZ25hbC52YWx1ZU9mJTIwJTNEJTIwZnVuY3Rpb24oKSUyMCU3QiUyMHJldHVybiUyMHRoaXMudmFsdWUlM0IlMjAlN0QlNUNuJTIwJTIwc2lnbmFsLnRvU3RyaW5nJTIwJTNEJTIwZnVuY3Rpb24oKSUyMCU3QiUyMHJldHVybiUyMCU2MCUyNCU3QnRoaXMudmFsdWUlN0QlNjAlM0IlMjAlN0QlNUNuJTIwJTIwc2lnbmFsLnRvSlNPTiUyMCUzRCUyMGZ1bmN0aW9uKCklMjAlN0IlMjByZXR1cm4lMjB0aGlzLnZhbHVlJTNCJTIwJTdEJTVDbiU1Q24lMjAlMjByZXR1cm4lMjAlM0NkaXYlMjBzdHlsZSUzRCU3QiU3QmRpc3BsYXklM0ElMjAnZmxleCclMkMlMjBmbGV4RGlyZWN0aW9uJTNBJ2NvbHVtbiclN0QlN0QlM0UlNUNuJTIwJTIwJTIwJTIwJTNDYnV0dG9uJTIwb25DbGljayUyNCUzRCU3QigpJTNEJTNFc2lnbmFsLnZhbHVlJTJCJTJCJTdEJTNFQ2xpY2slM0MlMkZidXR0b24lM0UlNUNuJTIwJTIwJTIwJTIwJTNDU3ViJTIwc2lnbmFsJTNEJTdCc2lnbmFsJTdEJTIwJTJGJTNFJTVDbiUyMCUyMCUyMCUyMCUzQ0NhbGMlMjB3aG8lM0QlNUMlMjJhcHAtc2lnJTVDJTIyJTIwY291bnQlM0QlN0JzaWduYWwlN0QlMjAlMkYlM0UlNUNuJTIwJTIwJTNDJTJGZGl2JTNFJTVDbiU3RCklM0IlMjIlN0QlMkMlN0IlMjJwYXRoJTIyJTNBJTIyJTJGZW50cnkuc2VydmVyLnRzeCUyMiUyQyUyMmNvZGUlMjIlM0ElMjJpbXBvcnQlMjAlN0IlMjByZW5kZXJUb1N0cmluZyUyQyUyMFJlbmRlck9wdGlvbnMlMjAlN0QlMjBmcm9tJTIwJyU0MGJ1aWxkZXIuaW8lMkZxd2lrJTJGc2VydmVyJyUzQiU1Q25pbXBvcnQlMjAlN0IlMjBSb290JTIwJTdEJTIwZnJvbSUyMCcuJTJGcm9vdCclM0IlNUNuJTVDbmV4cG9ydCUyMGRlZmF1bHQlMjBmdW5jdGlvbiUyMChvcHRzJTNBJTIwUmVuZGVyT3B0aW9ucyklMjAlN0IlNUNuJTIwJTIwcmV0dXJuJTIwcmVuZGVyVG9TdHJpbmcoJTNDUm9vdCUyMCUyRiUzRSUyQyUyMG9wdHMpJTNCJTVDbiU3RCU1Q24lMjIlN0QlMkMlN0IlMjJwYXRoJTIyJTNBJTIyJTJGcm9vdC50c3glMjIlMkMlMjJjb2RlJTIyJTNBJTIyaW1wb3J0JTIwJTdCJTIwQXBwJTIwJTdEJTIwZnJvbSUyMCcuJTJGYXBwJyUzQiU1Q24lNUNuZXhwb3J0JTIwY29uc3QlMjBSb290JTIwJTNEJTIwKCklMjAlM0QlM0UlMjAlN0IlNUNuJTIwJTIwcmV0dXJuJTIwKCU1Q24lMjAlMjAlMjAlMjAlM0NodG1sJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQ2hlYWQlM0UlNUNuJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTNDdGl0bGUlM0VIZWxsbyUyMFF3aWslM0MlMkZ0aXRsZSUzRSU1Q24lMjAlMjAlMjAlMjAlMjAlMjAlM0MlMkZoZWFkJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQ2JvZHklM0UlNUNuJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTIwJTNDQXBwJTIwJTJGJTNFJTVDbiUyMCUyMCUyMCUyMCUyMCUyMCUzQyUyRmJvZHklM0UlNUNuJTIwJTIwJTIwJTIwJTNDJTJGaHRtbCUzRSU1Q24lMjAlMjApJTNCJTVDbiU3RCUzQiU1Q24lMjIlN0QlNUQ%3D

manucorporat:
Yes! you are passing a signal as prop, right? so in the other side you get the same type!

wmertens:
So we had a call and I'd like to clarify my posistion. For simple components, there's no difference between a signal and its value already:

```tsx
export const ShowCount = component$((props: {count: number}) =>
  <span>{props.count}</span>)
```
```tsx
export const ShowCount = component$((props: {count: Signal<number>}) =>
  <span>{props.count}</span>)
```

With this PR, I'd like to extend this likeliness a little further
```tsx
export const ShowCountIfBig = component$((props: Signal<number>) =>
  props.count > 10 && <span>{props.count}</span>)
```

Note that the `Signal` type now is both the Signal class and its value, this makes TS happy about coercion.

Whenever you use a signal as its value, that will be properly tracked, and the code is very readable and easy to maintain.

If you don't declare your prop type as Signal, TS will complain when you try to pass a signal.

The coercion will work for all operations except `===`, `!==`, and `typeof`. IMHO having to declare the prop as a Signal means that you are aware of this.

---

Alternatively, all coercion could be disallowed, but we should then still have these helpers:
```js
{
  // prevent accidental use as value
  valueOf() {
    throw new TypeError('Cannot coerce a Signal, use `.value` instead');
  }
  toString() {
    return `[Signal ${JSON.stringify(this.value)}]`;
  }
  toJSON() {
    return {signal: this.value};
  }
}
```

manucorporat:
Let's do it! but let's update to be like this:
```tsx
  toString() {
    return `[Signal ${String(this.value)}]`;
  }
  toJSON() {
    return {value: this.value};
  }
```

wmertens:
Ok I'll change it. With the throw for valueOf, right?

Wout.

wmertens:
Done - IMHO there should be a unit test for these but there aren't any for Signal either so I don't know where to put them :-)

manucorporat:
Recently changed the lexicalscope test,can you try to resovle the conflicts it? i will merge afterwards!

wmertens:
@manucorporat done, the e2e on macos failed due to network during npm i

wmertens:
@manucorporat done, the e2e on macos failed due to network during npm i

PR #1721, opened 2022-10-13
Remove emotion from qwik-react

MarkLyck:
# What is it?

- [X] Feature / enhancement
- [X] Bug
- [ ] Docs / tests

# Description

This removes the @emotion dependencies from qwik-react.

# Use cases and why

qwik-react should not dictate any specific CSS-in-JS style library and it is not needed in order to get react components to render in Qwik.

By proxy it also fixes: https://github.com/BuilderIO/qwik/issues/1508 but only because there's an import related bug with emotion (unrelated to the idea of removing it entirely from qwik-react).

I'm guessing this change is going to require additional work as it would likely break imports of react components using emotion CSS. I believe there should be another plugin or package to solve that issue though, rather than solving it inside qwik-react. I don't use emotionCSS and have not actually tested what happens when you import a React component that does use it. But I'm hoping to at least start the discussion that this particular CSS in JS solution doesn't belong in the `qwik-react` package, and hopefully figure out where (if anywhere) it does belong.

I'd be happy to help develop that solution as well, but I'm assuming there's a reason it was put in here in the first place, so I would love to hear what that reason was and maybe I am entirely wrong about this. 🙏

I started a discussion here as well: https://github.com/BuilderIO/qwik/discussions/1719 But they don't seem to get much attention.

# Checklist:

### DIFF
--- a/packages/qwik-react/package.json
+++ b/packages/qwik-react/package.json
@@ -19,9 +19,6 @@
   },
   "devDependencies": {
     "@builder.io/qwik": "workspace:*",
-    "@emotion/react": "11.10.4",
-    "@emotion/server": "11.10.0",
-    "@emotion/styled": "11.10.4",
     "@mui/material": "5.10.9",
     "@types/react": "18.0.21",
     "@types/react-dom": "18.0.6",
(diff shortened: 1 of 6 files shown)


manucorporat:
this was a great PR! but got in conflict with a big refactor of the react integration! Thanks for your effort!

Issue #1746, opened 2022-10-16
The useLocation().query object values should return string or Array<string>


tuurbo:
### Qwik Version

qwik 0.11.0, qwik city 0.0.112

### Operating System (or Browser)

chrome

### Node Version (if applicable)

16.14.2

### Which component is affected?

Qwik City

### Expected Behaviour

The url `example.com?brand=amd&brand=intel&category=cpu` contains two "brand" query params and one "category" query param. 

useLocation().query should return `{ brand: ['amd', 'intel'], category: 'cpu' }`

### Actual Behaviour

useLocation().query returns `{ brand: 'intel', category: 'cpu' }`, only returning the last "brand" query param.

### Additional Information

demo https://stackblitz.com/edit/qwik-starter-7bcp3c?a=1&file=src%2Froutes%2Findex.tsx

adamdbradley:
The `query` object is now a `URLSearchParams` as of https://github.com/BuilderIO/qwik/pull/2385

Thanks for opening the issue!

PR #1759, opened 2022-10-16
Rewrite get query logic

mmarkelov:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Please see #1746. 
There is a problem with current approach when we try to get array values, it will be always one single value. 
I'm not sure if it should handle this case, but if it should be fixed, I guess that there will be maybe some breaking changes(but not sure).
If you are ok with this solution, I could try to fix TS errors 

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/library/qwik-city-component.tsx
+++ b/packages/qwik-city/runtime/src/library/qwik-city-component.tsx
@@ -90,8 +90,7 @@ export const QwikCity = component$<QwikCityProps>(() => {
   const getRouteLocationQuery = (searchParams: URLSearchParams) => {
     const query: RouteParams = {};
     for (const key of searchParams.keys()) {
-      const values = searchParams.getAll(key);
-      query[key] = values.length === 1 ? values[0] : values;
+      query[key] = searchParams.getAll(key).join(',');
     }
     return query
   }
--- a/packages/qwik-city/runtime/src/library/qwik-city-component.tsx
+++ b/packages/qwik-city/runtime/src/library/qwik-city-component.tsx
@@ -16,6 +16,7 @@ import type {
   MutableRouteLocation,
   PageModule,
   RouteNavigate,
+  RouteParams,
 } from './types';
 import {
   ContentContext,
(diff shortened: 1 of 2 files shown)


manucorporat:
I think this PR makes sense, only worried about the breaking change in terms of types and that now it’s not easy to easily get a query value because it’s always needed to handled the string or array of strings

mmarkelov:
> I think this PR makes sense, only worried about the breaking change in terms of types and that now it’s not easy to easily get a query value because it’s always needed to handled the string or array of strings

Alternatively, it can be another way to handle. Keep it as the string, but if we got an array we can translate it string with join.
But IDK how we should parse it in `getPathnameForDynamicRoute` function

tuurbo:
> I think this PR makes sense, only worried about the breaking change in terms of types and that now it’s not easy to easily get a query value because it’s always needed to handled the string or array of strings

Just a thought, but to keep the current simplicity of not having to handle a string or array of strings. Would something like this be possible? Call the query param as you currently do and return a string (current behavior). Call `.getAll('...')` and always return a array of strings.



URL: `example.com?brand=amd&brand=intel&category=cpu`

```ts
location.query.category // 'cpu'
location.query.getAll('category') // ['cpu']

location.query.brand // 'intel'
location.query.getAll('brand') // ['amd', 'intel']
```

--- OR ----

Support multiple ways to access the query params similar to Angular
```ts
location.query.category // 'cpu'
location.queryMap.get('category') // 'cpu'
location.queryMap.getAll('category') // ['cpu']

location.query.brand // 'intel'
location.queryMap.get('brand') // 'intel'
location.queryMap.getAll('brand') // ['amd', 'intel']
```

mmarkelov:
@tuurbo I'm ok with your suggestions. I can implement this kind of logic

tuurbo:
or here is an even simpler way the query could be accessed
```
// Always return string
location.query.brand // 'intel' 

// Always return string array
location.queryMap.brand // ['amd', 'intel']
```

what do you think @manucorporat  @mmarkelov ?

manucorporat:
how about query and queryAll?
```
// Always return string
location.query.brand // 'intel' 

// Always return string array
location.queryAll.brand // ['amd', 'intel']
```

tuurbo:
That works for me

manucorporat:
however i dont like the idea what we have two root objects for the same concept, i can see developers getting confused... what if we just exposed directly the `URLSearchParams`?

It would require a little change in qwik to support serializing URLSearchParams, but should be possible!
and deprecate `query`

cc @adamdbradley

zanettin:
Thanks a lot @mmarkelov for taking the time to create this PR 🙏  as @adamdbradley stated in the linked issue, this was fixes in https://github.com/BuilderIO/qwik/pull/2385. I'll therefore close this PR for now. Please feel free to open a new one if you see any further improvements in that or any other regard 🎉

PR #1827, opened 2022-10-20
Improve event types

brunocrosier:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Improvement on https://github.com/BuilderIO/qwik/pull/1559 

# Use cases and why

Automatically infer the correct types for events based on the HTML element it is on, instead of having to manually add a type assertion (which breaks type-safety)


manucorporat:
Merged!!! thanks a lot for the amazing work!

PR #1841, opened 2022-10-22
fix: Partytown Snippet inner html

nnelgxorz:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Per a discussion with a user on discord, the Partytown component in the cli sets the Partytown snippet in the script body instead of in dangerouslySetInnerHTML. This was preventing Partytown from working.

# Checklist:

### DIFF
--- a/starters/features/partytown/src/components/partytown/partytown.tsx
+++ b/starters/features/partytown/src/components/partytown/partytown.tsx
@@ -14,6 +14,5 @@ export interface PartytownProps extends PartytownConfig {}
  * You can pass setting with props
  */
 export const QwikPartytown = (props: PartytownProps): any => {
-  const innerHTML = partytownSnippet(props);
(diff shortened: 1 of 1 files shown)


shairez:
As always, great work @nnelgxorz !

I wonder if there's a way to add a regression test here to make sure this bug doesn't come back 🤔

nnelgxorz:
Thanks! We might be able to add the Partytown component into one of the e2e tests, if there's a way to check that Partytown is running. Maybe @adamdbradley could suggest a way to check that Partytown is operational on the page.

shairez:
Going to merge this as a fix in the meantime
Thanks!

I think it'll be good if we'll add also tests for all the starters (in a different PR)
I might be able to do it after I come back from the conf, but in the meantime if someone wants to do it - you'll rewarded eternal happiness and richness by the testing gods (it's like the demo gods, but for testing) 😊

PR #1843, opened 2022-10-22
remove `preventDefault()` from event type

brunocrosier:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

I realized that the event type should not have `preventDefault` on it due to https://qwik.builder.io/docs/components/events/#prevent-default


manucorporat:
Ohh this is actually a great improvement!

PR #1857, opened 2022-10-23
feat: deprecate `preventDefault()`

brunocrosier:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Instead of removing `preventDefault()` from the event type entirely, this will provide guidance on how to achieve the intended result by using the `preventdefault:{event}` attribute

# Screenshot

image

### DIFF
--- a/packages/qwik/src/core/render/jsx/types/jsx-qwik-events.ts
+++ b/packages/qwik/src/core/render/jsx/types/jsx-qwik-events.ts
@@ -3,8 +3,24 @@ export interface SyntheticEvent<T = Element, E = Event>
 
 interface BaseSyntheticEvent<E = object, C = any, T = any> {
   nativeEvent: E;
+  /**
+   * @deprecated 
+   * 
+   * Qwik cannot reliably capture `event.currentTarget`
+   * 
+   * Instead, you can use `event.target`
+   * 
+   * or 
+   * 
+   * ```tsx
+   * onClick$={(event, currentEvent) => {
+   * // use currentEvent here
+   * }}
+   * ``` 
+   * 
+   */
   currentTarget: C;
-  target: T;
+  target: C;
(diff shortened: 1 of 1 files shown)


packages/qwik/src/core/render/jsx/types/jsx-qwik-events.ts
```diff
@@ -12,6 +12,29 @@ interface BaseSyntheticEvent<E = object, C = any, T = any> {
...
+   *
+   * @example
+   * ```tsx
+   * <a
+   *  href="/about"
+   *  preventdefault:click // This will prevent the default behavior of the "click" event.
+   *  onClick$={(event) => {
+   *    // event.preventDefault() would not work here
+   *    // because of Qwik's async nature
+   *    trackAnalyticsEvent(event.currentTarget)
```
langbamit, line 29: you mean `event.target`?

manucorporat: yeah, i think event.currentTarget is not a good idea in Qwik either, since we capture the event, it's likely to be incorrect. I would rather use `event.target` or the second argument of the callback:

```
onClick$={(event, currentEvent) => {
```

Maybe we can set event.currentEvent as deprecated as well, and recommend the alternative

brunocrosier: sounds good, have updated in latest commit https://github.com/BuilderIO/qwik/pull/1857/commits/fcdbd53458f421cb766acf7a869da06be073f125

preview: 
image

zanettin:
@manucorporat may i hand this over to u?

manucorporat:
We decided to not even include `preventDefault()` at all in the type! Even more strict than deprecated!

Issue #1851, opened 2022-10-22
Allow appending multiple Set-Cookie Headers


nnelgxorz:
### Is your feature request related to a problem?

The current headers implementation uses an append function that concatenates header values together with a ", ". This works for all headers except Set-Cookie.

### Describe the solution you'd like

There are a few options I can think of:

 - create special functions for each headers method specifically for cookies: `getCookie`, `setCookie`, `appendCookie`.
 - keep the existing headers methods, but check for the header value internally and store cookies separately. (if (name === 'Set-Cookie) {...}`
 - keep everything the same internally, but special case Set-Cookie to values on the ", " before creating the response and generating valid Set-Cookie headers for each value.
 - same as above, but use node-fetch's .raw() to get the Set-Cookie headers as an array.

### Describe alternatives you've considered

![cookie](https://user-images.githubusercontent.com/9634080/197366397-51cb8b2f-8300-4980-9ded-b119cf9c797a.jpg)

### Additional context

https://github.com/whatwg/fetch/issues/973 - Discussion around formalizing a solution for Set-Headers
https://github.com/withastro/astro/pull/3092/files - Astro PR using the adding multiple Set-Cookies for Netlify
https://github.com/sveltejs/kit/pull/3502/files - Svelte PR adding multiple Set-Cookies

manucorporat:
I think good cookie support is super important, wanna take into preparing a PR most likely inspired by sveltekit and astro?

This would be such a massive improvement!!

nnelgxorz:
Definitely! I was going to take it on yesterday, but it looks like Qwik City e2e tests are broken. I'm guessing because of the refactor Adam did on the middlewares/servers.

nnelgxorz:
PR is ready for review!

PR #1864, opened 2022-10-24
feat: First Class Cookie Support

nnelgxorz:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

The current headers implementation concatenates values with the same header using ", ". This works for all headers except Set-Cookie where commas are not allowed.

Other frameworks appear to rely on node-fetch having a raw() function that returns the non-concatenated header values. This seems like a poor assumption as we should not rely on a third party package being available inside various edge environments.

This PR creates a Cookie class that is accessible on the RequestContext. It allows a type-safe API for getting/setting cookies and allows us to dodge the header concatenation issue entirely.

Inspired by this Astro RFC: https://github.com/withastro/astro/pull/4876

Fixes: #1851 

# Checklist:


langbamit:
I think you can't use `', '` here

```Set-Cookie: id=a3fWa; Expires=Wed, 21 Oct 2015 07:28:00 GMT```

The `<cookie-name>` and `<cookie-value>` can't contains whitespace and comma but the `Expires` attribute can.

**Comment by nnelgxorz** [comment]
Great catch, @langbamit! I think Svelte is using a special parser. I'll look into a better solution. Thanks!

**Review by ScottAgirs — COMMENTED** [review]
Nois, this is exciting!! ") Pls don't hate me for this question, but - any estimate if the could be a merge in near future? 😃

**Comment by nnelgxorz** [comment]
> Nois, this is exciting!! ") Pls don't hate me for this question, but - any estimate if the could be a merge in near future? smiley

Definitely the near future. :+1:

**Comment by nnelgxorz** [comment]
Hey @manucorporat. I'm unable to run yarn api.update locally after rebasing because of this error:

```bash
packages/qwik/src/optimizer/src/plugins/vite.ts:252:11 - error TS2322: Type '{ outDir: string; rollupOptions: { input: string[]; preserveEntrySignatures: "exports-only"; output: OutputOptions; treeshake: { moduleSideEffects: false; }; onwarn: (warning: RollupWarning, warn: WarningHandler) => void; }; modulePreload: { ...; }; dynamicImportVarsOptions: { ...; }; }' is not assignable to type 'BuildOptions'.
  Object literal may only specify known properties, and 'modulePreload' does not exist in type 'BuildOptions'.

252           modulePreload: {
              ~~~~~~~~~~~~~~~~
253             polyfill: false,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
254           },
    ~~~~~~~~~~~

  node_modules/vite/dist/node/index.d.ts:2448:5
    2448     build?: BuildOptions;
             ~~~~~
    The expected type comes from property 'build' which is declared here on type 'UserConfig'


Found 1 error in packages/qwik/src/optimizer/src/plugins/vite.ts:252
```

Any ideas? I could revert the change where I removed the cookies field from the Netlify Context and unblock this for merging.

packages/qwik-city/middleware/request-handler/cookie.unit.ts
```diff
@@ -0,0 +1,91 @@
...
+      key: 'k',
+      value: '11',
+      options: { expires: 'Wed, 21 Oct 2015 07:28:00 GMT' },
+      expect: 'k=11; Expires=Wed, 21 Oct 2015 07:28:00 GMT',
+    },
+    {
+      key: 'l',
+      value: '12',
+      options: { expires: new Date('Wed, 21 Oct 2015 07:28:00 GMT') },
+      expect: 'l=12; Expires=Wed, 21 Oct 2015 07:28:00 GMT',
```
Mouvedia, line 68: You should add a test for version 0 cookies.
i.e. if you have spaces it should be replaced by %20
e.g. `'l=foo%20bar; Expires=Wed, 21 Oct 2015 07:28:00 GMT'`

PR #1866, opened 2022-10-25
docs: add aria/rel attributes to docs header

timscodebase:
# What is it?

- [x] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
The header was missing standard aria and rel tags

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


manucorporat:
I think the linter is complainign because we are using single quotes, instead of double. Should be fixed by running:

yarn fmt

shairez:
@timscodebase can you please run `yarn fmt` or prettier --fix for the specific file? 

Thanks!

starters/apps/basic/src/components/header/header.tsx
```diff
@@ -8,23 +8,23 @@ export default component$(() => {
   return (
     <header>
       <div class="logo">
-        <a href="https://qwik.builder.io/" target="_blank">
+        <a rel='noopener' aria-label='logo' href="https://qwik.builder.io/" target="_blank">
```
ptu14, line 11:
```suggestion
        <a href="https://qwik.builder.io/" target="_blank" rel="noopener">
```

`href` is the most important attribute in `<a>`, so let's give it the first place. Also use double qoutes.

when it comes to `aria-label`, I think the `"logo"` itself doesn't mean anything. I think the link can have `"Qwik homepage"` aria-label.

svg itself can have `<title>Qwik logo</title>` inside and `role="img"` but it's irrelevant.

zanettin:
Hi @timscodebase 👋 
Thanks a lot for your input 🙏  do you have a chance to implement the mentioned changes against the current state? Otherwise leave a quick note, so anybody else can have a look at it 🙏

PR #1874, opened 2022-10-25
feat: add unit testing support

leifermendez:
# What is it?
### @builder.io/qwik/testing

The main idea is to create a layer/library to expose differences methods for writing unit testing in Qwik.
Below it you have an example of how it should look
- [ X ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
### @builder.io/qwik/testing

The main idea is to create a layer/library to expose differences methods for writing unit testing in Qwik.
Below it you have an example of how it should look

User API:
-  createDOM: Generate a DOM for every unit test and expose `screen`,`render`.`userEvent`  
-  screen: Return a DOM
-  render: Is the same function that inside @builder.io/qwik
-  userEvent: Is a wrapper friendly of `trigger`  same function that inside `'./element-fixture'` 

```ts
//vite.config.ts
import { qwikVite } from '@builder.io/qwik/optimizer';
import { qwikCity } from '@builder.io/qwik-city/vite';
import tsconfigPaths from 'vite-tsconfig-paths';

  return {
    plugins: [qwikCity(), qwikVite(), tsconfigPaths()],
    define: {
      'globalThis.qTest': true,
      'globalThis.qDev': true,
    },
  };
});
```

```jsx
// card.test.tsx

import { createDOM } from '@builder.io/qwik/testing';
import { test, expect } from 'vitest';
import Card from './card.tsx';

test(`[Card Component]: Only render`, async () => {
  const { screen, render } = createDOM();
  await render(<Card />);
  expect(screen.outerHTML).toCointain('Counter_0');
});

test(`[Card Component]: Click counter +1`, async () => {
  const { screen, render, userEvent } = createDOM();
  await render(<Card />);
  expect(screen.outerHTML).toCointain('Counter_0');
  await userEvent('button.btn-counter', 'click');
  expect(screen.outerHTML).toCointain('Counter_1');
});
```

Image

PD: Thnks @manucorporat  for the support!

# Checklist:

- [x ] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [x ] I have performed a self-review of my own code
- [x ] I have made corresponding changes to the documentation
- [ x] Added new tests to cover the fix / functionality


packages/qwik/src/testing/library.ts
```diff
@@ -0,0 +1,37 @@
...
+  setTestPlatform();
+  const host = new ElementFixture().host;
+  return {
+    render: function (jsxElement: JSXNode) {
+      return renderIn(host, jsxElement);
+    },
+    screen: host,
+    userEvent: function (element: HTMLElement | string | Element, event: string) {
+      if (typeof element === 'string') return trigger(host, element, event);
+      const classListValue = parseClassWithElement(element);
```
manucorporat, line 33: i am a bit confused by this

leifermendez: > I think we should modified `trigger` function to receive both CSS selector and element instead of.

But `trigger` already has this function. 

```tsx
// card
 <button class='block'>Hello!</button >
```


```tsx 
// test
 await userEvent('.block', 'click');
```

langbamit: I mean
```ts
export async function trigger(
  root: Element,
  selector: string,
  eventNameCamel: string
): Promise<void> {
  for (const element of Array.from(root.querySelectorAll(selector))) {
```
to

```ts
export async function trigger(
  root: Element,
  selector: string | Element, // ArrayLike<Element> | Iterable<Element>
  eventNameCamel: string
): Promise<void> {
const elements = typeof selector === 'string' ? Array.from(root.querySelectorAll(selector)) : [selector]
  for (const element of elements) {
```

So we can call `trigger` directly without weird part like `parseClassWithElement`

Btw, `parseClassWithElement` fail with this case below:
```tsx
// buttons
<button>Button 1</button>
<button>Button 2</button>
```
```ts
// test
const button = root.querySelector("button:nth-child(2)")
await userEvent(button, 'click'); // selector = 'BUTTON'
```

P/s: Sorry, I'm not a native English speaker

leifermendez: I get it the point. Yeah I think that is possible without complication. Let me try. Also, I'm working to move some functions

leifermendez: @manucorporat Do you know or have any idea about it?

I have two buttons but `document.querySelector` only work when I use `document.querySelector('button:nth-child(4)')`  

image

leifermendez: but the way @langbamit thanks for your suggestion. Added 🙌

manucorporat:
one unit test is failing, i am happy to merge this as it is (phase 1)

leifermendez:
> one unit test is failing, i am happy to merge this as it is (phase 1)

:) yeah! , I don't know Why failed the pipeline ?

leifermendez:
@manucorporat I updated the `component.unit.test` with the new implementation and work fine it locally but the pipeline doesn't pass

leifermendez:
image

manucorporat:
I know why it does not pass, it's tricky, tomorrow lets talk in our scheduled meeting and figure it out

leifermendez:
everything pass 🙌

PR #1975, opened 2022-11-03
Patch 1

JeanMeche:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x ] Docs / tests

# Description

Little update to the docs to use the new ssg adapter.

### DIFF
--- a/starters/apps/basic/src/routes/index.tsx
+++ b/starters/apps/basic/src/routes/index.tsx
@@ -85,7 +85,7 @@ export default component$(() => {
         </tr>
         <tr>
           <td>
-            <code>npm run qwik add static-node</code>
+            <code>npm run qwik add static</code>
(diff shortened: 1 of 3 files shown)


starters/apps/basic/src/routes/index.tsx
```diff
@@ -85,7 +85,7 @@ export default component$(() => {
         </tr>
         <tr>
           <td>
-            <code>npm run qwik add static-node</code>
+            <code>npm run qwik add static</code>
```
shairez, line 88: I think this should be `npx qwik add static` (also the next fix)

shairez:
Thanks @JeanMeche !

Added a comment

manucorporat:
Thank you very much!

PR #1987, opened 2022-11-04
fix: depend upon wrangler@latest

petebacondarwin:
The current dependency of `beta` is hopelessly out of date.


manucorporat:
Thank you very much Pete!

PR #1991, opened 2022-11-04
perf: drastically improve resuming perf (phase 1)

manucorporat:
# What is it?

Resuming already is pretty fast, but we were using some DOM API in a very unoptimal way, createTreeWalker is extremelly fast since it can walk the DOM at C++ speed, but we implemented a heavy acceptNode() function, which is a footgun.

Instead we implement the same thing in a different way, that least to a massive 14x improvement for a big DOM.
https://twitter.com/manucorporat/status/1588506463986933761

This is not the end, resuming does already a lot of stuff eagarly that it does not really have to do, but it was easy to implement it this way.

Follow up PR will improve perf in another 10x factor, getting the total improvement of around x100 times fast, ie, from 1 second to around 1ms

### DIFF
--- a/packages/qwik/src/core/container/pause.ts
+++ b/packages/qwik/src/core/container/pause.ts
@@ -233,8 +233,6 @@ export const _pauseFromContexts = async (
       id = getQId(el);
       if (!id) {
         console.warn('Missing ID', el);
-      } else {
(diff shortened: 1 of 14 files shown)


manucorporat:
Needs new e2e tesra for containers before merge

PR #2005, opened 2022-11-06
perf(resume): O(1) algorithm (phase 2)

manucorporat:
# What is it?

Resuming is a cheap algorithm, by definition. But current it was implemented in a naive way, where all the job is done eagarly, this is not a requirement. New algorithm can only restore the state that is actually required for the given interaction.

Leading to a O(1) perf characteristic, regarless of amount of DOM, or data!

## Next

Phase 3 will consist in microoptimizations and memory usage improvements!
Screenshot 2022-11-06 at 17 04 54

Screenshot 2022-11-06 at 17 44 14

### DIFF
--- a/packages/qwik/src/core/api.md
+++ b/packages/qwik/src/core/api.md
@@ -541,8 +541,6 @@ export interface SnapshotMetaValue {
     // (undocumented)
     h?: string;
     // (undocumented)
-    r?: string;
-    // (undocumented)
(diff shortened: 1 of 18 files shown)


nnelgxorz:
Whoa! Awesome work @manucorporat!

manucorporat:
Needs to e2e test involving DOM removal

PR #2027, opened 2022-11-07
feat: Expose event types

nnelgxorz:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

We now have awesome event types that work great in inline event handlers. But because these types are not exported, people are unable to use them when creating QRLs that are not inline.

This PR exposes those types for developers.

# Checklist:

### DIFF
--- a/packages/qwik/src/core/api.md
+++ b/packages/qwik/src/core/api.md
@@ -304,6 +304,42 @@ export type MountFn<T> = () => ValueOrPromise<T>;
 // @alpha @deprecated (undocumented)
 export const mutable: <T>(v: T) => T;
 
+// @public (undocumented)
+export type NativeAnimationEvent = AnimationEvent;
+
+// @public (undocumented)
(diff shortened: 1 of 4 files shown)


manucorporat:
LGTM! cc @adamdbradley

manucorporat:
One thought, `@builder.io/qwik` is getting bloated with JSX and HTML types, i am wondering if we should remove all this exports from `@builder.io/qwik` and export everything from `@builder.io/qwik/jsx-runtime`

manucorporat:
Nah, the motivation would be to not have a million things in qwik/core hurting discoverability of other APIs. To be honest i am not sold on putting everything in jsx-runtime either

manucorporat:
Let's ship this, and maybe we can interate later, dont want to block this PR!

Thanks again for the amazing work Glenn! you are amazing

Issue #1729, opened 2022-10-14
useSignal does not seem to work for nested objects and arrays (unlike useStore with `{recursive: true}`)


ramkumarkb:
### Qwik Version

0.11.0

### Operating System (or Browser)

Mac OS 10.15.7 (Catalina) 

### Node Version (if applicable)

16.13.0

### Which component is affected?

Qwik Runtime

### Expected Behaviour

`useSignal` does not seem to work for Arrays and nested Objects. However, `useStore` works as it has the `{recursive: true}` and it able to track changes across nested objects and arrays.

### Actual Behaviour

`useSignal` seems to work only for simple values. 

### Additional Information

Gist (with `useStore`)  that works vs  Gist (with `useSignal`) that does not seem work - 

https://gist.github.com/ramkumarkb/d02ee17d559cffe948d8e61b4879733b

manucorporat:
useSignal() is not recursive, and it wont be, only track if signal.value changes, not a tested properties. Use `useStore() + recursive`

shairez:
TODO: add this to the docs

PR #2035, opened 2022-11-07
tutorial: replace useRef with useSignal (fixes #1810)

### LINKED ISSUE #1810: [Tutorial] Use of Deprecated Functions useRef()
I am working through the tutorial and I noticed that `useRef` is used in multiple places. The app is declaring it as deprecated, and should be replaced with `useSignal`.

zanettin:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description
fixes #1810 

# Checklist:


forresst (commented):
I had wanted to make the modification in this tutorial, but I had noticed that useSignal had "two features". One that replaces useRef, the other was more obscure to me. Even if your work is very interesting, I still wonder about the two functionalities.

Maybe you should answer/explain in this tutorial my 2 questions to be clearer (but I confess that I am not able to answer these questions myself in a clear way):
- If useSignal is similar to useStore, why did you create useSignal ? (Explanation of the differences, advantages, disadvantages etc...)
- Why when you use useSignal as useRef, you have to declare with <Element> : const outputRef = useSignal<Element>(); ? (This seems a bit magic, an explanation would be a plus)

zanettin:
Thanks for your feedback and the questions 🙏  I am also still learning qwik and i am not part of the core team 😄 
i guess @manucorporat could answer your question with ease. What i can tell you is, that `useRef` and `useStore` were around first. Then `useSignal` was introduced and also replaced `useRef` usage. A very good intro into the topic can be found here: https://youtu.be/Mi7udzhcCDQ?t=4998
I'll try to enhance the tutorial a bit. IMO the tutorial should not replace the Docs which is also missing for `useSignal` but also requested to be created 👍

zanettin:
@shairez @manucorporat:  would that be smth to merge or should i improve it somewhere?

shairez:
Thanks @zanettin !
LGTM

@manucorporat WDYT?

washanhanzi:
Another difference between `useSignal` and `useStore` should be included in the documentation as mentioned here #1729.

zanettin:
thanks @washanhanzi for your input 🙏 i thought the tutorial shouldn't be too long but the docs would need to include all of this specialities. i am also not sure tbh, if this is smth that will persist or if useSignal already supports nested objects and arrays with current version. if it is okay for you, we should mention this in the docs and not on the tutorial page to keep it smaller. but still, i am open for votes/inputs 😃

zanettin:
@manucorporat @adamdbradley anything missing here? 👼

manucorporat:
All good! sorry for the delay! going through all PRs! 
Thank you so much for the work! hope to see more coming :)

PR #2090, opened 2022-11-11
fix: context treeshaking with slot

### LINKED ISSUE #2087: Context via Slot works on server but not on client resume
### Qwik Version

0.12.1

Chrome

_No response_

Qwik Runtime

### Expected Behaviour

I'm using context and it just works on the server, so it should work everywhere

### Actual Behaviour

This code shows the problem in the playground. When you click the button, it crashes the runtime

```tsx
import {
  component$,
  Slot,
  createContext,
  useContextProvider,
  useContext,
  useStore,
  useSignal,
} from "@builder.io/qwik";

export const Ctx = createContext("symbols");

export const Symbol = component$(({ id }) => {
  console.log("render Symbol", id);
  const s = useContext(Ctx);
  return <p>Symbol {id}, context value: {s.t}</p>
});

export const Provider = component$(() => {
  console.log("render Provider");
  const s = useStore({ t: "yes" });
  useContextProvider(Ctx, s);
  return (
      <Slot />
  );
});
export const Sub = component$(() => <>
      <Symbol id="A" />
</>)

export const App = component$(() => {
  const t=useSignal(0)
  return (
    <Provider>
      <Sub/>
      <Symbol id="B" />
      <button onClick$={()=>t.value++}>Click me</button>
      {!!t.value && <Symbol id="C" />}
    </Provider>
  );
});
```

### Additional Information

_No response_

manucorporat:
fixes #2087

### DIFF
--- a/packages/qwik/src/core/render/ssr/render-ssr.ts
+++ b/packages/qwik/src/core/render/ssr/render-ssr.ts
@@ -42,7 +42,6 @@ import { EMPTY_OBJ } from '../../util/flyweight';
 import {
   createContext,
   HOST_FLAG_DYNAMIC,
-  HOST_FLAG_DYNAMIC_SLOT,
   HOST_FLAG_NEED_ATTACH_LISTENER,
   QContext,
   Q_CTX,
@@ -626,7 +625,6 @@ const renderNode = (
     elCtx.$parent$ = rCtx.$cmpCtx$;
     elCtx.$slotParent$ = rCtx.$slotCtx$;
     if (dynamicChildren) {
-      elCtx.$flags$ |= HOST_FLAG_DYNAMIC_SLOT;
       if (hostCtx) {
         addDynamicSlot(hostCtx, elCtx);
       }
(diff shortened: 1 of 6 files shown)


packages/qwik/src/core/render/ssr/render-ssr.ts
```diff
@@ -915,10 +926,20 @@ export const escapeAttr = (s: string) => {
...
+const listenersNeedId = (listeners: Listener[]) => {
   return listeners.some((l) => l[1].$captureRef$ && l[1].$captureRef$.length > 0);
 };
 
+const addDynamicSlot = (hostCtx: QContext, elCtx: QContext) => {
+  let dynamicSlots = hostCtx.$dynamicSlots$;
+  if (!dynamicSlots) {
+    hostCtx.$dynamicSlots$ = dynamicSlots = [];
+  }
+  if (!dynamicSlots.includes(elCtx)) {
```
wmertens, line 938: O(n), is there a case that this can cause performance issues?

manucorporat: N should be small! HashMaps get faster as N get bigger, for small N, array usually performs better. 
Also hashmap better if in-order removing is important.

PR #2110, opened 2022-11-12
feat: Serializable Derived Signals

manucorporat:
Allow qwik runtime and optimizer to inline functions, this functions can still capture lexical scope and used to prevent downloads to whole template during rerendering.


wmertens:
So you have to manually mark them with $$? Will there be some heuristic that automatically does this for $ functions as well?

DustinJSilk:
Looks like this might fix https://github.com/BuilderIO/qwik/issues/3081 🥳

PR #2124, opened 2022-11-13
docs(resumability): description of non serializable entities

voluntadpear:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

This PR mentions some of the classes that are indeed serializable by Qwik, updating previously outdated bits of the documentation.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/concepts/resumable/index.mdx
+++ b/packages/docs/src/routes/docs/concepts/resumable/index.mdx
@@ -83,13 +83,17 @@ Limitations of JSON which Qwik solves:
 - JSON produces DAG. DAG stands for Directed Acyclic Graph, which means that the object which is being serialized can't have circular references. This is a big limitation because the application state is often circular. Qwik ensures that when the graph of objects gets serialized, the circular references get properly saved and then restored.
 - JSON can't serialize some object types. For example, DOM references, Dates, etc... Qwik serialization format ensures that such objects can correctly be serialized and restored. Here is a list of types that can be serialized with Qwik:
   - DOM references
-  - Dates (not yet implemented)
-  - Function closures (if wrapped in QRL).
+  - Promises
+  - Function closures (if wrapped in QRL)
+  - Dates
+  - `URL` objects
+  - `Map` and `Set` instances
 
 Limitations of JSON that Qwik does not solve:
 
 - Serialization of classes (`instanceof` and prototype)
-- Serialization of `Promise`s, Streams, etc...
+  - Although some built-in classes, such as [Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date), [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL), [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) are supported.
+- Serialization of Streams, etc...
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/concepts/resumable/index.mdx
```diff
@@ -83,13 +84,17 @@ Limitations of JSON which Qwik solves:
 - JSON produces DAG. DAG stands for Directed Acyclic Graph, which means that the object which is being serialized can't have circular references. This is a big limitation because the application state is often circular. Qwik ensures that when the graph of objects gets serialized, the circular references get properly saved and then restored.
 - JSON can't serialize some object types. For example, DOM references, Dates, etc... Qwik serialization format ensures that such objects can correctly be serialized and restored. Here is a list of types that can be serialized with Qwik:
   - DOM references
-  - Dates (not yet implemented)
-  - Function closures (if wrapped in QRL).
+  - Promises
```
shairez, line 87: well technically, promises are wrapped inside of "resources", so I think this should probably come with this disclaimer

shairez:
thanks @voluntadpear 
added a comment

packages/docs/src/routes/docs/concepts/resumable/index.mdx
voluntadpear: Thank you @shairez! I've added a link to the resources doc from this bit

voluntadpear:
Thank you @shairez! Please let me know how it looks now 🙂

packages/docs/src/routes/docs/concepts/resumable/index.mdx
shairez: Thanks! I would add something like

Promises (wrapped in "Resources" which is the link)

It'll make it more explicit IMO

shairez:
Thanks @voluntadpear !
Good idea to add the link
sorry to bug you about this, suggested a small improvement, I think after that we're good to go

packages/docs/src/routes/docs/concepts/resumable/index.mdx
voluntadpear: My doubt is that although the standard is to wrap promises under resources, it is technically possible to have a promise serialized without it, as can be seen in the revised serialization example: https://qwik.builder.io/docs/cheat/serialization/, that works correctly.

What do you think? If it would serve as a way to encourage people to always use `useResource` for promises regardless, I can word it as you suggested. Let me know 🙂

shairez: Let's bring in the big guns... @manucorporat what do you think?

manucorporat: promises can be serialized in isolation, as long as they are resolved

shairez: Thanks @manucorporat 
1. What do you mean by "in isolation"?
2. Do you think it's a good idea to have a disclaimer about it in this doc?

manucorporat: i think it's ok! i meant that a promise can be serialized without involving a resource, but it's good we put a reference there, they are mostly used together

manucorporat:
ship it!

packages/docs/src/routes/docs/concepts/resumable/index.mdx
shairez: cool thanks
merging

PR #2126, opened 2022-11-14
fix: argument of type 'unknown' is used in useStyles$

LiKang6688:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [x] Docs / tests

# Description

I started meeting the issue by creating a doc site starter app.
Then I found that inconsistent coding happens in other places as well.

# Use cases and why

Head:
Screenshot 2022-11-14 at 01 11 06

Current:
the type issue is solved by importing css or scss using ?inline

# Checklist:


manucorporat:
Love this change! it will actually be only right way in Vite 4! so good to be on top of it already! Thank youu!

PR #2130, opened 2022-11-14
feat: add playwright support

manucorporat:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests


starters/features/playwright/playwright.config.ts
```diff
@@ -0,0 +1,40 @@
...
+    {
+      name: 'chromium',
+      use: {
+        ...devices['Desktop Chrome'],
+      },
+    },
+    // {
+    //   name: 'webkit',
+    //   use: {
+    //     ...devices['Desktop Safari'],
```
Mouvedia, line 29: What was the problem you encountered with Safari?

PR #2157, opened 2022-11-15
fix(eslint-plugin-qwik): handle linting tuple types properly for valid-lexical-scope rule

jessezhang91:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
`TupleType` does not have a `symbol` property which caused the line: `const symbolName = type.symbol.name;` to throw an error.

When a tuple type is used, `eslint-qwik-plugin` would error out with:
```
TypeError: Cannot read properties of undefined (reading 'name')
Occurred while linting [REDACTED].tsx:1
Rule: "qwik/valid-lexical-scope"
    at r3 ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:69352)
    at wdr ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:70273)
    at r3 ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:70109)
    at Pdr ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:68153)
    at Adr ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:68005)
    at [REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:67003
    at Array.forEach (<anonymous>)
    at p ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:66124)
    at Array.forEach (<anonymous>)
    at p ([REDACTED]/node_modules/eslint-plugin-qwik/index.js:424:67171)
```

I have updated the code to check if it's a tuple type and check the inner types (retrieved using `checker.getTypeArguments(...)`). I found this out just by brute-force grepping of the Typescript codebase and have no idea if this is the right way to do it but it works! 😅

With this change, I see these eslint errors now:
```
export type Props = {
  foobar: [string, number, Function],
};
```

```
  32:18  error  Identifier ("props") can not be captured inside the scope (useWatch$) because "props.foobar" is an instance of the "Function" class, which is not serializable. Check out https://qwik.builder.io/docs/advanced/optimizer for more details   
```

# Use cases and why

n/a

# Checklist:

### DIFF
--- a/packages/eslint-plugin-qwik/qwik.unit.ts
+++ b/packages/eslint-plugin-qwik/qwik.unit.ts
@@ -269,6 +269,19 @@ export default component$(() => {
   };
 });
 `,
+      `
+        import { component$ } from "@builder.io/qwik";
+        
+        export interface Props {
+          serializableTuple: [string, number, boolean];
+        }
+
+        export const HelloWorld = component$((props: Props) => {
+          return (
+            <button onClick$={() => props.serializableTuple}></button>
+          );
+        });
+      `,
     ],
     invalid: [
       {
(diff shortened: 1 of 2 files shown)


jessezhang91:
Yup will do! Somehow I completely missed that there were tests!

PR #2158, opened 2022-11-15
docs: Use useClientEffect$ in debounced input example

boohba:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description
I doubt anyone wants to debounce on the server. Therefore, `useClientEffect$` is more appropriate than `useWatch$`.

# Checklist:


manucorporat:
This is not correct! `useClientEffect$()` is almost never the correct solution, been working on the docs to reflect this more.

useClientEffect$() comes with a big overhead to run something in the browser, this example does not require to run any code eagarly in the browser.

manucorporat:
useWatch runs before render, it renders first in the server, then in the client it has not need to run again.

PR #2192, opened 2022-11-20
feat: add human readable URL parsing/encoding of JS data structures

mhevery:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

This is a prerequisite to have human-readable rich data encoded in URL search params.

Example: `/contact/123?name=John&age=30` 
If `/contact/[:id]` then we should expect to get: `{id: 123, name: "John", age:30}`

Notice that the above parser correctly guessed the types of the extracted JS. In addition to flat data, we should be able to encode complex data of object literals and arrays. Example:

```typescript
testEncoding('example', () => {
  equal(
    encodeParamsToUrl('/contact/[contactId]/', {
      contactId: 123,
      dob: new Date(Date.parse('1980-01-01 GMT')),
      predicate: { name: 'John', age: 42, alias: ['Jon', 'Johnny'] },
    }),
    '/contact/123/?dob=@315532800000&predicate={age=42&alias=[Jon&Johnny]&name=John}'
  );
});
```

In the above case the URL can be parsed and it will correctly retrieve the correct data:
```typescript
{
      contactId: 123,
      dob: new Date(Date.parse('1980-01-01 GMT')),
      predicate: { 
        name: 'John', 
        age: 42, 
        alias: ['Jon', 'Johnny'] 
     },
}
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/url-params-encoding.ts
+++ b/packages/qwik-city/runtime/src/url-params-encoding.ts
@@ -1,0 +1,262 @@
+/**
+ * Encode a params into a URL query string.
+ *
+ * The encoding is done to remain human readable, but at the same time can deal with complex types
+ * such as `Date`, `Object` and `Array`.
+ *
+ * @public
+ * @param template - URL path which may contain slugs: `/foo/:bar`
+ * @param params - Key/value map of parameters to encode into the URL
+ * @returns URL string
+ */
+export function encodeParamsToUrl(template: string, params?: Record<string, any>): string {
+  const keys = Object.keys(params || {}).sort();
+  let out = template
+    .split('/')
+    .map((part) => {
+      if (isSlug(part)) {
+        let key = part.slice(1, -1);
+        if (key.startsWith('...')) {
+          key = key.slice(3);
+          return encodeValue(get(key));
+        } else {
+          return encodePathSegment(get(key));
+        }
+      } else {
+        return part;
+      }
+    })
+    .join('/');
+
+  out += keys.length ? '?' + encodeKeyValuePairs(params, keys) : '';
+  return out;
+
+  function get(key: string): any {
+    if (params) {
+      const index = keys.indexOf(key);
+      if (index !== -1) {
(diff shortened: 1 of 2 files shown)


packages/qwik-city/runtime/src/url-params-encoding.ts
```diff
@@ -0,0 +1,262 @@
...
+  const keys = Object.keys(params || {}).sort();
+  let out = template
+    .split('/')
+    .map((part) => {
+      if (isSlug(part)) {
+        let key = part.slice(1, -1);
+        if (key.startsWith('...')) {
+          key = key.slice(3);
+          return encodeValue(get(key));
+        } else {
```
hamatoyogi, line 22: maybe a bit nit picky but you can remove all `else` keywords here as you have `return`s for both else clauses...

packages/qwik-city/runtime/src/url-params-encoding.ts
```diff
@@ -0,0 +1,262 @@
...
+function isOpen(ch: number): boolean {
+  return ch === CharCode.openBrace || ch === CharCode.openBracket;
+}
+
+function isSep(ch: number): boolean {
+  return ch === CharCode.ampersand || ch === CharCode.equal;
+}
+
+function decodeValue(value: string): any {
+  debugger;
```
reemardelarosa, line 133: Can we cleanup this debugger?

zanettin:
A little bit dust on it but good as new 😄 @mhevery what is the command? 👼

mhevery:
This PR I think is a dead end, but I would like to keep the source around as we may use it...

wmertens:
Note that [jsurl2](https://www.npmjs.com/package/jsurl2) has the same aim and has a bunch of tests and is used in production for years now

wmertens:
Closing because it's a dead end, keeping the branch around.

PR #2194, opened 2022-11-20
fix(qwik city cloudflare pages adapter): added all q-data.json files …

### LINKED ISSUE #2193: Cloudflare Pages _routes.json generation stops caching of files.
### Qwik Version

v0.13.3 (City: v0.0.123)

Firefox

_No response_

Qwik City

### Expected Behaviour

q-data.json files should be cached.

### Actual Behaviour

There is no caching of any request for q-data.json files while deployed on cloudflare.

### Additional Information

This is solved by adding the "/*/q-data.json" exclusion to the _routes.json file, although there is still a noticeable delay between pressing a link and the page transitioning.

EamonHeffernan:
…to the generated _routes.json

fix #2193

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Added the rule "basePathname + '*/q-data.json'" to the code that generates the _routes.json file if there is not an existing _routes.json  file in the public folder.

# Use cases and why

q-data.json files are otherwise not cached while hosting on cloudflare pages.

# Checklist:

### DIFF
--- a/packages/qwik-city/adaptors/cloudflare-pages/vite/index.ts
+++ b/packages/qwik-city/adaptors/cloudflare-pages/vite/index.ts
@@ -40,7 +40,11 @@ export function cloudflarePagesAdaptor(opts: CloudflarePagesAdaptorOptions = {})
         const routesJson = {
           version: 1,
           include: [basePathname + '*'],
-          exclude: [basePathname + 'build/*', basePathname + 'assets/*'],
+          exclude: [
+            basePathname + 'build/*',
+            basePathname + 'assets/*',
+            basePathname + '*/q-data.json',
+          ],
         };
         await fs.promises.writeFile(routesJsonPath, JSON.stringify(routesJson, undefined, 2));
       }


manucorporat:
Amazing job! great fix!

packages/qwik-city/adaptors/cloudflare-pages/vite/index.ts
```diff
@@ -40,7 +40,11 @@ export function cloudflarePagesAdaptor(opts: CloudflarePagesAdaptorOptions = {})
         const routesJson = {
           version: 1,
           include: [basePathname + '*'],
-          exclude: [basePathname + 'build/*', basePathname + 'assets/*'],
+          exclude: [
+            basePathname + 'build/*',
+            basePathname + 'assets/*',
+            basePathname + '*/q-data.json',
```
petebacondarwin, line 46: This is not really the correct resolution here.
Adding these `q-data.json` paths to the `exclude` means that Pages will "never" make a request to the dynamic "Pages Function" to generate the data for a route.
This may work when using SSG if the `q-data.json` files are also generated alongside the `*.html` files, but that requires that the data is static and never changes.
In practice I would expect that the data might change over time and so statically generating all of it is not correct.

Instead of the approach here, each route should be able to set its own caching expiration for data requests via headers on the response. This way each endpoint can be cached for a suitable amount of time that appropriate for how volatile that data is.

Using response headers means that each layer of the network can cache as appropriate - whether that is Cloudflare's cache, an intermediate network cache, or the browser (or service worker).

petebacondarwin: The symptom of this, in case anyone is experiencing it, is that the q-data.json requests are always 404 responses.

We had to workaround this problem by specifying a _routes.json file in our project manually that does not exclude these data requests.

Issue #2202, opened 2022-11-20
qwik-city: add events to be able to react to the client side navigation


jwickers:
### Is your feature request related to a problem?

It would be nice to be able to react to the client-side navigation as currently, the page does not give any indication of loading which can lead to a bad user experience when trying to load a slow page, or on a slow network.

This should react when using the `Link` component or the `useNavigate` hook.

### Describe the solution you'd like

Simple events could be dispatched on navigation start and end.

This could then be used in the App layout, where navigation loader could be a top progress bar to indicate something is loading:

```js
  useClientEffect$(() => {
    console.log("adding document event listeners ...");
    const onNavStart = (e:Event) => {
      console.log("qnavigatestart event", e);
      document.getElementById("navigationLoader")?.classList.add("navigating");
    }
    const onNavEnd = (e:Event) => {
      console.log("qnavigateend event", e);
      document
        .getElementById("navigationLoader")
        ?.classList.remove("navigating");
    }
    document.addEventListener("qnavigatestart", onNavStart);
    document.addEventListener("qnavigateend", onNavEnd);
    return () => {
      console.log("removing document event listeners ...");
      document.removeEventListener("qnavigatestart", onNavStart);
      document.removeEventListener("qnavigateend", onNavEnd);
    }
  });
```

### Describe alternatives you've considered

Using a store or signal to store if the navigation is on-going, but AFAICT this did not work well due to render buffering.
Raw events would also be easier to integrate with other libraries.

manucorporat:
Should be possible now to track useLocation() isNavigating, to know when it's actively navigating!

jwickers:
Yes, I saw this was added when I looked at the code source today after the update. 
This could work for some use cases, will have to test it.

Not sure if there is a good way to capture start / end for UI updates and potentially prevent navigation just using this.

Currently, there is another issue with the URL not being updated when using `navigate` though.

PR #2203, opened 2022-11-20
feat: add qnavigatestart and qnavigateend events #2202

jwickers:

# What is it?

- [X] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Implement the events for issue #2202

# Use cases and why

- makes better UX possible as developpers can now hook up to those events and provide UI feedback like a loader, message if the load is long, etc ...

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/client-navigate.ts
+++ b/packages/qwik-city/runtime/src/client-navigate.ts
@@ -89,6 +89,23 @@ const scrollToHashId = (doc: Document, hash: string) => {
   return elm;
 };
 
+export const dispatchNavigationStartEvent = (data: RouteNavigate) => {
+  if (typeof document !== 'undefined')
+    document.dispatchEvent(
+      new CustomEvent('qnavigatestart', {
+        detail: data,
+      })
+    );
+};
+export const dispatchNavigationEndEvent = (data: RouteNavigate) => {
+  if (typeof document !== 'undefined')
+    document.dispatchEvent(
+      new CustomEvent('qnavigateend', {
+        detail: data,
+      })
+    );
+};
+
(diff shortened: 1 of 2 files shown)


manucorporat:
While i like the idea, this needs a bit of design work before get merged! this would be the first event based API to it sets a precedence, also it global, dispatched in the document, which is not ideal... I agree this is a problem to solve though

jwickers:
Yes, normally I'd also prefer using a store, like the context used by useNavigate for example, and this is what sveltekit does for example https://kit.svelte.dev/docs/modules#$app-stores-navigating

But I couldn't make that work in qwik because I think of the DOM update buffering causing the DOM not to react in real time in that case until the navigation is finished.
If there was a way to expose a store with real-time reactivity this would be an alternative.

Right now using global events seems to be the best way to implement this.

jwickers:
Also I think that using the events we could also make the navigationstart cancellable, so similar to the beforeunload event one could listen to that and on some pages ask user for confirmation before leaving and call prevent default.
In that case qwik could abort the client side navigation.

I don't think this is as easily doable with a store?

zanettin:
Hi @jwickers 👋 
Sorry for the silence on this PR 🙈 i really like the idea and i guess many devs coming from other frameworks are used to these events for showing loaders, etc.

@manucorporat : do you have smth on the radar/roadmap for this or a rough idea of how this could be implemented? also happy to give a hand for this feature ✋ 😄

gioboa:
@jwickers is this PR closed by #3244?

jwickers:
I'm not using this PR as we did something else that works well enough using the current APIs, so yeah I'll close it.

gioboa:
Thanks 👍

PR #2208, opened 2022-11-20
feat: allow recursive class definitions

wmertens:
+ merge class and className

This allows easy forwarding of class props.

# What is it?

- [x] Feature / enhancement

# Description

class and className now accept recursive objects and both are honored.

# Checklist:


wmertens:
I couldn't find a test for the non-SSR className merging but I did a search for className and I think I caught all uses.

manucorporat:
This is looking good! we need to add some tests for the vdom render, maybe some e2e too (i guess we can do that one later)

manucorporat:
Also, not sure i like the idea of merging className and class, it can explode in some many ways, like fine grained updates with signals

wmertens:
@manucorporat can you give an example of what will break?

Also note that right now, class and className overwrite eachother, whichever is last wins, so I think that's worse.

PR #2230, opened 2022-11-21
feat: allow recursive class definitions

wmertens:
My previous PR #2208 closed by accident when I pushed main.

# What is it?

- [x] Feature / enhancement

# Description

class and className now accept recursive objects and both are honored.

# Checklist:


packages/qwik/src/core/render/dom/visitor.ts
```diff
@@ -1001,23 +1001,27 @@ export const setProperties = (
       continue;
     }
 
+    if (prop === 'className') {
```
manucorporat, line 1004: why was this condition moved early? it can be problematic

Issue #2290, opened 2022-11-26
[✨] Parse query params in RequestEvent


mlynch:
### Is your feature request related to a problem?

For API handlers, `GET` query parameters aren't parsed and provided on the `RequestEvent` argument.

### Describe the solution you'd like

Would be really nice to have those parsed like `req.query` in Express.js

### Describe alternatives you've considered

Parsing URL manually

adamdbradley:
Related https://github.com/BuilderIO/qwik/pull/2352

PR #2293, opened 2022-11-27
feat: parse request query string into a dictionary

tuoitrevohoc:
…ring into a dictionary

re #2290

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

For API handlers, GET query parameters aren't parsed and provided on the RequestEvent argument.

# Use cases and why

After the change, query parameters can be accessed from 

```
request.query['fieldname']
```

# Checklist:


adamdbradley:
One API which @mhevery and I talked about was instead of the `query` values being either a string or an array of strings, it would instead have a `getAll()` added. So every property would also be the first (or only) value for that key. And if the developer knows there can be multiple, they'd instead use `query.getAll('key')`.

```typescript
// https://example.com/api?food=pizza&topping=pepperoni&topping=bacon

const food = query.food; // pizza
const topping = query.topping; // pepperoni
const toppings = query.getAll('topping'); // ['pepperoni', 'bacon']
```

Thoughts?

jwickers:
> One API which @mhevery and I talked about was instead of the `query` values being either a string or an array of strings, it would instead have a `getAll()` added. So every property would also be the first (or only) value for that key. And if the developer knows there can be multiple, they'd instead use `query.getAll('key')`.

This would be in line with `url.searchParams.get` which also returns a single value and in general the developer would know if a value is meant to be used as one or the other.
It's also nice because you can have an array even if a single value was given.
But it also kinda hides the ability to account for bad/unexpected inputs.

Looking at https://nodejs.org/api/querystring.html which the PR used it would set them as the actual needed type from whatever the query string was instead so I assume would be typed as `string | string[]` meaning you actually have to check the type of every param and decide what to do with that which may be a safer approach.

I think `getAll` is better DX, as it directly gives you an array if you want to deal with multiple values, can be checked with `array.length` so runtime validation is easy, don't require to test `isArray` then convert to array.

nnelgxorz:
I like the getAll api. I think most of the time you're looking for a single value, so making everyone check whether they have a single value or an array for every query param seems unnecessary. Better to explicitly ask for the array if you want it.

tuurbo:
#1759

manucorporat:
I would prefer `get()` and `getAll()`, i dont wanna deal with some user doing:
```
?getAll=value
```

adamdbradley:
@tuoitrevohoc I think that's a pretty good plan, but one thing we want to make sure is that we can maintain the types if we can. Like keeping a number a number, whereas URLSearchParams turns everything into a string. It's a pretty simple API, so thinking we can just reimplement URLSearchParams. Thoughts?

tuoitrevohoc:
I don’t think we should ’detect' if the query is a number or a string. unless we have api like

’query.getNumber(“param”)’

manucorporat:
yeah! i think we should not try to be too smart inferring type, query is a string, now it's up to developers to interpret that string as a number or not.

To give an example, if we infer that `price=0.1`, it will generate the wrong number, since floating aritmatic can not represent the value, and we are effectively losing information in between.

adamdbradley:
No i don't mean auto-inferring the types, but rather leaving in the possibility that we can maintain the types the developer chose, using some of the work we've been discussing around generating route types. So when the developer sets a certain param is a number, the `get('key')` already knows it's a number and the runtime returns the number. This would be in contrast to the dev always having to remember the different types for each param throughout the site. If we're already using typescript, I'd be nice to also provide this for URL params too. However, this is just an idea to maybe work on for another day. 

Regardless, I think just going with the URLSearchParams API is is a good idea, and possibly provide a `getNumber()` for convenience.

adamdbradley:
Related https://github.com/BuilderIO/qwik/pull/2352/files

adamdbradley:
Added URLSearchParams to qwik's serializer so we can better use it directly.

adamdbradley:
Thanks for spearheading this @tuoitrevohoc! We've made some modifications and applied them in this PR! Thanks again https://github.com/BuilderIO/qwik/pull/2385

Issue #2296, opened 2022-11-28
[🐞] Inconsistent handling of some (pseudo)boolean attributes like draggable in SSR vs CSR


jwickers:
### Which component is affected?

Qwik Runtime

### Describe the bug

A link with `draggable="false"` correctly renders with `draggable="false"` in SSR but renders as `draggable="true"` in CSR.
A link with `draggable={false}` incorrectly renders with no `draggable` attribute in SSR but renders correctly as `draggable="false"` in CSR.

Not only is this confusing behavior, as a result there is no way to render consistently in both SSR and CSR.

I understand that boolean attributes usually render with "truthy" values, however in this case the issue is that `draggable` is not actually a boolean attribute, but an enumerated one (see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable)

This bug might affect other attributes?

### Reproduction

https://stackblitz.com/edit/qwik-starter-rb5cxd

### Steps to reproduce

Run `npm install` then `npm start`.

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @builder.io/qwik: 0.14.1 => 0.14.1 
    @builder.io/qwik-city: 0.0.127 => 0.0.127 
    vite: 3.2.4 => 3.2.4
```

manucorporat:
Feels like a type issue! no? the values it should accept should be `"true"` or `"false"` as string, not boolean

jwickers:
Never mind I think I found it, will make a PR for this one line fix.

jwickers:
That fixes the CSR case for me which was the most annoying. 
I don't know if you want to look at the type issue here so that this only accepts `string`?

jwickers:
Awesome, if yours gets released soon I'll test it in my project.

PR #2298, opened 2022-11-28
fix: draggable must be set as attribute (#2296)

jwickers:
Because draggabe is also a property on the element, it must be set as
an attribute instead of a property as it is an enumerable and not a true
boolean.

# What is it?

- [X] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

The draggable attribute is also an attribute on the element in the DOM, but in HTML this is not a true boolean value but an enumerable instead.

See developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable

In the JSX we then set is as the string `"false"` for elements like A or IMG that are draggable by default, or `"true"` for other elements that aren't.
In SSR this is currently handled correctly but nodes that are added during CSR need the exception handling rule so that the behavior is consistent.

# Use cases and why

- 1. Setting `draggable="false"` on elements like A or IMG

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/dom/visitor.ts
+++ b/packages/qwik/src/core/render/dom/visitor.ts
@@ -848,6 +848,7 @@ export const PROP_HANDLER_MAP: Record<string, PropHandler | undefined> = {
   form: forceAttribute,
   tabIndex: forceAttribute,
   download: forceAttribute,
+  draggable: forceAttribute,
   [dangerouslySetInnerHTML]: setInnerHTML,
   innerHTML: noop,
 };


manucorporat:
qwik prioritizes properties over attributes, since it's most performance and more close to react semantic!

I have a different fix: https://github.com/BuilderIO/qwik/pull/2299
involves different types, and fix in SSR!

jwickers:
Awesome, this is even better then.

manucorporat:
Merged the other PR! but added u as co-author! :) 
https://github.com/BuilderIO/qwik/commit/ff60963961c6559d64b85bef67f5b440cdfbd606

Thank you so much for your PR and always amazing repo cases :)

Issue #2282, opened 2022-11-26
[📖] Add document for Hooks


Nefcanto:
### Suggestion

I searched your docs and could not find a page explaining hooks and their rules. For example, is it necessary to be prefixed with `use` just as the same as the react? Is it essential to use hooks only on the top level of a component?

I just found [this](https://qwik.builder.io/tutorial/composing/use/) tutorial.

But I think it's better to have a hooks page too specifically if Qwik hooks have differences from React hooks.

the-r3aper7:
i will start working on it

nnelgxorz:
I think all the hooks (mostly) there in the docs, but spread out through the Component API sections (Lifecycle, resource, context, etc)

the-r3aper7:
i think we should have separate section named `use-` hooks. WDYT?

nnelgxorz:
I think the docs are always a work in progress and if you have a good idea for improving them, go for it! 😄

the-r3aper7:
yeah i will make PR very soon busy in college right now.

manucorporat:
We have some docs here: https://qwik.builder.io/docs/components/lifecycle/#use-method-rules

but i agree! this deserve its own section, maybe adding something new in advanced ?

the-r3aper7:
#2301

PR #2301, opened 2022-11-28
docs: added useOn hooks

the-r3aper7:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

This should solve #2282 

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


shairez:
I think this should be part of the "Lifecycle" page no?

Maybe we should rename it in the menu to "Lifecycle (hooks)" to make it clearer

the-r3aper7:
cool I will make the changes by tonight.

shairez:
thanks buddy!

shairez:
on second thought - not sure it's related to lifecycle
it looks like utils for adding event listeners in general

@manucorporat WDYT?

manucorporat:
wondering if this is events

manucorporat:
for me this are just events, in fact it's the same as doing: `<div onClick`, but programatically, so it can be composed with hooks

the-r3aper7:
so, where should we add this?

the-r3aper7:
@shairez  @manucorporat i have made the changes see it's matches the context of events page.

zanettin:
@manucorporat  @shairez just tested the changes locally and they LGTM 👍 thanks a lot @the-r3aper7 🙏  hope we get this merged soon since it is a great addition for the docs imo.

shairez:
thanks @the-r3aper7 and @zanettin !
merging

PR #2310, opened 2022-11-29
chore: remove className

manucorporat:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests


shairez:
it's still there... but we're just removing it from the docs etc right?

manucorporat:
yes! only docs

PR #2348, opened 2022-12-02
refactor: use class prop without casting to any

langbamit:
# What is it?

- [ x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Remove casting when use class prop

# Use cases and why

Before https://github.com/BuilderIO/qwik/pull/2230

```jsx
// Component.jsx
import clsx from 'clsx'

<div class={clsx(["docs", props.class])}>Some text</div>
```

 After https://github.com/BuilderIO/qwik/pull/2230:
```jsx
// Component.jsx
// no more external package for class merging

<div class={["docs", props.class] as any}>Some text</div>
```
But if my `props.class` is a signal, the class result always `docs untrackValue`, after this PR type-check will complain and I can fix easier.

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/execute-component.ts
+++ b/packages/qwik/src/core/render/execute-component.ts
@@ -9,11 +9,10 @@ import type { RenderContext } from './types';
 import { ContainerState, intToStr } from '../container/container';
 import { fromCamelToKebabCase } from '../util/case';
 import { qError, QError_stringifyClassOrStyle } from '../error/error';
-import { seal, qDev } from '../util/qdev';
+import { seal } from '../util/qdev';
 import { EMPTY_ARRAY } from '../util/flyweight';
 import { SkipRender } from './jsx/utils.public';
 import { handleError } from './error-handling';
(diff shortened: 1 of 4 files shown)


packages/qwik/src/core/render/execute-component.ts
```diff
@@ -119,12 +121,14 @@ export const pushRenderContext = (ctx: RenderContext): RenderContext => {
...
 
-export const serializeClass = (
-  obj: string | { [className: string]: boolean } | (string | { [className: string]: boolean })[]
-): string => {
+export const serializeClass = (obj: ClassList): string => {
   if (!obj) return '';
   if (isString(obj)) return obj.trim();
 
+  if (qDev) {
+    verifySerializable(obj);
```
langbamit, line 129: Is it necessary?

manucorporat: probably not! we are learning that this check is better done at the `jsx()` level, so when it crashes the stacktrace is actually helpful for the developer!

PR #2355, opened 2022-12-03
Don't generate _routes.json for cloudflare

zhuhaow:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Currently, a `_routes.json` is generated by Cloudflare Pages adapter to exclude static assets.

However, this is not necessary since the Pages by default will exclude all static assets (https://developers.cloudflare.com/pages/platform/functions/routing/#function-invocation-routes)

Besides, currently the generated `_routes.json` is not correct since the service worker script is not excluded.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/qwikcity/adaptors/cloudflare-pages/index.mdx
+++ b/packages/docs/src/routes/qwikcity/adaptors/cloudflare-pages/index.mdx
@@ -85,34 +85,6 @@ export { onRequest } from '../server/entry.cloudflare-pages';
 - [Cloudflare Pages Functions](https://developers.cloudflare.com/pages/platform/functions/)
 - [Function Invocation Route Config](https://developers.cloudflare.com/pages/platform/functions/function-invocation-routes/)
 
-### Cloudflare Pages Function Invocation Routes
-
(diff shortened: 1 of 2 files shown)


manucorporat:
I understand the opposite:
>However, once you add Functions on a Pages project, all requests by default will invoke your function. In the spirit of continuing to offer unlimited free static requests, you can exclude these static routes by creating a _routes.json file.

zhuhaow:
> I understand the opposite:
> 
> > However, once you add Functions on a Pages project, all requests by default will invoke your function. In the spirit of continuing to offer unlimited free static requests, you can exclude these static routes by creating a _routes.json file.

And the next line reads:

> This file will be automatically generated if a functions directory is detected in your project when you publish your project with Pages CI or Wrangler.

I can see the document here is super confusing, just as the doc for most other CF products, but here the actual rule is: if there is no `_routes.json` file, we serve static first, then fallback to functions, if there is a `_routes.json`, follow the config there.

zhuhaow:
😓 You are right.

I got it wrong since Qwik City is very considerate, generates all known static paths, and serves them correctly, thus everything still works fine after I remove `_routes.json`.

Thanks for the great work.

PR #2359, opened 2022-12-04
fix(cli): build.types trigger condition

utherpally:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
After this changed https://github.com/BuilderIO/qwik/commit/7d6c810b695e564dc6171bbeba29ecd409cc92dc, qwik CLI never run `build.types` script

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/cli/build/run-build-command.ts
+++ b/packages/qwik/src/cli/build/run-build-command.ts
@@ -61,7 +61,7 @@ export async function runBuildCommand(app: AppCommand) {
 
   let typecheck: Promise<Step> | null = null;
 
-  if (buildTypes && buildTypes.startsWith('tsc')) {
+  if (buildTypes && buildTypes.startsWith(`${pkgManager} run tsc`)) {
     let copyScript = buildTypes;
     if (!copyScript.includes('--pretty')) {
       // ensures colors flow throw when we console log the stdout


manucorporat:
Maybe we can remove the condition `.startWith` entirely, i am not sure the new condition is correct

zanettin:
Hi @utherpally 👋  so sorry for the late reply to this PR 🙈 and thank you for creating it 🙏 

just had a look at it and i am confused tbh 😆 

### current state
Having a look at these lines: https://github.com/BuilderIO/qwik/pull/2359/files#diff-b89049f3a8803c4a6647b7ae179f73201c4c2cfd03adc1bd4530185be22a529cL17-L32
We check in the `getScript()` fn, if the given name (in our case `build.types` L32) exists in the `package.json`.
This is the case, so we return for pnpm users `pnpm run build.types`. So the value of our variable `buildTypes` is now set to `pnpm run build.types`.

Later in the code (L64) we check the value again and would expect, that it starts with `tsc` or in your version with `pnpm run tsc` which can't be the case 😄 it is still `pnpm run build.types` and not the value of this key from the pkg.json file. 

### what i would do
L23: add
```
const BUILD_TYPES_KEY = 'build.types';
```
L32: change to
```
const buildTypes = getScript(BUILD_TYPES_KEY);
```
L64: change to 
```
if(buildTypes && pkgJsonScripts[BUILD_TYPES_KEY].startsWith('tsc') { 
  ... 
}
```
### quick explanation
- we store a const of the key in the pkg.json file which executes the types def
- then we try to get the cmd incl. packagemanager to display in a log
- next we check if we have a key found in pkg.json AND if the **value** of this key starts with `tsc`

hope i've missed nothing 😅 

once again a huge thanks for your debugging work! really apprechiated 🙏

zanettin:
Hi @utherpally 
Issue was fixed in the meantime with https://github.com/BuilderIO/qwik/pull/3005.
Thanks again for creating this PR and hope to see others from you 🙏

PR #2375, opened 2022-12-04
feat: add config for dev inspector

zanettin:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

This PR makes the new dev inspector configurable by giving the devs the options to disable the inspector or assign different hot keys than the default `Alt` key.

# Use cases and why

- More flexibility especial if the `Alt` key is already used on an app

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/plugin.ts
@@ -52,6 +52,10 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
     transformedModuleOutput: null,
     vendorRoots: [],
     scope: null,
+    inspectorConfig: {
+      enabled: true,
+      hotKeys: ['Alt'],
+    },
   };
 
   const init = async () => {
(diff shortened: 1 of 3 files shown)


manucorporat:
This is amazing! It was in our todo list! So happy you are taking in this! 

I wonder if we should instead have a new option called devTools: and inside we can have clickToSource for example:

disables completely:
```
devTools: {
  clickToSource: false,
}
```

configure:
```
devTools: {
  clickToSource: {
    keys: [“control”]
  }
}
```

Thoughts?
Most likely we will keep adding more dev tools, and it would be nice to have everything in the same place to enhance discoverability

zanettin:
@manucorporat thanks a lot for your feedback 🙏 wasn't even able to test the changes locally 🤣 

> I wonder if we should instead have a new option called devTools: and inside we can have clickToSource
ah yes! of course! great idea. will change it accordingly 👍  

thanks for the early feedback 🙏

zanettin:
@manucorporat changes applied 👍

manucorporat:
love it!

manucorporat:
merged! great job man!

Issue #2351, opened 2022-12-02
[🐞] trailing slash added to route params


robisim74:
### Which component is affected?

Qwik City (routing)

### Describe the bug

When you have a parameter in the root, not followed by other segments, a trailing slash is added to the parameter:

`{ params: { id: '1234/' } }`

### Reproduction

https://stackblitz.com/edit/qwik-starter-zgnacc

### Steps to reproduce

After the app has started, click on `Go to id 1234` and look in the Terminal (Ok, just hover the mouse pointer over the button without pressing it... what an amazing improvement is this?!)

Thanks

### System Info

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    @builder.io/qwik: 0.15.0 => 0.15.0 
    @builder.io/qwik-city: 0.0.128 => 0.0.128 
    vite: 3.2.4 => 3.2.4
```

adamdbradley:
Related https://github.com/BuilderIO/qwik/pull/2352

adamdbradley:
Fixed in https://github.com/BuilderIO/qwik/pull/2385

PR #2385, opened 2022-12-06
feat: standardize params/query api

adamdbradley:
- Standardizes `params` and `query` for Request handlers, Document Head, and useLocation()
- `params` is a `Record<string,string>` (pathname slugs)
- `query` is an instance of `URLSearchParams` (search querystrings)
- Catchall routes will not have a trailing slash, even if there is one


adamdbradley:
Closes #2351
Closes #2290
Closes #2293

packages/qwik-city/runtime/src/types.ts
```diff
@@ -287,9 +287,18 @@ export interface RequestEvent<PLATFORM = unknown> {
   response: ResponseContext;
   url: URL;
 
-  /** URL Route params which have been parsed from the current url pathname. */
+  /**
+   * URL Route params which have been parsed from the current url pathname.
+   * Use `query` to instead retrieve the query string search params.
+   */
   params: RouteParams;
```
manucorporat, line 294: maybe let's rename RouteParams to `PathParams`, i am refering to the interface

packages/qwik-city/runtime/src/types.ts
```diff
@@ -287,9 +287,18 @@ export interface RequestEvent<PLATFORM = unknown> {
...
 
+  /**
+   * URL Query Strings (URL Search Pararms).
```
manucorporat, line 297:
```suggestion
   * URL Query Strings (URL Search Params).
```

saeedhbi:
**useLocation** hook query doesn't have anything in production mode. It just has a **{}** value. I'm not sure what the problem is but it's working in dev mode. All _trailingHash_ and configurations are the same.

PR #2392, opened 2022-12-07
feat: add static file handler to node middleware

DustinJSilk:
# What is it?

- [X] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

I've added a first draft function to the node middleware that allows us to stream static files.

Qwik already handles routing and does a lot of upfront optimisations. There should not be a need to rely on large, slow libraries like express or even fastify to serve static files. These libraries also come with a lot of unnecessary dependencies.

With this change we can start deploying apps on raw node http servers.

# Use cases and why

- 1. Smaller builds with fewer dependencies
- 3. Quicker startup times with prebuilt static paths

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/api.md
+++ b/packages/qwik-city/middleware/node/api.md
@@ -16,6 +16,7 @@ import type { ServerResponse } from 'node:http';
 export function createQwikCity(opts: QwikCityNodeRequestOptions): {
     router: (req: IncomingMessage, res: ServerResponse, next: NodeRequestNextFunction) => Promise<void>;
     notFound: (req: IncomingMessage, res: ServerResponse, next: (e: any) => void) => Promise<void>;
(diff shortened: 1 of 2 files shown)


DustinJSilk:
An entry file, without express, can then look something like this:

```typescript
import { createQwikCity } from '@builder.io/qwik-city/middleware/node';
import qwikCityPlan from '@qwik-city-plan';
import { createServer } from 'node:http';
import render from './entry.ssr';

// Create the Qwik City express middleware
const { router, notFound, staticFile } = createQwikCity({
  render,
  qwikCityPlan,
});

const server = createServer();

server.on('request', (req, res) =>
  staticFile(req, res, () => {
    router(req, res, () => {
      notFound(req, res, () => {});
    });
  })
);

server.listen(process.env.PORT ?? 8080);
```

DustinJSilk:
Closing this as this code will get merged in with this PR: https://github.com/BuilderIO/qwik/pull/2395

PR #2395, opened 2022-12-07
feat: add Google Cloud Run integration

DustinJSilk:
This relies on [this PR](https://github.com/BuilderIO/qwik/pull/2392). You may see some the same code in both.

# What is it?

- [X] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

A starter and integration to easily get started with deploying apps to Google Cloud Run.

Firebase and Cloud Functions users may also use this since Cloud Functions is built on top of Cloud Run. This means users can get more control.

Cloud Functions / Firebase also uses slower 3rd party libraries. This only uses Qwik which should give a performance bonus.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


manucorporat:
Wow Dustin! You are becoming the integration master! Love this

manucorporat:
Actually! Some of the ideas here could be used for a vanilla-node starter without express

DustinJSilk:
Aahah ye i'm finding my rhythm!

So i've been thinking about it, it might be worth taking these ideas into a vanilla-node starter, and then the Cloud Run integration doesn't include an entry file, it's up to the user to add either express/vanilla/etc.

An extra step for the user but less duplication.

I'm happy to adjust it, what do you think? @manucorporat

manucorporat:
i like the idea of end products, like google cloud, developers think in this terms! if this PR is good, we can merge and iterate from there

DustinJSilk:
Great, lets get it in and iterate with feedback!

adamdbradley:
awesome!

PR #2396, opened 2022-12-07
fix(header.css): make header menu sticky. This effects the landing, m…

ShaggyDude:
…edia and showcase + darkmode

# What is it?

- [ *] Feature / enhancement
- [ *] Bug
- [ ] Docs / tests

# Description

Makes menu sticky to top on landing, media, and, showcase at all sizes and doesn't break darkmode

# Use cases and why

- 1. Menu should stick to the top

# Checklist:

### Please see [loom](https://www.loom.com/share/f9052ba35cb54d75b4cf8ea6064c1bdb)

### DIFF
--- a/packages/docs/src/components/header/header.css
+++ b/packages/docs/src/components/header/header.css
@@ -1,16 +1,7 @@
 .header-container {
   height: calc(var(--header-height));
-  margin-bottom: 6px;
-  background: var(--bg-color);
-  position: relative;
-  z-index: 100;
-}
-
-@media (min-width: 1024px) {
-  .header-container {
-    background: var(--bg-header-color);
-    backdrop-filter: blur(23px) saturate(4.5);
-  }
+  background: var(--bg-header-color);
+  @apply backdrop-blur-xl backdrop-saturate-200 sticky top-0 mb-2 z-20;
 }
 
 .fixed-header .header-container {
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/header/header.css
```diff
@@ -1,16 +1,7 @@
...
-  z-index: 100;
-}
-
-@media (min-width: 1024px) {
-  .header-container {
-    background: var(--bg-header-color);
-    backdrop-filter: blur(23px) saturate(4.5);
-  }
+  background: var(--bg-header-color);
+  @apply backdrop-blur-xl backdrop-saturate-200 sticky top-0 mb-2 z-20;
```
dario-piotrowicz, line 4: out of curiosity, isn't `@apply` a deprecated rule that shouldn't be used? 🤔

ShaggyDude: No, it isn't but it is recommended to just use inline classes instead but the current css isn't written the 'tailwind way'.
I would love to clean up the css further and make it all inline tailwind classes but that would be a larger effort and I'm not sure if the core team is 100% on board with doing it that way?

https://tailwindcss.com/docs/reusing-styles

Thanks,

adamdbradley: To that argument, I'd rather use just standard css.

ShaggyDude: I'd suggest just removing tailwind from the project for consistency then?

Issue #1477, opened 2022-09-26
Routing to a component with `onPost` is causing server dependencies to be loaded into the client


meenie:
### Qwik Version

0.9.0

### Operating System (or Browser)

Chrome

### Node Version (if applicable)

18

### Which component is affected?

Qwik City

### Expected Behaviour

When using the `<Link />` component to route to another component that has `onPost` (or any other `on___` methods) defined, the libraries imported and utilized within `onPost` _should not_ be imported into the client.

### Actual Behaviour

When using the `<Link />` component to route to another component that has `onPost` (or any other `on___` methods) defined, the libraries imported and utilized within `onPost` _are_ imported into the browser. If those libraries are node only, then it will break the application. It also bloats the memory of the browser having to load extra libraries that aren't even needed.

On a side note, when you directly load the route in the browser URL bar, then the libraries are _not_ imported. This only happens when you use `<Link />` to route to this component.

### Additional Information

You can reproduce the issue here: https://stackblitz.com/edit/qwik-starter-nmrofc?file=src%2Froutes%2Fflower%2Ffoo.ts,src%2Froutes%2Fflower%2Findex.tsx

When clicking on the button to route to the Flower, notice the console message that shows up in the browser. This should not happen.  If you load the `/flower` page directly, this message does not appear in the browser console.

cmbartschat:
Adding on here, even just a function defined within the same file as the `onGet` like so:

```
export const makeData = () => {
  // Logic here
};

export const onGet: RequestHandler<Data> = async () => {
  return makeData();
};
```

I'm seeing that the `makeData` function is being loaded on the client after following a `<Link>`. 

I'm even seeing this with the Express integration running `npm run build && npm run serve` which I'm pretty sure is a production build.

meenie:
I believe the issue here is that when you load the page with the `onGet()`, the server SSRs the page and no JS is loaded.  And even if Qwik needs to load the component, it's only loading the `component$` chunk that was created, so all is still fine because it still has yet to download the `onGet()`.  But when you use the `<Link />` component to go to another route, it loads the would route file, including the `onGet()`. And because of that, the browser tries to resolve any dependencies imported into the route for the `onGet()` to work and if any of those dependencies are Node-only, it will blow up the browser.

cayter:
[...]

Issue #1476, opened 2022-09-26
The contents of `useServerMount$()` are bundled with client side JS


meenie:
### Qwik Version

0.9.0

### Operating System (or Browser)

Chrome

### Node Version (if applicable)

18

### Which component is affected?

Qwik Optimizer (rust)

### Expected Behaviour

When you run `npm run build`, the contents of the `useServerMount$()` hook should not be present in the bundles created for the client side. The contents should only be present in the Node bundle.

### Actual Behaviour

When you run `npm run build`, the contents of the `useServerMount$()` are present in the client side JS bundles.

### Additional Information

You can reproduce here: https://stackblitz.com/edit/qwik-starter-ixah1m

Stop the server when it boots up and run `npm run build` and search for `find me`. You will notice that the code is both in `entry.express.mjs` (Good!) and also in one of the files under `./dist/build` (Bad!).

manucorporat:
Is it bad if it never load anyway?

meenie:
If it never loaded, then no, probably not. But if the file does get loaded and you are using a node-only library, that will break the app.

wtho:
@meenie does it get loaded and break for you?

meenie:
@wtho, when you use `qwik-city` to route to a component that is using `useServerMount$()`, it does break with this message:
```
Code(22): Components using useServerMount() can only be mounted in the server, if you need your component to be mounted in the client, use "useMount$()" instead
```

You can see with this example: https://stackblitz.com/edit/qwik-starter-zye1es?file=src%2Froutes%2Findex.tsx,src%2Froutes%2Fflower%2Findex.tsx

meenie:
@manucorporat 
> Is it bad if it never load anyway?

I've been thinking about this and I think it's still a bit of a security risk to expose backend code when you shouldn't need to.  Specifically, business logic that's a trade secret. Or exposing what libraries your app uses in the backend to potential hackers trying to find vulnerabilities.

manucorporat:
Fixed! https://github.com/BuilderIO/qwik/pull/2253

jwickers:
FYI I ran into the same issue but when running with the DEV server, I see this was fixed in the build so I ported a similar fix for the dev.

PR #2404, opened 2022-12-09
fix: server symbols should be stripped in the transform for clients

jwickers:

# What is it?

- [ ] Feature / enhancement
- [X] Bug
- [ ] Docs / tests

# Description

Server symbols were not stripped by the optimizer when generating the JS for the client.
Related to issues #1476 #1477

This adds the missing `stripExports` and `stripCtxName` to the transform when not in the `ssr` context, similar to how the transform in the `buildStart` was setup.
This also add the stripping of `onStaticGenerate` for both.

# Use cases and why

JS served to the client for client side rendering should not include any server side only code as it also pulls imports of server modules and causes a runtime error.

# Checklist:


jwickers:
I have this fix working for my case, one thing I'm not sure about here is that typescript is not happy with the `ctx.ssr` .
If anyone knows where that field comes from, what would be a cleaner way to detect the SSR or if casting is fine here?

jwickers:
Any idea @manucorporat ?

manucorporat:
Thank you so much for the PR!!

PR #2423, opened 2022-12-12
fix: preview server should use the configured outDir

dmitry-stepanenko:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
The SSR build that is compiled to a non-default folder cannot be used because the `outDir` for the ssr build is hardcoded.

# Use cases and why
To reproduce:
- run `npm create qwik@latest`
- use the following for the qwikVite plugin in `vite.config.ts`: `qwikVite({  ssr: { outDir: "the-other-dir" } })`
- run `npm run preview`
- notice an error stating `Unable to find output "server/entry.preview" module.`

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/vite-server.ts
+++ b/packages/qwik/src/optimizer/src/plugins/vite-server.ts
@@ -153,7 +153,7 @@ export async function configureDevServer(
 
 export async function configurePreviewServer(
   middlewares: Connect.Server,
-  opts: NormalizedQwikPluginOptions,
+  ssrOutDir: string,
   sys: OptimizerSystem,
   path: Path
 ) {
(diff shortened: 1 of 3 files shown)


packages/qwik/src/optimizer/src/plugins/vite-server.ts
```diff
@@ -153,22 +153,22 @@ export async function configureDevServer(
 
 export async function configurePreviewServer(
   middlewares: Connect.Server,
-  opts: NormalizedQwikPluginOptions,
+  ssrOutDir: string,
```
shairez, line 156: question: what is the reason not to pass all the other options and to replace it with only this one?

thanks

dmitry-stepanenko: `ssrOutDir` is not present within the options object, `opts` themselves were used only to grab `opts.rootDir`, which is not longer needed. 

And it doesn't feel reasonable to include `ssrOutDir` into the `NormalizedQwikPluginOptions`, because it already contains `outDir` property, which is resolved either as `clientOutDir` or `ssrOutDir`

shairez: thanks @dmitry-stepanenko !

PR #2463, opened 2022-12-18
docs: replace the entire parent value to update the store

Pika-Pool:
updating a nested store value directly doesn't reflect changes in the UI. Assign a new object to the parent value to trigger rerender

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Docs show an example where a nested store value is updated. Updating a nested store value directly doesn't reflect changes in the UI. eg:
```js
const store = useStore$({
  mouse: {x: 0, y: 0}
});

// ...
store.mouse.x =  1;
//...
```
The above would not trigger a rerender, and the UI still shows the value `0` for `mouse.x`

# Use cases and why

Instead, assign a new object to `store.mouse`:
```js
store.mouse = { x: 1, y: 0};
```
This would trigger a UI update

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/components/events/index.mdx
+++ b/packages/docs/src/routes/docs/components/events/index.mdx
@@ -78,8 +78,8 @@ export const EventExample = component$(() => {
     <button
       window:onScroll$={(e) => (store.scroll = window.scrollY)}
       document:onMouseMove$={(e) => {
-        store.mouse.x = e.x;
-        store.mouse.y = e.y;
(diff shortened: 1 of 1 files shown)


shairez:
thanks for the fix @Pika-Pool !

It could work by adding the `recursive` flag to the store... but I wonder if we should promote it as a practice in the docs...

shairez:
I'll merge this anyway, and we'll keep thinking about it :)

PR #2471, opened 2022-12-19
docs: fix example passing number as value for `onClick$`

Pika-Pool:
`onClick$` requires a function

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description
the `onClick$` property of a JSX element needs a function, which is executed when the element is clicked

in the example, a number was being passed, instead of a function that changes the state

# Use cases and why

a function that changes the state should be passed to `onClick$`, as expected from the example

# Checklist:


manucorporat:
Thank you for the fix!

PR #2521, opened 2022-12-26
Fix plugin scoping for namespaced packages

JanJakes:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

This fixes a bug introduced in https://github.com/BuilderIO/qwik/pull/2398.

# Use cases and why

For namespaced packages (such as `@org/package`), the `scopePrefix` added in https://github.com/BuilderIO/qwik/pull/2398 will contain a `/` which will result in an error:

```bash
[vite-plugin-qwik] ENOENT: no such file or directory, open '/.../@org/package-vite-plugin-qwik-q-manifest.json'
```

Additionally to the fix, I changed the prefix to a suffix so the `vite-plugin-qwik` part is the actual prefix and these files are colocated. If needed, I can drop that change.

All that said, package names are optional, which means that in some edge cases, parallel builds can still conflict. An alternative might be to use a hash of something (`rootDir` maybe?).

# Checklist:


manucorporat:
The PR is not passing! mind updating the tsconfig or not use replaceAll?

Issue #2073, opened 2022-11-10
onProxy needed to support same-origin for existing and stand-alone API servers


n8sabes:
### Is your feature request related to a problem?

It is impractical for Qwik/Qwik-city to require all supporting API servers to be integrated into the same code base. In my case, I have an api server that has evolved over several years and it would be highly impractical to migrate all this this code, plus test cases into a Qwik project.

Both the Qwik front-end and API server must run on the same `protocol+domain+port` to achieve `same-origin`.

Over the last week, I (we in the Qwik community on Discord) have discussed and tried to find solutions to workaround this problem. There are currently bugs that prevent a vite server to work (see #2004) due to priority of plugins. The use of middleware fell flat, and the use of `onRequest` was painful and ultimately could not support both html body and json responses (as needed by the Sandbox and graphql endpoint).

Ultimately, it was not possible to support an Apollo GraphQL server on `/api` that had both the Apollo Sandbox with introspection front-end **AND** the graphQL query processor endpoints.

Please see several long Discord conversations about proxy and reverse-proxy. One such sub-chat, [reverse proxy middleware](https://discord.com/channels/842438759945601056/1039247505063424110).

### Describe the solution you'd like

`onProxy` that passes through everything seamlessly, just like [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware).

This allows for any route to be reverse-proxied to an external/existing backend while preserving `same-origin`.

### Describe alternatives you've considered

**EDIT:** Alternatives deleted on Nov 12, 2022. [SEE THREE WORKAROUNDS, BELOW](https://github.com/BuilderIO/qwik/issues/2073#issuecomment-1312660924)

### Additional context

[http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) was the most successful alternative. [SEE "METHOD A", BELOW](https://github.com/BuilderIO/qwik/issues/2073#issuecomment-1312660924)

thekip:
I set up qwik-city with express adapter, and set up poxy using it
```typescript
import { createProxyMiddleware } from 'http-proxy-middleware';

// Create the Qwik City middleware
const { router, notFound } = createQwikCity({
  render,
  qwikCityPlan,
});


// Create the express server
const app = express();

// Proxy request to CMS: admin + graphql
app.use('/cms', createProxyMiddleware({ target: process.env.PAYLOAD_CMS_ENDPOINT, changeOrigin: true }));

// Use Qwik City's page and endpoint router
[...]

Issue #2004, opened 2022-11-06
Vite server proxy does not work with Qwik-city / Qwik plugins


n8sabes:
### Qwik Version

0.12.1

### Operating System (or Browser)

Chrome 106.0.5249.119 (Official Build) (arm64)

### Node Version (if applicable)

v18.11.0

### Which component is affected?

Qwik Rollup / Vite plugin

### Expected Behaviour

To enable same-origin cookies to an existing api backend, a proxy is necessary. All suggested techniques/workarounds have been tried without success. Please see [Discord Thread](https://discord.com/channels/842438759945601056/937421500879867994/1039071149155422268) summary.

### Actual Behaviour

Proxy does not forward and same-origin is not possible.

### Additional Information

Vite server proxy works **without** plugins:
```
export default defineConfig(() => {
  return {
    server: {
      proxy: {
        '/api': {
          target: 'http://localhost:4000/api',
          changeOrigin: true,
          secure: false,
          rewrite: (path) => path.replace(/^\/api/, '')
        },
      }
    },
    build: { sourcemap: true },
    plugins: [
      // qwikCity(),
      // qwikVite(),
      // tsconfigPaths(),
      // qwikReact()
    ],
    preview: {
      headers: {
        "Cache-Control": "public, max-age=600",
      },
    },
  };
});
```

Vite server proxy does **NOT** work with plugins:
```
export default defineConfig(() => {
  return {
    server: {
      proxy: {
        '/api': {
          target: 'http://localhost:4000/api',
          changeOrigin: true,
          secure: false,
          rewrite: (path) => path.replace(/^\/api/, '')
        },
      }
    },
    build: { sourcemap: true },
    plugins: [
      qwikCity(),
      qwikVite(),
      tsconfigPaths(),
      qwikReact()
    ],
    preview: {
      headers: {
        "Cache-Control": "public, max-age=600",
      },
    },
  };
});
```

Using the above configuration with the target set to your favorite url (on the same protocol) is the fastest way to reproduce this issue. (e.g. http://fubar.tld)

n8sabes:
See #2073 with three workarounds, and explanation of limitations w/ vite server and plugins.

n8sabes:
There is a bug with the vite server that causes an infinite reload loop.

1. Trigger a call to the local vite server proxy
2. Reload the page, or update the source to trigger an HMR refresh

The page goes into an infinite reload loop and causes the server to generate the following error infinitely with every page reload:

```shell
[...]

PR #2542, opened 2023-01-02
Fix Qwik-City hang & enable vite server proxies

n8sabes:
@manucorporat, thanks for reviewing this fix and making any necessary changes.

# What is it?

- [ ] Feature / enhancement
- [√] Bug
- [ ] Docs / tests

# Description

Fix to Qwik-City bug #2004 that blocks backend access to vite server proxies required for `same-origin` cookies. Without this fix, client-side browsers hang waiting on a response from Qwik-City in DEV mode.

# Use cases and why

This fix is critical to enable backend api servers that must set `same-origin` cookies via route proxies, such as required for JWT authentication. `Same-origin` cookies are necessary due to the new highly restrictive 3rd-Party security model in browsers. Without this fix, there is no viable workaround in such as the many proposed in issue #2073 — see workarounds [Here](https://github.com/BuilderIO/qwik/issues/2073#issuecomment-1312660924).

# Reproduction 

*I don't believe this can be reproduced in a StackBlitz, but fortunately it's an easy two-step process to reproduce it locally.*

**SETUP**

1. `pnpm create qwik` 
2. Copy `vite.config.ts`, below. *This is directly from vite docs.*

**RUN IT**

A. Run `pnpm preview` and navigate to `localhost:4173/foo` — **WORKS**

B. Run `pnpm dev` and navigate to `localhost:5173/foo ` — **Route not found**

C. Run `pnpm dev` with local qwik build and navigate to `localhost:5173/foo ` — **HANGS**

```
// vite.config.ts

import { qwikVite } from '@builder.io/qwik/optimizer';
import { qwikCity } from '@builder.io/qwik-city/vite';
import tsconfigPaths from 'vite-tsconfig-paths';

  return {
    server: {
      proxy: {
        '/foo': {
          target: 'http://jsonplaceholder.typicode.com',
          changeOrigin: true,
          rewrite: (path) => path.replace(/^\/foo/, ''),
        },
      },
    },
    plugins: [
      qwikCity(),
      qwikVite(),
      tsconfigPaths()
    ],
    preview: {
      headers: {
        'Cache-Control': 'public, max-age=600',
      },
    },
  };
});
```

# Checklist:

- [√] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [√] I have performed a self-review of my own code

### DIFF
--- a/packages/qwik-city/buildtime/vite/dev-server.ts
+++ b/packages/qwik-city/buildtime/vite/dev-server.ts
@@ -63,8 +63,6 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
         return;
       }
 
-      const serverRequestEv = await fromNodeHttp(url, req, res, 'dev');
-
       await updateBuildContext(ctx);
 
       for (const d of ctx.diagnostics) {
@@ -79,6 +77,7 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
       const routeResult = matchRouteRequest(matchPathname);
       if (routeResult) {
         // found a matching route
+        const serverRequestEv = await fromNodeHttp(url, req, res, 'dev');
         const { route, params } = routeResult;
 
         // use vite to dynamically load each layout/page module in this route's hierarchy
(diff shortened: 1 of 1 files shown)


manucorporat:
Thank you a lot for the PR!

PR #2624, opened 2023-01-12
fix: node static files missing content-type headers

DustinJSilk:
# What is it?

- [ ] Feature / enhancement
- [X] Bug
- [ ] Docs / tests

# Description

Node middleware doesn't currently set the Content-Type header on static files which can prevent files from rendering.

I've added some of the more common files extensions, taken from [nginx](https://github.com/nginx/nginx/blob/master/conf/mime.types)

# Use cases and why

Browsers can't always determine the mime type, also if there is a`X-Content-Type-Options` header present.
The only way around this would be to allow MIME type sniffing which is a security risk.

# Checklist:


shairez:
thanks a lot for this @DustinJSilk ! 🙏

PR #2627, opened 2023-01-12
style: enforce consistent import order

DustinJSilk:
# What is it?

- [X] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Add a prettier plugin to enforce consistent sort orders.

# Use cases and why

- 1. Less lines will change with each commit
- 3. Avoid big changes when contributors have unexpected editor settings (like me)

# Checklist:


DustinJSilk:
@manucorporat if this is something you'd like in, it would be a good idea to keep it up to date with the main branch before merging so that any new commits are also formatted

DustinJSilk:
Haha fair enough.

If it helps, the plugin just calls [organizeImports](https://github.com/microsoft/TypeScript/blob/main/src/services/organizeImports.ts#L74) from the TypeScript API.

@manucorporat

zanettin:
I like this very much @DustinJSilk 🙏  makes things def. easier to find. 
i just had a look at the performance of `pnpm prettier.fix`:

| # | current | with sort order fix | change |
| -- | --- | -- | -- |
| 1 | `14.5s` | `23.2s` | `+62%` |
| 2|  `14.8s` | `22.1s` | `+67%` |
| 3 | `14.7s` | `22s` | `+66%` |

i thought there is smth similar for eslint (https://www.npmjs.com/package/eslint-plugin-simple-import-sort or https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md) which may perform better?

@manucorporat would you like to have smth in this direction? my point is, that it make files more aligned and easier to find things in import heave files. that's why we use it internally on our repos.

DustinJSilk:
The reason I chose this library was because it uses the official typescript sorting method rather than a custom implementation. I’d be fine to use something else with better performance if that’s a concern, though I’ve never felt the need to optimise prettier.

An alternative approach is to only sort imports on staged files with husky where performance is negligible, and ignore the import order during the GitHub workflow checks if that’s where the performance concern is.

zanettin:
> The reason I chose this library was because it uses the official typescript sorting method rather than a custom implementation.

Totally see your point. Just wanted to talk about this quickly 👍

>  I’ve never felt the need to optimize prettier.

always depends on the code base size 🤣 

> An alternative approach is to only sort imports on staged files with husky where performance is negligible, and ignore the import order during the GitHub workflow checks if that’s where the performance concern is.

Sounds both good to me 👍 🙏

gioboa:
@DustinJSilk This PR has a ton of conflicts because it's very old.
So I think the best thing to do is reopen a new fresh PR with this feature.
I close this one. Thanks @DustinJSilk for your help.

PR #2647, opened 2023-01-15
fix:consistent type imports warning

harshmangalam:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

fix consistent type import warning

### DIFF
--- a/starters/apps/e2e/src/components/events/events.tsx
+++ b/starters/apps/e2e/src/components/events/events.tsx
@@ -1,4 +1,4 @@
-import { component$, useStore, PropFunction } from '@builder.io/qwik';
+import { component$, useStore, type PropFunction } from '@builder.io/qwik';
 
 export const Events = component$(() => {
   const store = useStore({
--- a/starters/apps/e2e/src/components/factory/utils.tsx
+++ b/starters/apps/e2e/src/components/factory/utils.tsx
@@ -1,4 +1,4 @@
-import { component$, implicit$FirstArg, QRL } from '@builder.io/qwik';
+import { component$, implicit$FirstArg, type QRL } from '@builder.io/qwik';
 
 export function factoryQrl<P extends Record<string, any>>(componentQRL: QRL<(props: P) => any>) {
   return component$((props: P) => {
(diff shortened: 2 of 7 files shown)


harshmangalam:
Vs code warn when we import type without type keyword.

zanettin:
Wow thanks a lot for the quick reply 🔥 🙏 

I just checked it on latest `main` branch on my side vscode doesn't complain about it on the imports 🙈 
Bildschirmfoto 2023-02-16 um 21 10 51

tsc: **4.9.5**

harshmangalam:
Your eslint is disabled hence you are not getting any warning.

harshmangalam:
entry ssr tsx - Untitled (Workspace) - Visual Studio Code 17_02_2023 21_59_13

Typescript  v4.9.5

zanettin:
Thanks again for your feedback 🙏  

My eslint is working and reports all issues as it should except the one you refer here 🙈 
The mentioned rule `@typescript-eslint/consistent-type-imports` is not part of the `recommended` config of  `@typescript-eslint@5.52.0` which is used on the default generated `.eslintrc.cjs` nor in any other referenced configs/inline config sets. 

these presets are defined and as far as i can see non of them have the rule included (as far as i can see) 🙈 
- `eslint:recommended`: https://github.com/eslint/eslint/blob/main/conf/eslint-recommended.js
- `@typescript-eslint/recommended`: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts
- `qwik/recommended`: https://github.com/BuilderIO/qwik/blob/main/packages/eslint-plugin-qwik/index.ts

can it be, that you use a global version of eslint in vscode which uses another config file?
maybe we can get a step further when you do these steps on your app:
- run `pnpm lint` (or npm / yarn) 😄 
- if the error is reported proceed, otherwise vscode uses another config
- replace the `lint` script within the `package.json` file with `"lint": "eslint src/**/*.ts* --debug"`
- search in the generated output for `eslintrc:config-array-factory Loading JS config file` and check if your root file is used
- search also for `consistent-type-imports`

would be very interested to hear what the outcome of this is 🙏  thanks again for your help!

harshmangalam:
@zanettin got your points.

zanettin:
Hi @harshmangalam 
Just wanted to ask if you already had the time to check, where this rule is coming from 👼 Would be very interested to hear what the outcome is. Thanks again 🙏

harshmangalam:
Hey I  follow your process now everything  is working fine.

zanettin:
So glad to hear 🙏  Thanks anyways for creating this PR and doing the debugging! hope to see soooonish another PR from you 😃

PR #2694, opened 2023-01-22
feat(runtime): add a key attribute to <Fragment>

### LINKED ISSUE #2565: [✨] Keyed Fragments

When rendering multiple children inside an `Array.map` and a wrapper element is undesirable, there seems to be no way to specify a `key`.

React, in comparison allows putting the `key` on a `Fragment`.
https://reactjs.org/docs/fragments.html#keyed-fragments

Quik also has a `Fragment` component, but it does not have a `key` prop.

Add a `key` prop to `Fragment`.

```tsx
{items.map(item => (
  <Fragment key={item.id}>
    <div>{item.foo}</div>
    <div>{item.bar}</div>
  </Fragment>
)}
```

- Using a wrapper element, but sometimes this is undesirable.
- Putting the key on the first child of the fragment (item.foo in the example above) seems to avoid errors, but I'm not sure if that's a false-negative on the tooling side and whether it actually works as intended.

zanettin:
added a key attribute for <Fragment> to enable the usage within loops etc

fix #2565

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
Currently it is not possible to add a `key` prop on fragments which forbids the usage within a loop for example. This PR fixes #2565 where keyed fragments have been requested.

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/jsx/jsx-runtime.ts
+++ b/packages/qwik/src/core/render/jsx/jsx-runtime.ts
@@ -174,7 +174,12 @@ export const isValidJSXChild = (node: any): boolean => {
 /**
  * @public
  */
-export const Fragment: FunctionComponent<{ children?: any }> = (props) => props.children as any;
+export const Fragment: FunctionComponent<{ children?: any; key?: string }> = (props) => {
+  if (props.key) {
+    return jsx(Fragment, { children: props.children }, props.key);
+  }
+  return props.children;
+};
 
 interface JsxDevOpts {
   fileName: string;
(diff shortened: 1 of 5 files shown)


manucorporat:
`key` is never part of the props in the JSX transform:

```
jsx(Fragment, {}, key);
```

shairez:
thanks @zanettin !

@manucorporat can we merge? or is something still missing?

packages/qwik/src/core/render/jsx/jsx-runtime.ts
```diff
@@ -174,7 +174,12 @@ export const isValidJSXChild = (node: any): boolean => {
 /**
  * @public
  */
-export const Fragment: FunctionComponent<{ children?: any }> = (props) => props.children as any;
+export const Fragment: FunctionComponent<{ children?: any; key?: string }> = (props) => {
+  if (props.key) {
```
manucorporat, line 178: This code is not needed, only the updated types!

zanettin:
all good 😂👍

PR #2698, opened 2023-01-23
feat: add node middleware connecting functions

DustinJSilk:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Makes it easier to use a vanilla node middleware with an express-like syntax.

Node middleware now exports an `app` object which can be used like this:

```ts
// Individually
app.use(helmet())

// Or as an arrray
app.use([
  staticFile,
  router,
  notFound,
])

// Add an error handler
app.onError((err, req, res) => {
  const code = (res.statusCode = Number(err.code || err.status) || 500)
  res.end(STATUS_CODES[code])
})

// Run the chain of middleware in an httpserver or return a middleware in `entry.preview.ts`
const server = (req: IncomingMessage, res: ServerResponse<IncomingMessage>) => {
  app.run(req, res)
}

export default server
```

# Use cases and why

It can be difficult for the average developer to write their own methods for chaining middleware

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/api.md
+++ b/packages/qwik-city/middleware/node/api.md
@@ -18,11 +18,14 @@ export function createQwikCity(opts: QwikCityNodeRequestOptions): {
     notFound: (req: IncomingMessage, res: ServerResponse, next: (e: any) => void) => Promise<void>;
     staticFile: (req: IncomingMessage, res: ServerResponse, next: (e?: any) => void) => Promise<void>;
     app: {
-        use: (fn: NodeMiddleware) => void;
+        use: (fn: NodeMiddleware | NodeMiddleware[]) => void;
         run: (req: IncomingMessage, res: ServerResponse<IncomingMessage>) => void;
         onError: (callback: OnErrorHandler) => void;
     };
 };
+
+// @alpha (undocumented)
+export type NodeMiddleware = (req: IncomingMessage, res: ServerResponse<IncomingMessage>, next: (err: any) => void) => void;
 
 // @alpha (undocumented)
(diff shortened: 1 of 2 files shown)


manucorporat:
I am confused about this, qwikcity should not replace express

DustinJSilk:
Thanks for taking a look @manucorporat !

QwikCity already handles routing and static files. Does it not already replace express then?
Adding express to a Qwik City app feels like an extra dependency and routing system on top of QwikCity which already does almost everything.

This change should make it easy to use a vanilla node setup for better performance, and let Qwik City handle all the routing without express or another third party library.

What do you think? Im happy to drop it and just keep it in my own repositories

manucorporat:
yeah! i recall that, i am not sure where the static handling was added to the node middleware, but we were thinking it was a mistake, the static should be handled by the express static plugin... still deciding about it

manucorporat:
Express is heavy and slow agreed, we were looking at polka: https://github.com/lukeed/polka
it's express compatible and very low overhead

DustinJSilk:
Yes static handling can happen outside of qwik, but then it would be helpful to export the isStaticFile function as that has a compiled map of static files for faster lookup.

This PR actually copies how Polka loops over middleware - that was my inspiration haha

mhevery:
@DustinJSilk Thanks for the PR, but from a discussion with @manucorporat, it sounds like it is not the direction we would like to go. We are going to close it. If you feel otherwise, let us know, and let's restart the discussion.

DustinJSilk:
Not a problem @mhevery

PR #2722, opened 2023-01-26
docs: server functions

manucorporat:


packages/docs/src/routes/qwikcity/endpoints/index.md
```diff
@@ -0,0 +1,152 @@
...
+    staleWhileRevalidate: 31536e3,
+  });
+};
+```
+
+### Sending a JSON body
+
+```tsx
+export const onGet: RequestHandler = async ({ json }) => {
+  throw json(200, { hello: 'world' });
```
VinSpee, line 80: Is the new syntax to actually `throw` each of these (`text`, `json`, etc?)

packages/docs/src/routes/qwikcity/loader/index.md
```diff
@@ -0,0 +1,164 @@
...
+    </section>
+  );
+});
+```
+
+## Loader context
+
+Just like request handlers such as `onRequest` and `onGet`, loaders have access to the `RequestEvent` object, that includes information about the current request.
+
+This information comes handly when the loaders needs to conditionally return data based on the request, or it needs to override the response status, headers or body manually.
```
VinSpee, line 144:
```suggestion
This information comes in handy when the loader needs to conditionally return data based on the request, or it needs to override the response status, headers or body manually.
```

packages/docs/src/routes/qwikcity/loader/index.md
```diff
@@ -0,0 +1,164 @@
...
+    time: Date.now();
+  }
+});
+```
+
+## When to use a loader?
+
+A loader should be used when you need to provide some server-side data to your Qwik Components. For example, if you need to fetch some data from a database or an API, you can use a loader to do that.
+
+You should not use a loader to create a REST API, for that you better use a [Endpoint](/qwik-city/), which allows you to have tight control over the response headers and body.
```
VinSpee, line 164:
```suggestion
You should not use a loader to create a REST API, for that you’d be better off using an [Endpoint](/qwik-city/), which allows you to have tight control over the response headers and body.
```

VinSpee (changes requested):
Some typos / tweaks to wording. I’m very excited about these docs / the upcoming release, looking good!

manucorporat:
Thanks for the review! unfortunately i am moving so many things, i cant apply the feedback!
I will get it to a much better spot today

packages/docs/src/routes/qwikcity/loader/index.mdx
```diff
@@ -0,0 +1,164 @@
...
+title: Qwik City - loader$()
+contributors:
+  - manucorporat
+---
+
+# `loader$()`
+
+Loaders allow data to flow from the server to the Qwik Components. For this reason, Qwik Loaders should be always understood in the context of **loading data in the server that is later consumed by a Qwik Component**, if you want to create a RESTful endpoint, please check the [endpoints](/qwikcity/endpoints/index.mdx) guide instead.
+
+The behave like RPC server-side functions that can be invokated by Qwik Components during rendering.
```
Futhark, line 11: The -> They?

PR #2743, opened 2023-01-28
fix(scoped styles): fix pseudo-element logic

cmbartschat:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

I noticed that the fix here: https://github.com/BuilderIO/qwik/pull/2738 commented out some unit tests which I was able to bring back by tweaking the logic.

# Explanation

The reason this is the correct fix in my mind is that the lookahead logic is incrementing `idx` by the length of the lookahead string, so by the time we get into this block, `idx` is no longer what it was at the start of this iteration. This allows us to flush the intended number of characters.

# Checklist:


manucorporat:
Thank you a lot! great PR!

PR #2746, opened 2023-01-28
docs: problem and solution on resource tutorial are no longer equal

mimafogeus2:
# What is it?

- [ ] Feature / enhancement
- [X] Bug
- [X] Docs / tests

# Description

In the tutorial Resource section, the initial problem's code is similar to the solution. This PR replaces the solution parts with placeholders.

### DIFF
--- a/packages/docs/src/routes/tutorial/introduction/resource/problem/app.tsx
+++ b/packages/docs/src/routes/tutorial/introduction/resource/problem/app.tsx
@@ -1,4 +1,4 @@
-/* eslint-disable no-console */
+// @ts-ignore: Unused import
 import { component$, useStore, Resource, useResource$ } from '@builder.io/qwik';
 
 export default component$(() => {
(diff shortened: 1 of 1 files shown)


zanettin:
Hi @mimafogeus2 👋 
Thanks for creating this PR 🙏 
I just had a look at the initial state and the solution and they are different atm. We wanted to have the tutorials focusing on the topic and not make it a to hard to solve challenge 😆 

What do you think? Should we still remove more from the question or is the diff big enough to get new devs into the topic?
Bildschirmfoto 2023-02-18 um 13 16 20

mimafogeus2:
Hey

I just looked at the `BuilderIO:main` branch and it seems like the problem and solution are similar.
What am I missing? Am I looking at the wrong branch?

zanettin:
OMG 😆  sry that was my bad 🙈  we have two identical pages in the docs 🙈 

https://qwik.builder.io/tutorial/introduction/resource/
https://qwik.builder.io/tutorial/reactivity/resource/

same wording but not the same problem and solutions assigned...
would you mind to re-use the values from this problem? https://github.com/BuilderIO/qwik/blob/main/packages/docs/src/routes/tutorial/reactivity/resource/problem/app.tsx

Thanks a lot and sorry again for the confusion!

zanettin:
Hi @mimafogeus2 👋 
Wanted to ask quickly if you would have time for the mentioned adjustment 👼  if not, pls leave me a quick note 🙏  thanks again for creating this PR 🙏

PR #2757, opened 2023-01-30
DX: Flip loader$ generic arguments

nnelgxorz:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Currently, when defining a loader$, you have to first specify the PLATFORM generic you can specify the RETURN generic, like this:

```typescript
export const productLoader = loader$<unknown, Product>(() => {
    // loader code
});
```

This is a rough edge, because you will likely always want to set the RETURN generic everytime, and maybe set the PLATFORM generic or maybe not.

This change moves the RETURN generic to the first position, and defaults the PLATFORM field to unknown if not specified.

```typescript
export const productLoader = loader$<Product>(() => {
    // loader code
});
```

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/api.md
+++ b/packages/qwik-city/runtime/src/api.md
@@ -231,10 +231,10 @@ export interface LinkProps extends AnchorAttributes {
 }
 
 // @alpha (undocumented)
-export const loader$: <PLATFORM, B>(first: (event: RequestEventLoader_2<PLATFORM>) => B) => ServerLoader<B>;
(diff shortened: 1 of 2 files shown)


manucorporat:
Curious, why would you ever do this:
```
export const productLoader = loader$<Product>(() => {
    // loader code
});
```
?

nnelgxorz:
Sorry I missed your question. The issue came up for me when I was prototyping and wanted the loader$ to return some hard-coded data that adhered to an interface. When I went to type the loader to ensure I was getting autocomplete for keys it wasn't working and I couldn't figure out why until I looked at the source code.

PR #2761, opened 2023-01-30
fix: creating project in current folder

RoadRoller01:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

no need to remove selected folder, because it may have setup things or its open in the current terminal so just overwrite it.

# Use cases and why

- 1. aple to create project in cuurent folder 
- 2. don't replece selected folder.

# Checklist:

### DIFF
--- a/packages/create-qwik/create-interactive.ts
+++ b/packages/create-qwik/create-interactive.ts
@@ -48,8 +48,6 @@ export async function runCreateInteractiveCli() {
 
   const outDir: string = getOutDir(projectNameAnswer.outDir.trim());
 
-  let removeExistingOutDirPromise: Promise<void> | null = null;
-
   if (fs.existsSync(outDir)) {
     const existingOutDirAnswer = await prompts(
       {
@@ -61,7 +59,7 @@ export async function runCreateInteractiveCli() {
         )}" already exists. What would you like to do?`,
         choices: [
           { title: 'Do not overwrite this directory and exit', value: 'exit' },
-          { title: 'Overwrite and replace this directory', value: 'replace' },
+          { title: 'Overwrite this directory', value: 'overwrite' },
         ],
         hint: '(use ↓↑ arrows, hit enter)',
       },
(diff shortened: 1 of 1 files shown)


zanettin:
Hi @RoadRoller01 👋 
Thanks a lot for your time to create this PR 🙏 very much appreciated! 
Just a question regarding the made changes: if we don't remove the existing content it could lead to unwanted left-overs from a previous try (legacy code from older versions). imo to have a clean starter, we should go with a complete flushed directory. if the user likes to keep the state, he could select another name or rename the legacy directory. what do you think?

RoadRoller01:
if someone create it a directory and cd to it, he can not do init, try this by your self so you can understand.

zanettin:
Thank you very much for your reply 🙏 can you give me a more detailed reproduction please?
Below you find a quick video of what i've tried based on your description from above but it works as i would expect it. So i guess i didn't get your explanation correctly.


https://user-images.githubusercontent.com/3241476/220759447-df6d85ca-9a1a-4188-997b-efac4923f265.mp4

RoadRoller01:
Thank you also for your cool reply, if the current directory was foo, and you tried to type `.`  for the new project location it will crash, thank you for your cooperation.

zanettin:
aaaah! gotcha 🙏  

what do you think about this change then?
checking if the answer equals `.` or `./`. if so, we remove all existing files and continue. the rest would stay as it currently is.
so a 2nd condition within the `replacement` path.

RoadRoller01:
no, just remove all files like this 
`rm -r ~/foo/*` not like this 
`rm -r ~/foo/`,
but still i see that there is no need to remove all files,
because the devlapor may drop  some images or any setup things, meaning just print warning that there is files and it will be ignored.

zanettin:
i see what you mean, but i am not sure if this is a common use-case tbh 😄 

i mean if i have an existing app with stuff in it, why would i re-run the CLI? i'd run it to create a brand new version and then migrate my existing stuff and make sure that i adapt to the latest configs, right? if i want to have the latest version of the framework i'd run `pnpm upgrade -i -L`. what i would love to have on the other hand is an option to apply certain codemods against my app after upgrading to a newer version. like renaming all my `useClientEffects` to `useBrowserVisibleTask`, etc 😄

zanettin:
Hi @RoadRoller01 👋 
Let me close this PR for now and we see if there will be more requests / use-cases in that regard 👍 Thank you very much for your time and effort in creating. this PR 🙏  very appreciated! I hope to see another one from you soon 😃

PR #2818, opened 2023-02-03
feat(qwik): customize error overlay

### LINKED ISSUE #2710: [✨] Enhanced Error Overlay [DX]

Enhance the current error overlay to increase DX

Astro v2 just launched a [new error overlay](https://github.com/withastro/astro/pull/5495) which provides additional informationen and a cleaner overview. To implement a custom overlay we would need to "opt-out" of the vite provided web component (`vite-error-overlay`) and create a custom solution.

things we could included if we want to implement this:
- qwikified look and feel (use qwik colors, fonts ,etc)
- show more lines of code in the trace
- highlight the row that contains the error
- reference the docs for certain issues
- "open in editor" link instead of the current file path link

Keep the existing one and lower the maintenance work 😄 

### Additional context

_No response_

zanettin:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Qwikify the vite dev server error overlay. Fixes #2710.

## PR state after review session
B1

## Terminal output (stays unaffected)
Bildschirmfoto 2023-02-04 um 20 44 05

# Info
- Used Icons are from https://www.svgrepo.com

# Todos
- [x] Get feedback regarding design
- [x] Wait till this PR https://github.com/vitejs/vite/pull/11830 gets merged
- [x] Wait till qwik uses version `4.2.0` of vite which includes the changes from above

# Follow-ups
- Remove vite code after https://github.com/vitejs/vite/pull/11970 got merged

# Checklist:


packages/qwik-city/buildtime/vite/dev-server.ts
```diff
@@ -194,6 +194,9 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
 
       next();
     } catch (e) {
+      if (res.writableEnded) {
```
zanettin, line 197: we've extended the `qwik` dev server middleware to handle errors as well instead of forwarding the to the vite error middleware. this because we are otherwise unable to ship the needed CSS to modify the overlay on SSR errors. The new error handler terminates the response with an `.end()` call. After that point it does no longer make sense to follow the `next(e)` path at all because an `header already sent` error would occur.

packages/qwik/src/optimizer/src/plugins/vite-error.ts
```diff
@@ -0,0 +1,210 @@
+import type { RollupError } from 'rollup';
+import type { ViteDevServer } from 'vite';
+import { yellow, magenta, cyan } from 'kleur/colors';
+import strip from 'strip-ansi';
+
+/**
```
zanettin, line 6: i think about creating another PR on the vite repo to have an option to pass in some custom markdown into the vite error handler so we can remove the whole logic again on our side.

packages/qwik/src/optimizer/src/plugins/vite-error.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+
+vite-error-overlay::part(tip):before {
+  content: "Not sure how to solve this? Visit https://qwik.builder.io or connect with the community on Discord.";
```
zanettin, line 206: just an idea. please let me know what u think about it...

manucorporat: yeah if this works! i think it's ok!

manucorporat:
THIS IS AMAZING!! love it, this will deserve a good tweet when we merge it

packages/qwik/src/optimizer/src/plugins/vite-error.ts
zanettin: it does but w/o a link 😄 and can't style the url

shairez:
great job @zanettin !

Looks much more readable

BTW, for errors, I'd love to see the color "red" because it makes it easier to scan for it in the console.

Plus, not sure if it's possible, but syntax highlighting would also be great, but that's not a much

manucorporat:
I dont think code highlighting is an easy thing to add, i would not block this PR for that, we never had it before, but i would say, i would like to see the code with a more vibrant color, the yellow for vite was cooler.

manucorporat:
My only take would be to add a little bit of color, the red @shairez suggested would be nice and some other

zanettin:
thanks a lot @manucorporat and @shairez for your feedbacks 🙏 

the concern @shairez mentioned was about the terminal output which would stay unaffected by now.
i'll try to add some more colors to the overlay and post an update asap 🙏 

regarding the formatting/highlighting: yes we are limited as long as we'd like to keep the vite error overlay web component since the error is rendered within a `<pre>` tag. if we would like to do everything on our side (other PR imo as well) we could do that which would give us complete flexibility 👍

zanettin:
@manucorporat @shairez here are some other variations 👨‍🎨 😄  
pls let me know in which direction i should go or if one of them match your expectations 🙏  i've also updated the description of the PR and added a screenshot of the terminal output.

# B1
B1

# B2
B2

# B3
B3

# C1
Bildschirmfoto 2023-02-04 um 21 26 55

shairez:
thanks @zanettin !

I prefer B2

zanettin:
> thanks @zanettin !
> 
> I prefer B2

thanks @shairez 🙏 maybe i should play around a bit with the aggressivity of the red color? give it a bit of opacity or so 👍 
imo a bit less "qwik"-style than the other options but it is an error overlay after all 😄

shairez:
I think it's OK not to have the error message with the branded blue color, because it's an error after all :)

Maybe the icons could be with that blue or something, but it's not a biggie IMO

zanettin:
totally with u @shairez 🙏 let me try some variations tomorrow evening 👍

zanettin:
# B4
Bildschirmfoto 2023-02-05 um 23 12 13
tiny adjusted B2 version (reduced red, removed blue completely from the screen)

nnelgxorz:
I think it looks great. 

I personally prefer not having the error message header as red and reserving the red color for highlighting the actual error in the code, [like this](https://elm-lang.org/assets/blog/error-messages/0.16/context.png). Giant red text just makes me feel like I'm being yelled at. 

I would bet that it stresses out beginners as well. Just my two cents. 👍

shairez:
but you ARE being yelled at @nnelgxorz !! 😅😂

adamdbradley:
Love it, and like the idea of less red and more qwik-ish colors. But think we ship it and see how it feels and iterate.

zanettin:
thanks for all the feedbacks 🙏 set the current state of this PR to the B1 version. hope that is a good starting point. if you have other wishes/requirements, pls let me know 🙏 
otherwise i'll wait till the PR on the vite repo got merged (was approved twice now 🎉 ) and we've upgraded to latest vite version for final testings.

manucorporat:
B1 is cool!

packages/qwik/src/optimizer/src/plugins/vite-error.ts
```diff
@@ -0,0 +1,135 @@
...
+  --vertical-box-spacing: 15px;
+  --box-padding: 20px;
+  --box-margin: 0 0 var(--vertical-box-spacing) 0;
+  --box-background: rgba(0, 0, 0, 0.5);
+  --box-border-radius: 8px;
+}
+
+vite-error-overlay::part(backdrop) {
+  background: rgb(2 11 17 / 60%);
+  backdrop-filter: blur(20px) brightness(0.4) saturate(3);
```
intellix, line 22: this causes my GPU to use like 5gb of memory and it makes my M4 MBP fans go insane, which I think I've never heard until this line was presented to my browser :P

wmertens: all from a little blur? Weak :stuck_out_tongue: ;)

PR #2824, opened 2023-02-04
fix: package.json type field setting error

cunzaizhuyi:
# What is it?

- [ ] Feature / enhancement
- [X] Bug
- [ ] Docs / tests

# Description

when user viteConfig of ssr.format is `cjs`, the dist package.json's type field should be set to 'commonjs' instead of 'module'.
 
error is in the following:
```
ReferenceError: exports is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension and '/Users/vscodeProject/qwik-demo3/server/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/buildtime/vite/plugin.ts
+++ b/packages/qwik-city/buildtime/vite/plugin.ts
@@ -41,7 +41,7 @@ export function qwikCity(userOpts?: QwikCityVitePluginOptions): any {
   let mdxTransform: MdxTransform | null = null;
   let rootDir: string | null = null;
   let qwikPlugin: QwikVitePlugin | null;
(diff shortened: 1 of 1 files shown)


manucorporat:
can you run `pnpm fmt` at the root to solve the lint issues?

cunzaizhuyi:
> can you run `pnpm fmt` at the root to solve the lint issues?

finished

manucorporat:
lets do it! merged! thank you!!

PR #2825, opened 2023-02-04
Tutorial: Best practices refactor.

mimafogeus2:
# What is it?

- [ ] Feature / enhancement
- [X] Bug
- [X] Docs / tests

# Description

In the Implicit Template Updates tutorial, the solution's callbacks only do side effects, and are not wrapped in curly braces. Since arrow functions without curly braces return their value, it is customary to wrap them in curly braces to show that the return value is insignificant/void.

### DIFF
--- a/packages/docs/src/routes/tutorial/reactivity/template/solution/app.tsx
+++ b/packages/docs/src/routes/tutorial/reactivity/template/solution/app.tsx
@@ -13,10 +13,10 @@ export default component$(() => {
   console.log('Render: <App>');
   return (
     <>
-      <button onClick$={() => store.countA++}>a++</button>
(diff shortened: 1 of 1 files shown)


zanettin:
Hi @mimafogeus2 

Thank you very much for submitting this PR 🙏 
I totally see your point. Imo it is a well known practice to write it the way it currently is (w/o the curlies) 👍
Just as an example how it's recommended on React: https://beta.reactjs.org/learn/responding-to-events#adding-event-handlers

Hope it is okay, when i close this PR?
Thanks again for your time and help improving the docs 🙏

mimafogeus2:
Hey @zanettin 
Thank you for taking the time for this

I think this is also an anti-pattern in the React docs, but this is a very minor issue so I won't push it

PR #2835, opened 2023-02-05
fix: set cookies on dev server

genki:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

It seems there is no code that sets `Set-Cookie` response header in the dev-server.
So cookies set in server side such as `onGet` or `loader$` had never sent to client.

# Checklist:


genki:
@adamdbradley Unfortunately, it doesn't work correctly #2835.

https://github.com/BuilderIO/qwik/blob/main/packages/qwik-city/middleware/node/http.ts#L65-L71

I guess there can be a path that `getWritableStream` of `serverRequestEv` is never being called.
If so, the cookie headers also never set.

genki:
I have confirmed at 14f3256dab95ae8f5ab336826cff28f6995e8225 and confirmed the `getWritableStream` is not called.

For an example, some page in my app (testing with 14f3256dab95ae8f5ab336826cff28f6995e8225) which is jumped by clicking `<Link />` set `Set-Cookie` headers but when reload it by browser or HMR then never set.
This is reprodusable.

adamdbradley:
Thanks, will keep looking into it

genki:
@adamdbradley Thanks!

adamdbradley:
Would you be able to provide an example of how you're setting the cookie to make sure we're adding the right tests? Thanks

genki:
@adamdbradley This is it :)
https://github.com/genki/qwik-test/blob/cookie-test/src/routes/index.tsx

zanettin:
This PR contains changes of https://github.com/BuilderIO/qwik/pull/2852 as well

genki:
This PR has mistakenly chosen the main branch. I will fix it and retry.

PR #2852, opened 2023-02-07
fix: load `.env` files on dev server

### LINKED ISSUE #2834: [🐞] env arg of loader$ is lacking .env variables at the first time after run dev server

Qwik City (routing)

### Describe the bug

When the first access to dev server, `env` of `loader$(({env}) => { ... }` is not filled with values of `.env`.
After the second time the `env` is filled expectedly.

### Reproduction

https://stackblitz.com/edit/qwik-starter-4gsffl?file=src/routes/index.tsx

### Steps to reproduce

I am sorry the reproduction code above is not working because the environment is old then there's no `env` arg.

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M2
    Memory: 63.28 MB / 24.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 19.5.0 - /opt/homebrew/bin/node
    npm: 9.3.1 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 109.0.5414.119
    Firefox: 109.0
    Safari: 16.2
  npmPackages:
    @builder.io/qwik: file:../clone/qwik/packages/qwik => 0.17.4 
    @builder.io/qwik-city: file:../clone/qwik/packages/qwik-city => 0.1.0-beta13 
    undici: 5.14.0 => 5.14.0 
    vite: 4.0.4 => 4.0.4
```

### Additional Information

_No response_

genki:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Load env variables from `.env` by calling `loadEnv` of the Vite in order to prepare `process.env` before the wakeup of the dev-server so that it can use `.env`.

cf.) https://vitejs.dev/config/#using-environment-variables-in-config

# Use cases and why

Fixes #2834

# Checklist:


DustinJSilk:
@manucorporat can we get this merged? 🙂

manucorporat:
i am confused why we need to do this manually, and it's not done by Vite automatically

genki:
@manucorporat 
Yeah, I think it's a kind of workaround.
The initialization of qwik is quicker than the Vite's expect.
The future Vite may have other options, but currently we have to call `loadEnv()` as documented.

DustinJSilk:
@manucorporat any chance we could find a solution to this?
This also means we can initiate anything as a script is parsed, it has to be lazily initiated when requested and the variabled still wont exist on the first load

zanettin:
Hi @genki 👋 
Thanks a lot for creating this PR 🙏  would you mind to push an empty commit to re-trigger the CI? it somehow hangs 😄 🙏

genki:
@zanettin  Ok I will

shairez:
@genki @zanettin just FYI for next time we can just close and reopen the PR and it will retrigger the CI checks

DustinJSilk:
Just thinking about this a bit more, i wonder if we should rather provide a hook for developers to initialise code that is guaranteed to run before any onRequest handlers or loader$s.

That would give flexibility to solve other similar issues such as initialising telemetry, which currently cant be done before the first request. And leave it to the user to call loadEnv themselves.

genki:
@DustinJSilk 
I think it is not a good design that the framework has every users put `loadEnv()` manually.
This pain comes from the not good design of the Vite.
We have to wait the fix with a patch like this, before making a big structure on the sand.

PR #2863, opened 2023-02-07
docs: add `<RenderOnce>` REVERT

zanettin:
Reverts BuilderIO/qwik#2848

As discussed with @shairez , we remove that part again since it is a more internal API.


shairez:
thanks @zanettin

Issue #2797, opened 2023-02-02
[🐞] Cross-site form submission incorrect origin match error


tuurbo:
### Which component is affected?

Qwik Runtime/city

### Describe the bug

On my local dev environment (eg: https://beta.mysite.loc/) when I submit a form using `action$` I get error "Cross-site POST form submissions are forbidden". If I console log `request.headers.get('origin')` and `url.origin` within the `securityMiddleware` function https://github.com/BuilderIO/qwik/blob/v0.17.2/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts#L274...
`request.headers.get('origin')` returns the correct https url https://beta.mysite.loc and 
`url.origin` returns incorrect http url http://beta.mysite.loc

Example component
```tsx
import { component$ } from '@builder.io/qwik';
import { action$ } from '@builder.io/qwik-city';

export const signinAction = action$(async () => {
  console.log('this is never called!');
});

export default component$(() => {
  const signIn = signinAction.use();

  return (
    <>
      <form
        action={signIn.actionPath}
        method="POST"
      >
        <label>
          <span>Email</span>
          <input
            name="email"
            type="text"
          />
        </label>

        <label>
          <span>Password</span>
          <input
            name="password"
            type="password"
          />
        </label>

        <button>Login</button>
      </form>
    </>
  );
});
```

![qwik-error](https://user-images.githubusercontent.com/996899/216439877-3ea9dd2a-42bd-4ddb-90fd-a481c69ff168.png)

### Reproduction

I dont think you can replicate on stackblitz but here is a link anyways https://stackblitz.com/edit/qwik-starter-uyvonq?file=src%2Froutes%2Findex.tsx

### System Info

```shell
 System:
    OS: Windows 10 10.0.19045
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i9-12900HK
    Memory: 12.93 GB / 31.68 GB
  Binaries:
    Node: 18.12.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (109.0.1518.69)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @builder.io/partytown: ^0.7.5 => 0.7.5
[...]

PR #2864, opened 2023-02-07
Fixes issue on HTTP/2 headers for resolving origin

genki:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Fixes remaining issue of #2797 for about dev server.
Because of HTTP/2 headers differ from 1.1, some changes were needed to `getUrl()`.
I have made it used `http2.Http2ServerRequest` if the req object comes from HTTP/2.

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/http.ts
+++ b/packages/qwik-city/middleware/node/http.ts
@@ -1,4 +1,5 @@
 import type { IncomingMessage, ServerResponse } from 'node:http';
+import { Http2ServerRequest } from 'node:http2';
 import type {
   ServerRequestMode,
   ServerRequestEvent,
@@ -7,6 +8,9 @@ import type {
 const { ORIGIN, PROTOCOL_HEADER, HOST_HEADER = 'host' } = process.env;
 
 function getOrigin(req: IncomingMessage) {
+  if (req instanceof Http2ServerRequest) {
+    return `${req.scheme}://${req.authority}`;
+  }
   const headers = req.headers;
   const protocol =
     (PROTOCOL_HEADER && headers[PROTOCOL_HEADER]) ||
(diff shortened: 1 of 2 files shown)


packages/qwik-city/middleware/node/http.ts
```diff
@@ -7,6 +8,9 @@ import type {
 const { ORIGIN, PROTOCOL_HEADER, HOST_HEADER = 'host' } = process.env;
 
 function getOrigin(req: IncomingMessage) {
+  if (req instanceof Http2ServerRequest) {
```
manucorporat, line 11: Not sure it's ok to reoverride completely the other logic, since it's useful for proxy support

genki: I think the responsibility of `getOrigin()` is limited to obtain the origin string from the request object with no side effect.
I guess the proxy support can be provided in `getUrl()`, because the proxy needs other parts of URL not only the origin.

manucorporat: The origin might not be the one from the request, in some cases the backend is behind a reverse proxy and the origin of the request is not relevant at all.

manucorporat: Imo we should respect the HOST_HEADER when it's not Host, and the PROTOCOL_HEADER

genki: I see, you meant not about the forward proxy but the reverse proxy.
Yes, I will fix it :)

genki: @manucorporat Updated PR.
Changed the default value of the HOST_HEADER to be conditional to http version, if HTTP/2 then `:authority` else `host`.

manucorporat:
now its perfect! merging!

PR #2886, opened 2023-02-08
fix: code allows sitemapOutFile to be null but type is wrong

### LINKED ISSUE #2872: [🐞] code allows sitemapOutFile to be null but type is wrong

Qwik City (routing)

### Describe the bug

Quote from types.ts:

```
   /**
   * File system path to write the `sitemap.xml` to. Defaults to `sitemap.xml`
   * and written to the root of the `outDir`. Setting to `null` will prevent
   * the sitemap from being created.
   */
  sitemapOutFile?: string;
```

Code confirms it can be null: https://github.com/BuilderIO/qwik/blob/b1ec486944b38383c8177bc1ad16e0b028f4970e/packages/qwik-city/static/node/node-main.ts#L43

So the type should allow null too.

### Reproduction

NA

### System Info

"@builder.io/qwik-city": "0.1.0-beta14",

###  Additional Information

No response

felixsanz:
Fix #2872

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

See #2872

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/static/api.md
+++ b/packages/qwik-city/static/api.md
@@ -32,7 +32,7 @@ export interface StaticGenerateRenderOptions extends RenderOptions {
     maxWorkers?: number;
     origin: string;
     outDir: string;
-    sitemapOutFile?: string;
+    sitemapOutFile?: string | null;
 }
 
 // @alpha (undocumented)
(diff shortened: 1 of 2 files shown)


manucorporat:
makes a lot of sense! i would love to merge this! 

the CI is failing because there is a change in types, so it's needed to run pnpm build.full to update them

felixsanz:
> makes a lot of sense! i would love to merge this!
> 
> the CI is failing because there is a change in types, so it's needed to run pnpm build.full to update them

thanks, I forgot that.

done!

Issue #2873, opened 2023-02-08
[🐞] useResource$ fails with relative paths -- Failed to parse URL


n8sabes:
### Which component is affected?

Qwik City (routing)

### Describe the bug

@shairez brought this to my attention.

 After debugging it appears: `useResource$` fails with relative paths -- Please see attached repo.

ERROR: `Failed to parse URL`

WORKS:
1. `useResource$` works with absolute paths
2. initial `<Link>` to relative path

**DOES NOT WORK WITH RELATIVE PATHS:**
1. Reload
2. `window.location` navigation
3. First nav to route from browser

### Reproduction

https://github.com/n8sabes/qwik_broken_route_lab

### Steps to reproduce

~3mins to reproduce with attached [repo](https://github.com/n8sabes/qwik_broken_route_lab).

1. `git clone https://github.com/n8sabes/qwik_broken_route_lab.git`
2. `cd qwik_broken_route_lab`
3. `pnpm i`
4. `pnpm dev`
5. Lunch browser and click red buttons

Could [undici.setGlobalOrigin(origin)](https://github.com/nodejs/undici#undicisetglobaloriginorigin) not be set on initial load?

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M1
    Memory: 137.58 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 19.4.0 - ~/.nvm/versions/node/v19.4.0/bin/node
    npm: 9.2.0 - ~/.nvm/versions/node/v19.4.0/bin/npm
  Browsers:
    Chrome: 109.0.5414.119
    Firefox Developer Edition: 110.0
    Safari: 16.2
  npmPackages:
    @builder.io/qwik: github:builderio/qwik-build => 0.17.2
    @builder.io/qwik-city: github:builderio/qwik-city-build => 0.1.0-beta11
    undici: 5.16.0 => 5.16.0
    vite: 4.0.4 => 4.0.4
```

### Additional Information

Cc: @manucorporat

n8sabes:
The local repo is better to reproduce the issue, but provided a StackBlitz for convenience, [HERE](https://stackblitz.com/edit/qwik-starter-fku325?file=README.md).

manucorporat:
I dont think this is solvable, without patching fetch which is not something i want to do.
[...]

PR #2924, opened 2023-02-10
Fix RouteLocation by including essential fields

n8sabes:
# What is it?
This resolves missing essential fields in RouteLocation.

- [√] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
Add essential fields to RouteLocation so `useLocation()` can provide access to necessary fields.

# Use cases and why
Without these fields, the developer is required to write redundant code to extract the sub-fields from `href` (i.e. `protocol`, `host`, `port`) that should be easily available. Coincidentally, this also helps resolve the missing fields from `useLocation()` discussed in-part in Issue #2873.

**BEFORE:**
```typescript
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly href: string;
  readonly pathname: string;
  readonly query: URLSearchParams;
  readonly isNavigating: boolean;
}
```

**AFTER:**
```typescript
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly href: string;
  readonly origin: string;
  readonly protocol: string;
  readonly host: string;
  readonly hostname: string;
  readonly port: string;
  readonly pathname: string;
  readonly query: URLSearchParams;
  readonly isNavigating: boolean;
}
```

# Checklist:

- [√] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [√] I have performed a self-review of my own code

### DIFF
--- a/packages/docs/src/routes/qwikcity/pages/index.mdx
+++ b/packages/docs/src/routes/qwikcity/pages/index.mdx
@@ -144,7 +144,7 @@ export default component$(() => {
 });
 
 export const head: DocumentHead = ({ params, getData }) => {
-  const product = getData(getProductData);
+  const product = getData(useGetProductData);
   return {
     title: `Product "${product.title}"`,
     meta: [


manucorporat:
i am not sure i like this change much, it's a lot of duplicated state, could we add `origin` instead of host, port, and schema?

PR #2939, opened 2023-02-11
refactor: useClientEffect to useOnVisibleTask

manucorporat:
## There are many problems with useClientEffect$():

- people have pre-conceived notions of what it is and they abuse it.
- They tend to put event listeners in there instead of putting it in useOn().
- a lot of community feedback that it is confusing.
- implies that useTask$() and useClientEffect$() are way different things.
- It is valuable to have a well named "dangerous" method which causes slowness as it is easy to search for. 
- useEffect is already being abused by React developers, in Qwik the same problems can be solved more efficiently with `useTask` or `useResource`

## Design constrains

- It's a different type of Task (happens at the different stage of the rendering pipeline), not just a Task but in the browser
- Should be clear that it's costly
- Should be clear that it's browser only
- Should be clear that happens after the component is rendered, and it's available in the DOM
- Should be clear that the DOM is there, and it's safe to use DOM APIs, without `window` check
- Should be clear that it's not only about rendering, Tasks can be executed independently, so PostRender is not a good fit.
- Should not contain the word `Effect`

## Why `useOnVisibleTask`

- `use`: like other APIs, it's a `use` method, meaning it creates some internals state and needs to be declared at the root level of a component$
- `On`: under the hood, `useOnVisibleTask` creates a event handler, similar to `onClick`, but the event is is triggered automaticlaly, like document-load or IntersectionObserver. `on` also implies that it's a browser only thing, since events only happen in the browser. Also signals the `cost` of this API.
- `VisibleTask:` One word, it's a different type of Task that happens after the component is rendered, there the `Visible` part. Also can behave like a Task, using the `track` API to reexecute when tracked state changes. Also `Visible` also implies that only happens happens in the browser, things are not visible in the server, there is no DOM in the server.

### But, it's so long

This is on purpose, this is an API that usually should not be used by developers, most of the time, the same can be achieved without eagarly running JS in the browser, using useTask.

The wordy name introduces a little bit of friction to favor `useTask`

### DIFF
--- a/packages/qwik/src/core/api.md
+++ b/packages/qwik/src/core/api.md
@@ -899,6 +899,12 @@ export interface Tracker {
 // @alpha (undocumented)
 export const untrack: <T>(fn: () => T) => T;
 
+// @public
+export const useBrowserVisibleTask$: (first: TaskFn, opts?: OnVisibleTaskOptions | undefined) => void;
+
+// @public
+export const useBrowserVisibleTaskQrl: (qrl: QRL<TaskFn>, opts?: OnVisibleTaskOptions) => void;
+
 // @alpha @deprecated
 export const useCleanup$: (first: () => void) => void;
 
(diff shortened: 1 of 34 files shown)


packages/qwik/src/core/api.md
```diff
@@ -942,6 +936,12 @@ export const useOn: (event: string | string[], eventQrl: QRL<(ev: Event) => void
 // @alpha
 export const useOnDocument: (event: string | string[], eventQrl: QRL<(ev: Event) => void>) => void;
 
+// @public
+export const useOnVisibleTask$: (first: TaskFn, opts?: UseEffectOptions | undefined) => void;
```
zanettin, line 940: just that we don't forget to rename the `UseEffectOption` interface/type as well

manucorporat: Done!

PR #2968, opened 2023-02-14
docs: add specific type to Context example

kerbelp:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Example in docs does not work without a specific type cast for `createContext`. I think that updating the docs with my changes will save some time for less experienced TS developers in the future (such as my self).

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/components/context/index.mdx
+++ b/packages/docs/src/routes/docs/components/context/index.mdx
@@ -32,8 +32,12 @@ import {
   createContextId,
 } from '@builder.io/qwik';
 
+export interface IMyContext {
+    count: number
+}
+
 // Create a new context descriptor
-export const MyContext = createContextId('my-context');
+export const MyContext = createContextId<IMyContext>('my-context');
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/components/context/index.mdx
```diff
@@ -53,7 +57,7 @@ export const Parent = component$(() => {
 
 export const Child = component$(() => {
   // Get reference to state using MyContext
-  const state = useContext(MyContext);
+  const state = useContext<IMyContext>(MyContext);
```
shairez, line 60: you don't actually need to add it here because it's inferred from the `MyContext` type already

shairez:
thanks @kerbelp 

added a comment

packages/docs/src/routes/docs/components/context/index.mdx
kerbelp: Thanks, updated PR.

PR #2983, opened 2023-02-15
Add Playground as bug category

samijaber:
# What is it?

Add the playground as a bug category, as its currently not represented by any of the existing ones.


adamdbradley:
thx

PR #2985, opened 2023-02-15
allow `OnRenderFn` be async and use `Promise`s

revintec:
allow `component$(...)` to use async functions as callback, thus simplifying server code
this would also align its signature more with `useBrowserVisibleTask$(...)`(originally `useClientEffect$(...)`

# What is it?

- [X] Feature / enhancement

# Description

fixes https://github.com/BuilderIO/qwik/issues/2765

# Checklist:

- [x] Added new tests to cover the fix / functionality -- only changes function signature, in a backward compatible way

### DIFF
--- a/packages/qwik/src/core/api.md
+++ b/packages/qwik/src/core/api.md
@@ -373,7 +373,7 @@ export type NoSerialize<T> = (T & {
 export const noSerialize: <T extends object | undefined>(input: T) => NoSerialize<T>;
 
 // @public (undocumented)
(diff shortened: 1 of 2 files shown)


manucorporat:
it's not a practice we recommend! in fact we might error if you detect promises at component$, why do you need to await inside a component$ callback?

revintec:
@manucorporat sorry for replying late, let's continue discussion here https://github.com/BuilderIO/qwik/issues/2765

PR #2990, opened 2023-02-16
Updated action$() documentation

markovickosta:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

An example in docs does not work without a ".use()" when using the action.
I also changed a few action names to stay consistent.

# Why

- I got confused while working on a personal project because I was missing .use() and trying just to call the action.

# Checklist:


manucorporat:
The docs are actually correct! you should udpate to latest :)

markovickosta:
@manucorporat Thank you for the response. I'm on version 0.17.5, where can I check the change log for the new versions?

PR #3002, opened 2023-02-16
feat: add nonce support

DustinJSilk:
fixes https://github.com/BuilderIO/qwik/issues/2663

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/index.ts
+++ b/packages/qwik-city/middleware/node/index.ts
@@ -34,8 +34,7 @@ export function createQwikCity(opts: QwikCityNodeRequestOptions) {
   const router = async (
     req: IncomingMessage,
     res: ServerResponse,
-    next: NodeRequestNextFunction,
-    nonce?: string
+    next: NodeRequestNextFunction
   ) => {
     try {
       const serverRequestEv = await fromNodeHttp(getUrl(req), req, res, 'server');
--- a/packages/qwik-city/middleware/node/index.ts
+++ b/packages/qwik-city/middleware/node/index.ts
@@ -34,7 +34,8 @@ export function createQwikCity(opts: QwikCityNodeRequestOptions) {
   const router = async (
     req: IncomingMessage,
     res: ServerResponse,
-    next: NodeRequestNextFunction
+    next: NodeRequestNextFunction,
+    nonce?: string
   ) => {
     try {
       const serverRequestEv = await fromNodeHttp(getUrl(req), req, res, 'server');
--- a/packages/qwik-city/middleware/request-handler/api.md
+++ b/packages/qwik-city/middleware/request-handler/api.md
@@ -148,12 +148,8 @@ export interface ResolveValue {
 // @alpha (undocumented)
 export interface ServerRenderOptions extends RenderOptions {
     // Warning: (ae-forgotten-export) The symbol "ContentSecurityPolicy" needs to be exported by the entry point index.d.ts
-    //
-    // (undocumented)
     contentSecurityPolicy?: ContentSecurityPolicy;
-    // (undocumented)
     headers?: Record<string, string>;
-    // (undocumented)
     nonce?: boolean;
     // (undocumented)
     qwikCityPlan: QwikCityPlan;
--- a/packages/qwik-city/middleware/request-handler/api.md
+++ b/packages/qwik-city/middleware/request-handler/api.md
@@ -147,6 +147,14 @@ export interface ResolveValue {
 
 // @alpha (undocumented)
 export interface ServerRenderOptions extends RenderOptions {
+    // Warning: (ae-forgotten-export) The symbol "ContentSecurityPolicy" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    contentSecurityPolicy?: ContentSecurityPolicy;
+    // (undocumented)
+    headers?: Record<string, string>;
+    // (undocumented)
+    nonce?: boolean;
     // (undocumented)
     qwikCityPlan: QwikCityPlan;
     // (undocumented)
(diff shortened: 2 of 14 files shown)


manucorporat:
Interesting, i like how this does not interfere with existing code! love the direction!

manucorporat:
so developers would have to generate their own nonce?
how does this work? it's one for the whole server?

DustinJSilk:
yes they would have to generate it themselves.
They would need to generate it for each new request, and pass it into the render function.

Here is an example i've put together: https://github.com/DustinJSilk/qwik-nonce/blob/main/src/entry.preview.tsx

manucorporat:
could this implemented within qwikcity? as a plugin?

manucorporat:
I understand this solves the problem for you, but we can't change the signature of express middleware so work around to pass a nonce, this needs to work in cloudflare, netlify, vercel...

manucorporat:
I propose to use `requestEv.sharedMap` as a connection point, it could be added a plugin that adds the nonce for every request, and sets the headers and qwik will use it.

How about some well known keys, like:
`@nonce`

DustinJSilk:
> I understand this solves the problem for you, but we can't change the signature of express middleware so work around to pass a nonce, this needs to work in cloudflare, netlify, vercel...

Yep that makes sense!

> ... it could be added a plugin that adds the nonce for every request, and sets the headers and qwik will use it.

Ok, Im happy to work on something like this.
I'll have another crack at it some time today/tomorrow and we can take it from there

Thanks!!

DustinJSilk:
@manucorporat let me know if this is closer to what you had in mind!

[Ive updated the example as well](https://github.com/DustinJSilk/qwik-nonce/blob/main/src/entry.preview.tsx)

manucorporat:
Unfortunately, i didnt mean this. I dont want to add any `nonce` option. instead i wanna rely on `sharedMap` fucntionality to pass a custom nonce to the qwik middleware

manucorporat:
ie, no extra code to handle nonce, it will be implemented by a user plugin to add it

/src/routes/plugin@nonce.ts
```tsx
export const onRequest = (ev) => {
  const nonce = createNonce();
  ev.sharedMap.set('@nonce', nonce):
  ev.headers.set(....)
}
```

packages/qwik-city/middleware/request-handler/request-event.ts
```diff
@@ -31,11 +32,27 @@ export function createRequestEvent(
...
   resolved: (response: any) => void
 ) {
-  const { request, platform, env } = serverRequestEv;
+  const { request, platform, env, contentSecurityPolicy } = serverRequestEv;
 
   const cookie = new Cookie(request.headers.get('cookie'));
-  const headers = new Headers();
+  const headers = new Headers(serverRequestEv.headers);
   const url = new URL(request.url);
+  const nonce = serverRequestEv.nonce ? randomBytes(16).toString('base64') : undefined;
```
manucorporat, line 40: all this changes should be reverted

packages/qwik-city/middleware/request-handler/api.md
```diff
@@ -151,15 +155,21 @@ export interface ServerRenderOptions extends RenderOptions {
 
 // @alpha
 export interface ServerRequestEvent<T = any> {
+    // (undocumented)
+    contentSecurityPolicy?: ContentSecurityPolicy;
```
manucorporat, line 159: no changes needed

packages/qwik-city/middleware/node/index.ts
```diff
@@ -31,7 +31,15 @@ export function createQwikCity(opts: QwikCityNodeRequestOptions) {
     next: NodeRequestNextFunction
   ) => {
     try {
-      const serverRequestEv = await fromNodeHttp(getUrl(req), req, res, 'server');
+      const serverRequestEv = await fromNodeHttp(
+        getUrl(req),
+        req,
```
manucorporat, line 36: revert all this, no needed

DustinJSilk:
aaaah, i see. Sorry, i didnt fully follow.
Thanks for the time! I'll get this corrected.

DustinJSilk:
Clever! I didnt realise that was actually possible. 
Ive made some updates.

Thanks!

manucorporat:
Amazing feature my friend!!

tzdesign:
@DustinJSilk How did you implement it now? 
It would be nice to document. I created a plugin@nonce.ts and saw actually the @nonce in the map in all the routes, but the script tags lags in the attribute.

Didn't find anything about it in the docs or in discord.
image

Or is this Script-Tag only?  How we are doing with qwiks own script tags or I missing something? Because I would prefer not having unsafe-inline or even unsafe-eval in the headers. 

Am I missing something?

DustinJSilk:
You can see a basic example of the `plugin@nonce.ts` file [here](https://github.com/DustinJSilk/qwik-nonce/blob/main/src/routes/plugin%40nonce.ts)

The sharedMap is a way to share data across all your loaders and endpoints. It's also accessible by the Qwik runtime.
Adding a field `@nonce` to the sharedMap will be automatically picked up by Qwik and added to any internal script tags.

It's then up to you to add that nonce to any scripts in your code base.
The default ones are in the root.tsx and the router-header

Qwik will add the nonce from the sharedMap to the serverData, making it available to your components at render time. You can grab it with this line:

`const nonce = useServerData<string | undefined>("nonce");`

[Here](https://github.com/DustinJSilk/qwik-nonce/blob/main/src/root.tsx) is an example of it being added to the ServiceWorker script. I recommend then passing that into your router-head and adding the nonce to the scripts there

DustinJSilk:
Your browser is likely just hiding the nonce. Try and view the raw page source to see if the nonce exists everywhere.
Or check the console for CSP errors.

tzdesign:
image

You're right! WTF? Chrome

I need to investigate further, CSP is still reporting this vite-error-overlay:
image

Obviously it will not be in production, but I want CSP to work in dev so that all devs know what is happening.


Thank you so much for your support.

DustinJSilk:
Unfortunately I can’t help with that. On my dev environment I don’t use a nonce, I just check it works on a preproduction environment

PR #3010, opened 2023-02-17
feat: improve eslint messages

manucorporat:


packages/eslint-plugin-qwik/qwik.unit.ts
```diff
@@ -417,7 +417,9 @@ export default component$(() => {
           );
         });`,
         errors: [
-          'JSX attributes that end with $ can only take an inlined arrow function of a QRL identifier. Make sure the value is created using $()',
+          {
+            messageId: 'unvalidJsxDollar',
```
shairez, line 421: correction: should be "invalid"

manucorporat: Done!

PR #3074, opened 2023-02-22
fix: test mode

### LINKED ISSUE #3063: [🐞] Error: Missing URL when button clicked in v0.19.0 with vitest

Qwik Runtime

### Describe the bug

After upgrading to the latest version of Qwik, vitest can't successfully click a button with a QRL parameter. It fails with the following error: `Error: Missing URL` which is being thrown here: https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/testing/platform.ts#L20

### Reproduction

https://github.com/DustinJSilk/qwik-btn/blob/main/src/components/button/button.spec.tsx

### Steps to reproduce

The following code can be found in the above repository.

- Clone & install deps
- Run `pnpm test.unit`

```tsx
export const MyButton = component$((props: { onClick$: QRL<() => any> }) => {
  return <button onClick$={props.onClick$}></button>
})

test(`[MyButton]: should fire onClick$ event`, async () => {
  const { render, userEvent } = await createDOM()
  const signal = { value: 0 }

  await render(
    <QwikCityMockProvider>
      <MyButton onClick$={() => signal.value++}></MyButton>
    </QwikCityMockProvider>,
  )

  await userEvent('button', 'click')
  expect(signal.value).toEqual(1)
})
```

### System Info

```shell
System:
    OS: macOS 12.0.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 234.66 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 7.13.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 109.0.5414.119
    Firefox: 110.0
    Safari: 15.1
  npmPackages:
    @builder.io/qwik: 0.19.0 => 0.19.0 
    @builder.io/qwik-city: 0.3.0 => 0.3.0 
    undici: 5.20.0 => 5.20.0 
    vite: 4.1.3 => 4.1.3
```

### Additional Information

_No response_

manucorporat:
fixes #3063

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/plugin.ts
@@ -109,10 +109,12 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
 
     opts.debug = !!updatedOpts.debug;
 
+    updatedOpts.target === 'test';
     if (
       updatedOpts.target === 'ssr' ||
       updatedOpts.target === 'client' ||
-      updatedOpts.target === 'lib'
+      updatedOpts.target === 'lib' ||
+      updatedOpts.target === 'test'
     ) {
       opts.target = updatedOpts.target;
     } else {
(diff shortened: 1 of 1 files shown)


packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -109,10 +109,12 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
 
     opts.debug = !!updatedOpts.debug;
 
+    updatedOpts.target === 'test';
```
shairez, line 112: @manucorporat I think this should be removed right?

PR #3092, opened 2023-02-22
fix: q:key containing double quotes does not break elements on server side render

christinatka:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Fixes: https://github.com/BuilderIO/qwik/issues/3091

Using strings that contain double quotes as the key for elements breaks on server-side render. Expect a string with double quotes to still work as a valid q:key.

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/ssr/render-ssr.ts
+++ b/packages/qwik/src/core/render/ssr/render-ssr.ts
@@ -655,7 +655,7 @@ This goes against the HTML spec: https://html.spec.whatwg.org/multipage/dom.html
       }
     }
     if (key != null) {
-openingElement += ' q:key="' + escapeAttr(key) + '"';
+      openingElement += ' q:key="' + escapeAttr(key) + '"';
     }
     if ('ref' in props || useSignal || listeners.length > 0) {
       if ('ref' in props || useSignal || listenersNeedId(listeners)) {
--- a/packages/qwik/src/core/render/ssr/render-ssr.ts
+++ b/packages/qwik/src/core/render/ssr/render-ssr.ts
@@ -655,7 +655,7 @@ This goes against the HTML spec: https://html.spec.whatwg.org/multipage/dom.html
       }
     }
     if (key != null) {
-      openingElement += ' q:key="' + key.replace(/"/g, '&quot;') + '"';
+openingElement += ' q:key="' + escapeAttr(key) + '"';
     }
     if ('ref' in props || useSignal || listeners.length > 0) {
       if ('ref' in props || useSignal || listenersNeedId(listeners)) {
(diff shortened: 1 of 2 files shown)


zanettin:
Hi @christinatka 👋 
Thank you very much for taking the time to contribute 🙏  Really appreciated!
Just thinking about if we should replace a quote or just strip it away since it shouldn't be set by purpose imo 😄

jessezhang91:
@zanettin we discovered this yesterday while using dynamic values that contained a double quote. our original solution was to just url encode the values but we noticed that the browser rendered value worked properly (escaped quote in html) while the server did no encoding.

I think stripping quotes would also work but would then require a change to the browser rendered approach as well while this just fixes server rendered to be consistent with the browser rendered.

zanettin:
Hi @jessezhang91 👋 
Thanks a lot for your reply 🙏 
Just had some time to debug the mentioned issue. IMO the current client output in that case is incorrect as well and i can't see an escaped value on the client neither 😞

### client
Bildschirmfoto 2023-02-24 um 22 24 50

### server
Bildschirmfoto 2023-02-24 um 22 23 54

so from my pov we need to tackle both ends, unless you're already a step further with your debugging.
thanks once again 🙏

jessezhang91:
@zanettin i _think_ the value in the client one that you showed there is escaped. otherwise the red style would be treated as an attribute rather than part of the q:key string.

if you take a look at the fixed ssr version (from this PR) it looks the same as the client one from your screenshot

packages/qwik/src/core/render/ssr/render-ssr.ts
```diff
@@ -655,7 +655,7 @@ This goes against the HTML spec: https://html.spec.whatwg.org/multipage/dom.html
       }
     }
     if (key != null) {
-      openingElement += ' q:key="' + key + '"';
+      openingElement += ' q:key="' + key.replace(/"/g, '&quot;') + '"';
```
zanettin, line 658: just have seen that there is already a fn which can be used for the escaping:

```suggestion
openingElement += ' q:key="' + escapeAttr(key) + '"';
```

zanettin (commented):
Hi @jessezhang91 👋 

Jup u'r right! reproduced it now locally and tested the MPA and SPA outputs. It works and the produced markup is valid (just looks strange in the dev tools).
Just added a tiny input. Found a fn which does the escaping already within this file so we could reuse this one as well 👍

packages/qwik/src/core/render/ssr/render-ssr.ts
manucorporat: @zanettin is right here! we should use the escapeAttr() function

zanettin:
Hi @christinatka 👋 
Thanks a lot for the adjustments 🙏 
can you please also run `(p)npm fmt` to format the code? CI is complaining 👼 🙏

christinatka:
@zanettin fixed the indent!

zanettin (approved):
thanks a lot @christinatka 🙏  and welcome in the contributor family 🤩 . as soon as the CI is happy, i'll merge this pr 👍

Issue #2959, opened 2023-02-13
[✨] Consider making { eagerness: "deep" } the default for useStore


nnelgxorz:
### Is your feature request related to a problem?

New users very often hit problems when updating values in a store. 

The issue is that useStore will only updated the top level values by default. This is good for performance, but is not intuitive for new users who expect updating nested values to 'just work'.

To make matters worse, there is currently no feedback to users that updating nested values without setting the `{ eagerness: "deep" }` will not cause an update. This leads to confusion and frustration.

### Describe the solution you'd like

Make `{ eagerness: 'deep' }` the default and add notes to the docs about changing to different values for performance improvements.

### Describe alternatives you've considered

A linting rule or optimizer warning that alerts users that their deep update will not work as expected.

zanettin:
love the linter approach! imo changing the default is suboptimal due to the fact that we would "encourage" users to take the slower solution. but i totally see the point and have made this mistake as well when i started 👼

tzdesign:
If you can lint this out I’m fine es well. But we have 3 states in a big app. All deep=true.

zanettin:
guess most of the devs will end with a few states and use signals for the rest, true. imo we should just somehow make sure, they stay on the "right" path or leave it because of reasons and not accidentally 😅

manucorporat:
yeah... i think the default is ok, in fact signals should be prefered

manucorporat:
the `deep` comes with a lot of cost, most of the times it would be better without it, but requires more work for developers

cmbartschat:
Is it worth typing data from a shallow store as read-only below the top level? It would mark any mutations on non-deep data as invalid at the type level, not sure if that's too severe or breaks use cases, but it will definitely catch this earlier for people.

```
[...]

PR #3132, opened 2023-02-24
[draft] feat: set nested data in useStore to read only by default

cmbartschat:
RE #2959

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

As seen in #2959 , there is interest in preventing shallow stores from being mutated, except at the top level. This prevents subtle bugs with reactivity, but preserves the default behavior of only installing a proxy at the root of the store.

# Why is this still a draft?

- [ ] Decide if signal should also be equivalently readonly
- [ ] Align on expected behavior for types like `Element`/`Set`
- [ ] Make sure there isn't any code in docs/starters that expects deeply mutable stores
- [ ] Confirm that this is considered a breaking change

# Checklist:


manucorporat:
We decided to make useStore() deep true by default!
https://github.com/BuilderIO/qwik/issues/2959

PR #3192, opened 2023-02-28
chore: release 0.20.0

manucorporat:


ahnpnl:
for some reasons `qwik-city` **0.5.0** was published into npm without this MR

PR #3200, opened 2023-02-28
feat: useComputed$()

manucorporat:
```tsx
export const ComputedRoot = component$(() => {
  const count = useSignal(0);
  const double = useComputed$(() => count.value * 2);
  const plus3 = useComputed$(() => double.value + 3);
  const triple = useComputed$(() => plus3.value * 3);
  const sum = useComputed$(() => double.value + plus3.value + triple.value);

  return (
    <div>
      <div>count: {count.value}</div>
      <div>double: {double.value}</div>
      <div>plus3: {plus3.value}</div>
      <div>triple: {triple.value}</div>
      <div>sum: {sum.value}</div>
      <button onClick$={() => count.value++}>Increment</button>
    </div>
  );
});
```

This API would be equivalent to a Signal + useTask + track + auto initial value, without this API you will need to write:
```tsx
export const ComputedRoot = component$(() => {
  const count = useSignal(0);
  const doublePlusOne = useSignal(1); // notice we need to pass a dumb initial value
  useTask$(({track}) => {
    doublePlusOne.value = track(() => count.value) * 2 + 1;
  });

  return (
    <div>{doublePlusOne.value}</div>
  );
});
```

with new API:

```tsx
export const ComputedRoot = component$(() => {
  const count = useSignal(0);
  const doublePlusOne = useComputed$(() => count.value * 2 + 1);
  return (
    <div>{doublePlusOne.value}</div>
  );
});
```

NOTICE:
- Auto tracking
- No need to create a signal with a bogus value
- One single API
- Much less nested and boilerplate
- Faster!


jherr:
Love it!

solamichealolawale:
Hi, I am building for SSG and I am currently facing this computed issue. So my question right now is, can I still go for the useTask$, useSignal & track version for SSG? Since useTask$ will run both in server and client and I am not sure of the way forward now. Thanks in advance

phcoliveira:
This might be what I have been looking for as a readonly reactive value. Svelte has writable and readable stores, and I was missing the latter.

So I presume that something like this would be equivalent to readable stores:
```
const writable = useSignal(10);
const readable = useComputed$(() => writable.value);

useContextProvider(MyContext, readable);
```

What do you think, @manucorporat?

PR #3209, opened 2023-03-01
refactor(qwik): add css properties to jsx html attributes

huijiewei:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

add css properties to jsx html attributes.

# Use cases and why

none

# Checklist:

### DIFF
--- a/packages/qwik/package.json
+++ b/packages/qwik/package.json
@@ -133,6 +133,9 @@
     "testing/index.d.ts",
     "testing/package.json"
   ],
+  "dependencies": {
+    "csstype": "^3.1.1"
+  },
   "devDependencies": {
     "@builder.io/qwik-dom": "2.1.19",
     "kleur": "4.1.5"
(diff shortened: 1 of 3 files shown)


manucorporat:
Adding a new dependency is not an easy decision for us, happy to add this functionality, but i would like to avoid the extra dependency

huijiewei:
> Adding a new dependency is not an easy decision for us, happy to add this functionality, but i would like to avoid the extra dependency

css-type only typescript type define.

```
import type * as csstype from 'csstype';
```

dist will drop all.

gioboa:
@huijiewei thanks for your help. `csstype` is a dependency in our codebase right now. The PR is old and there are few conflicts to solve, the best way to proceed is to create a new fresh PR with the latest codebase. 
Are you still interested in it?

mhevery:
Thank you for your contribution. I am going to close this as a similar PR https://github.com/BuilderIO/qwik/pull/4577 has already landed and has added this dependency.

PR #3211, opened 2023-03-01
fix(doc-search): unable to prevent default

vedantnn71:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
The doc search component's {meta|ctrl}k keybinding was not working correctly. The default browser action was triggered along with the opening of `DocSearch` component.

# Use cases and why
To prevent the default browser behavior while opening `DocSearch` component.

- 1. Current behavior - the `DocSearch` component opening along with the default browser behavior. (for eg, focusing of address bar in chromium based browsers)
- 2. Expected behavior - prevent default browser behavior while opening `DocSearch` component.

# Checklist:

### DIFF
--- a/packages/docs/src/components/docsearch/doc-search.tsx
+++ b/packages/docs/src/components/docsearch/doc-search.tsx
@@ -1,5 +1,11 @@
 import type { SearchClient } from 'algoliasearch/lite';
-import { component$, useStore, useStyles$, useSignal } from '@builder.io/qwik';
+import {
+  component$,
+  useStore,
+  useStyles$,
+  useSignal,
+  useBrowserVisibleTask$,
+} from '@builder.io/qwik';
 import type { DocSearchHit, InternalDocSearchHit, StoredDocSearchHit } from './types';
 import { ButtonTranslations, DocSearchButton } from './doc-search-button';
 import { DocSearchModal, ModalTranslations } from './doc-search-modal';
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/docsearch/doc-search.tsx
```diff
@@ -1,5 +1,5 @@
 import type { SearchClient } from 'algoliasearch/lite';
-import { component$, useStore, useStyles$, useRef } from '@builder.io/qwik';
+import { component$, useStore, useStyles$, useRef, useBrowserVisibleTask$ } from '@builder.io/qwik';
```
zanettin, line 2: since we're already on this file: could we replace the deprecated `useRef` with `useSignal` als well? 👼

packages/docs/src/components/docsearch/doc-search.tsx
```diff
@@ -66,43 +66,43 @@ export const DocSearch = component$((props: DocSearchProps) => {
...
-          (event.key === 'k' && (event.metaKey || event.ctrlKey)) ||
-          // The `/` shortcut opens but doesn't close the modal because it's
-          // a character.
-          (!isEditingContent(event) && event.key === '/' && !state.isOpen)
-        ) {
-          // FIXME: not able to prevent
-          // event.preventDefault();
+  useBrowserVisibleTask$(() => {
+    window.addEventListener('keydown', (event) => {
+      event.preventDefault();
```
zanettin, line 71: with this in place, you're not able to type anything into the search form since all key strokes are prevented atm 😄

zanettin (changes requested):
Hi @vedantnn71 👋 
Thank you very much for creating this PR 🙏  i can't reproduce the issue you try to fix with this PR but i guess i am on another env than you're testing on.
just left you two comments regarding the impl 👼 
thanks once again 🙏

packages/docs/src/components/docsearch/doc-search.tsx
vedantnn71: I have fixed that :)

vedantnn71:
Hey @zanettin 👋🏻 

Thank you for reviewing my pr. My bad, now only the {meta|ctrl}+k keybinding's default action is prevented.

Also, the issue is that in linux or windows machines, the `DocSearch` component's keybinding triggers the default browser behavior along with the opening of `DocSearch` component. See:

https://user-images.githubusercontent.com/70624701/222329626-2e57ea58-8bf2-459f-9dfe-42253a0ce9f5.mp4

zanettin:
Hi @vedantnn71 👋 
The solution is working. I just pushed the mentioned changes regarding `useRef` => `useSignal` changes. I guess it's ready to merge but would love to get a feedback from anybody else because of the other changes. Thanks once again 🙏

zanettin (approved):

vedantnn71:
ty for approving the changes @zanettin, I tried it running locally and it is working fine. :)

zanettin:
Hi @vedantnn71 👋 
Just had an idea of how we could keep the current handling w/o leveraging `useVisibleTask`:
```
  return (
    <div
      class="docsearch"
      preventdefault:keyDown={!state.isOpen}                   // <= PREVENT DEFAULT the qwik way
      window:onKeyDown$={(event) => {
        function open() {
          // We check that no other DocSearch modal is showing before opening
          // another one.
          if (!document.body.classList.contains('DocSearch--active')) {
            state.isOpen = true;
          }
        }
        
        ...
        
```
can you pls try that on your end as well and if it works update the PR? looked good on my end so far 👍

vedantnn71:
Hello @zanettin, tested it out in the qwik way like you suggested, everything seems to be working fine with that.

pushed the changes :)

packages/docs/src/components/docsearch/doc-search.tsx
```diff
@@ -64,43 +64,45 @@ export const DocSearch = component$((props: DocSearchProps) => {
...
     <div
       class="docsearch"
+      preventdefault:keyDown={!state.isOpen}
       window:onKeyDown$={(event) => {
-        function open() {
-          // We check that no other DocSearch modal is showing before opening
-          // another one.
-          if (!document.body.classList.contains('DocSearch--active')) {
-            state.isOpen = true;
+        window.addEventListener('keydown', (event) => {
```
zanettin, line 74: imo this line should be obsolete now since we already have a `keyDown` listener registered on L73 😃

zanettin (commented):
thanks once again @vedantnn71 🙏  just a tiny finding 👼

packages/docs/src/components/docsearch/doc-search.tsx
vedantnn71: ty for pointing out, my bad tho it doesn't work again when using ctrl-k to close the doc search component :(

I think we should bind it to window only rather than the div

https://user-images.githubusercontent.com/70624701/222942166-55cd1194-0272-4276-9a1e-4276391a9228.mp4

zanettin: oh no 🤣  i mean we could try to debounce the update but this is a bit over the top. i am also not sure, why this fn needs so many triggers. you can open it atm using `cmd + k` or `/`. to close it you can hit `cmd + k` and `esc`.
just compared it with https://vercel.com/docs and it has the same behavior as on your video.
for me the important thing is, that you can open it with that shortcut. the goal is not to close it right away again but finding the page you're looking for 😄 so maybe we could remove `cmd + k` from the close cmd as well.

vedantnn71: yeah i agree with you, i think it would be better to just rm `cmd + k` binding 😄

zanettin: Hi @vedantnn71 sry for the late answer 🙈 
thanks for the replies 🙏  imo we should remove the `window.eventListener` as mentioned above and also the `event.preventDefault()` on line 90. and then we should be defenetely good to go 🚀

vedantnn71: ty @zanettin, rmed prevent default 😄

zanettin (approved):
thank u very much @vedantnn71 🎉  let's merge this 🙏

PR #3232, opened 2023-03-02
The 404 page on root dir does not render

keuller:
# What is it?

- [ ] Feature / enhancement
- [X] Bug Fix
- [ ] Docs / tests

# Description

According to docs when we create `/404.tsx` on root dir it should be rendered, but when we use SSR with Express it does not happen, even if Qwik generates `404.html` file.

Related issue: [1560](https://github.com/BuilderIO/qwik/issues/1560)

# Use cases and why

- 1. Now the behavior is exact what we expect according to docs.

# Checklist:


manucorporat:
Let's get this review by @adamdbradley !

PR #3237, opened 2023-03-02
docs(useLocation): extend docs with deeply nested child routes

jakovljevic-mladen:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Extend docs by explaining `...` routes syntax with an example.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/qwikcity/api/index.mdx
+++ b/packages/docs/src/routes/qwikcity/api/index.mdx
@@ -75,9 +75,9 @@ Use the [`useLocation()`](/qwikcity/api/index.mdx#uselocation) function to retri
 ```tsx
 export interface RouteLocation {
   readonly params: RouteParams; // Record<string,string>
-  readonly href: string;
-  readonly pathname: string;
-  readonly query: Record<string, string>;
+  readonly href: string; // @deprecated Use `url` instead
+  readonly pathname: string; // @deprecated Use `url` instead
+  readonly query: Record<string, string>; // @deprecated Use `url` instead
   readonly isNavigating: boolean;
   readonly url: URL;
 }
@@ -91,11 +91,12 @@ The return value of [`useLocation()`](/qwikcity/api/index.mdx#uselocation) is si
 
 Assume you have:
 
-- Route: `https://example.com/sku/[skuId]`
+- File: `src/routes/sku/[skuId]/index.tsx`
 - User navigates to: `https://example.com/sku/1234`
-- Then the `skuId` can be retrieved through `useLocation().params.skuId`.
+- Then the `skuId` can be retrieved through `useLocation().params.skuId`
 
 ```tsx
+// src/routes/sku/[skuId]/index.tsx
 import { component$ } from '@builder.io/qwik';
 import { useLocation } from '@builder.io/qwik-city';
 
@@ -116,8 +117,40 @@ The above code would generate:
 
 ```html
 <h1>SKU</h1>
-<p>pathname: /sku/1234</p>
+<p>pathname: /sku/1234/</p>
 <p>skuId: 1234</p>
+```
+
+You can also use `...` syntax to indicate deeply nested child routes. For example:
+
+- File: `src/routes/post/[...all]/index.tsx`
+- User navigates to: `https://example.com/post/2023/id/title`
+- Then the `all` can be retrieved through `useLocation().params.all`
+
+```tsx
+// src/routes/post/[...all]/index.tsx
+import { component$ } from '@builder.io/qwik';
+import { useLocation } from '@builder.io/qwik-city';
+
+export default component$(() => {
+  const loc = useLocation();
+
+  return (
+    <>
+      <h1>POST</h1>
+      <p>pathname: {loc.url.pathname}</p>
+      <p>all: {loc.params.all}</p>
+    </>
+  );
+});
+```
+
+The above code would generate:
+
+```html
+<h1>POST</h1>
+<p>pathname: /post/2023/id/title/</p>
+<p>all: 2023/id/title</p>
 ```
 
 > Notice that `useLocation` is a read-only API, you should never attempt to mutate the values of the `loc` object returned. Instead look into the `useNavigate()` API.
--- a/packages/docs/src/routes/qwikcity/api/index.mdx
+++ b/packages/docs/src/routes/qwikcity/api/index.mdx
@@ -74,12 +74,9 @@ Use the [`useLocation()`](/qwikcity/api/index.mdx#uselocation) function to retri
 
 ```tsx
 export interface RouteLocation {
-  readonly params: RouteParams; // Record<string,string>
-  readonly href: string; // @deprecated Use `url` instead
-  readonly pathname: string; // @deprecated Use `url` instead
-  readonly query: Record<string, string>; // @deprecated Use `url` instead
-  readonly isNavigating: boolean;
+  readonly params: RouteParams; // Readonly<Record<string, string>>
   readonly url: URL;
+  readonly isNavigating: boolean;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


packages/docs/src/routes/qwikcity/api/index.mdx
```diff
@@ -75,9 +75,9 @@ Use the [`useLocation()`](/qwikcity/api/index.mdx#uselocation) function to retri
 ```tsx
 export interface RouteLocation {
   readonly params: RouteParams; // Record<string,string>
-  readonly href: string;
-  readonly pathname: string;
-  readonly query: Record<string, string>;
+  readonly href: string; // @deprecated Use `url` instead
```
zanettin, line 78: can we use the current typing here which is:
```
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly url: URL;
  /**
   * @deprecated Please use `url` instead of href
   */
  readonly href: string;
  /**
   * @deprecated Please use `url` instead of pathname
   */
  readonly pathname: string;
  /**
   * @deprecated Please use `url` instead of query
   */
  readonly query: URLSearchParams;
  readonly isNavigating: boolean;
}
```
but w/o the deprecated keys 😄

zanettin (commented):
Hi @jakovljevic-mladen 👋 
Great updates 🙏  thanks a lot! left two comments and would love to hear your feedback 🙏

packages/docs/src/routes/qwikcity/api/index.mdx
```diff
@@ -116,10 +117,42 @@ The above code would generate:
 
 ```html
 <h1>SKU</h1>
-<p>pathname: /sku/1234</p>
+<p>pathname: /sku/1234/</p>
 <p>skuId: 1234</p>
 ```
 
+You can also use `...` syntax to indicate deeply nested child routes. For example:
```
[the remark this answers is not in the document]
jakovljevic-mladen: Yes, sort of. The page you mentioned never showed an example of using `params` that represent deeply nested routes (that are located within folder that starts with `[...`). I thought that the API Reference page should cover the details, though I'm not sure did the right job. Maybe propose something? 🙂

BTW, I somehow read that the type of `params` filed was `Readonly<Record<string, any>>` (instead of `Readonly<Record<string, string>>`) which got me confused (by my mistake) that `params` properties can have any value. This is why I wanted to document this so that other people don't get confused as well, but they probably won't since `params.anyParamHere` will always be `readonly string`.

packages/docs/src/routes/qwikcity/api/index.mdx
jakovljevic-mladen: > but w/o the deprecated keys 😄

How? 😄 

1:

```ts
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly url: URL;
  /**
   * @deprecated Please use `url` instead of href
   */
  readonly href: string;
  /**
   * @deprecated Please use `url` instead of pathname
   */
  readonly pathname: string;
  /**
   * @deprecated Please use `url` instead of query
   */
  readonly query: URLSearchParams;
  readonly isNavigating: boolean;
}
```

2:

```ts
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly url: URL;
  /**
   * Please use `url` instead of href
   */
  readonly href: string;
  /**
   * Please use `url` instead of pathname
   */
  readonly pathname: string;
  /**
   * Please use `url` instead of query
   */
  readonly query: URLSearchParams;
  readonly isNavigating: boolean;
}
```

or 3:

```ts
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly url: URL;
  readonly isNavigating: boolean;
}
```

or maybe something like this (4):

```ts
export interface RouteLocation {
  readonly params: Readonly<Record<string, string>>;
  readonly url: URL;
  readonly isNavigating: boolean;

  // deprecated...
  readonly href: string;
  readonly pathname: string;
  readonly query: URLSearchParams;
}
```

zanettin (commented):
Thank you very much for the additions @jakovljevic-mladen 🙏 

try to sum it up here what my suggestion would be:

- https://qwik.builder.io/qwikcity/api/#uselocation
  - update the API docs with your proposal number 3 🎉 
- https://qwik.builder.io/qwikcity/routing/#dynamic-route-segments
  - extend this part to also reflect your suggested changes. on this page the deeply nested catch all is described as well and maybe we can expand it a bit to make it clearer to everybody
  
what do you think?

jakovljevic-mladen:
>what do you think?

@zanettin, agreed. Will update the code.

packages/docs/src/routes/qwikcity/api/index.mdx
```diff
@@ -74,12 +74,9 @@ Use the [`useLocation()`](/qwikcity/api/index.mdx#uselocation) function to retri
...
 ```tsx
 export interface RouteLocation {
-  readonly params: RouteParams; // Record<string,string>
-  readonly href: string;
-  readonly pathname: string;
-  readonly query: Record<string, string>;
-  readonly isNavigating: boolean;
+  readonly params: RouteParams; // Readonly<Record<string, string>>
   readonly url: URL;
+  readonly isNavigating: boolean;
```
jakovljevic-mladen, line 79: Switched positions of `url` and `isNavigating` since `url` is gonna be much more used property, plus [it](https://github.com/BuilderIO/qwik/blob/9cbb12d7915865ed1b3dfb12bf74f45cd56c9e84/packages/qwik-city/runtime/src/types.ts#L60) is positioned before [`isNavigating`](https://github.com/BuilderIO/qwik/blob/9cbb12d7915865ed1b3dfb12bf74f45cd56c9e84/packages/qwik-city/runtime/src/types.ts#L73) in the interface declaration.

packages/docs/src/routes/qwikcity/routing/index.mdx
```diff
@@ -50,8 +50,8 @@ src/
 Special named directories with square brackets, such as `[paramName]` and `[...catchAll]` can be used to match route segments which are dynamic:
 
 ```
-src/routes/blog/index.tsx → /blog/:slug (/blog/hello-world)
```
jakovljevic-mladen, line 53: This wasn't aligned with line 62.

packages/docs/src/routes/qwikcity/routing/index.mdx
```diff
@@ -50,8 +50,8 @@ src/
...
-src/routes/user/[username]/index.tsx → /user/:username/settings (/user/foo/)
```
jakovljevic-mladen, line 54: `/settings` part never existed, so I removed it. Also, example `(/user/foo/)` never had this route part.

packages/docs/src/routes/qwikcity/routing/index.mdx
```diff
@@ -50,8 +50,8 @@ src/
...
+src/routes/blog/index.tsx → /blog
+src/routes/user/[username]/index.tsx → /user/:username (/user/foo)
 src/routes/post/[...all]/index.tsx → /post/* (/post/2020/id/title)
```
jakovljevic-mladen, line 55: I'd replace `/post/*` part with something more standard than it is now. Not sure how though. Maybe:

```
/post/:all (/post/2020/id/title)
```

But that does not indicate catch-all routes that can be deeper than one level. Or:

```
/post/:...all (/post/2020/id/title)
```

though this one ☝️ looks ugly. Or maybe:

```
/post/[:all] (/post/2020/id/title)
```

[Next.js](https://nextjs.org/docs/routing/dynamic-routes#optional-catch-all-routes) uses this:

```
/post/[[...all]] (/post/2020/id/title)
```

Though, I'd rather go with `:` instead of `...`:

```
/post/[[:all]] (/post/2020/id/title)
```

@zanettin any suggestion? 😄

packages/docs/src/routes/qwikcity/routing/index.mdx
```diff
@@ -62,7 +62,7 @@ src/
     │   └── index.tsx         # https://example.com/blog
     ├── post/
     │   └── [...all]/
-    │       └── index.tsx     # https://example.com/post/2020/id/title/2020/12/03
```
jakovljevic-mladen, line 65: Example above (line 55) used route example without additional date `(/post/2020/id/title)`, so I removed it from here.

packages/docs/src/routes/qwikcity/routing/index.mdx
zanettin: good question 🤔 i'd go with v1 (`:all`). in the code you still have a single string containing the whole path. guess the notations in brackets could be misleading. just my thoughts.

zanettin: agree 👍 let's keep the wildcard for now. should be clear in the full context 🙏

zanettin:
thanks once again @jakovljevic-mladen 🙏  let's merge this 🚀

PR #3242, opened 2023-03-02
fix(qwik-auth): keep the trailing slash in the default callback url

ulic75:


zanettin:
Hi @ulic75 👋 
Would you mind to push an empty commit to re-trigger the CI? smth is blocking it 👼 🙏

ulic75:
forced an empty commit this should be good now.

Issue #3149, opened 2023-02-25
[🐞] actions from the Form are vanished in the history.


genki:
### Which component is affected?

Qwik City (routing)

### Describe the bug

When to make a search form with the form component, it seemed there is no way to record the action in the browser history.
Suppose a man has searched words "Foo", "Bar" and then "Baz" and wanted to back to the result of "Foo".
But if he clicked the back button then the browser shows the page everything has not happened yet.
I am wondering on it is designed behavior or a bug.

### Reproduction

https://stackblitz.com/edit/qwik-starter-ry1wpp?file=src/routes/index.tsx

### Steps to reproduce

Please go to the reproduction link and click the `Open in New Tab` button at the top rigth corner.
And then type some words and click "submit" button. You can find no history has made.
This is not depending on if `reloadDocument` is set or not.

### System Info

```shell
System:
    OS: macOS 13.1
    CPU: (8) arm64 Apple M2
    Memory: 77.92 MB / 24.00 GB
    Shell: 3.5.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 19.5.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.3.1 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 110.0.5481.100
    Firefox: 110.0
    Safari: 16.2
  npmPackages:
    @builder.io/qwik: file:../clone/qwik/packages/qwik => 0.19.2 
    @builder.io/qwik-city: file:../clone/qwik/packages/qwik-city => 0.4.0 
    undici: ^5.20.0 => 5.20.0
```

genki:
I think it is best if there can be a way to store the states made by `useStore` or `useSignal` and viewing parameters such as the scroll position in the history and restore it when the browser goes back.

gioboa:
Hi @genki, thanks for opening this issue.
A lot of things happened during this months, with the latest version of Qwik is it still a valid issue?

gioboa:
@genki you can use [useNavigate](https://qwik.builder.io/docs/api/#usenavigate) to add your query strings to the history.
with routeAction$ I didn't find the way to redirect to the same path + query strings
So I'm closing this issue for now, feel free to re-open it if it's still an issue for you.

Issue #2162, opened 2022-11-16
Adding and removing query params from url with useNavigate scrolls to top


jweb89:
### Qwik Version

qwik-city:0.0.122 qwik:0.13.3 stackblitz is different version due to stackblitz reloading issues

### Operating System (or Browser)

Browser chrome

### Node Version (if applicable)

16

### Which component is affected?

Qwik City

### Expected Behaviour

When adding or removing query params with ? using useNavigate the page gets scrolled to top

### Actual Behaviour

When adding or removing query params with ? using useNavigate the page should not scrollToTop

### Additional Information

https://stackblitz.com/edit/qwik-starter-ps4otn?file=src%2Fcomponents%2Ftest%2Findex.jsx

To test this it's probably easier to copy the stackblitz url in a new window. The qwik and qwik-city versions are a little older but I had issue with the server reloading in stackblitz on the latest. In my local environment the issue still exists with the latest versions qwik-0.13.3 and qwik-city:0.0.122

the-zimmermann:
@manucorporat this is fixed in 0.18.1.

PR #3244, opened 2023-03-03
[✨] Critical DX/UX improvement to SPA navigation in Qwik City

### LINKED ISSUE #2202: qwik-city: add events to be able to react to the client side navigation

billykwok:
# What is it?

- [x] Feature / Enhancement
- [x] Bug / Fixes
- [x] Docs / tests

# Overview

This is a relatively straightforward change that addresses a number of interrelated issues around SPA navigation in Qwik City. Some are already mentioned in other issues (#2162, #2050, #3149, #2202, #2203), while some are newly reported here.

### Support MPA-like and customizable scroll restoration in SPA

**Before**: Qwik City always scrolls to the top upon SPA navigation, unless there is a hash segment in the URL. This is not consistent with how browsers handle MPA scroll restoration on `popstate` and creates bad UX when users return to a page with a resetted scroll position. Moreover, the behavior is hard-coded into the Qwik City navigation logic, even though different applications have their own needs (e.g. app-like and page-like UIs should restore scroll positions differently).

**After**: The default behavior now mimics how browsers handle scroll restoration in MPA - scroll to the **top** for **new pages** (e.g. `<Link />`), **but** to the **last-visited scroll position** for **`popstate` navigation** (e.g. back/next button). Moreover, users can customize scroll restoration with the built-in scroll restorer `toTopAlways` and `toLastPositionOnPopState` or even their own implementation. The 'before' behavior is equivalent to `toTopAlways`.

- This provides feature parity against React Router and Remix
- This provides customizability for users to implement #2162 in userland

### Race condition when restoring scroll positions

**Before**: Scroll position is restored after the DOM is 'settled' in order to synchronize content update and scroll position update. However, the current implementation relies on a series of hard-coded `setTimeout()`, which very often results in either early scrolling or late scrolling.

**After**: 'DOM Settled' is accurately detected using `useLocation()` and `useVisibleTask$()`. Scroll and content updates are now always in-sync.

- This provides feature parity against React Router and Remix
- This fixes #2050 and #3149

### Prevent full-page reload when `popstate` happens before manual navigation

**Before**: Qwik City registers a temporary `popstate` event listener to reload the page if there is any `popstate` navigation before the first `pushState` is called. This essentially turns all SPAs into MPA whenever users refresh the page. This also makes page navigation inconsistent before and after the first `pushState`. In addition, there are unnecessary `popstate` listener registration/unregistration, because the temporary listener will be replaced by a real listener anyway after the first `pushState`.

**After**: Instead of a temporary `popstate` listener that does full-page refreshes, Qwik City now directly sets up the real `popstate` listener that updates the page without full-page refreshes.

- This makes Qwik City consistent with all other SPA frameworks in the market.
- This also makes page transition animation across `'popstate'` possible.

### Improve hash navigations

**Before**: Hash navigation is neither updating the URL hash nor written into the session history. The 'hashchange' event is also never fired.

**After**: The new behavior now mimics native behavior - clicking `<a href="#some-hash" />` now updates the page URL hash without triggering unnecessary re-rendering. It also gets written into session history like what browsers would natively do. The 'hashchange' event is also fired.

- This fixes the hash issue mentioned in #3691.

# Backward-compatible API Changes
```tsx
const navigate = useNavigate();
// New signature that accepts an options object
type Options = { type: 'form' | 'link' | 'popstate'; forceReload: boolean; };
const defaultOptions: Options = { type: 'link', forceReload: false };
navigate('/some/pathname', defaultOptions);

// Old signature still works
navigate('/some/pathname', true /* equivalent { forceReload: true } */);
navigate('/some/pathname', false /* equivalent { forceReload: false } */);
```

```tsx
import {
  QwikCityProvider,
  RouterOutlet,
  ServiceWorkerRegister,
  toTopAlways
} from '@builder.io/qwik-city';
import { RouterHead } from './components/router-head/router-head';

export default function Root() {
  return (
    <QwikCityProvider>
      <head>
        <meta charSet="utf-8" />
        <link rel="manifest" href="/manifest.json" />
        <RouterHead />
      </head>
      <body>
        //            ⬇️ New (optional): toTopAlways or toLastPositionOnPopState (default)
        <RouterOutlet restoreScroll$={toTopAlways} />
        <ServiceWorkerRegister />
      </body>
    </QwikCityProvider>
  );
}
```

# Demo

### Scroll restoration after `popstate`

Before the change:
- always to the top

https://user-images.githubusercontent.com/8078716/225979618-ce4e7ec9-996b-4258-915d-1b8dc8799a86.mp4

After the change:
- scroll to the last-visited positions

https://user-images.githubusercontent.com/8078716/225979653-01ea348c-6915-45be-a520-e7cd8fef3f6b.mp4

### `popstate` after reload

Before the change:
- full-page reload the second time

https://user-images.githubusercontent.com/8078716/225979767-131b7914-4b1f-453d-818d-e6ce3a71755a.mp4

After the change:
- re-rendering

https://user-images.githubusercontent.com/8078716/225979800-d8e88838-a56f-44e0-9960-209271830857.mp4

# Checklist


zanettin:
Hi @billykwok 👋  thank you very much for creating this PR 🙏 

I just have a question based on the description: 
>Currently, pages are always scrolled to the top (or hashed element if specified in URLs) upon navigation. While this makes sense for <Link />-triggered navigation, popstate-triggered navigation fails to restore previous scroll positions. This does not match the native behavior of browsers nor that of competing frameworks.

you mention, that the restore for MPA (popstate) navigations does not work atm, which i can't confirm so far. At least not on the available browsers 😄 The `<Link>` component driven navigations are the SPA (pushstate) ones which you also show in the video.

### MPA navigation scroll position restore

https://user-images.githubusercontent.com/3241476/222718122-f5b9a9a5-d70e-4c3a-aba8-df2f0917d680.mp4

Is the description wrong in that case or do i miss something? 👼 
Thanks again for your time and contribution 🙏

billykwok:
Thanks for checking out my change. It is intended for addressing SPA scrolling. When I said popstate, I was referring to popstate navigation in SPA (e.g. re-rendering due to pressing back button after pushState). MPA does not need manual scroll restoration because browsers handle it by default.

billykwok:
Thanks for checking out my change. It is intended for addressing SPA scrolling. When I said popstate, I was referring to popstate navigation in SPA (e.g. re-rendering due to pressing back button after pushState). MPA does not need manual scroll restoration because browsers handle it by default. Hope this clarifies the change.

billykwok:
Converted to draft for now as I found a new way to implement the change that is more flexible and robust.
Will update it with the new implementation in the coming days.

billykwok:
@adamdbradley
The new implementation is now complete. Please take a look and let me know if you have any questions. Thank you in advance.

billykwok:
@zanettin @adamdbradley 
I know that the core team is probably quite busy at the moment and has its own priorities, but I have already needed to resolve merge conflicts more than a couple of times. I would really appreciate it if there is anyone who can at least spend a few minutes looking at this. I'm happy to make changes if the approach of this PR does not align with the team's vision.
Thank you in advance.

billykwok:
Now there seems to be a bug in CI that prevents the GitHub API token from being read.
I think clearing the GitHub Actions may resolve the issue.

zanettin:
thanks again @billykwok 🙏 i try to bring some attention to this PR. and sorry for the delay. the team is hard at work 👼

shairez:
thanks @billykwok for this thoughtful PR!

I say - to save your time, don't merge conflicts until @adamdbradley or @manucorporat will review this change.
And once they'll do, we'll try to give the final reviews more focus so it'll get merged as soon as it's approved.

That way it won't take more of your time, and it'll give the team a chance to carefully review it without the pressure of the other things they're currently trying to get done.

Sounds reasonable?

manucorporat:
Wonder if we can ship this without new APIs:
        <SinglePageNavigation restoreScroll$={toTopAlways} />

can it be part of the QwikCity provider?

billykwok:
Thanks for the suggestion. Yes, this can be done, for example, by putting `<SinglePageNavigation />` inside `<RouterOutlet />` and rendering it conditionally based on an MPA/SPA prop. What do you think about the following API?
```ts
<body>
  <RouterOutlet
    navigation="spa"
    restoreScroll$={toLastPositionOnPopstate} // omitted when navigation="mpa"
  />
</body>
```

billykwok:
I just made a change to directly integrate `<SinglePageNavigation />` into `<RouterOutlet />`. Now the public API should be simpler.

billykwok:
Thank you. Having `navigation` set to `'spa'` will result in a `popstate` listener being registered and a marker for "dom update settled" detection being rendered. All SPAs work like MPA until JS is loaded.
If `navigation` is set to `'mpa'`, the listener and marker code will never be run. Such behavior is suitable for complete MPA like the Qwik Doc site.

Regarding automatic flipping to SPA with `<Link />`, a few versions ago, Qwik City had a similar feature. I personally think that it sounds like a good idea but is hard to implement correctly. Some considerations include:
- SPA can still have pages that have no `<Link />` (i.e. the leaf nodes of the site tree)
- Most of the time, whether a page/site is SPA or MPA is known at SSR time or even compile time. The runtime cost of determining SPA vs MPA becomes pure overhead.
- Users who write their own `<Link />` implementation using `useNavigate()` need to be covered as well.

shairez:
thanks @billykwok 

I agree with @jordanw66 
I think we should strive for a MPA that turns into SPA as I really believe that most MPA (which aren't small documentation websites) will want to become SPAs (to preserve better UX)

billykwok:
This PR does not make it harder to turn MPA into SPA.

Looks like the `navigation` props cause some confusion. The `'mpa'` option is meant for optimizing sites that for sure don't want to be SPAs. Any other sites should be configured with `'spa'`, which makes a site both MPA-first and SPA-enhanced.

If pure MPA is not a use case that Qwik wants to optimize for, I am happy to remove the `navigation` option and include SPA-specific initialization logic even for MPA sites.

billykwok:
I just realized Qwik has been able to avoid running most of my SPA initialization all along. The `popstate` and SPA-specific listeners will only be downloaded and run when the first event arrives (which will never happen for MPA). It also removes the need for opting in/out MPA/SPA. The SPA functionalities are loaded incrementally and naturally as users interact with the page.

I pushed another (and probably the last one) update to this PR to reflect this change.

@jordanw66 
Regarding the edge case that you mentioned for leaf nodes/dead ends, it has been one of the motivations of this PR from the very beginning. It replaces the MPA refresh with a soft SPA re-render after manual refresh at a dead end.

billykwok:
Would you mind clarifying your question? As far as I know, the View Transition API leaves the DOM update control to the users.

If you are implying that View Transition API may be used for detecting DOM 'settled' time, it probably can't. It relies on users to tell how and when the DOM is settled so that it can set up the corresponding pseudo-DOM for transition. It is not related to the DOM settled detection in this PR, which is caused by the asynchronous nature of Qwik and the inability to tell directly when the DOM is updated after a rerun of the component function.

If you are implying that View Transition API can be used for enriching navigation in the future, absolutely! However, I guess Qwik (instead of Qwik City) may be a better place to implement this. Only Qwik has information about when the DOM change is flushed to the actual DOM, and the View Transition API requires this level of control.

manucorporat:
So, i would like to push the limits a bit more, i dont like users to choose between SPA and MPA, qwik is kind of designed to blur the lines a bit. Is it really necessary for developers to choose spa? what's the disadvantage?

manucorporat:
I would really like to merge this API, i am a bit worry about a regression in initial loaded JS and the mutation observer code, not entirely sure, why we need it. But not a fan of that part.

packages/qwik-city/runtime/src/form-component.tsx
```diff
@@ -105,21 +107,23 @@ export const GetForm = component$<FormProps<undefined, undefined>>(
               params.append(key, value);
             }
           });
-          nav('?' + params.toString(), true).then(() => {
```
manucorporat, line 108: This is changing when the event is getting dispatched!

packages/qwik-city/runtime/src/router-outlet-component.ts
```diff
@@ -2,59 +2,43 @@ import {
...
-    let cmp: JSXNode | null = null;
-    for (let i = contentsLen - 1; i >= 0; i--) {
-      cmp = jsx(value[i].default, {
-        children: cmp,
-      });
     }
-    return cmp;
+
+    return jsxs(Fragment, {
+      children: [contentCmp, jsx(SinglePageNavigation, { restoreScroll$ })],
```
manucorporat, line 41: why another component? i am not sure i like the extra layer of indirection

packages/qwik-city/runtime/src/single-page-navigation-component.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  useOnDocument(
+    'qnav',
+    event$(async (e: Event) => {
+      const detail = (e as NavigationStatusEvent).detail;
+      switch (detail.status) {
+        case 'pending': {
+          // set up a single-use MutationObserver asynchrously and
+          // observe the data-key attribute of the marker for DOM update detection
+          settled.value = new Promise((resolve, reject) => {
```
manucorporat, line 49: i would like to understand this. why do we need a mutation observer?

billykwok: I think both Point 3 in the PR and the comment above this line provide an explanation. Could you clarify which part you don't understand?

packages/qwik-city/runtime/src/router-outlet-component.ts
billykwok: Sure, I will inline that into the RouterOutlet component

packages/qwik-city/runtime/src/qwik-city-component.tsx
```diff
@@ -170,21 +171,32 @@ export const QwikCityProvider = component$<QwikCityProps>(() => {
...
             qwikCity.menus,
             qwikCity.cacheModules,
             url.pathname
           );
         }
-        loadedRoute = await loadRoutePromise;
+        try {
+          loadedRoute = await loadRoutePromise;
+        } catch (e) {
+          dispatchNavigationStatusEvent(type, prevUrl, url, 'rejected');
```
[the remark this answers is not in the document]
billykwok: If you take a closer look, this branch is only executed when `isServer` is `false`.

packages/qwik-city/runtime/src/form-component.tsx
billykwok: It does look like so, but if you look at how `nav` (i.e. `goto` in `QwikCityProvider`) in detail, you will find that it is actually not an async function coz it neither `await` anything nor returns a Promise. The `submitcompleted` event is dispatched at the same time as it was originally intended.

packages/qwik-city/runtime/src/router-outlet-component.ts
billykwok: done

packages/qwik-city/runtime/src/single-page-navigation-component.tsx
billykwok: This is now moved to `router-outlet-component.ts`

billykwok:
> I would really like to merge this API, i am a bit worry about a regression in initial loaded JS and the mutation observer code, not entirely sure, why we need it. But not a fan of that part.

In terms of initial JS load, I don't have numbers yet, but I don't think this would have any impact since all codes are either executed on the server or after users click the back/forward button or `<Link />`. If anything, the change actually reduced the initial load by replacing the 'eagerly-loaded' `'qinit'` listener with the 'lazily-loaded' `'popstate'` and `'qnav'` listeners in `<RouterOutlet />`.

For MutationObserver, I think both Point 3 in the PR and the comment provide an explanation. Could you clarify which part you don't understand?

I would love to find another way to reliably detect when Qwik flushes DOM update to the actual DOM tree. Since Qwik does not expose such information, a short-lived MutationObserver is so far the only solution I can think of.

Any suggestion is welcome.

packages/qwik-city/runtime/src/form-component.tsx
manucorporat: i see, I would say it's a bug of useNavigate then, i would love to wait for navigation to end

billykwok: I just removed the whole 'qnav' and 'qsettled' event change to simplify the change further. This should be addressed now.

billykwok:
Thanks for following up. I appreciate the pursuit of avoiding the initial JS load at all costs. However, there is probably no simple way to achieve the intended result of this PR with Qwik's current design/APIs.

While I agree that the current implementation may not fully align with Qwik's vision, the UX issues in scroll restoration mentioned in the original post should still be fixed ASAP. There might be some clever tricks that I'm not aware of (perhaps through some unknown internal APIs). Unfortunately, I've decided to move on from this.

For the hard refresh issue specifically, it may not be that big of a concern after all. Since Qwik embraces and optimizes MPA more than most JS frameworks (except Astro), maybe triggering a hard page refresh on `popstate` is acceptable for most Qwik adopters (unfortunately I am not among this crowd).

Qwik has introduced so many innovative solutions to improve the initial page load experience, and it has so much potential IMO. I am happy to revisit it one day. But until then, if someone cares enough about the SPA Scrolling UX in Qwik and has a solution that fits into Qwik's vision, please feel free to pick this up and fix it once and for all.

I am closing this PR for now. Thank everyone who has put time into it.

manucorporat:
Hey! actually i am working on it, i think it's possible

manucorporat:
We will get it merged!

manucorporat:
Do you think we could merge and iterate, seems like this PR would fix some issues + add a cool feature, what if we leave for a second PR:

- Fix flickering
- Fix the popstate (very tricky one)

manucorporat:
let me try to find him in discord!

thejackshelton:
@billykwok this is amazing. Thank you so much for the contribution.

PR #3295, opened 2023-03-07
fix(qwik-auth): add @auth/core dep optimization

ulic75:
second attempt

not sure how to test this.  I think it's right now, but I'm not 100%.  The previous attempt result in nothing new being injected into the vite.config.js when adding the integration.


zanettin:
Hi @ulic75 👋 

So, gave it a spin and i guess we're missing smth on the vite config side. i did these things to test it within the qwik repo on latest:

- exec `pnpm cli`
- keep default app name
- choose basic app
- let it install
- run `cd qwik-app`
- run`pnpm i` (not needed)
- run `pnpm qwik add`
- select auth
- confirm install

## result
- `package.json` file got two new inserts for `@authjs/core@0.5.1` and `@builder.io/qwik-auth@0.0.4` ✅ 
- `vite.config.ts` looks the same as before. ❓ 
- file `src/routes/plugin@auth.ts` was added ✅ 

can you show me how the exepected output would look like on the `vite.config.ts` file so i can have a look at it? thanks a lot 🙏

ulic75:
I expect it would look something like this:
```js
  return {
    plugins: [qwikCity(), qwikVite(), tsconfigPaths()],
    preview: {
      headers: {
        'Cache-Control': 'public, max-age=600',
      },
    },
    optimizeDeps: {
      include: ['@auth/core'],
    },
  };
});
```

zanettin:
Hi @ulic75 
Sry for the late reply 🙈 took me some time to get into the code first 😆 i've pushed a change to this PR which should fix the issue described above. Can you validate it on your side? Thanks again 🙏

ulic75:
Oh, it expects a string?  I'm not sure how to test it.  But if it just needs it in a string format I think it looks good to me.

zanettin:
Yes 😆  just took me an hour to figure this out tbh 😆 
Regarding the testing: I do it that way -> https://github.com/BuilderIO/qwik/pull/3295/#issuecomment-1468896644
Let me know if it works on your side 🙏

ulic75:
Maybe I'm missing something.  I get the same results.  But won't this still just pull the config from the current starter code which is not functional?  Do I need to build qwik locally and use it?

ulic75:
If it tried a result like I posted above I saw just go for it. Can't make it any worse :)

manucorporat:
To be honest i am not sure the code works without testing myself, cc @adamdbradley could we make this work?

ulic75:
Okay.  So I tested it by modifying `node_modules/@builder.io/qwik/starters/features/auth/package.json` before running qwik add and selecting auth.  The change you made does indeed add the `optimizedDeps` object to the vite.config.js.

We'll probably want to create another PR at some point those because it doesn't seem to merge with an existing options.  For example I put in a fake optimizeDeps in my vite file and then adding auth overwrote it completely.

zanettin:
thanks @adamdbradley & @manucorporat for having a look at this 🙏  and thx @ulic75 for your tests as well 🙏 the point you've mentioned @ulic75 about the overwrite is true.

ulic75:
How would people feel about adding this instead since the merging doesn't work correctly?  It's referencing non-existent docs, but I can work on writing those up.

```json
"nextSteps": {
  "lines": [
    " Please add `optimizeDeps: { include: ['@auth/core'] }`",
    " to your vite.config.ts file.",
    " Have a look at the docs for more info: ",
    " https://qwik.builder.io/integrations/authjs/"
  ]
}
```

PR #3307, opened 2023-03-08
perf: massive rendering perf refactor

manucorporat:
* [x] Generate fn string
* [x] Use inlineFn in children transform 
* [x] Skip inlineFn when call expr for children
* [x] TrackSignal uses RenderEvent
* [x] Common class for all signals
* Move signal operation after render
* Optimized inlinedQrl in server 
* Review optimized bundle for server 
* Optimize useMethod().value
* SignalDeroved is never leaky of the args
* Immutable signals are not leaky 
* [x] Prefer inlineFn over wrapSign
* [x] Render functional cmp key
* [x] Remove getKeys()
* [x] Remove getProps entirely
* [x] Simplify update class
* [x] Make style and class update similar 
* Don’t set property on update for signals 
* Remove listeners clean up
* [x] Collapse function cmp key with Qwik component 
* Don’t emit custom key when cmp it’s branch free
* Only emit key on root JSX node
* [x] Split mutable, immutable and children 
- Event handlers are mutable 
- One spread makes everything mutable 
* JSX flags
- Dynamic direct children
- Static sub tree
- Variable props (spread)
- Dynamic listeners


packages/docs/src/components/header/header.tsx
```diff
@@ -90,7 +90,7 @@ export const Header = component$(() => {
           </li>
           <li>
             <a href="/community/groups" class={{ active: pathname.startsWith('/community') }}>
-              <span>Community</span>
+              <span>df</span>
```
hamatoyogi, line 93: This seems unintentional @manucorporat

packages/docs/src/routes/examples/apps/visibility/clock/app.tsx
```diff
@@ -30,34 +30,38 @@ export const Clock = component$(() => {
...
+      store.second = now.getSeconds() * (360 / 60);
+      store.minute = now.getMinutes() * (360 / 60);
+      store.hour = now.getHours() * (360 / 12);
     };
     update();
     const tmrId = setInterval(update, 1000);
     return () => clearInterval(tmrId);
   });
 
+  jsx();
```
hamatoyogi, line 50: what's this?

zanettin:
Just got off a call with Google, and they won't be increasing the Core Web Vitals score to 200 to showcase the effects of these changes. 😢

PR #3338, opened 2023-03-12
docs: add `bunx create-qwik` command

cunzaizhuyi:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [X] Docs / tests

# Description

I've tried `bunx create-qwik` and `bun start`, it works!
so, we should add `bunx create-qwik` command into docs.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ As users interact with the site, only the necessary parts of the site load on-de
 npm create qwik@latest
 # or
 pnpm create qwik@latest
-# or
 yarn create qwik@latest
+bunx create-qwik@latest
 ```
 
 - Understand the difference between [resumable and replayable](https://qwik.builder.io/docs/concepts/resumable/) applications.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -18,8 +18,8 @@ As users interact with the site, only the necessary parts of the site load on-de
 npm create qwik@latest
 # or
 pnpm create qwik@latest
-# or
 yarn create qwik@latest
+bunx create-qwik@latest
```
golamrabbiazad, line 22: Bun is still under development and doesn't support windows native build. 
Tomorrow, someone could raise or open a PR and say hey let's remove it from the docs. 🙂

zanettin:
Thanks for creating this PR @cunzaizhuyi and thx as well for your review @golamrabbiazad 🙏 
Leave the dicission to the core team 😃

adamdbradley:
Personally I'm of the opinion we actually only write out `npm create`. That's in the base install for nodejs. If you already installed and use yarn, you know it's different command and how it's different from npm already. If you use pnpm, you know it's a different command, etc. I think this list will only continue to grow and make things more confusing for new users, so if anything I think we reduce it to just `npm`.

PR #3344, opened 2023-03-13
docs: improved emoji shuffle

LockingReal:
according to my idea, the current changes will not cause badcase, which is in line with expectations,
and can reduce invalid assignments.🥰
@golamrabbiazad


golamrabbiazad:
will you give an example of how this can reduce invalid assignments?

LockingReal:
> will you give an example of how this can reduce invalid assignments?
```
        const demoArray = [2,5,6,7,3098];
        [demoArray[3],demoArray[3]] = [demoArray[3],demoArray[3]];
        console.log(demoArray);
```
if i ==j ,array not change,we don't need to assign

golamrabbiazad:
Yes, it doesn't for your example case but that is different from what I'm saying.

 So, the shuffleArray function should return shuffle values for each function call, there are no static things in the sense. 
idx I'm getting based on the array and randIdx I'm getting from randomize value and it is bound to the array. These values are changing dynamically.

[shuffleArray](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAJgUwLYgIICd0EMCeMC8MA2gEwA0MArBQGwUDsFAzAAwCcAHALoDcAUHwBmAVzDAoAS3AwIAC2GDBAGwQZsOABRZMuAFwwwwpACME6IlwCUMAN58YDmIJDoYGlbAlwAHgRjb1ADoVMABzKFkYAFoYAEYeGC9fAD4YFgSkqKjrO0c8mFBIWGwwOABJHz8AWSwIwOUQFw0aupK4ECQNawAqNySYAGo4y0t+fMciANwiJK4KKZwiNorvLi4-SZ1F5Z85-y2Z3d57RwBfE4d0BChhdDB99X5zvkKIEBVgkFCNOQVlVS2GkQKDUuBGPCAA) example, if you could run it, hope you understand that array is shuffled.

Thanks

LockingReal:
@mhevery cc ❤

LockingReal:
@adamdbradley cc

LockingReal:
@manucorporat I don't understand why you both ignored me. I'm so sad...

shairez:
Thanks for the PR @LockingReal !

Can you fix the conflicts please? That way I could merge

Thanks 🙏

LockingReal:
> 



> 

done~

LockingReal:
> Thanks for the PR @LockingReal !
> 
> Can you fix the conflicts please? That way I could merge
> 
> Thanks 🙏

done~

LockingReal:
> Thanks for the PR @LockingReal !
> 
> Can you fix the conflicts please? That way I could merge
> 
> Thanks 🙏


There should be no problem ~ 😁

PR #3360, opened 2023-03-14
docs: fix Adding response headers example

himorishige:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

Thanks for the great framework.

Sorry for the very small point🙏

There seems to be a small error in the sample code([Adding response headers](https://qwik.builder.io/qwikcity/endpoints/#adding-response-headers)).
After referring to the type information, I believe the correct code is as follows.

```ts
// docs Adding response headers
headers.set("X-My-Custom-Header" : "Hello World");
// -->
headers.set("X-My-Custom-Header", "Hello World");
```

```ts
interface Headers {
    // ...
    set(name: string, value: string): void;
    // ...
}
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/qwikcity/endpoints/index.mdx
+++ b/packages/docs/src/routes/qwikcity/endpoints/index.mdx
@@ -48,7 +48,7 @@ export const onGet: RequestHandler = async ({ status }) => {
 
 ```tsx
 export const onPost: RequestHandler = async ({ headers }) => {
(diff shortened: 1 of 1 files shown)


zanettin:
Thx a lot @himorishige 🙏  eagle eyes 🦅 😄

PR #3372, opened 2023-03-14
docs(packages/docs/readme.md): replace yarn with pnpm

mindmind:
If user try to use scripts from packages/docs/README.md, then he will see:

image


shairez:
thanks @mindmind !

PR #3417, opened 2023-03-17
feat: cli git init prompt

adamdbradley:


packages/create-qwik/create-interactive.ts
```diff
@@ -102,10 +108,33 @@ export async function runCreateInteractiveCli() {
...
   const s = spinner();
 
-  s.start('Creating App');
+  s.start('Creating App...');
   const result = await createApp(opts);
   s.stop('Created App 🐰');
 
+  if (gitInitAnswer) {
+    if (fs.existsSync(join(outDir, '.git'))) {
+      console.log(`Git has already been initialized`);
```
zanettin, line 117: just a short input 👼 
using `console.log` would drop the indent which clack creates with the line on the left. you could use `log.error` (`import {log} from '@clack/prompts`) from clack itself to have a "cleaner" flow 👼

zanettin: list of supported log types: https://github.com/natemoo-re/clack/blob/main/packages/prompts/src/index.ts#L576:L602

Issue #3321, opened 2023-03-10
[🐞] Dev mode failing since v0.20.0


EamonHeffernan:
### Which component is affected?

Qwik Runtime

### Describe the bug

Since v0.20.0 dev mode fails with this error.

I am unable to reproduce the bug. however the message is

```
1:16:29 PM [vite] Internal server error: require() of ES Module C:\Users\Admin\Documents\Full-Stack\<project>\frontend\node_modules\@builder.io\qwik\build\index.mjs not supported.
Instead change the require of C:\Users\Admin\Documents\Full-Stack\<project>\frontend\node_modules\@builder.io\qwik\build\index.mjs to a dynamic import() which is available in all CommonJS modules.
      at Object.<anonymous> (C:\Users\Admin\Documents\Full-Stack\<project>\frontend\node_modules\@builder.io\qwik\core.cjs:17:13)
 ```

To get dev server to run (not without issues such as `Uncaught (in promise) SyntaxError: ambiguous indirect export: isServer`, modify `node_modules\@builder.io\qwik\build\package.json` and change `"main": "index.mjs",` to `"main": "index.cjs",`)

### Reproduction

.

### Steps to reproduce

pnpm dev

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 3.86 GB / 15.93 GB
  Binaries:
    Node: 19.4.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.2.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (110.0.1587.63)
    Internet Explorer: 11.0.19041.1566
```

doksara:
Any news on this?

EamonHeffernan:
I poked around in debug mode and the issue seems to be caused by the module map for @qwik-serializer including an error. This was the `mod` object that vite generated that includes the error.
Sorry for the image format. I couldn't easily serialize it.

![image](https://user-images.githubusercontent.com/16983388/224605412-35cf662e-676e-45cb-8aa8-28f6df16b5be.png)

EamonHeffernan:
Replacing the import for isServer with in platform.ts
`const isServer = /*#__PURE__*/ (() =>
[...]

PR #3421, opened 2023-03-18
fix(dev mode esm error): fixes error that require() cannot be used when in dev mode

EamonHeffernan:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Fixes the error that is detailed in #3321. This is a bandaid fix as I need to update to avoid the security issues. I am not sure of the underlying issue, however this uses the code that is in that file and seems to work just fine for now. Happy to work to solve the actual issue, but I think this is worth fixing in the mean time.

# Use cases and why

Detailed in #3321

# Checklist:


manucorporat:
This is not likely a good fix! replacing the import can cause other issues. We need a repo case for this issue, many developers are building qwik apps without this issue. Can we provide a simple repo that reproduces the issue?

EamonHeffernan:
@NexonTreehouse found the issue in #3321

PR #3433, opened 2023-03-19
feat: bind:value for input

manucorporat:
```tsx
  const value = useSignal('');
  return (
    <>
      <input bind:value={value} />
      <div>Value: {value}</div>
    </>
  );
});
```

### DIFF
--- a/packages/qwik/src/core/render/jsx/types/jsx-generated.ts
+++ b/packages/qwik/src/core/render/jsx/types/jsx-generated.ts
@@ -1,3 +1,4 @@
+import type { Signal } from '../../../state/signal';
 import type { DOMAttributes, ClassList } from './jsx-qwik-attributes';
 interface HTMLWebViewElement extends HTMLElement {}
 export type Booleanish = boolean | `${boolean}`;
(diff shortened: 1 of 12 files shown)


shairez:
that's awesome!

question: 
do we bind to the signal or the value of the signal?

if so, the `{value}` should probably be `{signal.value}` no?

tleperou:
To the signal as I understood this one https://twitter.com/manucorporat/status/1637429164255879168?t=FJf423aLapRduYIbl9c2Lg&s=19.

Which events proceed with binding; input, change etc?

shairez:
yes saw this now

IMO it's kind of confusing... as in anywhere else we're using `someSignal.value`... or else we'll be able to use "just the signal" everywhere and then it makes sense

manucorporat:
I don’t understand what do you mean with {signal.value}?

manucorporat:
Everywhere else you are passing the value, in this case you pass the signal, because you need two way, exactly the same as ref={}

shairez:
Yes to pass it to `bind:value` it makes sense to pass the signal.

What I didn't realize is that you can do `{signal}` and it will work 😅
I always passed the `signal.value`

Here's an example:
https://stackblitz.com/edit/qwik-starter-qu7vul?file=src%2Froutes%2Findex.tsx

Albuquerquess:
Boa tarde. A implementação do bind:value foi removida do repositório?

subhendukundu:
We will need to update the lint as well?

leighwilliams:
Working fine for text input but `bind:checked` has an issue. Seems to be using `input.value` instead of `input.checked`.

https://stackblitz.com/edit/qwik-starter-5w2pos?file=package.json,src%2Froutes%2Findex.tsx

1. Not checked - No value
Screen Shot 2023-03-31 at 10 20 11 pm
2. Checked - 'on' instead of true
Screen Shot 2023-03-31 at 10 20 23 pm
3. Unchecked - Value remains 'on'
Screen Shot 2023-03-31 at 10 20 38 pm

I'll open an issue if it's not already being addressed.

PR #3441, opened 2023-03-20
fix(azure-swa): SSR build should target Node.js environment

derkoe:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

Azure Static Web Apps support a full Node.js environment. The Vite config for the starter should target Node as default.

All external modules are packed inside the bundle so that they do not have to be declared in the package.json.

# Use cases and why

See https://github.com/BuilderIO/qwik-city-e2e/pull/6

# Checklist:

### DIFF
--- a/starters/adapters/azure-swa/adapters/azure-swa/vite.config.ts
+++ b/starters/adapters/azure-swa/adapters/azure-swa/vite.config.ts
@@ -16,8 +16,7 @@ export default extendConfig(baseConfig, () => {
       },
     },
     ssr: {
-      target: 'webworker',
-      noExternal: true,
+      noExternal: /.*/,
     },
     plugins: [azureSwaAdapter()],
   };


starters/adapters/azure-swa/adapters/azure-swa/vite.config.ts
```diff
@@ -16,8 +16,7 @@ export default extendConfig(baseConfig, () => {
       },
     },
     ssr: {
-      target: 'webworker',
-      noExternal: true,
+      noExternal: /.*/,
```
manucorporat, line 19: why /.*./? we dont do this in other node deployments

derkoe: The other option would be to generate a `azure-functions/render/package.json` that includes all necessary runtime dependencies and `npm install` those. I've tried this as well but this results in bigger bundles. Bigger bundles mean slower start times for the Azure Functions. 

The config ` ssr: { noExternal: /.*/ }` inlines all third-party dependencies but still uses the Node.js provided modules. This works quite well for my apps so far. See https://github.com/derkoe/qwik-todos/ for an example.

derkoe: This also fixes the e2e tests https://github.com/BuilderIO/qwik-city-e2e/pull/6

manucorporat: Ok! let's do it!

manucorporat: Thanks for the explanation and all the great work with Azure!!

PR #3475, opened 2023-03-22
`useStore(...)` proxy now handles `delete proxy[k]`

revintec:
# What is it?

- [X] Bug

# Description

previously proxy objects don't implement `deleteProperty` handler, thus:
1. the implicit default implementation allows deletion of any key, including symbolic ones, which could be problematic
2. deleting a key won't trigger subscribers' update callback, which could be considered a BUG

# Use cases and why

```javascript
const store=useStore({key:1})
componentCall(store) // internally uses `store.key`
delete store.key // this is valid javascript and store usage, but it won't trigger reactivity inside `componentCall`
```

# Checklist:

### DIFF
--- a/packages/qwik/src/core/state/store.ts
+++ b/packages/qwik/src/core/state/store.ts
@@ -97,8 +97,8 @@ export class ReadWriteProxyHandler implements ProxyHandler<TargetType> {
 
   // no need to handle `immutable` and `recursive` for now
   deleteProperty(target: TargetType, prop: string | symbol): boolean {
-    if(!delete target[prop])return false;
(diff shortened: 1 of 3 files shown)


revintec:
about the failing check, it seems I don't have the permission to run it
`Error: Parameter token or opts.auth is required`

manucorporat:
Curious why not handle immutable for now? I would be great to have some unit test too

manucorporat:
We should add a new component here:
https://github.com/BuilderIO/qwik/blob/main/starters/apps/e2e/src/components/render/render.tsx

that tests this, maybe a onClick that deletes it, a e2e test that validates the correct behaviour:
https://github.com/BuilderIO/qwik/blob/main/starters/e2e/e2e.render.spec.ts#L258

PR #3484, opened 2023-03-22
fix: tfoot inside table

ofavre:
`<tfoot>` is also a permitted direct child of `<table>`.

Sources:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#try_it:~:text=Flow%20content-,Permitted%20content,-In%20this%20order
- https://html.spec.whatwg.org/multipage/tables.html#the-table-element

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

I was surprised to see the following error when writing the following HTML in my component:

```html
<table>
  <thead>…</thead>
  <tbody>…</tbody>
  <tfoot>…</tfoot>
</table>
```

```
The <table> element requires that its direct children to be '<tbody>' or '<thead>', instead, '<tfoot>' was rendered.
```

The HTML spec and MDN describe what are permitted direct children of `<table>`, they include the `<thead>` and `<tbody>` for sure but also `<tfoot>` and some other tags I never heard of.

As a workaround in my project I tried placing `<tfoot>` inside `<tbody>` but I then for the following error: `Uncaught (in promise) Error: close not found` which was caused by the browser reordering misplaced tags and the `<!--/qv-->` node being moved outside the `<tbody>`.

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/ssr/render-ssr.ts
+++ b/packages/qwik/src/core/render/ssr/render-ssr.ts
@@ -623,7 +623,7 @@ This goes against the HTML spec: https://html.spec.whatwg.org/multipage/dom.html
       } else {
         if (flags & IS_TABLE && !tableContent[tagName]) {
           throw createJSXError(
-            `The <table> element requires that its direct children to be '<tbody>' or '<thead>', instead, '<${tagName}>' was rendered.`,
+            `The <table> element requires that its direct children to be '<tbody>', '<thead>' or '<tfoot>', instead, '<${tagName}>' was rendered.`,
             node
           );
         }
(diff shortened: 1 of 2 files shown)


ofavre:
The HTML spec also lists `<caption>` and `<colgroup>`.
`<tr>` are also accepted as direct children of `<table>` if there are no `<tbody>`.

manucorporat:
Exactly! now that we fix it, we should add `caption` and `colgroup`.

For `tr` is not true, this HTML:

```html
<table>
  <tr></tr>
  <tr></tr>
</table>
```

generated the following DOM:
```html
<table>
  <tbody>
    <tr></tr>
    <tr></tr>
  </tbody>
</table>
```

PR #3527, opened 2023-03-26
build: make root.tsx optional

shairez:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

Make build `root.tsx` for client mode building to make it easier for testing tools to run Qwik libraries.

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/plugin.ts
@@ -110,7 +110,7 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
     return optimizer.sys.path;
   };
 
-  const normalizeOptions = (inputOpts?: QwikPluginOptions) => {
+  const normalizeOptions = async (inputOpts?: QwikPluginOptions) => {
     const updatedOpts: QwikPluginOptions = Object.assign({}, inputOpts);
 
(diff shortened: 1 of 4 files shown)


mhevery:
@shairez can you better explain the motivation?

shairez:
Yes, it's something me and @manucorporat figured out when trying to build the cypress plugin.

I ran into some limitations when trying to fix this, but I need to complete this work when I'll have time

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -758,6 +759,16 @@ export const isDev = ${JSON.stringify(isDev)};
...
   }
 
+  async function isPathExist(path: string) {
+    const sys = getSys();
+    if (sys.env === 'node') {
+      const fs: typeof import('fs') = await sys.dynamicImport('node:fs');
+
+      return fs.existsSync(path);
+    }
+    return false;
```
wmertens, line 769: maybe better to assume they exist, so that it's like previous behavior?

PatrickJS:
oh I actually ran into this when I was working on Qwik Start. It should be optional

PatrickJS:
can we redo this PR

wmertens:
we're going to do this as part of the vite environment support

PR #3535, opened 2023-03-28
feat: angular integration

dmitry-stepanenko:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description
Adding the qwik-angular package to be able to use Angular components in Qwik applications

# Use cases and why
Expanding the ecosystem of frameworks compatible with Qwik.
Documentation is available here https://github.com/dmitry-stepanenko/qwik/tree/ds/qwikify-angular/packages/qwik-angular

This package includes the vite plugin, qwikify function and a new CLI starter. 

# Known issues
- material styles are loaded per file which greatly worsens the DX 
- Angular requires typescript@4 to be used 

# Checklist:


brandonroberts:
Nice work! In Angular v16, `renderApplication` won't take a component directly anymore https://github.com/angular/angular/pull/49463

gioboa:
This is an amazing feature 🚀

mhevery:
Hey @dmitry-stepanenko, this is a fantastic piece of work! Thank you!

We are figuring out the best way to integrate this and get it into the hands of our users. We have some ideas and we would like to hear your thoughts.

## Option 1: Merge this into our mono repo
As you can imagine merging things into our mono repo has to meet a very high bar:
- It has to be production ready. Now that we have announced v1.0 we have to be sure that what goes in is of sufficient high quality.
- The implication of the above is that we are now on the hook to maintain that code. We may not be the right experts for maintaining it. Issues will be filed against our repo, and we may not be the best people to fix this issue.
- We don't think this is scalable because we can imagine a world where many people will want to add many features to our monorepo, and we would be on the hook to maintain it. A more distributed way of doing this would be better. 

## Option 2: Merge this into Qwik Labs

- This would be better as it would mean that we don't have to worry that it is of v1.0 quality.
- However, it still means that we would be on the hook to maintain the code. Issues would be field with us etc...

## Option 3: Have a community organization
- Have a https://github.com/QwikDev organization that can publish to `@qwikdev/____` The nice part of the org is that it still has the "seal" of Qwik without passing on the cost to the Qwik core maintainers. 
- This way, the core Qwik team is not on the hook to support the feature long time (as most of the time, they are not the experts)
- Qwik core team is not involved in having to review the PR and make a judgment call if the quality is good enough (given that hey are not the experts in that area)

---

Would love to hear your thoughts.

dmitry-stepanenko:
Hey @mhevery, thank you for the feedback!

It's understandable that the monorepo approach might not scale well, especially with such big features as Angular integration. Apart from that, it is probably better to find a balance between keeping some of the features together or separating them since maintaining multiple repositories can also be a bit of a problem.

I'm pretty much up for option 3, although I would still be extremely happy to get some meaningful reviews from the core team on features like this one due to its complexity.

Additionally, if we start moving in this direction, there's one thing that might be a problem, which is keeping packages up to date and functional as Qwik itself can break certain things. Do you think it would be reasonable to set up in advance a testing system that would allow us to track any issues early? Something similar to what Vite does to test breaking changes against frameworks like Qwik or what Nx did recently for integrations like qwik-nx (https://blog.nrwl.io/introducing-nx-ecosystem-ci-ad0526d37f83).

mhevery:
> Hey @mhevery, thank you for the feedback!
> 
> It's understandable that the monorepo approach might not scale well, especially with such big features as Angular integration. Apart from that, it is probably better to find a balance between keeping some of the features together or separating them since maintaining multiple repositories can also be a bit of a problem.

Thanks for understanding.

> I'm pretty much up for option 3, although I would still be extremely happy to get some meaningful reviews from the core team on features like this one due to its complexity (and you specifically as probably there's no better expert in these two technologies combined).

Yes, we are happy to consult on things from time to time, just ping us on Discord.

I have created https://github.com/QwikDev/qwik-angular and given your maintainer rights. Please let me know your NPM username and I will get you publish rights for `@qwikdev/qwik-angular`



> Additionally, if we start moving in this direction, there's one thing that might be a problem, which is keeping packages up to date and functional as Qwik itself can break certain things. Do you think it would be reasonable to set up in advance a testing system that would allow us to track any issues early? Something similar to what Vite does to test breaking changes against frameworks like Qwik or what Nx did recently for integrations like qwik-nx (https://blog.nrwl.io/introducing-nx-ecosystem-ci-ad0526d37f83).

Yes, I think that is a great idea. We already have some e2e tests which we run so I think extending it to the community sounds like a good idea.

Are there any private APIs that you used and that you are concerned about?

dmitry-stepanenko:
> Yes, we are happy to consult on things from time to time, just ping us on Discord.

> I have created https://github.com/QwikDev/qwik-angular and given your maintainer rights. Please let me know your NPM username and I will get you publish rights for @qwikdev/qwik-angular

Thank you, I will let you know once I refactor the package to a standalone one. I'm going to do a few tweaks with an integration approach so that it works well while being standalone. It might take me a few days to find enough time to work on it.

> Are there any private APIs that you used and that you are concerned about?

The challenges mainly lie in the intricacies of Angular itself. As an example, I had to make certain updates in the qwik core, replacing instances of `value instanceof Promise` with `value && typeof value.then === 'function'` to ensure proper functionality when zone.js is added. Additionally, I've enabled decorators support for the qwik optimizer to not fail once it sees angular code.

These may seem like minor changes, but there is a potential for similar adjustments in the future. Given that regressions can be specific to the Angular environment, it becomes challenging to catch them in advance

dmitry-stepanenko:
This PR is no longer relevant. Changes are pulled into separate repos:

- Angular integration is now going to be set in a separate repo https://github.com/QwikDev/qwik-angular 
- Code generation capabilities are going to be extracted to the https://github.com/qwikifiers/qwik-nx repository (there's an opened PR at the moment of writing this)

PR #3550, opened 2023-03-29
docs: Make the introductoin to resources friendlier

steve8708:
We dump people into a complex example that adds a lot of cognitive overhead to make sense of

Let's instead show a very simple example first, so you can go "oh I see, fetch async stuff there, render it there", and then show the complex stuff

Also clean up the code formatting to logically group things a little easier to see what is going on in the more complex example

### DIFF
--- a/packages/docs/src/routes/docs/(qwik)/components/resource/index.mdx
+++ b/packages/docs/src/routes/docs/(qwik)/components/resource/index.mdx
@@ -62,28 +62,27 @@ import {
 
 export default component$(() => {
   const query = useSignal('busy');
-  const jokes = useResource$(
-    async ({ track, cleanup }) => {
-      track(query);
-      // A good practice is to use `AbortController` to abort the fetching of data if
-      // new request comes in. We create a new `AbortController` and register a `cleanup`
-      // function which is called when this function re-runs.
-      const controller = new AbortController();
-      cleanup(() => controller.abort());
+  const jokes = useResource$(async ({ track, cleanup }) => {
+    track(query);
+    // A good practice is to use `AbortController` to abort the fetching of data if
+    // new request comes in. We create a new `AbortController` and register a `cleanup`
+    // function which is called when this function re-runs.
+    const controller = new AbortController();
+    cleanup(() => controller.abort());
 
-      if (query.value.length < 3) {
-        return [];
-      }
-
-      const url = new URL('https://api.chucknorris.io/jokes/search');
-      url.searchParams.set('query', query.value);
-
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/(qwik)/components/resource/index.mdx
```diff
@@ -22,6 +22,38 @@ This method allows you to generate computed values asynchronously. The asynchron
...
 Just like all the `use-` methods, it must be called within the context of [`component$()`](/docs/(qwik)/components/overview/index.mdx#component), and all the [hook rules apply](/docs/(qwik)/components/lifecycle/index.mdx#use-method-rules).
 
+Here is a very simple example of using `useResource$()` to fetch a random joke:
+
+```tsx
+import { component$, useResource$, Resource } from '@builder.io/qwik';
+
+export default component$(() => {
+  const jokes = useResource$<{ value: string }[]>(
+    async ({ track, cleanup }) => {
```
zanettin, line 32: imo we could also remove the two params here to avoid distraction at the simplest implementation, right? 👼

packages/docs/src/routes/docs/(qwik)/components/resource/index.mdx
```diff
@@ -22,6 +22,38 @@ This method allows you to generate computed values asynchronously. The asynchron
...
+      const resp = await fetch('https://api.chucknorris.io/jokes/random')
+      return (await resp.json()) as { value: string };
+    }
+  );
+
+  return (
+    <Resource
+      value={jokes}
+      onPending={() => <>loading...</>}
```
zanettin, line 41: here we could use a `<span>` or `<div>`, etc instead. guess not all devs are aware of fragments.

zanettin (commented):
love it <3 thx a lot @steve8708 🙏

packages/docs/src/routes/docs/(qwik)/components/resource/index.mdx
steve8708: ah, yes agreed! will update

packages/docs/src/routes/docs/(qwik)/components/resource/index.mdx
steve8708: works for me ✅

steve8708:
I also realized we are explicitly typing things in these examples that are better off using inference (which is worth showing off anyway for the "full stack type safety" benefit), so that helped clean the code up too

zanettin (approved):
🎉 🙏

PR #3555, opened 2023-03-29
docs: fix and simplify resource example

steve8708:
found a couple more minor things to get in

cc @zanettin


zanettin:
thx a lot @steve8708 🙏

PR #3556, opened 2023-03-29
docs: make errors in the repl red so they are noticeable

steve8708:
currenly it is not clear when your code is erroring. You see no logs and the only indication is a tiny "1" in the far right of the screen

This will leave the user thinking the repl is broken, as it does not say "there was an error" in the preview

This makes the diagnostics tab turn red to make it more clear there is an error under there that you should look at

Screenshot 2023-03-29 at 12 36 22 PM


zanettin (approved):
another thank you 🙏

PR #3585, opened 2023-03-30
docs: More real world examples for route loader

steve8708:
Route loaders are for fetching data, so lets use real world examples so people properly understand


manucorporat:
This will be great for GPT!

steve8708:
That's definitely a big part of the idea!

PR #3606, opened 2023-03-31
docs: improved blog section in media

the-r3aper7:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

## Previously
Before

## Now
After

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/package.json
+++ b/packages/docs/package.json
@@ -43,7 +43,8 @@
     "undici": "5.21.0",
     "uvu": "0.5.6",
     "vite": "4.2.1",
-    "wrangler": "^2.13.0"
+    "wrangler": "^2.13.0",
+    "node-html-parser": "^6.1.5"
   },
   "author": "Builder.io Team",
   "bugs": {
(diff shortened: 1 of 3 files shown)


the-r3aper7:
Is there any way to optimize this?

zanettin:
Thank you for the adjustments 🙏  imo this looks great ❤️ 
just my thoughts regarding the implementation: what do you think about adding the values right away to the the blog list instead of 'crawling' them on each request?
so adding the `imgSrc` as well here: https://github.com/BuilderIO/qwik/blob/main/packages/docs/src/routes/(ecosystem)/media/index.tsx#L156

mhevery:
@the-r3aper7 This is a great improvement, but it can't be merged int the current state because `node-html-parser` is a node package, and the site runs at the edge (Cloudflare) where you can't have runtime dependencies on node packages (This works fine in local development because you are running in node, but will fail in Cloudflare.)

As @zanettin Could this be a build step which fetches these values so that we don't have to do it at runtime?

the-r3aper7:
> @the-r3aper7 This is a great improvement, but it can't be merged int the current state because `node-html-parser` is a node package, and the site runs at the edge (Cloudflare) where you can't have runtime dependencies on node packages (This works fine in local development because you are running in node, but will fail in Cloudflare.)
> 
> As @zanettin Could this be a build step which fetches these values so that we don't have to do it at runtime?

yeah I was busy for days i will make it as static input

PR #3609, opened 2023-03-31
docs: support highlighting specific lines of code

steve8708:
like the beta react docs do

https://react.dev/learn

image


manucorporat:
Amazing idea Steve! so much i copied it, but wanted to add it as a MDX plugin, so it can be used directly usign MD

PR #3631, opened 2023-04-02
Improve definition of `useTask$`

brunocrosier:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [X] Docs / tests

# Use cases and why

Current wording is:

> It runs on the server and the browser.

This could be understood to mean that `useTask$` always runs twice.


zanettin (commented):
thanks a lot for the correction 🙏

Issue #3664, opened 2023-04-05
[✨] Support View Transition API


swwind:
### Is your feature request related to a problem?

I came into this interesting API introduced in [this article](https://developer.chrome.com/docs/web-platform/view-transitions/). ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API), [Caniuse](https://caniuse.com/mdn-api_viewtransition))

View Transition API brings better transition animations for SPA, but not widely supported yet (Only on Chrome 111+ now).

The API is simple, we just need to use `document.startViewTransition()` to wrap any DOM changes in it. If the function does not exists, we can simply apply the DOM changes without it.

This should be easy to implement (Maybe?).

### Describe the solution you'd like

-

### Describe alternatives you've considered

-

swwind:
Dived into the source code a little bit, and I find it is pretty hard for me to implement this feature.

Tracking the DOM change after triggering a navigation is hard to follow, and even some part of code seems to be unfinished yet.

https://github.com/BuilderIO/qwik/blob/63f77a55b3a05cf314f3e7fb17937716943f184c/packages/qwik-city/runtime/src/client-navigate.ts#L85

swwind:
@jordanw66 it's hard to say this PR makes it better for supporting the view transition api.

The API itself supports async functions as the update DOM callback, so we can start the transition when we want to start a navigation, and resolves it when all the DOM changes was applied. This PR uses a custom `<script>` element and a MutationObserver to tracking the DOM changes, and unfortunately this will not work if we started `document.startViewTransition` function before that. The `document.startViewTransition` will cache all the DOM changes under the hood and `qsettled` event will never be triggered since no DOM actually changes before we resolve the view transtion.

mhevery:
Related https://github.com/BuilderIO/qwik/issues/3664 and https://github.com/BuilderIO/qwik/issues/2078

PR #3691, opened 2023-04-06
feat: View Transition API

swwind:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

#3664

# Use cases and why

This is very brute force implement for supporting View Transition API.

I'm not 100% sure this code doesn't introduce other problems, just for showing my solution.

https://user-images.githubusercontent.com/26316494/230405912-ef6bf8f9-a734-4ba7-bb56-6af0d87cb444.mp4

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/api.md
+++ b/packages/qwik-city/runtime/src/api.md
@@ -304,8 +304,11 @@ export interface QwikCityPlan {
     readonly trailingSlash?: boolean;
 }
 
-// Warning: (ae-forgotten-export) The symbol "QwikCityProps" needs to be exported by the entry point index.d.ts
-//
+// @public (undocumented)
+export interface QwikCityProps {
+    enableViewTransitionAPI?: boolean;
+}
+
 // @public (undocumented)
 export const QwikCityProvider: Component<QwikCityProps>;
 
--- a/packages/qwik-city/runtime/src/api.md
+++ b/packages/qwik-city/runtime/src/api.md
@@ -365,13 +365,7 @@ export interface RouteLocation {
 export type RouteNavigate = QRL<(path?: string, forceReload?: boolean) => Promise<void>>;
 
 // @public (undocumented)
-export const RouterOutlet: Component<RouterOutletProps>;
-
-// @public (undocumented)
-export interface RouterOutletProps {
-    // (undocumented)
-    enableViewTransition?: boolean;
-}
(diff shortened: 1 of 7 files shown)


swwind:
@manucorporat what do you think of this :thinking:

manucorporat:
I think this might be closer to the solution:
https://github.com/BuilderIO/qwik/pull/3684

We dont want to be blocking for too long, qwik already buffers the dom operations and commits them all in one go!

swwind:
@manucorporat I got your idea. I made some changes accordings to your #3684, and with more stuffs.

- add an option in `<RouterOutlet />` to enable this feature manually.
- only use view transition when `routeLocation` has changed. (#3684 seems to have view transition every render, this is ideal because we may not want to have animations in signal changes)
- disable view transition if new `routeLocation` has the same `pathname` to old one. (this happens in hash link jumps)

However, `<Link href="#xxx" />` seems to be broken but I think this is a problem due to an old DOM waiting problem and could be fixed in #3244 hopefully.

manucorporat:
Interesting approach... I think it's pretty low risk approach!

manucorporat:
wonder if this is a RouteOutlet o QwikCityProvider api? it really configures a global thing of qwikcity

manucorporat:
The RouterOutlet, is just an Outlet, not the render itself.

manucorporat:
Happy to merge an iterate from here!

swwind:
@manucorporat I moved code from RouterOutlet to QwikCityProvider, and now the API looks like this

```js
<QwikCityProvider enableViewTransitionAPI>
// ...
</QwikCityProvider>
```

In short, this code will set `document.__q_view_transition__` when the SPA `nav` function is called, and checks this mark inside Qwik to decide whether use ViewTransition API or not.

However, I still doesn't write anything when you click the Back button to go to prior page. There exists a large difference in popstate event handler between `main` and #3244, if that PR was supposed to be merged, I think it's better to wait for that interface to be stabilized.

manucorporat:
Gonna merge and do some minor fixes, I think we can enable by default (only for transitions) + disable the html root view-name layer with css. Seems like a no risk API to instruct the browser when the naviagtion happens, then developers can opt-in using CSS

manucorporat:
Amazing job!

manucorporat:
The other PR, i wanna merge, but it has a big API surface + big changes

PR #3706, opened 2023-04-07
feat: Add MDX rehypeSyntaxHighlight plugin options

### LINKED ISSUE #3603: [✨] Add extra syntax highlighters via. MDX config

Currently, I cannot seem to register extra languages to refractor as it is run as part of the Vite build process via. a plugin.

As you can see, this does register the tsx syntax but nothing else - https://github.com/BuilderIO/qwik/blob/6f3f0ef5a8eb472a9dab0e934a5327dd1005cfb2/packages/qwik-city/buildtime/markdown/syntax-highlight.ts

Allow specifying additional languages to be registered. 

This might be best done via. a 'setup' callback, which can

```ts
// vite.config.ts

import fsharp from "refractor/lang/fsharp";

export default defineConfig(async () => {
  return {
    plugins: [
      qwikCity({
        mdxPlugins: {
          // as well as a `boolean`, also allow taking a config object
          rehypeSyntaxHighlight: {
            enabled: true,
            setup: (refractor: Refractor) => {
              refractor.register(fsharp)
            }
          }
        }
      })
    ]
  }
```

This `setup` callback would be passed to the `rehypeSyntaxHighlight` plugin and called there.

I tried to import `refractor` and register new languages in the `root.tsx` as part of a `useTask$` but it had no effect.

### Additional context

_No response_

swwind:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

fixes #3603

Add plugin options for all 3 core plugins.

# Use cases and why

```jsx
// vite.config.mts

import fsharp from "refractor/lang/fsharp";

  return {
    plugins: [
      qwikCity({
        mdxPlugins: {
          remarkGfm: {
            singleTilde: false,
          },
          rehypeSyntaxHighlight: {
            setup(refractor) {
              refractor.register(fsharp)
            }
          },
          rehypeAutolinkHeadings: false,
        }
      })
    ]
  }
```

# Checklist:

### DIFF
--- a/packages/qwik-city/buildtime/markdown/mdx.ts
+++ b/packages/qwik-city/buildtime/markdown/mdx.ts
@@ -34,9 +34,17 @@ export async function createMdxTransformer(ctx: BuildContext): Promise<MdxTransf
 
   if (
     typeof coreMdxPlugins?.rehypeSyntaxHighlight === 'undefined' ||
-    coreMdxPlugins.rehypeSyntaxHighlight
+    (typeof coreMdxPlugins?.rehypeSyntaxHighlight === 'boolean' &&
+      coreMdxPlugins.rehypeSyntaxHighlight)
   ) {
     coreRehypePlugins.push(rehypeSyntaxHighlight);
+  }
+
+  if (
+    typeof coreMdxPlugins?.rehypeSyntaxHighlight === 'object' &&
+    coreMdxPlugins.rehypeSyntaxHighlight.enabled !== false
+  ) {
+    coreRehypePlugins.push([rehypeSyntaxHighlight, coreMdxPlugins.rehypeSyntaxHighlight]);
   }
 
   if (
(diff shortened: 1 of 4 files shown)


swwind:
This fix still contains a problem, types of `refractor`, `remark-gfm` and `rehype-autolink-headings` must be installed manually to have full functionality.

i.e. You must install `remark-gfm` **manually** to make `remarkGfm` option in `mdxPlugins` have type `boolean | import('remark-gfm').Options`, otherwise it will be `any` due to the bundle step doesn't bundle external types all together.

Despite this little problem, other functionalities just works fine.

manucorporat:
It's a big problem, no? if we merge this, it will break everybody's project because of missing types

swwind:
> It's a big problem, no? if we merge this, it will break everybody's project because of missing types

@manucorporat I think it's possible to inline all the types to handle this problem, but I think the original solution requires more API design thinking.

The original issue wants to have setup function for configuring `refractor`, but `qwik-city` bundled every dependency into a single package, and user manually installed `refractor` package version may differ than our bundled version. Combining pieces of code from different versions of a package has a high potential for bugs and vulnerabilities.

PR #3739, opened 2023-04-10
Update README.md

steve8708:
Minor tweaks to README


steve8708:
did in github UI, not passing prettier, will redo in code so can run prettier locally

PR #3744, opened 2023-04-11
fix home dir creation when running pnpm create qwik@latest

iseth:
# What is it?

- [x] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description

When creating a new qwik project and it prompts for a project name/path it suggests to use `./project_name` but if you use `~/Projects/qwik` it will make a `~` folder in the current dir. 

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/README.md
+++ b/packages/docs/README.md
@@ -1,135 +1,74 @@
-# Qwik Docs Site ⚡️
+<br>
+<p align="center">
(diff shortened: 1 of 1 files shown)


adamdbradley:
Good idea! Would you be able to fix the build by importing `os`? Then we can merge it, thanks

iseth:
Good catch just added it. Also manu suggested to use `~/` instead of `~` just in case someone wants to have a folder with the `~` for any reason lol.

PR #3799, opened 2023-04-16
fix(eslint-plugin-qwik): remove warning for bigint now that it can be serialized

steve8708:
via #3767


manucorporat:
Great fix!

PR #3807, opened 2023-04-17
Improve use 'use-not-root' behavious in es-lint plugin

Wimpert:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests

# Description

I expanded the `eslint-plugin-qwik` to support the following cases:

```ts

export const useMyCustom = () => {
  return myExpression() ??  useContext();
}

export const useSession3 = () => useContext()?.value;

```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/eslint-plugin-qwik/qwik.unit.ts
+++ b/packages/eslint-plugin-qwik/qwik.unit.ts
@@ -57,22 +57,6 @@ export const useSession5 = () => useContext().value + 10;
 
       `
 export const useSession1 = () => {
-  myExpression() ?? useContext();
-}
-
-export const useSession2 = () => {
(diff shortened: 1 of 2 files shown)


manucorporat:
I don't think this should be allowed:
```
export const useMyCustom = () => {
  return myExpression() ??  useContext();
}
```

Wimpert:
> I don't think this should be allowed:
> 
> ```
> export const useMyCustom = () => {
>   return myExpression() ??  useContext();
> }
> ```
> 
> In that example, `useContext()` is under a branch, just like an if statement

@manucorporat you are right, I removed support for that case.

PR #3808, opened 2023-04-17
docs: fix redirects

adamdbradley:


packages/docs/public/_routes.json
```diff
@@ -1,5 +1,5 @@
 {
   "version": 1,
-  "include": ["/*"],
+  "include": ["/"],
```
manucorporat, line 3: this will break Qwik AI in all pages but the front page

PR #3813, opened 2023-04-18
feat: server$ this will be defined most of the times

### LINKED ISSUE #3751: [🐞] `server$` RequestEvent is undefined during SSR

Qwik City (routing)

### Describe the bug

The `server$` function has an undefined RequestEvent attached to `this` when rendered on the server inside a `useTask$` or `useResource$`.

Understandably, it's just calling the function and not in fact making a request, however, this means we don't have access to necessary information to complete the server function, such as:

- `sharedMap` which is where i store session data
- `cookies` that I need to pass on with my RPC calls
- other `headers` such as a traceparent

Is it possible to still attach these to the function during SSR, the same event from the page request would make sense.

Example:

```tsx
const request = server$(async function () {
  console.log(this) // <- undefined
})

  useTask$(async ({ track }) => {
    await request()
  })
})
```

### Reproduction

https://github.com/DustinJSilk/qwik-server-scope

### Steps to reproduce

- Install & run
- see the logged `undefined` in the console
- click the button and see it is no longer undefined

### System Info

```shell
N/A
```

### Additional Information

_No response_

manucorporat:
fixes #3751

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/env-variables/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/env-variables/index.mdx
@@ -18,7 +18,7 @@ Build-time variables are powered by [Vite](https://vitejs.dev/guide/env-and-mode
 
 Notice that build-time variables should be considered as **public**, since they will be hardcoded in the browser build, which can be easily read by anyone.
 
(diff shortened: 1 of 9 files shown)


DustinJSilk:
Amazing thank you!!

KenAKAFrosty:
yay!! I'm very excited for this too thank you!!

PR #3860, opened 2023-04-20
Docs: The onSubmitCompleted$ handler

thejackshelton:
Tried my best to add a section about the onSubmitCompleted$ event handler in the docs.

Can someone explain to me the need for 

{15-18} /firstName/#a /lastName/#b /email/#c

At the top of the code examples? I left it out since I didn't quite understand what it meant. 

I hope my documentation changes can help! A friend in the discord came across this event handler in the type definitions and it was the perfect solution to my EditForm component in my todo app. Here's the live version to see how it works

https://qwik-todo-app.netlify.app/

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/action/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/action/index.mdx
@@ -131,48 +131,42 @@ Here's an example of the `onSubmitCompleted$` handler used to edit an item in a 
 ```tsx title="src/components/EditForm.tsx"
 import { component$, type Signal, useSignal } from '@builder.io/qwik';
 import { Form } from '@builder.io/qwik-city';
-import styles from './editform.module.css';
-import { Check, X } from 'example-icons';
-import type { listItem } from '../../routes/index';
-import { useEditFromListAction } from '../../routes/index';
+import { type ListItem, useEditFromListAction } from '../../routes/index';
 
 export interface EditFormProps {
   item: listItem;
   editingIdSignal: Signal<string>;
 }
 
-export const EditForm = component$(
+const EditForm = component$(
   ({ item, editingIdSignal }: EditFormProps) => {
     const editAction = useEditFromListAction();
 
     return (
-      <div class={styles.wrapper}>
+      <div>
         <Form
           action={editAction}
-          class={styles.form}
           onSubmitCompleted$={() => {
             editingIdSignal.value = '';
           }}
           spaReset
         >
           <input
-            ref={inputRef}
-            class={styles.edit}
             type="text"
             value={item.text}
             name="text"
             id={`edit-${item.id}`}
           />
-          {/* Include the hidden input field for the "id" attribute */}
+          {/* Sends item.id with form data on submission. */}
           <input type="hidden" name="id" value={item.id} />
-          <button class={styles.done} type="submit">
-            <Check />
+          <button type="submit">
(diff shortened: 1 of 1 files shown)


PatrickJS (approved):

packages/docs/src/routes/docs/(qwikcity)/action/index.mdx
```diff
@@ -122,6 +122,69 @@ export default component$(() => {
 
 In the example above, the `addUser` action is triggered when the user clicks the button. The `action.submit()` method returns a `Promise` that resolves when the action is done.
 
+## Actions with Event Handlers
+
+We can use the `onSubmitCompleted$` event handler after an action is executed successfully and returns some data. It is particularly useful when you need to perform additional tasks, such as resetting UI elements or updating the application state, once an action has been completed successfully.
+
+Here's an example of the `onSubmitCompleted$` handler used to edit an item in a EditForm component of a todo app.
```
hamatoyogi, line 129: I think the example has a bit too much out of scope code (like styles and such) as well as missing context of the parent passing down the signal.

AFK right now, but ideally the example would be the bare minimum code.

I'll see if I can get around to showing what I mean soon.

PatrickJS: yeah it might make sense to make a super simple version then link to an example thats more complex

manucorporat: Given that there is no docs at all about this, i am happy to merge and iterate, at least we have something

thejackshelton: > Given that there is no docs at all about this, i am happy to merge and iterate, at least we have something

Should I edit and remove the styles? Along with taking out the Icons and replacing X with "Cancel" for example?

thejackshelton:
@manucorporat I've gone ahead and stripped whatever I could to keep it as minimal as possible. I was able to get it down to 45 lines and checked that the edit functionality still works on my localhost after making the following changes.

thejackshelton:
@hamatoyogi let me know if there's anything I could've done further here! I think you mentioned getting some time to show what you meant earlier.

packages/docs/src/routes/docs/(qwikcity)/action/index.mdx
```diff
@@ -122,6 +122,63 @@ export default component$(() => {
...
+
+```tsx title="src/components/EditForm.tsx"
+import { component$, type Signal, useSignal } from '@builder.io/qwik';
+import { Form } from '@builder.io/qwik-city';
+import { type ListItem, useEditFromListAction } from '../../routes/index';
```
hamatoyogi, line 134: I think having the full context, e.g. having ListItem and useEditFromListAction in the same file would be better.

Otherwise, looking good!

thejackshelton: > I think having the full context, e.g. having ListItem and useEditFromListAction in the same file would be better.
> 
> Otherwise, looking good!

Gotcha. The useEditFromListAction and listItem is about another 25 lines. Should I add only those two parts? Or the entire file with the other actions removed?

PR #3885, opened 2023-04-22
Redirect docs fix - index.mdx

mattriese:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [x] Docs / tests

# Description

`redirect()` is on the `response` object, not the `RequestEvent` directly.  Also, the type definition file says (in a comment) that the default status code is 307, but here it says it's 302. When I try a redirect, I only see 200 status codes in the network tab (both when I set it to 301 and when I don't pass in a code, expecting the default), but maybe I'm doing it wrong. 

Also, the params of `redirect()` are reversed. 

Also, it would be nice if there was a brief note about why we are supposed to `throw` the response.redirect() return value, since it's not an `Error` object. I've never seen that pattern.

my `onGet`:
```
export const onGet: RequestHandler<BlogData> = async ({ params, response }) => {
  const res = await fetch("http://localhost:3000/blogs/" + params.id);
  if (!res.ok) {
    throw response.redirect("/", 301);
  }
  const data = await res.json();
  return data;
};
```

# Use cases and why

I am just starting to learn Qwik and this was the first docs page I needed help from. 
This is my first PR to a open source project :)

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/guides/redirects/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/guides/redirects/index.mdx
@@ -17,24 +17,24 @@ import type { RequestEvent } from '@builder.io/qwik-city';
 import { checkAuthorization } from '../auth'; // Your authorization code
 import type { DashboardData } from '../types'; // Your types
 
-export const onGet = async ({ cookie, redirect }: RequestEvent) => {
+export const onGet = async ({ cookie, response }: RequestEvent) => {
   const isAuthorized = checkAuthorization(cookie.get('cookie'));
 
   if (!isAuthorized) {
     // User is not authorized!
     // throw the redirect response to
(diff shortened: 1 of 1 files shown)


adamdbradley:
Sorry the version you're using may be outdated. The `redirect` method is on the Request event:
https://github.com/BuilderIO/qwik/blob/7fd6b9f5479559454922a18ade1195708b7fb66d/packages/qwik-city/middleware/request-handler/types.ts#L88-L94 and is `redirect(statusCode: RedirectCode, url: string)`

Here's an example of it in our e2e tests: https://github.com/BuilderIO/qwik/blob/7fd6b9f5479559454922a18ade1195708b7fb66d/starters/apps/qwikcity-test/src/routes/(common)/(auth)/sign-in/index.tsx#L27

PR #3892, opened 2023-04-23
fix(qwik): update object based vite config properly

dmitry-stepanenko:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests

# Description
1. Vite's `defineConfig` supports both function and object as a config, right now only functions are supported.
This means that configs defined like this will not be updated
```ts
export default defineConfig({
    plugins: [
      qwikVite(),
    ]
});
```

2. If plugin already exists with "plugins" array, it should not duplicate it
# Use cases and why
1. Users can update configs to the object-based one and qwik should still be able to update it.
2. Should not add duplicated values for the plugins

# Checklist:


manucorporat:
This is awesome! thank you very much!

PR #3938, opened 2023-04-26
fix(jsx types): make `QwikIntrinsicElements` extend `IntrinsicElements`, not `IntrinsicHtmlElements`

### LINKED ISSUE #3934: [✨] Make `QwikIntrinsicElements` extend `IntrinsicSVGElements`, or otherwise expose SVG element types

I'm trying to extend the `<svg>` builtin, and I'd like to be able to accept the same props and forward them on. I can't seem to make it happen, however, because [`QwikIntrinsicElements`](https://github.com/BuilderIO/qwik/blob/73ef1f9b6a2d7bb69d7f9fa4f2a360e69ab99339/packages/qwik/src/core/render/jsx/types/jsx-qwik-elements.ts#L23) doesn't extend [`IntrinsicElements`](https://github.com/BuilderIO/qwik/blob/73ef1f9b6a2d7bb69d7f9fa4f2a360e69ab99339/packages/qwik/src/core/render/jsx/types/jsx-generated.ts#L1168) but rather [`IntrinsicHtmlElements`](https://github.com/BuilderIO/qwik/blob/73ef1f9b6a2d7bb69d7f9fa4f2a360e69ab99339/packages/qwik/src/core/render/jsx/types/jsx-qwik-elements.ts#L23) I don't have access to those types.

Modify the `QwikIntrinsicElements` type to extend `IntrinsicElements`

I hadn't really considered any -- I guess another alternative is to expose `IntrinsicSVGElements` directly.

### Additional context

_No response_

caass:
# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Previously, it was impossible to access JSX types for SVG elements because `QwikIntrinsicElements` extended `IntrinsicHTMLElements`, rather than `IntrinsicElements` (which itself extends both `IntrinsicHTMLElements` and `IntrinsicSVGElements`.

fix #3934

# Use cases and why

Previously:

```tsx
// has type `QwikCustomHTMLAttributes<QwikCustomHTMLElement>`
type WeaklyTyped = QwikJSX.IntrinsicElements["svg"];
```

Now:
```tsx
// has type `SVGProps<SVGSVGElement>`
type StronglyTyped = QwikJSX.IntrinsicElements["svg"];
```

# Checklist:

    -  There were a couple test failures after implementing this, since now all of a sudden props for SVG were being typechecked where they weren't before (they were falling through to `QwikCustomHTMLAttributes`) -- if you want me to add more tests, let me know :)

### DIFF
--- a/packages/qwik/src/core/render/jsx/types/jsx-qwik-elements.ts
+++ b/packages/qwik/src/core/render/jsx/types/jsx-qwik-elements.ts
@@ -1,4 +1,9 @@
-import type { HTMLAttributes, IntrinsicHTMLElements, ScriptHTMLAttributes } from './jsx-generated';
+import type {
+  HTMLAttributes,
+  IntrinsicElements,
+  SVGAttributes,
+  ScriptHTMLAttributes,
+} from './jsx-generated';
 
 interface QwikScriptHTMLAttributes<T> extends ScriptHTMLAttributes<T> {
   events?: string[];
@@ -8,7 +13,12 @@ interface QwikCustomHTMLAttributes<T> extends HTMLAttributes<T> {
   [key: string]: any;
 }
 
+interface QwikCustomSVGAttributes<T> extends SVGAttributes<T> {
+  [key: string]: any;
+}
+
 interface QwikCustomHTMLElement extends HTMLElement {}
+interface QwikCustomSVGElement extends SVGElement {}
 
 /**
  * @public
@@ -20,7 +30,9 @@ export interface QwikIntrinsicAttributes {
 /**
  * @public
  */
-export interface QwikIntrinsicElements extends IntrinsicHTMLElements {
+export interface QwikIntrinsicElements extends IntrinsicElements {
   script: QwikScriptHTMLAttributes<HTMLScriptElement>;
-  [key: string]: QwikCustomHTMLAttributes<QwikCustomHTMLElement>;
+  [key: string]:
+    | QwikCustomHTMLAttributes<QwikCustomHTMLElement>
+    | QwikCustomSVGAttributes<QwikCustomSVGElement>;
 }
--- a/packages/qwik/src/core/api.md
+++ b/packages/qwik/src/core/api.md
@@ -516,15 +516,17 @@ export interface QwikFocusEvent<T = Element> extends SyntheticEvent<T, NativeFoc
     target: EventTarget & T;
 }
 
-// Warning: (ae-forgotten-export) The symbol "IntrinsicHTMLElements" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "IntrinsicElements" needs to be exported by the entry point index.d.ts
 //
 // @public (undocumented)
-export interface QwikIntrinsicElements extends IntrinsicHTMLElements {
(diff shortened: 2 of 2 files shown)


packages/qwik/src/core/render/jsx/types/jsx-generated.ts
caass, line 1: I went through and changed all of these props from `kebab-case` to `camelCase`, and that wound up fixing most of the test failures. I'm not sure if that was the right thing to do, but it seems like it...

packages/qwik/src/core/render/jsx/types/jsx-generated.ts
```diff
@@ -1155,6 +1155,8 @@ export interface SVGAttributes<T> extends AriaAttributes, DOMAttributes<T> {
   xmlBase?: string | undefined;
   xmlLang?: string | undefined;
   xmlns?: string | undefined;
+  // TODO: proper xml namespacing support
+  'xmlns:xlink'?: string | undefined;
```
caass, line 1159: I'd never heard of xml namespacing before, but it seems like actually you would want some type like ```
"xmlns" in Type ? { [`xmlns:${key}` for key in keyof SVGAttributes]: SVGAttributes[key] } : {}
``` and `&` it...or something...

File: packages/qwik/src/core/render/jsx/types/jsx-qwik-elements.ts [hunk 2]
```diff
@@ -20,7 +30,9 @@ export interface QwikIntrinsicAttributes {
 /**
  * @public
  */
-export interface QwikIntrinsicElements extends IntrinsicHTMLElements {
+export interface QwikIntrinsicElements extends IntrinsicElements {
   script: QwikScriptHTMLAttributes<HTMLScriptElement>;
-  [key: string]: QwikCustomHTMLAttributes<QwikCustomHTMLElement>;
+  [key: string]:
+    | QwikCustomHTMLAttributes<QwikCustomHTMLElement>
+    | QwikCustomSVGAttributes<QwikCustomSVGElement>;
```
caass, line 37: This probably isn't accurate, I think doing this you'll be able to pass SVG attributes to HTML elements...which I think we don't want?

packages/qwik/src/core/render/jsx/types/jsx-generated.ts
manucorporat: We cant merge this PR, since the current mixed kebab-case, camelCase is the correct SVG api.
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute

yes, svg attributes are case sensitive and it's a mess

caass: Gotcha, I went ahead and reverted those changes.

gioboa:
@caass thanks for your great help. are you still on it?

packages/qwik/src/core/render/jsx/types/jsx-qwik-elements.ts
mhevery: This comment concerns me....

Also could we have some tests which dave `satisfies` to verify that TS is now happy with then new types?

gioboa:
I close this PR because it's stale.
Thanks @caass for your contribution. if you would like to go on with it, feel free to reopen.

PR #3974, opened 2023-04-29
chore(adapters entry file comments): slightly update/fix the adapters entry file initial comments

dario-piotrowicz:
@adamdbradley I've noticed you fixed the comments in https://github.com/BuilderIO/qwik/pull/3959 (I was about to do it as well but you beat me to it! 😁)

I think you've missed a broken URL, also a couple of the adapters don't have the comment (which they should I guess for consistency sake)

Opening this PR with such minor changes 🙂 

PS: unimportant but I've also added the `/docs/` path to all the urls just for consistency sake (if the version without the `docs` is preferred let me know and I can update the url to be consistent with that)

### DIFF
--- a/starters/adapters/azure-swa/src/entry.azure-swa.tsx
+++ b/starters/adapters/azure-swa/src/entry.azure-swa.tsx
@@ -1,3 +1,12 @@
+/*
+ * WHAT IS THIS FILE?
+ *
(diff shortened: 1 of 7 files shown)


manucorporat:
Thank you very much!

PR #4004, opened 2023-05-01
docs: fix header positioning

steve8708:
links were not centered the way the are supposed to be

### DIFF
--- a/packages/docs/src/components/header/header.css
+++ b/packages/docs/src/components/header/header.css
@@ -52,7 +52,7 @@ header .mobile-menu {
 
 header .menu-toolkit {
   @apply hidden lg:flex;
-  @apply pt-0 pb-6 lg:pt-0;
+  @apply pt-0 lg:pt-0 pb-0;
   height: 100%;
   align-items: center;
 }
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/header/header.css
```diff
@@ -52,7 +52,7 @@ header .mobile-menu {
 
 header .menu-toolkit {
   @apply hidden lg:flex;
-  @apply pt-0 pb-6 lg:pt-0;
+  @apply pt-0 lg:pt-0 pb-0;
```
zanettin, line 55: just out of curiosity: wouldn't we get the same result by deleting this line?

steve8708: I wanted to make as few edits as possible, so I first just removed pb-6, but then saw there was padding soming from elsewhere, so then added `pb-0`

I could've just changed `pb-6` to `pb-0` for a cleaner diff, but the change purely was to remove the bottom padding

zanettin: got it 🙏  just thought since all values are set to `0` now, we could remove the whole line since it does not affect anything. or at least i was not able to see any different behavior 👼  thanks for this adjustment 🙏 🚀

PR #4007, opened 2023-05-01
docs: fix header styles

steve8708:
light:
Screenshot 2023-05-01 at 1 42 34 PM
dark:
Screenshot 2023-05-01 at 1 42 31 PM

### DIFF
--- a/packages/docs/src/components/header/header.css
+++ b/packages/docs/src/components/header/header.css
@@ -13,6 +13,21 @@
   }
 }
 
+.header-container.home-page-header {
+  background: transparent !important;
+  color: white;
+}
+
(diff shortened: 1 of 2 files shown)


zanettin:
guess the labels of the screenshots are swapped 👼 and do you want to change the background color of the search slot as well to a darker bg? imo the white one does not fit that well. but maybe just me feeling 👼 

thanks again 🙏

steve8708:
@zanettin I'm totally game for whatever, basically was just making as few edits as needed to get an MVP out (the big glaring issue first being the bg color, and then the dark/light icon being the wrong color). definitely we can tweak the search bars too, no strong preference there on nmy side, that's just the default colors for those modes, and has a slight benefit I guess of keeping them the same so when you click around to other pages they have some consistency, but I don't think that's hugely important

PR #4047, opened 2023-05-03
Update Tailwind documentation

NickClark:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

The current documentations claims that there are required extra steps, which no longer appears to be the case. Some other details updated as well.


adamdbradley:
thanks!

Issue #4034, opened 2023-05-03
[🐞] `server$` doesn't work in unit test when binding to an event handler of DOM element


ahnpnl:
### Which component is affected?

Qwik Runtime

### Describe the bug

When using `server$` for a DOM event handler, for example `click` event handler, `server$` throws error
```
➤ YN0000: QWIK ERROR ReferenceError: SubmitEvent is not defined
➤ YN0000:     at /node_modules/@builder.io/qwik-city/index.qwik.mjs:759:30
➤ YN0000:     at Array.map (<anonymous>)
➤ YN0000:     at serverQrl_stuff_wOIPfiQ04l4 (/node_modules/@builder.io/qwik-city/index.qwik.mjs:758:31)
➤ YN0000:     at invoke (/builds/dist-dev/tsc-out/packages/qwik/src/core/use/use-core.js:60:6)
➤ YN0000:     at /builds/dist-dev/tsc-out/packages/qwik/src/core/qrl/qrl-class.js:63:18
➤ YN0000:     at invokeQRL (/builds/dist-dev/tsc-out/packages/qwik/src/core/qrl/qrl-class.js:82:23)
➤ YN0000:     at async Promise.all (index 0)
```

The bug occurs with Qwik version **>=0.104.0**

### Reproduction

https://github.com/ahnpnl/my-qwik-app/tree/bug/server-function

### Steps to reproduce

- Check out the repo URL, branch `bug/server-function`
- Install deps with `pnpm`
- Run `pnpm run test.unit`
- Observe the error

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1
    Memory: 108.77 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.6.5 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Firefox: 111.0.1
    Safari: 16.4
  npmPackages:
    @builder.io/qwik: 1.0.0 => 1.0.0 
    @builder.io/qwik-city: 1.0.0 => 1.0.0 
    undici: 5.22.0 => 5.22.0 
    vite: 4.3.3 => 4.3.3
```

manucorporat:
Note to self: we need to make sure isBrowser===false during testing

PR #4056, opened 2023-05-04
fix: test mode force isServer=true

manucorporat:


manucorporat:
Should also fix:
#4034

PR #4088, opened 2023-05-06
fix: no need for prop function

manucorporat:
This PR allows to not need `PropFunction` to create custom events:
```tsx
interface CmpProps {
  onClick$?: () => void;
}

export const Cmp = component$<CmpProps>((props) => {
  return (
    <div>
      <h1>CHILD</h1>
    </div>
  )
});
```


shairez:
Wow that's amazing! 😊

motss:
Does it make sense to pass `props.onClick$` to another component with custom `onClick$`?

```tsx
interface CmpProps {
  onClick$?: () => void;
}

export const Cmp = component$<CmpProps>((props) => {
  return (
    <div>
      <h1>CHILD</h1>
     <CmpProps onClick$={props.onClick$} /> {/* Type conflict between onClick$ and props.onClick$ */}
    </div>
  )
});
```

PR #4106, opened 2023-05-08
feat: HTMLFragment

manucorporat:


manucorporat:
Fix the difference between SSR and browser, virtual element rendering
Screenshot 2023-05-08 at 12 53 21

PR #4114, opened 2023-05-08
fix: export ClassList for easier typing of components with a `class` prop

steve8708:
a common pattern in React is to take className as a prop, like

```tsx
function MyComponent(props: { className?: string }) {
  return <div className={`${props.className || ''} my-class`}>...</div>
}

<MyComponent className="some-class" />
```

But Qwik supports a more elegant API for classes, but it is a bit awkward to work with the types for as the actual type is not exported.

So today I do this:

```tsx
import type  { HTMLAttributes } from '@builder.io/qwik'

function MyComponent(props: { class?: HTMLAttributes<HTMLElement>["class"] }) {
  return <div class={[props.class, 'my-class']}>...</div>
}
```

but it would be much nicer to just do

```tsx
import type  { ClassList } from '@builder.io/qwik'

function MyComponent(props: { class?: ClassList }) {
  return <div class={[props.class, 'my-class']}>...</div>
}
```

This PR exports ClassList so we can do this

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Checklist:

### DIFF
--- a/packages/docs/src/routes/api/qwik/api.json
+++ b/packages/docs/src/routes/api/qwik/api.json
@@ -79,6 +79,20 @@
       "mdFile": "qwik.resourcectx.cache.md"
     },
     {
+      "name": "ClassList",
+      "id": "classlist",
+      "hierarchy": [
+        {
+          "name": "ClassList",
+          "id": "classlist"
+        }
+      ],
+      "kind": "TypeAlias",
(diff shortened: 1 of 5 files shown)


manucorporat (approved):
love it!

shairez:
great PR @steve8708 ! this will be useful in Qwik UI as well

BTW, you can run `pnpm fmt.staged` and it'll fix the prettier error just for the staged changes (more convenient than running on everything)

PR #4119, opened 2023-05-08
docs: add qwik-auth route protection example

ulic75:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Add an example of how to provide route protection

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/integrations/authjs/index.mdx
+++ b/packages/docs/src/routes/docs/integrations/authjs/index.mdx
@@ -230,3 +230,18 @@ AUTH_SECRET=
 
 4. The application is now ready to impliment authentication using Auth.js.
 5. Enjoy!
+
+### Route Protection
+
+Session data can be accessed via the route `event.sharedMap`.  So a route can be protected and redirect using something like this located in a `layout.tsx` or page `index.tsx`:
+
+```tsx
+export const onRequest: RequestHandler = (event) => {
+  const session: Session | null = event.sharedMap.get('session');
+  if (!session || new Date(session.expires) < new Date()) {
+    throw event.redirect(302, `/api/auth/signin?callbackUrl=${event.url.href}`);
+  }
+};
+```
+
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/integrations/authjs/index.mdx
```diff
@@ -230,3 +230,18 @@ AUTH_SECRET=
...
+```tsx
+export const onRequest: RequestHandler = (event) => {
+  const session: Session | null = event.sharedMap.get('session');
+  if (!session || new Date(session.expires) < new Date()) {
+    throw event.redirect(302, `/api/auth/signin?callbackUrl=${event.url.href}`);
+  }
+};
+```
+
+> Note: If placed in a layout.tsx ensure that the redirect destination does not share the same layout.tsx or a redirection loop could occur.
```
zanettin, line 247: thank you @ulic75 🙏 
should we check the current location as well on the expression to avoid a potential loop? like when you're already on the sign in page, do not redirect?

ulic75:
Seems like a good idea. I'm out of town for a while.

packages/docs/src/routes/docs/integrations/authjs/index.mdx
ulic75: I'm back from my trip.  So in thinking about your question, I guess maybe that would be an implementation detail of the particular site/app.  The `/api/auth/...` routes are dynamically created by the plugin@auth and won't run an onRequest event located in the layout.tsx.  Now if the developer was redirecting to an actual route location they built then yes, they wouldn't want to add some kind of additional check.

KaizenProgrammer:
This was a great help to me...took me a while to find this since it's still an open PR. But it works as expected for my route protections. I'm using a narrow layout-app.tsx and the index@app.tsx convention and it appears to work correctly.

zanettin (approved):
Hi @ulic75 👋  
imo we could merge this. objections?

ulic75:
> Hi @ulic75 wave imo we could merge this. objections?

Yeah, let's do it.

PR #4160, opened 2023-05-10
[DX] - Add autocompletion for event names for useOnX

barel-mishal:
Overview
This pull request introduces an enhancement to the Qwik framework that improves the developer experience by providing autocompletion for well-known event names when using the useOn, useOnDocument, and useOnWindow functions. The change maintains support for custom event names as strings, offering flexibility to developers while making it easier to work with predefined event names.

The key features of this PR include:

Implementation of the CustomEventType type using LiteralUnion in scripts/util, which combines well-known event names with the string type.
Updates to the useOn, useOnDocument, and useOnWindow functions to utilize the CustomEventType.
Improved developer experience with autocompletion for well-known event names.

This enhancement contributes to Qwik's core functionality and is expected to be valuable for the entire Qwik community (Thanks chat-gpt that help create this pr and provide big excitement about my changes 🙈 ). 

I was wandering for the place to put my changes eventually decided put them in scripts/util and import the CustomEventType.
But still wander if that a good place please let me know if you are think there is better place more appropriate.


manucorporat:
Hey! thanks a lot for putting this together, maybe this https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/core/render/jsx/types/jsx-qwik-events.ts
would be a better spot, maybe we can even reuse something we got there

manucorporat:
or here: https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/core/render/jsx/types/jsx-qwik-attributes.ts#L23

packages/qwik/src/core/use/use-on.ts
```diff
@@ -18,7 +19,7 @@ import { useInvokeContext } from './use-core';
  * @public
  */
 // </docs>
-export const useOn = (event: string | string[], eventQrl: QRL<(ev: Event) => void> | undefined) =>
+export const useOn = (event: CustomEventType, eventQrl: QRL<(ev: Event) => void> | undefined) =>
```
JorgeVV, line 22: `event`'s type should be `CustomEventType | CustomEventType[]`. Same for `useOnDocument` and `useOnWindow`'s `event` parameter.

barelmishal: Right I am on it thanks for the feedback!

barel-mishal:
I did this changes:
1. move the code I have added into https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/core/render/jsx/types/jsx-qwik-events.ts @manucorporat like you suggested.
2. I have implemented the right type in useOn, useOnDocument and useOnWindow from CustomEventType to CustomEventType | CustomEventType[]

manucorporat:
Might need to run `pnpm fmt` at the root of the project to fix the lint issues!! i think we will be able to merge then

barel-mishal:
> Might need to run `pnpm fmt` at the root of the project to fix the lint issues!! i think we will be able to merge then

I need some assistance because I'm not sure why my changes aren't showing up in the PR after running ```pnpm fmt``` and making some additional code improvements. I've pushed my changes, but they aren't visible in the PR. Can you help me understand why this might be happening?

barel-mishal:
> when doing `git push` make sure you push it to the right remote. not sure which one is in your case. maybe try `git push` alone

Ok that fine now!! I am explaining what the additional changes I did please approve or comment thanks a lot for your fest feedback

barel-mishal:
Removed the DomEvent type from jsx-qwik-events.ts by utilizing existing DOM event types in jsx-qwik-attributes.ts. This reduces code duplication by using the pre-existing list of DOM API events provided by Qwik. As a result, the code becomes more maintainable and we can reduce the amount of code shipped.

manucorporat:
you might have to run:
```
pnpm api.update
```
becuase the public api has changed

barel-mishal:
> you might have to run:
> 
> ```
> pnpm api.update
> ```
> 
> becuase the public api has changed

Did it
Thanks for point me to this direction

barel-mishal:
This change is passing all the tests.

barel-mishal:
Please let me know is something is missing to complete the pr

PR #4196, opened 2023-05-13
fix: fix issue where RequestEvent.platform is overwritten in dev-server

DanielAdolfsson:
# Overview

In developer mode, RequestEvent.platform is always overwritten. `incomingMessage` is always undefined because `normalizeOptions` ensures that `opts.platform` is always set. 

As I don't know the purpose of being able to override the platform object using userOpts, I decided to go for `Object.assign` for this. Allowing us to override what we want, and leave the rest (that came from `fromNodeHttp`). I removed the surrounding `if`, as normalizeOptions already guarantees that it's true-ish.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


manucorporat:
Thank you very much! seems reasonable!

Issue #4199, opened 2023-05-13
[🐞] Unable to install into current directory


Arctomachine:
### Which component is affected?

Starters / CLI

### Describe the bug

Related: #3901

The solution introduced to that feature request is not sufficient. Whenever I try to create new project, I get following error in console
```
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: EBUSY: resource busy or locked, rmdir 'C:\Users\user\projects\qwik-test'] {
  errno: -4082,
  code: 'EBUSY',
  syscall: 'rmdir',
  path: 'C:\\Users\\user\\projects\\qwik-test'
}
```
It looks like ide is not allowing to delete its project settings folder while running.
Cannot reproduce it in online sandbox, only locally

### Reproduction

It has to be reproduced locally and manually

### Steps to reproduce

1. Open ide
2. Create new project from ide interface
3. Run `npm/yarn/pnpm create qwik@latest` in terminal
4. Select current `.` folder as destination
5. Agree to overwrite it when asked
6. Script fails with error mentioned above

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (12) x64 AMD Ryzen 5 2600 Six-Core Processor
    Memory: 12.89 GB / 23.95 GB
  Binaries:
    Node: 18.13.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.19.3 - C:\Program Files\nodejs\npm.CMD
```

manucorporat:
Seems like something is using the folder...
`EBUSY: resource busy or locked, rmdir`

do you have any editor or something open?

Arctomachine:
Yes, naturally

manucorporat:
The current idea works by removing the folder and creating a new one, vscode is locking the folder, so the CLI can not remove it.

This could be fixed by trying to avoid removing the folder and instead remove all the content, in here:
https://github.com/BuilderIO/qwik/blob/9be6588e6ff83ed704b0d117e71ae5a434b0d51a/packages/create-qwik/create-interactive.ts#L74

Would u be up to make a PR? :)

Arctomachine:
I can try to

Arctomachine:
How can I test this script locally to see how it works with my changes?

Arctomachine:
I too support for installing without deleting anything.

zanettin:
[...]

PR #4209, opened 2023-05-13
fix(create-interactive.ts): make script remove only contents of targe…

Arctomachine:
…t folder

Instead of removing entire target folder, which can be locked by editor, only remove contents of this folder

#4199

# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Makes installation script remove contents of target folder instead of deleting folder itself to prevent error when it is locked by editor.

# Use cases and why

- 1. When running installation script from inside open editor

# Checklist:

### DIFF
--- a/packages/create-qwik/create-interactive.ts
+++ b/packages/create-qwik/create-interactive.ts
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import fs from 'node:fs';
-import path, { join, relative } from 'node:path';
+import { join, relative } from 'node:path';
 import {
   text,
   select,
@@ -75,7 +75,7 @@ export async function runCreateInteractiveCli() {
         .readdir(outDir)
         .then((filePaths) =>
           filePaths.forEach((pathToFile) =>
-            fs.promises.rm(path.join(outDir, pathToFile), { recursive: true })
+            fs.promises.rm(join(outDir, pathToFile), { recursive: true })
           )
         );
     }
--- a/packages/create-qwik/create-interactive.ts
+++ b/packages/create-qwik/create-interactive.ts
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 import fs from 'node:fs';
-import { join, relative } from 'node:path';
+import path, { join, relative } from 'node:path';
 import {
   text,
   select,
@@ -61,7 +61,7 @@ export async function runCreateInteractiveCli() {
       )}" already exists and is not empty. What would you like to do?`,
       options: [
         { value: 'exit', label: 'Do not overwrite this directory and exit' },
-        { value: 'replace', label: 'Overwrite and replace this directory' },
+        { value: 'replace', label: 'Remove contents of this directory' },
       ],
     });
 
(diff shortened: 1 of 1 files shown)


manucorporat:
Looks good! cc @adamdbradley

packages/create-qwik/create-interactive.ts
```diff
@@ -71,7 +71,13 @@ export async function runCreateInteractiveCli() {
     }
 
     if (existingOutDirAnswer === 'replace') {
-      removeExistingOutDirPromise = fs.promises.rm(outDir, { recursive: true });
+      fs.promises
+        .readdir(outDir)
+        .then((filePaths) =>
+          filePaths.forEach((pathToFile) =>
+            fs.promises.rm(path.join(outDir, pathToFile), { recursive: true })
```
zanettin, line 78: thx a lot for the adjustments 🙏 will give it a spin asap 👍 

regarding the code changes: i guess you could write it just like this:
```suggestion
            fs.promises.rm(join(outDir, pathToFile), { recursive: true })
```
since `join` is already a named import 🙏

Arctomachine: Right, I will commit and push it now

packages/create-qwik/create-interactive.ts
```diff
@@ -71,7 +71,13 @@ export async function runCreateInteractiveCli() {
     }
 
     if (existingOutDirAnswer === 'replace') {
-      removeExistingOutDirPromise = fs.promises.rm(outDir, { recursive: true });
```
zanettin, line 74: just tested it and it looks like we don't need the `removeExistingOutDirPromise` anymore. In the current implementation, after confirming the deletion, the directory is removed bevor the type of the app is selected.

this is now also the case with this implementation. imo i would wait with the deletion till the type (qwik city, library, docs page, etc) is selected but @adamdbradley pls advice on this 👼 

thanks again for the adjustments 🙏

packages/create-qwik/create-interactive.ts
Arctomachine: Done

packages/create-qwik/create-interactive.ts
zanettin: ps: this is also the reason why the CI isn't happy atm

Arctomachine: Oh, now that I look at it, there indeed was that. Should I assign new code back to this variable?

zanettin: imo we could remove it everywhere and just move the code execution below tbe type selection to give the user another 'escape hatch'. but i'd wait for @adamdbradley or @manucorporat s feedback tbh 😃

manucorporat (changes requested):

packages/create-qwik/create-interactive.ts
```diff
@@ -71,7 +71,13 @@ export async function runCreateInteractiveCli() {
...
+      fs.promises
```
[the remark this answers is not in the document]
Arctomachine: Updated this and next one

manucorporat:
@jordanw66 it's a good question, but... i think it would introduce situations where the starter is broken and getting people confused. This tool is built around creating a new starter not adding things to a existing project.

So i think, removing is important in order to keep the tool reliable and reduce the scope

manucorporat (approved):

zanettin:
let's merge this 🚢  thx again @Arctomachine 🙏

PR #4214, opened 2023-05-14
Add support for server$ cancelation using AbortSignal

DanielAdolfsson:
# Overview

When using `yield` in a server function, it makes no sense that it will continue even after the client connection has ended. In addition, I found that I needed to be able to cancel long running server functions from the client-side somehow. 
This PR adds support for passing an AbortSignal object when calling the server function.

This is not a complete PR yet, I merely want to check the temperature and see if this is an acceptable addition to the framework. The code works now, but only tested with the dev server.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This PR introduces a couple of changes:
1. It adds a second signature to the QRL interface; to allow passing an AbortSignal object as a first argument to the server function.
2. Amends `serverQrl` to allow passing an AbortSignal - in which case it will be handed over to `fetch`. In essence, this just terminates the connection to the server.
4. Fixes the node middleware to ensure the writable stream is closed when the connection is closed.

# Use cases and why

I wanted to stream OpenAI generated text to the client by yielding new tokens. But quickly noticed that the server function would continue yielding tokens- even after leaving the page. This could be solved with a dedicated route, but I think being able to cancel long running server functions is not an outlandish idea.

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/http.ts
+++ b/packages/qwik-city/middleware/node/http.ts
@@ -87,13 +87,15 @@ export async function fromNodeHttp(
           res.on('close', () => controller.error());
         },
         write(chunk) {
-          return new Promise((resolve, reject) => res.write(chunk, cb => {
-            if (cb) {
-              reject(cb);
-            } else {
-              resolve();
-            }
-          }))
+          return new Promise((resolve, reject) =>
+            res.write(chunk, (cb) => {
+              if (cb) {
+                reject(cb);
+              } else {
+                resolve();
+              }
+            })
+          );
         },
         close() {
           res.end();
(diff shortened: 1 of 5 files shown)


manucorporat:
This is amazing!

DanielAdolfsson:
Might have been a little premature to merge it immediately, I haven't checked that I haven't broken anything. I'll do some more checking now then!

amit13k:
Thanks a lot for for merging this feature. I just tried using this with an AbortSignal.Timeout. After abort, I get "Uncaught (in promise)". It seems there is no way to properly handle that exception. Any suggestions on how to fix this ?

DanielAdolfsson:
You mean on the client side?

amitkarmakar13:
> You mean on the client side?
I tried this example,
https://stackblitz.com/edit/qwik-starter-6ajbb2?file=src%2Froutes%2Findex.tsx

I can seeing this in console, https://cdn.discordapp.com/attachments/1105635284781117532/1108622570170888274/image.png

DanielAdolfsson:
PR #4254 should fix this problem

DanielAdolfsson:
I think that's a good idea, actually.

PR #4224, opened 2023-05-14
QRL::bind() is QRL / Issue2972 rewritten

mesterum:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

[PR for issue: 4223](https://github.com/BuilderIO/qwik/issues/4223)

# Checklist:

- [x] Replaced the Issue2972 test with mine

### DIFF
--- a/packages/qwik/src/core/qrl/qrl-class.ts
+++ b/packages/qwik/src/core/qrl/qrl-class.ts
@@ -159,7 +159,9 @@ export const createQRL = <TYPE>(
     $captureRef$: captureRef,
     dev: null,
   };
-  const qrl = Object.assign(invokeQRL, methods);
+  Object.setPrototypeOf(methods, Function.prototype)
+  seal(methods)
+  const qrl = Object.setPrototypeOf(invokeQRL, methods);
   seal(qrl);
   return qrl as any;
 };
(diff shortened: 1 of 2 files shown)


mesterum:
I admit that this solution is half baked. It doesn’t hold over the wire. The bind arguments must be captured too.

packages/qwik/src/core/qrl/qrl-class.ts
```diff
@@ -159,7 +159,9 @@ export const createQRL = <TYPE>(
     $captureRef$: captureRef,
     dev: null,
   };
-  const qrl = Object.assign(invokeQRL, methods);
+  Object.setPrototypeOf(methods, Function.prototype)
+  seal(methods)
+  const qrl = Object.setPrototypeOf(invokeQRL, methods);
```
mhevery, line 164: Why do we have to do this? This has a big performance penalty.: https://stackoverflow.com/questions/52654781/why-does-using-object-setprototypeof-impact-performance-heavily

mhevery (changes requested):
1. It introduces a negative performance of using `Object.setProtypeOf`
2. It will not survive serialization over the wire.

mhevery:
If there is no progress on this PR in the next few weeks, I would suggest closing it.

mesterum:
I didn’t know about the penalty of using Object.setPrototypeOf(). Using
Object.create() is not possible be instead of the last line because the qrl
must be a callable function.
Why do I want the methods be in the prototype chain(PC). Because the
f.bind(…args) function creates another function with all the PC of f. If
methods object is not in PC of qrl, it will be not in qrl.bind().

mesterum:
1.
   2. It will not survive serialization over the wire.

I don’t know how to do it.

PR #4234, opened 2023-05-16
docs: Link reload does not work without href attribute

alimrb:
Link reload doesn't work without href

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/routing/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/routing/index.mdx
@@ -231,7 +231,7 @@ export default component$(() => {
 
   return (
     <div>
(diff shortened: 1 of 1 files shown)


zanettin:
Hi @alimrb 👋 
Thanks a lot for this adjustment 🙏  tbh i am not sure if this is not a bug 👼 
IMO the condition here should be adjusted: https://github.com/BuilderIO/qwik/blob/main/packages/qwik-city/runtime/src/link-component.tsx#L25

to:
`if(elm.href || reload) { ... }`

@adamdbradley can you confirm that or am i mistaking 👼

PR #4254, opened 2023-05-18
serverQrl: simplify event streaming, and get rid of unhandled rejections

DanielAdolfsson:
# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

When streaming events, we're using a transformer to turn a stream of bytes to a stream of events.
This transformer is only used for this specific purpose, and is a bit overkill for what it's supposed to do.
In addition, we need to pipe the data from the response body into the transformer. Any errors while transforming (e.g. an abort signal) will simply remain unhandled as we're ignoring the returned Promise from `pipeTo`.

This PR simplifies the event streaming code, and removes the transformer. The logic will now be done in `streamEvents`.

The user should now be able use abort signals without having unhandled rejections. The user will still have to catch the rejection on their end.

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/server-functions.ts
+++ b/packages/qwik-city/runtime/src/server-functions.ts
@@ -325,7 +325,9 @@ export const serverQrl: ServerConstructorQRL = (qrl: QRL<(...args: any[]) => any
 
         const contentType = res.headers.get('Content-Type');
         if (res.ok && contentType === 'text/event-stream') {
-          return streamEvents(res.body!, ctxElm ?? document.documentElement);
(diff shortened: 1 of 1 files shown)


DanielAdolfsson:
Might not be ideal to parse this manually, there are good libraries for parsing SSEs. The latest commit makes sure the SSE is properly parsed [according to spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream) (I hope - should add tests!). The new signature is `streamEvents(stream[, transform])`.

DanielAdolfsson:
Closing this PR in favor of #4270

PR #4290, opened 2023-05-22
server$ fixes, and improvements

### LINKED ISSUE #4268: [📖] Add docs about `AbortSignal` in `server$` API
### Suggestion

#4214 added support for `AbortSignal` in `server$` API
However, it's undocumented and not that clear how to use it.
[server$ - streaming responses](https://qwik.builder.io/docs/server$/#streaming-responses) seems like a good place where to add docs for this feature.

DanielAdolfsson:
# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

This PR improves cancellation when using server$ async generator.

- Simplify the streaming code, and use a simple newline delimited result with `qwik-json-stream` content type.
- Fix a bug that could lead to ERR_STREAM_DESTROYED.
- Add proper async generator cleanup. Now, whenever the client breaks out from an async generator for-loop (or otherwise call `return()` on the generator), the connection will be closed.
- Updates server$ documentation (fixes #4268)

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/http.ts
+++ b/packages/qwik-city/middleware/node/http.ts
@@ -86,16 +86,14 @@ export async function fromNodeHttp(
         start(controller) {
           res.on('close', () => controller.error());
         },
-        write(chunk) {
-          return new Promise((resolve, reject) =>
-            res.write(chunk, (cb) => {
-              if (cb) {
-                reject(cb);
-              } else {
-                resolve();
-              }
-            })
-          );
+        write(chunk, controller) {
+          res.write(chunk, (error) => {
+            if (error) {
+              // FIXME: Ideally, we would like to inform the writer that this was an error.
+              //        Not all writers seem to handle rejections, though.
+              // controller.error(error);
+            }
+          });
         },
         close() {
           res.end();
--- a/packages/docs/src/routes/docs/(qwikcity)/server$/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/server$/index.mdx
@@ -13,6 +13,12 @@ contributors:
 
 `server$` is a form of RPC (Remote Procedure Call) mechanism between the client and server, just like a traditional HTTP endpoint but strongly typed thanks to Typescript, and easier to maintain.
 
+Your new function will have the following signature:
+`([AbortSignal, ] ...): Promise<T>`
+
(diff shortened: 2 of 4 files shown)


packages/qwik-city/middleware/node/http.ts
```diff
@@ -86,16 +86,14 @@ export async function fromNodeHttp(
...
-                reject(cb);
-              } else {
-                resolve();
-              }
-            })
-          );
+        write(chunk, controller) {
+          res.write(chunk, (error) => {
+            if (error) {
+              // FIXME: Ideally, we would like to inform the writer that this was an error.
```
manucorporat, line 92: this does not seem right

manucorporat: i dont want to merge something that looks WIP, what do we do here?

DanielAdolfsson: The original code, before my previous PR.

```
write(chunk) {
   res.write(chunk);
}
```

We were basically throwing away errors from res.write(). Now, if we pass the error to the controller, Node will instead die because there is a floating promise somewhere that isn't taking care of errors. 

In any case, I wouldn't say it's a WIP. This fix is to make sure Node does not die by going back to how it was before. The note I made there is because people reading the code should be aware that errors disappear there. We can go back to it just being the above three lines if you like, but it would probably be a good idea to make a note somewhere that we are very much aware of what's going on there.

manucorporat:
There we go! I will try to add some warning and new tests later!

shairez:
great job @DanielAdolfsson ! thanks for this PR! 🙏

thekip:
@DanielAdolfsson it seems the issue is still not fixed. It's indeed doesn't crash with `ERR_STREAM_DESTROYED` but it crashes with 
```
 node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

In the same circumstances where it crashed before. I tried to debug and find a root of this promise, but had no luck.  
The issue is happend when server receives few concurrent requests  (i able to reproduce it hitting f5 quickly)

The promise is rejected with 'undefined' object.

PR #4314, opened 2023-05-23
fix: store `qsymbol` events for easier optimization

### LINKED ISSUE #4181: [✨] store `qsymbol` events in an array on `window`

If in application has `useVisibleTask` it can be hard to register `addEventListener("qsymbol"` early enough to see oll of the `qsymbol`s as they are leaded.

Since we already have `qsymbol` listener, it would be nice if the listeners would add the result into `window.qwiksymbols` array. this way it would be very easy for the app developer to see which symbols have been loaded up to this point.

```
const qwiksymbols: {name: string, timestamp: 0}[]
```

Maybe just do something like
```
addEventListener('qsymbol', (e) {
  window.qwidSymobls.push(e.details);
});
```

registering it yourself, but it is hard to ensure tha the registration happens before `useVisibleTask` wakes up.

### Additional context

_No response_

mhevery:
fix: store `qsymbol` events for easier optimization
    
    fix #4181

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/components/real-metrics-optimization/real-metrics-optimization.tsx
+++ b/packages/docs/src/components/real-metrics-optimization/real-metrics-optimization.tsx
@@ -6,12 +6,6 @@ export default (props: RealMetricsOptimizationProps) => (
   />
 );
 
-interface QSymbolEventDetail {
-  symbol: string;
-  element: HTMLElement;
-  reqTime: number;
-}
-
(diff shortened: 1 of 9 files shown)


mhevery:
closing in favor of Insights.

Issue #4319, opened 2023-05-24
Link may handle click events multiple times 


jwickers:
I have a page where a list of links is loaded and I switched them to use Link but I ran into this weird bug:
- I click on one of the links
- the Link causes the SPA navigation
- the links are changed according to the next page data which causes the links HREF to change in place
- the click event is then somehow triggered again
- this causes another immediate SPA navigation

Since there does not seem to be anything wrong with my application, I debugged the Link component to confirm it was handling the same event twice.
I use this little hack for now which solves the issue ( in `link-component.tsx` ):

```typescript
  const handleClick = event$(async (event: any, elm: HTMLAnchorElement) => {
    // >> check if the event was already handled
    if (elm.href && !event._qwik_nav_handled) {
      elm.setAttribute('aria-pressed', 'true');
      await nav(elm.href, reload);
      elm.removeAttribute('aria-pressed');
      // >> mark the event
      event._qwik_nav_handled = true;
    }
  });
``` 

Not sure if there is a better way to handle this.

jwickers:
Added PR #4320

jwickers:
Looks like this was caused by a third party lib emitting programmatic clicks in some cases, closing this.

PR #4320, opened 2023-05-24
#4319 prevent Link from handling the same click event multiple times

jwickers:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Prevents the Link component Click handler from handling the same event multiple times.
This may happen when a Link is updated due to the SPA navigation and the same DOM element reprocess the same click event with a different href.

# Checklist:


manucorporat (changes requested):
Thanks for the PR, it's just hiding a possibly real bug. we need to debug why it's happening, and fix it for everybody. this feels like a hack.

jwickers:
After more testing and debuggung i was a weird bug caused by a third party lib, not qwik. Closing this.

PR #4329, opened 2023-05-24
perf: faster resuming

manucorporat:
move style moving to the render pipeline


packages/qwik/src/core/container/resume.ts
```diff
@@ -115,7 +115,6 @@ export const resumeContainer = (containerEl: Element) => {
 
   const inlinedFunctions = getQwikInlinedFuncs(parentJSON);
   const containerState = _getContainerState(containerEl);
-  moveStyles(containerEl, containerState);
```
manucorporat, line 118: this is the expensive line removed

PR #4361, opened 2023-05-28
fix(vercel): [vercel-edge adapter] allow to use the `node` target

Schniz:
# Overview

Introduce a new `target` option for the Vercel Edge Functions adapter
to allow usage of `process.env`

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The current Vercel Edge Functions adapter uses `webworker` as the target, which
inlines an empty object on all `process.env` references.

This is problematic for packages that read `process.env` at runtime, such as
`@vercel/postgres`.

This change introduces a new config option for the adapter (`target`) which
allows to use `node` target, instead of `webworker`.

In the long term, I see this becoming the _default_ behavior, but for now, I
want to make sure that we don't break existing customers of the adapter: I'm
not sure how your versioning pattern works here.

This also fixes the `conditions` option to the values defined in Next.js

see https://github.com/orgs/vercel/discussions/2385

# Use cases and why

Using `@vercel/postgres` in a Qwik app deployed to Vercel.

# Checklist:


manucorporat:
Unfortunately this is not a good fix, i am in contact with the vite team to ship a fix in vite to prevent the processEnv replacement! we should use webworker + no process.env replacement

Vercel made is harder by choosing `process.env` a node api in a non-node environment. </rant?

Schniz:
Why is this fix not good?

* Vercel Edge Functions supports global Buffer, and several native node modules
* Vercel Edge Functions use process.env

manucorporat:
does vercel support several native node modules? maybe u right, looking

shairez:
@gioboa - Reopening it until @manucorporat could come back with an answer

Just to give context - @Schniz is one of the guys from Vercel who is actually working on the edge functions, so he is the perfect person IMO to help with the qwik-vercel integration 😊

Looking at the changes, and how it's only affecting this specific integration, is there any reason not to merge it? are there any potential issues it will create in the Qwik SSR process in vercel?

packages/qwik-city/adapters/vercel-edge/vite/index.ts
```diff
@@ -131,6 +134,14 @@ export interface VercelEdgeAdapterOptions extends ServerAdapterOptions {
    * come from a static file, rather than a server-side rendered response.
    */
   staticPaths?: string[];
+
+  /**
+   * Define the `target` proeprty in the `ssr` property in the `vite.config.ts` file.
+   *
+   * Defaults to `webworker` for not having a breaking change. But `node` will become the default
```
mhevery, line 141: I need some clarification here. How can `edge` have `node` as the default? I would expect that edge is `webworker` API not `node` API.

Schniz: I understand that it's confusing, but `webworker` does not fit 100% too, right? `process.env` exists on edge, but not on webworker environments. we have Buffer on global. We have some Node.js module support. It's not completely Node.js and not completely webworker, but something in between <picture data-single-emoji=":thonk:" title=":thonk:">:thonk:</picture>

Schniz:
my PR does not change the default behavior btw (as far as I remember <picture data-single-emoji=":lol:" title=":lol:">:lol:</picture>), it merely adds an experimental configuration that people can try out. if this works well for them, we can consider having that as a breaking change in the adapter in the future.

packages/qwik-city/adapters/vercel-edge/vite/index.ts
mhevery: In that case, would it not be better to set it to `webworker` and than add the missing parts from `node`. My worry is that if we make it `node`, than things that should be valid (because you think you have node) are not.

If I understand this correctly, this switch controls the `.d.ts` files used for compilation and it will under-restrictive.

mhevery:
Also since you are changing public API
```

❌ Use "pnpm api.update" to automatically update the .md files if the api changes were expected
 Error: Use "pnpm api.update" to automatically update the .md files if the api changes were expected
    at panic (file:///home/runner/work/qwik/qwik/scripts/util.ts:182:4)
    at createTypesApi (file:///home/runner/work/qwik/qwik/scripts/api.ts:187:5)
    at apiExtractor (file:///home/runner/work/qwik/qwik/scripts/api.ts:109:3)
    at build (file:///home/runner/work/qwik/qwik/scripts/build.ts:98:13)
Error: Process completed with exit code 1.
```

gioboa:
Hi @Schniz are you still on it or is it abandoned?

Schniz:
Hi, I don't have the time to do that right now (the whole Israel situation). I need to fix the `pnpm api.update` to update the docs. Feel free to continue if you want to... I think that it is pretty much settled that the team does not want to change from `webworker` to `node` so I'm not sure how we can continue contributing to this PR. I'd be happy to make sure Qwik works well for Vercel customers <picture data-single-emoji=":pray_parrot:" title=":pray_parrot:">:pray_parrot:</picture>

gioboa:
Thanks for your fast feedback, I will check it out.

PR #4406, opened 2023-06-04
docs: update the previous form state section

wtlin1228:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

https://github.com/BuilderIO/qwik/issues/3861

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/action/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/action/index.mdx
@@ -307,14 +307,16 @@ Thanks to Typescript type discrimination, you can use the `action.value.failed` 
 
 ## Previous form state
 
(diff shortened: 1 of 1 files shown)


manucorporat:
I dont think it's correct, and if it's the case it's a bug, it should not change if JS is enabled or not.

wtlin1228:
> I dont think it's correct, and if it's the case it's a bug, it should not change if JS is enabled or not.

Indeed, you're correct. The value of `action.formState` will be updated to the previous submit value when the submission is made through the `Form` component regardless of whether JavaScript is enabled or disabled.

barbalex:
Do I understand this correctly: `action.formData` is ONLY updated when the form is submitted?

So when using `action.submit` there is no way to get the current data of the form? Unless I build a parralel state universe to manage form state as is normal in react?

PR #4424, opened 2023-06-06
feat: image performance dev tools

manucorporat:
Screenshot 2023-06-06 at 18 37 56


packages/qwik/src/optimizer/src/plugins/image-size-runtime.html
```diff
@@ -0,0 +1,299 @@
...
+                url.searchParams.set('height', info.height);
+
+                await fetch(url, {
+                  method: 'POST',
+                });
+              };
+            }
+          } else if (tooBig) {
+            overlay.header = 'Perf: properly size image';
+            overlay.message = `The image is too big, <a href="https://developer.chrome.com/en/docs/lighthouse/performance/uses-responsive-images/" target="_blank" rel="noopener noreferrer">hurting performance</a>, it should be resized to the size it\'s displayed at. The image dimmensions are ${info.width} x ${info.height} but it\'s displayed at ${rect.width}x${rect.height}.</p>`;
```
mrkoreye, line 236:
```suggestion
            overlay.message = `The image is too big, <a href="https://developer.chrome.com/en/docs/lighthouse/performance/uses-responsive-images/" target="_blank" rel="noopener noreferrer">hurting performance</a>, it should be resized to the size it\'s displayed at. The image dimensions are ${info.width} x ${info.height} but it\'s displayed at ${rect.width}x${rect.height}.</p>`;
```

PR #4429, opened 2023-06-07
docs: qwiksand-box

manucorporat:
https://twitter.com/erikras/status/1666361545629024256

Screenshot 2023-06-07 at 14 44 40

### DIFF
--- a/packages/docs/src/components/header/header.tsx
+++ b/packages/docs/src/components/header/header.tsx
@@ -101,7 +101,7 @@ export const Header = component$(() => {
                 href="/examples/introduction/hello-world/"
                 class={{ active: pathname.startsWith('/examples') }}
               >
-                <span aria-aria-label='Qwik playground'>Qwik sandbox</span>
+                <span aria-aria-label='Qwik playground'>Qwik-sandbox</span>
               </a>
             </li>
             <li>
--- a/packages/docs/src/components/header/header.tsx
+++ b/packages/docs/src/components/header/header.tsx
@@ -101,7 +101,7 @@ export const Header = component$(() => {
                 href="/examples/introduction/hello-world/"
                 class={{ active: pathname.startsWith('/examples') }}
               >
-                <span>Playground</span>
+                <span aria-aria-label='Qwik playground'>Qwiksand-box</span>
               </a>
             </li>
             <li>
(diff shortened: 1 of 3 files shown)


erikras (approved):
👏 

giphy-25

packages/docs/src/components/header/header.tsx
```diff
@@ -101,7 +101,7 @@ export const Header = component$(() => {
                 href="/examples/introduction/hello-world/"
                 class={{ active: pathname.startsWith('/examples') }}
               >
-                <span>Playground</span>
+                <span aria-aria-label='Qwik playground'>Qwik-sandbox</span>
```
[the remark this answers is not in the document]
bolonio: You don't really need the `aria-label` attribute, and definitely not in a `<span>` tag. If you ever need to add it, please add it to the interactive element. I've kept the suggestion from @erikras 

```suggestion
                <span>Qwik Sandbox</span>
```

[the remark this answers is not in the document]
erikras: Sorry. Was a copy-paste mistake.

packages/docs/src/components/header/header.tsx
```diff
@@ -101,7 +101,9 @@ export const Header = component$(() => {
                 href="/examples/introduction/hello-world/"
                 class={{ active: pathname.startsWith('/examples') }}
               >
-                <span>Playground</span>
+                <span aria-aria-label="Qwik playground" class="qwiksand">
```
bolonio, line 104:
```suggestion
                <span class="qwiksand">
```

packages/docs/src/components/header/header.tsx
```diff
@@ -101,7 +101,9 @@ export const Header = component$(() => {
                 href="/examples/introduction/hello-world/"
                 class={{ active: pathname.startsWith('/examples') }}
```
bolonio, line 102:
```suggestion
                class={{ active: pathname.startsWith('/examples') }}
                aria-label="Qwik playground" 
```

bolonio (approved):

packages/docs/src/components/header/header.tsx
bolonio: I would stick with @erikras suggestion and remove directly the dash from the title, but it you still want to keep, please add the `aria-label` to the interactive element

manucorporat:
Thank you very much Adrian! i think we might not need any label as i named it "qwik sandbox" as before it was "qwik-sand box" which was confusing.

learnt new things today :)

PR #4433, opened 2023-06-07
feat: client info API

### LINKED ISSUE #2648: Qwik City feature: To be able to get the IP address of the initial client's request[✨]

Be good to be able to add logic to the app that is based on the IP address of the incoming request.

Be able to query the request's IP address (the originating client's IP address, using Headers from proxys, etc...)

N/A

### Additional context

svelte kit: https://github.com/sveltejs/kit/pull/4289

### LINKED ISSUE #4396: [✨] Support remote IP address in request object for loader and server$ functions

Currently, it is not obvious nor documented how to access the remote IP address of the requestor in a vendor agnostic way (e.g. via `remoteIp`, etc. This should be made easier, and documented.

Make the remote IP more easily accessible within the event object.

This only works on the dev server and will not work in preview as middleware is not supported.

```ts
// vite.config.ts

// Pass the remote IP through via a header (NOTE: should be able to mod the request object)
export const remoteIp = {
  configureServer(server) {
    server.middlewares.use((req, res, next) => {
      const remoteIp = req.headers['x-forwarded-for'] || req.connection.remoteAddress;
      req.headers['remoteIp'] = remoteIp;
      next();
    });
  },
};

// Add to plugins
plugins: [
  remoteIp,
  qwikCity({trailingSlash: false}),
  qwikVite(),
  tsconfigPaths(),
]
```

Usage
`event.request.headers.get('remoteIp')`

### Additional context

_No response_

manucorporat:
fixes #4396
fixes #2648


packages/qwik-city/middleware/node/http.ts
```diff
@@ -100,6 +102,13 @@ export async function fromNodeHttp(
         },
       });
     },
+    getClientInfo: () => {
+      return getClientInfo
+        ? getClientInfo(req)
+        : {
+            ip: req.socket.remoteAddress,
```
SoerenSilkjaer, line 109: Should this also account for x-forwarded-for?

`ip: req.headers['x-forwarded-for'] || req.socket.remoteAddress`

manucorporat: No, it would be a security vulnerability, our google cloud run starter will be configured to use `x-forwarded-for`
Notice that doing that unconditionally, even when the node server is not under a reverse proxy would give the user the availity to hide their IP.

SoerenSilkjaer: Yes you are right, that would require a configuration the developer can set, when they run the node server behind a proxy. Only then should it use the X-Forwarded-For header.

PR #4434, opened 2023-06-07
fix: 'Numberish' type used for width/height didn't allow %

KenAKAFrosty:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

The `Numberish` type which was used for heights and widths didn't allow for any % values like "100%"

# Use cases and why
The following should support the percentage value passed
image

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/jsx/types/jsx-generated.ts
+++ b/packages/qwik/src/core/render/jsx/types/jsx-generated.ts
@@ -2,7 +2,7 @@ import type { Signal } from '../../../state/signal';
 import type { DOMAttributes, ClassList } from './jsx-qwik-attributes';
 interface HTMLWebViewElement extends HTMLElement {}
 export type Booleanish = boolean | `${boolean}`;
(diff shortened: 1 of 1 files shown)


manucorporat:
i dont think `%` is allowed:

Use both [width](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#width) and [height](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#height) to set the intrinsic size of the image, allowing it to take up space before it loads, to mitigate content layout shifts.


width
The intrinsic width of the image in pixels. Must be an integer without a unit.

height
The intrinsic height of the image, in pixels. Must be an integer without a unit.

ulic75:
Ken's example is an iframe on MDN it says this:

> height
> The height of the frame in CSS pixels. Default is 150.

It doesn't explicitly say "without a unit", but I think it's implied here as well.

manucorporat:
Maybe we can revert to string | number to anything other than img

PR #4440, opened 2023-06-08
docs(adding content security policy to the advanced topics)

the-zimmermann:
docs advanced » content security policy

I added information about CSP and how to integrate it with qwik. Added code examples and routing structure for middleware (plugins)

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Adding information how to add content-security-policy to a qwik application using middleware (plugin)

# Use cases and why

Every large website with security in mind will need to add CSP.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/advanced/content-security-policy/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/advanced/content-security-policy/index.mdx
@@ -1,0 +1,103 @@
+---
+title: Content Security Policy
+contributors:
+  - tzdesign
+---
+
+# Content Security Policy
+
+## What is CSP?
+
+[Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) is an added layer of security that helps to detect and mitigate certain types of attacks, including Cross-Site Scripting (XSS) and data injection attacks. These attacks are used for everything from data theft, to site defacement, to malware distribution.
+
+## Steps to integrate
+
+### Add a route plugin
+
+To add the middleware plugin to your application, simply add a file named ```plugin@your-plugin-name.ts``` to the routes folder. 
+This file will be loaded with every request, allowing you to add headers, modify the response, and more.
+
+```bash  {3} /plugin@csp\.ts/ title="Add the plugin"
+src/
+└── routes/
+    ├── plugin@csp.ts         # The plugin which runs on every request (route middleware)
+    ├── contact/
+    │   └── index.mdx         # https://example.com/contact
+    ├── about/
+    │   └── index.md          # https://example.com/about
+    ├── index.mdx             # https://example.com/
+    │
+    └── layout.tsx            # This layout is used for all pages
+```
+
+```typescript title="src/routes/plugin@csp.ts"
+import type { RequestHandler } from "@builder.io/qwik-city";
+
+export const onRequest: RequestHandler = event => {
+  const nonce = Date.now().toString(36); // Your custom nonce logic here
+  event.sharedMap.set("@nonce", nonce);
+  const csp = [
+    `default-src 'self' 'unsafe-inline'`,
+    `font-src 'self'`,
+    `img-src 'self' 'unsafe-inline' data:`,
+    `script-src 'strict-dynamic' 'unsafe-inline' 'nonce-${nonce}'`,
+    `style-src 'self' 'unsafe-inline'`,
+    `frame-src 'self' 'nonce-${nonce}'`,
+    `object-src 'none'`,
+    `base-uri 'self'`,
+  ];
+
+  event.headers.set("Content-Security-Policy", csp.join("; "));
+};
+```
+
+### Add it to the service worker as well
+
+```tsx {12,22} /nonce/ title="src/root.ts"
+import { component$, useServerData } from "@builder.io/qwik";
+import {
+  QwikCityProvider,
+  RouterOutlet,
+  ServiceWorkerRegister,
+} from "@builder.io/qwik-city";
+import { RouterHead } from "./components/router-head/router-head";
+
+import "./global.css";
+
+export default component$(() => {
+  const nonce = useServerData<string | undefined>("nonce");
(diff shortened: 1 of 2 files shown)


tzdesign:
If you like to add me as contributor, please use tzdesign. During the pr I used the wrong account 🤦‍♂️😅

tzdesign:
@jordanw66 this is it, thanks!

I updated the PR. The https: part is definitely good, but will fail for everyone in preview.
However, your reasoning is totally understandable and we should definitely show good and complete examples in the docs.

It will be an advanced topic anyway.

packages/docs/src/routes/docs/(qwikcity)/advanced/content-security-policy/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+    `script-src 'unsafe-inline' https: 'nonce-${nonce}' 'strict-dynamic'`,
+    `style-src 'self' 'unsafe-inline'`,
+    `frame-src 'self' 'nonce-${nonce}'`,
+    `object-src 'none'`,
+    `base-uri 'self'`,
+  ];
+
+  event.headers.set("Content-Security-Policy", csp.join("; "));
+};
+```
```
tzdesign, line 52: @jordanw66  what do you think about this?
```suggestion
#### Example

> Please note that in dev mode the Vite scripts have no nonce and will report. For this reason, the example will not add csp in dev mode.
> These are very good defaults with backwards compatibility, but please adapt them to your needs.
> As this is an advanced topic, you should take a closer look at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) or [web.dev](https://web.dev/csp/) to get a better understanding of CSP.

```typescript title="src/routes/plugin@csp.ts"
import type { RequestHandler } from "@builder.io/qwik-city";
import { isDev } from "@builder.io/qwik/build";

export const onRequest: RequestHandler = event => {
  if (isDev) return; // Will not return CSP headers in dev mode
  const nonce = Date.now().toString(36); // Your custom nonce logic here
  event.sharedMap.set("@nonce", nonce);
  const csp = [
    `default-src 'self' 'unsafe-inline'`,
    `font-src 'self'`,
    `img-src 'self' 'unsafe-inline' data:`,
    `script-src script-src 'self' 'unsafe-inline' https: 'nonce-${nonce}' 'strict-dynamic'`,
    `style-src 'self' 'unsafe-inline'`,
    `frame-src 'self' 'nonce-${nonce}'`,
    `object-src 'none'`,
    `base-uri 'self'`,
  ];

  event.headers.set("Content-Security-Policy", csp.join("; "));
};
```
```

tzdesign:
@jordanw66 I'm new to vite and have no clue about the connection of vite and qwik.

It looks like the vite-server.ts adds plain scripts and maybe @manucorporat can answer the question of adding nonce to all the vite-scripts.

https://github.com/builderIO/qwik/blob/36391fd064cd9d9ced3ef82a2be7e360993c59d9/packages/qwik/src/optimizer/src/plugins/vite-server.ts#L312

Let me know what you thing about my suggestion.

tzdesign:
> @tzdesign Looks like script-src is accidentally repeated on the line but otherwise it looks perfect to me, great job!

@jordanw66 Copy/Paste is the source of all evil.

I also added nonce to RouterOutlet, I'm not sure if this is the right way to do it either, but since it's a component now, it should work. See: https://github.com/BuilderIO/qwik/pull/4440/commits/2be5a9781dba6e7c9dcfd9ae71696422b77e66e5

tzdesign:
@zanettin First I was sure this PR is just docs, how should I rename this one here or should I split it in two PRs?
@manucorporat maybe you have time to take a look at the RouterOutlet to see if this is the way to go?

zanettin:
imo we could keep it as it is within a single PR but i leave the decission to Manu 👍

tzdesign:
@Scott-MacD This is just a sample policy. You must implement the policy yourself. 

The nonce in the server data has a specific fixed name that is already necessary to follow because of all the script tags qwik creates. I was wondering why we have to add it to the service worker, it makes it much easier to forget.

I don't mind either way, my pr was mostly for having docs about CSP so new people don't have to go to issues to find out about adding it to every request.

There is also a feature request to add it to styles, but that's not trivial. Even though most docs say it's not really necessary, I still want it in qwik.

tzdesign:
Hi there,

thanks @DustinJSilk for fixing the popstate part.

What do you guys think about the docs? Is this a good first suggestion?
It's a complicated topic, so @jordanw66 and I agreed to add a disclaimer and links for more information.

Let me know if I need to do anything on my end.

tzdesign:
@jordanw66 this sounds awesome. I will also add you to the contributions section of the docs.

I would change the whole block to this if you agree (the advice part sounds a bit like we are parenting on the devs 😅).
I also moved the dev mode info to the bottom to prioritize the info.
> This template provides very permissive backward-compatible defaults. 
> It is highly recommended that you customize it to better suit your specific use case.
> As this is an advanced topic, you should take a closer look at [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) or [web.dev](https://web.dev/csp/) to get a better understanding of CSP.
> Please note that in dev mode the Vite scripts have no nonce and will report. For this reason, the example will not add csp in dev mode.

tzdesign:
@manucorporat who is reviewing docs? @jordanw66 I can't edit pr meta data, so it's up to the owners.

shairez:
Great job @tzdesign and @jordanw66 ! 
And thanks for the feedback @Scott-MacD 

The docs look great
One thing I'm confused about is this - https://github.com/BuilderIO/qwik/commit/2be5a9781dba6e7c9dcfd9ae71696422b77e66e5

is it related to a different PR?

tzdesign:
@shairez This is already merged into main, so I merged main and removed it from the changes.

@DustinJSilk took it out in a separate pr, which Manu already merged.

shairez:
Thanks @tzdesign !
Merging!
Great job

Issue #4408, opened 2023-06-05
[🐞] Incorrect behavior when passing an array as the first parameter to the useOnDocument hook


utherpally:
### Which component is affected?

Qwik Runtime

### Describe the bug

When using the `useOnDocument` hook and providing an array of strings as the first parameter, the expected behavior is to add event listeners for each event in the array to the entire document. However, the current implementation does not handle this scenario correctly and fails to attach the event listeners.

This work
```ts
useOnDocument("mousemove", handle);
useOnDocument("touchmove", handle);
```

This not work
```ts
useOnDocument(["mousemove", "touchmove"], handle);
```

### Reproduction

[...]

PR #4453, opened 2023-06-10
fix(use-on.ts): fixed useOn methods to pass correct eventName to _useOn

OrenSayag:
#4408

# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The issue was: when passing an array of events to the `useOnDocument` method, the events didn't work as expected in the browser.

### The useOnDocument method passes invalid param to \_useOn

```ts
export const useOnDocument = (
	event: PascalCaseEventLiteralType | PascalCaseEventLiteralType[],
	eventQrl: QRL<(ev: Event) => void> | undefined
) => {
	_useOn(`document:on-${event}`, eventQrl);
};
```

The problem is that the `${event}` is translated into `e1,e2` when the event is `[e1,e2]`.

Replication in js:

```js
const event = ['e1', 'e2'];
const res = `document:on-${event}`;
console.log(res); // document:on-e1,e2
```

The solution is to change the `event` parameter to pass an array of event names, instead of the malformed event string.

This is achieved by changing the creation of the eventName parameter that is passed to the method `useOn`:

```ts
export const useOn = (
	event: PascalCaseEventLiteralType | PascalCaseEventLiteralType[],
	eventQrl: QRL<(ev: Event) => void> | undefined
) => {
	const eventName = Array.isArray(event)
		? event.map((e) => `document:on-${e}`)
		: `document:on-${event}`;
	_useOn(eventName, eventQrl);
};
```

I added a method to the module which is reusable within the `useOn` methods in the module, and it solves the bug.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
+++ b/packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
@@ -1062,6 +1062,28 @@ renderSSRSuite('component useOn()', async () => {
   );
 });
 
+renderSSRSuite('component useOn([array])', async () => {
+  await testSSR(
+    <body>
+      <UseOnMultiple />
+    </body>,
+    `
+    <html q:container="paused" q:version="dev" q:render="ssr-dev">
+      <body>
+        <!--qv q:id=0 q:key=sX:-->
+        <div on:click="/runtimeQRL#_\n/runtimeQRL#_"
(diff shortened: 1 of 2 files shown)


manucorporat:
I will add an e2e test and i think we can merge!! awesome fix!

PR #4456, opened 2023-06-11
Reorganize .gitignore

szepeviktor:
# Overview

Organize git ignores in to categories.

[_source_](https://github.com/szepeviktor/debian-server-tools/blob/9abca88bda2b3805a7be0c9a535966d775da8d5f/.gitignore#L6-L12)

I'm highly unsure whether anchoring is correct and I've put ignores in the correct category.
Mainly in case of `/target/` and `**/server/`
Please help me.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

We could provide example `~/.gitignore` contents in CONTRIBUTING.md

```gitignore
# OS
.DS_Store
Thumbs.db

# IDE and local environment
/.history/
/.idea/
*. or .* ?
```

This is because one project cannot cover every OS, every IDE, every local env.
Please see https://git-scm.com/docs/gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,13 +5,13 @@
 
 # Dependencies
 *.node
+*.log
 /.pnpm-store/
 /.yarn/
 !/.yarn/releases
 /node_modules/
 
 # Application
-*.log
 /dist-dev/
 /etc/
 /lib/
--- a/.gitignore
+++ b/.gitignore
@@ -1,28 +1,29 @@
-.history
-.vscode/settings.json
-dist
-lib
-dist-dev
-node_modules
-tsc-out
-external
-*.
-**/*.log
-etc
-temp
-tsdoc-metadata.json
-**/.DS_Store
-src/napi/package-*
-target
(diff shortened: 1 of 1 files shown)


manucorporat:
I like the change, started to become a bit messy

szepeviktor:
You're welcome.

It made the write a mini howto.

zanettin:
looks like nested `/node_modules` and `dist` dirs are no longer excluded 😅  guess we should adjust that asap 😄 thanks for the cleanup @szepeviktor 🙏

szepeviktor:
> nested `/node_modules` and `dist` dirs

Actually I've felt a contradiction.
This gitignore file is for Qwik development and qwik-app development too. Am I right?

szepeviktor:
```gitignore
node_modules/
# and
dist/
```
No anchors.

zanettin:
the issue i just faced after pulling the changes in was, that all pnpm workspaces (`packages/xxx`) `node_modules` and `dist` folders where tracked 😅

Issue #4041, opened 2023-05-03
[🐞] aria-current not updating


GrandSchtroumpf:
### Which component is affected?

Qwik Runtime

### Describe the bug

I'm building a `PageLink` component that set `aria-current` to page when the current page is the one described by the href.
It works well except that the first anchor is not updated on page change. 
I suppose it has to do with the `useLocation`, but I'm not sure.

```jsx
import { component$, Slot, useComputed$ } from '@builder.io/qwik';
import { useLocation, Link } from '@builder.io/qwik-city';
import type { LinkProps } from '@builder.io/qwik-city';

const normalize = (pathname: string) => {
  return pathname[pathname.length - 1] === '/'
    ? pathname.slice(0, -1)
    : pathname;
};

export const PageLink = component$((props: LinkProps) => {
  const { url } = useLocation();
  if (!props.href) throw new Error('PageLink should have a href');
  const href = props.href;
  const current = useComputed$(() =>
    normalize(href) === normalize(url.pathname) ? 'page' : undefined
  );
  return (
    <Link {...props} aria-current={current.value}>
      <Slot />
    </Link>
  );
});
```

### Reproduction

https://stackblitz.com/edit/qwik-starter-frgqvy?file=src/components/page-link.tsx

### System Info

```shell
System:
    OS: Windows 10 10.0.22623
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 16.69 GB / 31.71 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 3.3.1 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1095.0), Chromium (112.0.1722.68)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @builder.io/qwik: 1.0.0 => 1.0.0
    @builder.io/qwik-city: 1.0.0 => 1.0.0
    undici: 5.22.0 => 5.22.0
    vite: 4.3.3 => 4.3.3
```

manucorporat:
Gonna look into this issue, but also thinking `<Link>` could do this built-in!

GrandSchtroumpf:
[...]

PR #4469, opened 2023-06-12
feat: aria-current on Link component

mtt-artis:
# Overview

fixes: #4041

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This pull request introduces a new feature that adds support for aria-current on the Link component. The aria-current attribute is a helpful accessibility feature that allows screen readers and other assistive technologies to indicate the current active link to users.

this is inspired by the `NavLink` component of react-router v6
https://github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L644

# Use cases and why

With this enhancement, when the Link component is rendered as the active link, the aria-current="page" attribute will be automatically applied, providing better accessibility and usability for users navigating through the application.

By incorporating aria-current support into the Link component, developers using Tailwind CSS or Unocss can easily style the active link state with `class="aria-[current=page]:bg-red"`

# Checklist:

### DIFF
--- a/packages/docs/src/routes/api/qwik-city/api.json
+++ b/packages/docs/src/routes/api/qwik-city/api.json
@@ -320,7 +320,7 @@
         }
       ],
       "kind": "Interface",
-      "content": "```typescript\nexport interface LinkProps extends AnchorAttributes \n```\n**Extends:** AnchorAttributes\n\n\n|  Property | Modifiers | Type | Description |\n|  --- | --- | --- | --- |\n|  [end?](#) |  | boolean | _(Optional)_ |\n|  [prefetch?](#) |  | boolean | _(Optional)_ |\n|  [reload?](#) |  | boolean | _(Optional)_ |",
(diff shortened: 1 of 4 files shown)


zanettin:
thank u very much @mtt-artis 🙏  just updated the api docs (using `pnpm api.update`) and formatted code (using `pnpm fmt`) 👍 so that the CI is happy as well 😄 

one question regarding the new `end` property on the link. shouldn't we name it a bit more specific? to give the devs a glue what it is used for?

manucorporat:
Right, i have to clue what `end` means in this context

mtt-artis:
Hello. 
Sorry for the delay.
I basically copied the end props from  https://reactrouter.com/en/main/components/nav-link#end. How would you like to rename it ?

zanettin:
Hi @mtt-artis 👋  thx for the clarification and the link. i've just updated the API docs and imo this is ready to be merged (cc: @manucorporat )

manucorporat:
I will merge after 1.2.0 worries me a bit the amount of new subscriptions it could create, wanna test it further

mtt-artis:
I liked the idea that Link was connected to the router. so it seemed fine to me to provide this service by default.
mabe checked if there is already an `aria-current` before create the compute ?

at the beginning I had the same problem as here: https://github.com/BuilderIO/qwik/issues/4041
and the problem does not seem to be solved

zanettin:
thanks again for this PR and the great discussion 🙏 also thx to @jordanw66 for the inputs and research 🎉 
since u already agreed to sunset this approach, i'll close this PR for now. feel free to reopen it if needed 👍

PR #4478, opened 2023-06-13
refactor: containerState for appendHeadStyle

manucorporat:


tzdesign:
Perfect @manucorporat!

It's basically the same as the script tag. 

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

The only thing I don't know is what happens with imports, but I would say they are part of the style.

I am still on vacation, but I can have a look at it next week.

PR #4479, opened 2023-06-13
feat: image vite transform

manucorporat:
- Zero runtime, zero JS
- Simplest API surface possible (zero props by default)
- Strongly typed (You cant referenced 404 images)
- Autocompletion supported by typescript
- Immutable cached images (hashed files)
- Optimized to webp
- Automatic width/height (prevent layout reflows)
- Automated srcSet generation (ideal image size)
- Extendable (add any `` attribute)
- Loading lazy and async decoding
- Lightweight, a single `` node in the HTML

https://www.loom.com/share/23c21ef8d4be464dab59eae6064c918a

```tsx
import Image from '~/media/image2.jpg?jsx';

  return (
    <div>
      <Image alt='Unsplash image' />
    </div>
  );
});
```

Output:
Screenshot 2023-06-13 at 20 54 00

### DIFF
--- a/packages/qwik-city/buildtime/vite/api.md
+++ b/packages/qwik-city/buildtime/vite/api.md
@@ -6,6 +6,7 @@
 
 import { CompileOptions } from '@mdx-js/mdx/lib/compile';
 import { ConfigEnv } from 'vite';
+import type { PluginOption } from 'vite';
 import { UserConfigExport } from 'vite';
 
 // @public (undocumented)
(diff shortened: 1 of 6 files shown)


manucorporat:
It could work, but CDN images should be able to deliver the srcset automatically, this is build-time...
I think something like unpic-qwik should be a better fit for external images

manucorporat:
https://github.com/ascorbic/unpic-img

gioboa:
https://github.com/qwikifiers/qwik-image

PR #4509, opened 2023-06-16
feat: frame-perfect and state-backed durable SPA scroll restoration

jordanw66:
# Overview

_**Update:** Included the durable scroll history changes because they required changes to scroll restoration API. The next PR to address SPA context recovery should be a transparent upgrade (no API changes) that can be merged post-release if it's not done/agreed upon before next week._

As mentioned here (https://github.com/BuilderIO/qwik/pull/3244#issuecomment-1551719502), this PR correctly implements the following:

- Properly disables the built-in browser scroll restoration when SPA is enabled.
    - It needs to be disabled during SPA, otherwise you will get pre-scrolling.
- Makes the manual scroll restoration synchronous with render.
  - Async has problems with potential for draws between render and when the event loop gets around to the scroll restore.
  - Simple, implemented in the same way alongside the View Transition.
- Makes the scroll history durable by storing it directly in history state instead of session storage.
  - This is a must to eventually recover scroll states on resuming sessions (desktop, mobile). Browsers natively recover MPA scroll sessions by storing scroll positions in the tab history. With session storage, it becomes impossible to replicate MPA.
  - Mobile browsers specifically can lose session storage very easily resulting in the loss of all scroll history.
  - Because state cannot be updated after a pop, it requires a scroll handler to be added when upgrading to SPA, but we are very careful to debounce reasonably here. Performance impact should be negligible.

The navigation API shipped live in Chromium 102+, in a future PR I can implement this to simplify SPA navigation/scroll in browsers that support it and leave the state-backed SPA as a fallback.

_**Note:** This PR does **not** address these issues:_

- Complete loss of SPA ctx on refresh and bfcache clear.
- Complete loss of scroll restore when losing SPA as above.

These will be addressed in a follow-up PR to make SPA navigation recoverable and function nearly identically to MPA in all circumstances, you should not be able to tell the difference.

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/api/qwik-city/api.json
+++ b/packages/docs/src/routes/api/qwik-city/api.json
@@ -460,7 +460,7 @@
         }
       ],
       "kind": "Interface",
-      "content": "```typescript\nexport interface QwikCityProps \n```\n\n\n|  Property | Modifiers | Type | Description |\n|  --- | --- | --- | --- |\n|  [restoreScroll$?](#) |  | PropFunction&lt;RestoreScroll&gt; | _(Optional)_ |\n|  [viewTransition?](#) |  | boolean | <p>_(Optional)_ Enable the ViewTransition API</p><p>Default: <code>true</code></p> |",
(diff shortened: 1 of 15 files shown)


manucorporat:
Impressed with the PR! good job, i will love to merge this before the next big release (next week!)

manucorporat:
This is amazing, planning to do some extra manual testing before release. I am not sure i am a big fan of the restoreScroll logic in core (refering to the logic added around view transitions), is it unavoidable?

PR #4550, opened 2023-06-21
fix: perfect hash scroll

jordanw66:
# Overview

This is my last PR re. the actual scroll experience, I think it now provides a near flawless UX. This should be a good baseline for this release. Next PR will be restoring SPA/scroll after refresh/reloading/session restore, I think you can add that in a point release after, should be a transparent upgrade. cc @manucorporat 

This PR fixes some big issues around hash scrolling (and others) that were present for a long time:
- Browsers natively re-scroll to hashes always when clicked. Now they do here.
- Anchors (not Links) with same-page hashes have numerous issues that differ by browser when SPA is enabled, we introduce a new default when upgrading to SPA by patching these events to redirect these all into the Link pipeline.
- Links/anchors with empty hashes (and same-page Link) should always scroll to top. Now they do here.
- Empty hashstrings create many edgecase problems for SPA and differ by browser, Link already removes these from its href. We patch these out for all same-page hash anchors, and also useNavigate to prevent devs from breaking SPA.
- Empty hashes on anchors will not navigate to /# (instead to /), but they do scroll, so they are functionally the same.
- etc.

Prior to this PR the use of `<a href="#...">` anywhere in SPA would effectively break SPA state/UX.

This was quite the rabbit hole to fix. I don't have access to a Mac but I have tested these changes extensively in Chromium, Firefox, and WebKitGTK, I am confident that this covers most or all of the edgecases.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/scroll-restoration.ts
+++ b/packages/qwik-city/runtime/src/scroll-restoration.ts
@@ -55,7 +55,10 @@ const scrollForHashChange = (fromUrl: URL, toUrl: URL): boolean => {
   return true;
 };
 
-const scrollToHashId = (hash: string) => {
+/**
+ * @alpha
+ */
+export const scrollToHashId = (hash: string) => {
   const elmId = hash.slice(1);
   const elm = document.getElementById(elmId);
   if (elm) {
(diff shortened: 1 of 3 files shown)


packages/qwik-city/runtime/src/scroll-restoration.ts
```diff
@@ -55,7 +56,10 @@ const scrollForHashChange = (fromUrl: URL, toUrl: URL): boolean => {
   return true;
 };
 
-const scrollToHashId = (hash: string) => {
+/**
+ * @alpha
```
manucorporat, line 60: no need for this comment, since it's not a public api

manucorporat:
At this point, i am trusting u here! gonna merge and test myself

KenParsons:
This is awsesome!

PR #4558, opened 2023-06-22
feat: bulletproof SPA recovery

jordanw66:
# Overview

This PR finally delivers the holy trinity of Qwik SPA, specifically:
- Every page arrived at through SPA can be determined by history alone.
- (Nearly) every page can recover the SPA context on popstate after a refresh, navigation away/back, browser session restore, etc. (relies on any Link if present, this can be improved, falls back to reload if not)
- Scroll is saved and restored for every single SPA page, even if the SPA context is lost and not recovered.
- Scroll history is saved directly in history state, impossible to lose, not subject to sessions or storage being lost. Seriously, when the nukes come for us all the last things standing will be twinkie factories, and your scroll states.
- Scroll restore and saving almost perfectly mimics browser native scroll restore. You literally cannot tell the difference, and if you find some small edgecase, it should easily be fixable under this setup.
- This level of SPA recovery doesn't come free, but we "keep it Qwik" by inserting a small shim to replace the old popstate handler. This shim checks if SPA needs to be rebooted, if so it scrolls then manually imports a standalone SPA init QRL to handle the rest. There is essentially 0 impact on MPA pages. The size is comparable to the old handler.
- This standalone SPA init QRL hands off to the useNavigate context once it's restored. It is never sent to users on MPA, instead we pre-cache when the user first upgrades to SPA, this way it's cached and ready for SPA reboots.
- @billykwok's custom `restoreScroll$` `@alpha` api has been removed. It was incompatible with these changes, but I do believe the code in this PR can be extended to provide more customization than any other framework. However, it will take some thought to implement and nail down the correct API. For now, we choose to deliver best-in-class MPA-like scroll experience by default.
- Browsers will natively remember and scroll on every single state. For further resiliency, we patch the `history` object during an SPA context so that devs who use `pushState` and `replaceState` will always maintain scroll history.
- Rewrote and expanded docs on `SPA Navigation` to be more informative and also added `Scroll Restoration` docs.
- Fixed a longtime issue where `isNavigating` should've been set to false _after_ render, instead it was being set before.

With this PR, the SPA experience in Qwik becomes indistinguishable from MPA.

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


DustinJSilk:
amazing work!

manucorporat:
At this point the implementation is way over my head, you know more than anybody how this works, but the work seems good and i like the reduced API surface, gonna merge before 1.2 release!

Can u fix the merge conflict?

shairez:
Wow this is some amazing job @jordanw66 ! 👏👏

(btw, if you are on discord / twitter / myspace, please DM me :)

manucorporat:
Improving e2e is a must a this point indeed, like @shairez said, if you are discord we can chat about a plan!
Once we have a good structure in place is easy to add new regression tests as issues are reported!

manucorporat:
Gonna merge!

PR #4564, opened 2023-06-22
fix(qwik-city buildtime): make generated file ids unique

### LINKED ISSUE #4504: [🐞] using menu.md and having a route named `menu` causes an error

Qwik Runtime

### Describe the bug

If I'm using a menu.md file for creating a menu using useContext and I also have a route named `menu` then I receive the following error when trying to run with `preview` and during a build.

```
Identifier 'Menu' has already been declared (Note that you need plugins to import files that are not JavaScript)
```

### Reproduction

https://stackblitz.com/edit/qwik-starter-kd1pse

### Steps to reproduce

npm run preview

### System Info

```shell
Stackblitz
```

### Additional Information

_No response_

hbendev:
By appending type to the id generated from the path - except for menus and layouts, because they already contain their type based on the file names.

fix #4504

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Generated module identifiers were not unique. If you defined a `/menu` route and a `menu.md` for the root, the build failed. The same could occur with layouts. If you had my `/my/layout` route and a layout defined for `/my`, the build would fail.

# Use cases and why

While this is an edge case, and there is a workaround described in the ticket, we can save some other devs from headaches.

# Checklist:

### DIFF
--- a/packages/qwik-city/buildtime/build-endpoints.unit.ts
+++ b/packages/qwik-city/buildtime/build-endpoints.unit.ts
@@ -5,14 +5,14 @@ const test = testAppSuite('Build Endpoints');
 
 test('endpoint', ({ assertRoute }) => {
   const r = assertRoute('/api/data.json');
-  equal(r.id, 'CommonApiData');
+  equal(r.id, 'CommonApiDataRoute');
   equal(r.pattern, /^\/api\/data\.json\/?$/);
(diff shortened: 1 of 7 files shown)


manucorporat:
This is an amazing fix!! Thank you so much!

Issue #4059, opened 2023-05-04
[🐞] Support for typescript suggestions in style properties of tags. 


adiathasan:
### Which component is affected?

Qwik Runtime

### Describe the bug

This is not a bug. ⚔️ 

Support for typescript suggestions in style properties of tags. Not good at memorizing the props.

Currently `Record<string | ....>` isn't helping much

`(property) HTMLAttributes<HTMLDivElement>.style?: string | Record<string, string | number | undefined> | undefined`

### Steps to reproduce

Just `command+space` suggestions not showing!!!

PR #4577, opened 2023-06-23
fix(qwik-core): add types for the style attribute

hbendev:
# What is it?

- [x] ~~Feature~~ / enhancement
- [x] Bug

# Description

Adds csstype as a devdependency, and changes the TS types of the `style` attribute. This adds autocomplete options, while still allowing all previously supported syntaxes, like the hyphen-properties and specifying css variables in style objects.

Improvement asked in #4059

# Use cases and why

Other UI frameworks/libraries have this

# Checklist:

- [x] I have generated corresponding changes to the documentation

### DIFF
--- a/packages/qwik/src/core/render/jsx/types/jsx-generated.ts
+++ b/packages/qwik/src/core/render/jsx/types/jsx-generated.ts
@@ -9,8 +9,19 @@ export type Numberish = number | `${number}`;
 /**
  * @public
  */
-export type CSSProperties = CSS.Properties<string | number> &
-  Record<string, string | number | undefined>;
+export interface CSSProperties
+  extends CSS.Properties<string | number>,
+    CSS.PropertiesHyphen<string | number> {
+  /**
+   * The index signature was removed to enable closed typing for style
+   * using CSSType. You're able to use type assertion or module augmentation
+   * to add properties or an index signature of your own.
+   *
+   * For examples and more information, visit:
+   * https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors
+   */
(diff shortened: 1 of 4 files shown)


hbendev:
Actually, I just found [Vue's definition](https://github.com/vuejs/core/blob/124160e417015ed55afc2e7a9207429ff4eaebb4/packages/runtime-dom/src/jsx.ts#L31C15-L31C15), which is even nicer. I'll change it soon.

packages/qwik/src/core/render/jsx/types/jsx-generated.ts
```diff
@@ -1,10 +1,28 @@
...
 import type { DOMAttributes, ClassList } from './jsx-qwik-attributes';
 interface HTMLWebViewElement extends HTMLElement {}
 export type Booleanish = boolean | `${boolean}`;
 export type Size = number | string;
 export type Numberish = number | `${number}`;
 
+/**
+ * @public
+ */
+export interface CSSProperties
```
hbendev, line 12: I think it makes sense to export it so library authors and qwik users can reuse it.

manucorporat:
My only hesitance is adding a dependency to qwik itself... i really wish we didnt have to do that, does Vue depend on csstype?

hbendev:
Yes, and [@react/types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3420adf4ca802933d360aced242b4a44ebca507d/types/react/index.d.ts#L1616C4-L1625C6) too.

manucorporat:
The fact that vue and react use it... makes me think, gonna check solid!

manucorporat:
Let's do it! solid too!

PR #4601, opened 2023-06-26
chore: change useStore link

GustavoMelloGit:
This new link takes the user straight to useStore section

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/tutorial/hooks/use-signal/index.mdx
+++ b/packages/docs/src/routes/tutorial/hooks/use-signal/index.mdx
@@ -7,7 +7,7 @@ contributors:
   - GustavoMelloGit
 ---
 
-Use `useSignal()` to store any single value similar to `useStore()`. `useSignal` is heavily optimized when it comes to re-rendering components. It is able to skip re-renderings of parent components even when the signal itself is defined in the parent. `useSignal` works with all [primitive types](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) as well as with not nested / flat objects. If you need to store arrays or complex objects use [useStore](https://qwik.builder.io/docs/components/state/#usestore) instead.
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/tutorial/hooks/use-signal/index.mdx
```diff
@@ -4,9 +4,10 @@ contributors:
   - manucorporat
   - adamdbradley
   - neonpie
+  - GustavoMelloGit
 ---
 
-Use `useSignal()` to store any single value similar to `useStore()`. `useSignal` is heavily optimized when it comes to re-rendering components. It is able to skip re-renderings of parent components even when the signal itself is defined in the parent. `useSignal` works with all [primitive types](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) as well as with not nested / flat objects. If you need to store arrays or complex objects use [useStore](../../../docs/(qwik)/components/state/index.mdx) instead.
+Use `useSignal()` to store any single value similar to `useStore()`. `useSignal` is heavily optimized when it comes to re-rendering components. It is able to skip re-renderings of parent components even when the signal itself is defined in the parent. `useSignal` works with all [primitive types](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) as well as with not nested / flat objects. If you need to store arrays or complex objects use [useStore](https://qwik.builder.io/docs/components/state/#usestore) instead.
```
manucorporat, line 10: This would be even better! since it would be checked at compiler time that the link exists!
```suggestion
Use `useSignal()` to store any single value similar to `useStore()`. `useSignal` is heavily optimized when it comes to re-rendering components. It is able to skip re-renderings of parent components even when the signal itself is defined in the parent. `useSignal` works with all [primitive types](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) as well as with not nested / flat objects. If you need to store arrays or complex objects use [useStore](/docs/(qwik)/components/state/index.mdx#usestore) instead.
```

manucorporat (changes requested):

manucorporat (approved):

PR #4617, opened 2023-06-26
docs: dynamic ogImage feature url format fix and clean up

mrhoodz:
# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

Minor fix to the url structure of the edge function that generates the ogImage by replacing " " with "-"

from:
`https://opengraphqwik.vercel.app/api/og?subtitle=Project Structure&title=Qwik City&level=1`

to:
`https://opengraphqwik.vercel.app/api/og?subtitle=Project-Structure&title=Qwik-City&level=1`

# Enhancement notes

- the dynamic ogImage feature will now work reliably on twitter
- the change in the edge function will not affect the ogImage output

# Checklist:


manucorporat:
Feels like the root of the problem is that we are generating a URL using string concatenation, let's do the right thing:
like this
```tsx
const url = new URL('https://builder.io/api/v2/content')

url.searchParams.set('model', model)
url.searchParams.set('locale', locale)
url.searchParams.set('text', text)
  
const res = await fetch(url.toString())
```
More info: https://www.builder.io/blog/new-url

mrhoodz:
> Feels like the root of the problem is that we are generating a URL using string concatenation

You are right, thank you so much for the resource 👌.
I've committed the changes to match the pattern from the article.
 @manucorporat

PR #4622, opened 2023-06-27
feat: scroll opt-out on nav() and Link

jordanw66:
# Overview

This simple PR adds ability for devs to opt-out of scroll-to-top for specific `nav()` and `Link`. This might be useful for moving between pages with mostly the same but slightly different content while maintaining the same positions, or let devs hack together their own scroll behavior on navigates with a visible task or something.

Might be a little while before I get around to custom scroll behavior, this should suffice for now.

Others have a similar api as well, see: https://reactrouter.com/en/main/components/scroll-restoration#preventing-scroll-reset

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


manucorporat:
In general it's a negated APIs:

rename to `scroll: boolean`

manucorporat:
Funny enough, Next's link component also uses `scroll`:
https://nextjs.org/docs/pages/api-reference/components/link#scroll

Let's go with that

manucorporat:
Looks good to me!

manucorporat:
Can merge already?

manucorporat:
We had the same debate years ago at ionic, usually the benefit of saving {} does not pay off the cost of negations, in practice, things end up negated multiple times... 

You are rocking it!!

tzdesign:
@jordanw66 did someone changed something in the routing again?
We had implemented this on long detail-pages, but with the one of the last versions, it start scrolling again for ```<Link scroll={false}/>```.

Issue #4496, opened 2023-06-14
[✨] prevent default functions


gederer:
### Is your feature request related to a problem?

Currently, it is not possible to implement conditional `preventDefault` behavior in Qwik components. In other frameworks, we would just call `e.preventDefault()` within an event handler when needed. But, because of the way Qwik is implemented, this is not possible in Qwik.

This poses serious limitations when a component author wants to reuse most of the browser's event handling behavior for a particular event, but needs to override some of this behavior within an event handler.

### Describe the solution you'd like

The proposed solution is prevent default functions. This would allow the component author to pass a function into `preventdefault:*` that would return `true` if the default should be prevented, and `false`, otherwise.

Example:

```tsx
<div preventdefault:keydown={(event) => {return event.key === "Enter" ? true : false}}></div>
```

In the above example, the `keyup` default behavior would be prevented if the user pressed the Enter key, but not if the user pressed any other key.

### Describe alternatives you've considered

The component author can use the existing `prevent default:*` attributes to prevent all default behavior, then reimplement all of the desired default behavior (as well any custom behavior) in the event handler. This approach is, however, both cumbersome, and not performant.

### Additional context

This problem has come up a number of times in the qwik-ui project.

### Related PRs

#4625

shairez:
thanks @gederer !

The function would probably have to be wrapped in `$()` or something

FlatMapIO:
I forcibly called `// @ts-ignore event.preventDefault()`  in a code that requires conditional preventDefault. It works, but I'm not sure if there are any hidden risks

shairez:
pasting here as well because we want to keep this issue opened - 


A workaround is to use something like @mhevery suggested -

```
const anchorElement = useSignal<Element>();
useVisibleTask$(() => {
  const handler = (event: Event) => {
    event.preventDefault();
    // do something here
  };
  anchorElement.value!.addEventListener('click', handler);
  return () => anchorElement.value!.removeEventListener('click', handler);
});
  
 ```

which works well for now

GregOnNet:
Hey together,

in my view, there is also an issue with `useOnWindow` and `preventDefault`.

The example below demonstrates a simple version of a Dialog-Component using `HTMLDialgo`.
By default the HTMLDialog closes when <kbd>ESC</kbd> is pressed on the keyboard.
Dialog-Component listens to keyboard-Events and uses `preventDefault` to avoid closing the HTMLDialog.
[...]

Issue #3871, opened 2023-04-21
[🐞] SPA routing is preventing browser defaults


tzdesign:
### Which component is affected?

Qwik Runtime

### Describe the bug

When you press cmd + click on a link, qwik uses preventDefault. 

The default browser experience would open a new tab. If you add shift to the party, it'll even open it directly.
I know there are several other shortcuts to open in a new window or private window.

Qwik should not prevent that kind of functionality.

### Reproduction

https://stackblitz.com/edit/qwik-starter-qads2x?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

Run it on stackblitz.

## Click flower app with cmd 
You just get routed

## Click todo app with cmd
You get to a new tab

### System Info

```shell
See stackblitz
```

manucorporat:
is it a thing in other metaframeworks like nextjs, when using Link?

tzdesign:
@manucorporat no I think you are preventing default.
In react-router and next it's only preventing default if shiftKey, meta and ctrl are not true.

I checked it with our internal apps and my personal nextjs page.

We discussed this with @hamatoyogi and @shairez, and I think we agreed that qwik should avoid preventing browser APIs as much as possible.

You can call it a feature request, but I would say it should have been implemented in the same way already.

tzdesign:
@jordanw66 I agree, trying to fake the default behavior is not an option.

I think that qwik suppresses the default behavior here is a mistake. However, I don't know how important it is to everyone to go to the qwik-loader here. 

Do you think it's justified to pursue it here? I personally use meta + shift + click every single day.

@manucorporat what is your opinion about it?

tzdesign:
@jordanw66 perfect to pointing to it. Your suggestion is awesome. The question is, how deep do we want to go.
I might want to implement it for all developers and pass the allowmodifiers as a type like that:

```ts
export type AllowModifiers = SingleOrArray<
  keyof Pick<QwikMouseEvent, 'altKey' | 'ctrlKey' | 'metaKey' | 'shiftKey'>
>;
```

This way it's connected to the event and the developer can be specific.
For link we set linkProps to all of them.

I have some space tomorrow and can check it out, but maybe @manucorporat can tell us if this is ok. We also can catch up and talk about it, as Manu wanted to write e2e tests for navigation?

gioboa:
This is solved by [sync$()](https://qwik.builder.io/docs/cookbook/sync-events/#example-sync-with-state)

[...]

PR #4625, opened 2023-06-27
feat: allow overriding preventDefault for modifier keys

jordanw66:
# Overview

In response to this issue: https://github.com/BuilderIO/qwik/issues/3871

The only place to preventDefault or not is in the loader, unfortunately this makes it impossible to preventDefault on anchors while also letting users use browser-native modifier keys (ctrl, shift, alt) to open new tabs, etc on these same links.

This impl. adds an attribute at the same level as `preventdefault:*` that allows overriding it for modifier keys. The reason for this is to allow devs to still preventDefault even on these modifier keys.

Given the widespread use of `preventdefault` and the niche use of modifier keys, I think keeping this check as false by default and opt-in only when needed makes more sense.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/qwikloader.ts
+++ b/packages/qwik/src/qwikloader.ts
@@ -52,7 +52,17 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
   const dispatch = async (element: Element, onPrefix: string, ev: Event, eventName = ev.type) => {
     const attrName = 'on' + onPrefix + ':' + eventName;
     if (element.hasAttribute('preventdefault:' + eventName)) {
-      ev.preventDefault();
+      if (
+        !element.hasAttribute('allowmodifiers') ||
+        !(
+          (ev as MouseEvent).ctrlKey ||
+          (ev as MouseEvent).metaKey ||
+          (ev as MouseEvent).shiftKey ||
+          (ev as MouseEvent).altKey
+        )
+      ) {
+        ev.preventDefault();
+      }
     }
     const ctx = (element as any)['_qc_'] as QContext | undefined;
     const qrls = ctx?.li.filter((li) => li[0] === attrName);
--- a/packages/qwik-city/runtime/src/link-component.tsx
+++ b/packages/qwik-city/runtime/src/link-component.tsx
@@ -14,6 +14,7 @@ export const Link = component$<LinkProps>((props) => {
   const clientNavPath = untrack(() => getClientNavPath(linkProps, loc));
   const prefetchDataset = untrack(() => getPrefetchDataset(props, clientNavPath, loc));
   linkProps['preventdefault:click'] = !!clientNavPath;
+  linkProps['allowmodifiers'] = !!clientNavPath;
   linkProps.href = clientNavPath || originalHref;
   const onPrefetch =
     prefetchDataset != null
@@ -21,9 +22,8 @@ export const Link = component$<LinkProps>((props) => {
           prefetchLinkResources(elm as HTMLAnchorElement, ev.type === 'qvisible')
         )
       : undefined;
-  const handleClick = event$(async (_: any, elm: HTMLAnchorElement) => {
-    if (!elm.hasAttribute('preventdefault:click')) {
-      // Do not enter the nav pipeline if this is not a clientNavPath.
+  const handleClick = event$(async (event: any, elm: HTMLAnchorElement) => {
+    if (!(event as PointerEvent).defaultPrevented) {
       return;
     }
 
--- a/packages/qwik/src/core/render/jsx/types/jsx-qwik-attributes.ts
+++ b/packages/qwik/src/core/render/jsx/types/jsx-qwik-attributes.ts
@@ -186,6 +186,11 @@ export interface QwikProps<T extends Element> extends PreventDefault<T> {
   ref?: Ref<T> | undefined;
 
   /**
+   * Allow modifier keys (ctrl, meta, shift, alt) to override preventdefault.
+   */
+  allowmodifiers?: boolean;
+
+  /**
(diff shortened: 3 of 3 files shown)


packages/qwik/src/qwikloader.ts
```diff
@@ -52,7 +52,17 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
...
+        !element.hasAttribute('allowmodifiers') ||
+        !(
+          (ev as MouseEvent).ctrlKey ||
+          (ev as MouseEvent).metaKey ||
+          (ev as MouseEvent).shiftKey ||
+          (ev as MouseEvent).altKey
+        )
+      ) {
+        ev.preventDefault();
+      }
```
tzdesign, line 65: You always deliver @jordanw66. This dispatch is for all events right?

So this might be a good check and shortens the code a bit.
```suggestion
      if (
        !element.hasAttribute('allowmodifiers') ||
        !(ev instanceof MouseEvent && (ev.ctrlKey || ev.metaKey || ev.shiftKey || ev.altKey))
      ) {
        ev.preventDefault();
```

I'm maybe thinking to generic here and you might be right, this is enough.
As of right now, I have no clue how to document this that everyone gets it. Maybe this is not necessary as react-router-link works in the same way I guess.

tzdesign (changes requested):

shairez:
Thanks @jordanw66 !

Somewhat related to #4496 (as it's about adding a change to the qwikloader, which should be done on very rare ocassions IMO)

tzdesign:
@shairez @jordanw66 this is even better. we can than fine-grain these options in Link or custom components.

It's not breaking if it's a union with boolean. @jordanw66 let me know if you need anything. I'm pretty free tomorrow.

packages/qwik/src/qwikloader.ts
tzdesign: @jordanw66 valid point, but what happens on keyup?

Speaking of performance, instanceof will be faster that the suggestion here: #4496.   
The alternative would be casting to unknown and then check with ```in```.

I guess you abort this one and create a new one? Let me know

shairez:
> @shairez Interesting, didn't see that request. That will require a slightly larger increase to loader, do you guys know which direction you want to go in? I could take a stab at implementing that if you want. Could replace this PR with more functionality.

Thanks! yes this will resolve so many future issues as well 😅

The only concern I had is about how to convey that this is a syncornous function to the developer... but maybe if it's withouth the `$` then it's fine :)

Let's wait to hear what @manucorporat @adamdbradley @mhevery have to say about this before proceeding with a PR

manucorporat (changes requested):
Gonna block this PR so it's not merged by mistake, i agree preventdefault need to be improved, but changing qwikloader is a very serious matter, and this would not solve all the problems.

Wonder if the solution is make preventdefault some sort of inlined function, but that would require some `new Function()` or `eval()`

manucorporat:
The optimizer can already inline functions, but it's implicit there is not way to signal this
wonder if you can do this:

```
preventdefault:click={$$((event) => {

})}
```

$$ would be a new compiler API, like $, but syncronous, and you cant have imports!
the optimizer already knows to do that, but it's not public

manucorporat:
Not a fan of `$$` and it would be very advanced API. Maybe:

`_inlined_fn_`

tzdesign:
@manucorporat let us know if you know how to move forward. I think synchronous would be crazy in this async context.

I still can't get my head around, why we can't use event.preventDefault() in the Link-Component. You pass it through in dispatch, but I haven't read all of this and it ends up beeing any at some point.

I really need to read more about the qwik loader 🥲

tzdesign:
@jordanw66 Thanks for the kick to my head,it works better now 😂

Your idea sounds good. Keeping track of them by referencing might be the best solution.

I have popcorn 🍿 now. Manu will find the right solution, he always does.

tzdesign:
@manucorporat @jordanw66 is this still ongoing?

The browser's default behavior should actually be the default. I would propose `disallowModifiers` should be a prop.
For me even @jordanw66 PR would mean I have to change already over 100 links in our project.


I am thinking about writing a custom link component because of this behavior.
If you hit the meta keys to open in a new silent tab or window to copy the link faster or compare products in or ecommerce shop there is no way we will leave the link like this.

DustinJSilk:
Hey guys, this issue has been stale for a long time, is it possible we could revive this and decide on a solution?
It's a poor experience to not be able to use default browser behaviour with links.

wmertens:
@manucorporat So we want custom preventDefault filtering. Do we ever want any other sync event processing? What if we want to have the same custom preventDefault on several event types?

You proposed but did not like `preventdefault:click={$$((event) => {})}`, but I think we could special-case this? So just `preventDefault:whatever={(ev) => ...}`. The optimizer can emit the callback as an inline function and replace it with some sort of IQRL (inlined function index)?

tzdesign:
@jordanw66 but is there a plan to implement it as default? Qwik is the only framework I know, overriding browser standards here.

gabrielgrant:
@tzdesign this is tracked by #3871  right? looks like @mhevery self-assigned that, so guessing he's planning to make use of `sync()` to solve that once he's merged the `sync()` feature itself

tzdesign:
> @tzdesign this is tracked by #3871  right? looks like @mhevery self-assigned that, so guessing he's planning to make use of `sync()` to solve that once he's merged the `sync()` feature itself

Oh gosh, I forgot to check my own issue 😅

Good to know. With @mhevery its in the best hands 🙌

PR #4628, opened 2023-06-27
fix(qwik-core): reuse Invoke Context between grouped tasks

### LINKED ISSUE #4417: [🐞] `this` is undefined in server$ function when called consequentially.

Qwik City (routing)

### Describe the bug

When two or more `$server` functions called one after each other the second called function would have `this === undefined`

This doesn't happen when they called concurrently in `Promise.all`

```ts
const serverFunctionA = server$(async function () {
  console.log('serverFunctionA', this);
  return true;
});

const serverFunctionB = server$(async function () {
  console.log('serverFunctionB', this); // <--- here `this` would be undefined
  return true;
});

  const resource = useResource$(async () => {
    const resultA = await serverFunctionA();
    const resultB = await serverFunctionB();

    return { resultA, resultB };
  });

  return (
    <Resource
      value={resource}
      onResolved={({ resultA, resultB }) => (
        <>
          {resultA} {resultB}
        </>
      )}
    />
  );
});
```

### Reproduction

https://stackblitz.com/edit/qwik-starter-jdcnqd?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: macOS 13.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 75.86 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.17.0 - ~/.nvm/versions/node/v16.17.0/bin/node
    Yarn: 3.5.1 - ~/.nvm/versions/node/v16.17.0/bin/yarn
    npm: 8.15.0 - ~/.nvm/versions/node/v16.17.0/bin/npm
    Watchman: 2023.02.27.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 113.0.5672.126
    Firefox: 95.0.2
    Safari: 16.4
  npmPackages:
    @builder.io/qwik: 1.1.4 => 1.1.4 
    @builder.io/qwik-city: ~1.1.4 => 1.1.4 
    @builder.io/qwik-react: ^0.5.0 => 0.5.0 
    undici: 5.21.0 => 5.21.0 
    vite: 4.2.1 => 4.2.1
```

### Additional Information

_No response_

hbendev:
# Overview

When multiple `server$` functions are called from a `useResource$` hook, the invocation's context is dropped by `useResource` after the first function resolved.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

fixes #4417

# Use cases and why

One would expect they can write and use separate `server$` functions, and reuse them inside a single `<Resource />` 

# Checklist:

### DIFF
--- a/packages/qwik/src/core/qrl/qrl-class.ts
+++ b/packages/qwik/src/core/qrl/qrl-class.ts
@@ -146,7 +146,7 @@ export const createQRL = <TYPE>(
             context.$event$ = this as Event;
           }
           emitUsedSymbol(symbol, context.$element$, start);
-          if (maybeAsync && isServerPlatform()) {
+          if (maybeAsync) {
             return asyncInvoke.call(this, context, f, ...(args as Parameters<typeof f>));
           }
           return invoke.call(this, context, f, ...(args as Parameters<typeof f>));
--- a/packages/qwik/src/core/qrl/qrl-class.ts
+++ b/packages/qwik/src/core/qrl/qrl-class.ts
@@ -146,7 +146,7 @@ export const createQRL = <TYPE>(
             context.$event$ = this as Event;
           }
           emitUsedSymbol(symbol, context.$element$, start);
-          if (maybeAsync) {
+          if (maybeAsync && isServerPlatform()) {
             return asyncInvoke.call(this, context, f, ...(args as Parameters<typeof f>));
           }
           return invoke.call(this, context, f, ...(args as Parameters<typeof f>));
--- a/packages/qwik/src/core/qrl/qrl-class.ts
+++ b/packages/qwik/src/core/qrl/qrl-class.ts
@@ -9,6 +9,7 @@ import {
   newInvokeContext,
   newInvokeContextFromTuple,
   tryGetInvokeContext,
+  asyncInvoke,
   type InvokeContext,
   type InvokeTuple,
 } from '../use/use-core';
@@ -48,7 +49,8 @@ export type QRLInternalMethods<TYPE> = {
   getCaptured(): unknown[] | null;
   getFn(
     currentCtx?: InvokeContext | InvokeTuple,
-    beforeFn?: () => void
+    beforeFn?: () => void,
+    maybeAsync?: boolean
   ): TYPE extends (...args: any) => any
     ? (...args: Parameters<TYPE>) => ValueOrPromise<ReturnType<TYPE>>
     : // unknown so we allow assigning function QRLs to any
@@ -124,7 +126,8 @@ export const createQRL = <TYPE>(
   function invokeFn(
     this: unknown,
     currentCtx?: InvokeContext | InvokeTuple,
-    beforeFn?: () => void | boolean
+    beforeFn?: () => void | boolean,
+    maybeAsync?: boolean
   ) {
     return (...args: QrlArgs<TYPE>): QrlReturn<TYPE> => {
       const start = now();
@@ -143,6 +146,9 @@ export const createQRL = <TYPE>(
             context.$event$ = this as Event;
           }
           emitUsedSymbol(symbol, context.$element$, start);
+          if (maybeAsync) {
+            return asyncInvoke.call(this, context, f, ...(args as Parameters<typeof f>));
+          }
           return invoke.call(this, context, f, ...(args as Parameters<typeof f>));
         }
         throw qError(QError_qrlIsNotFunction);
--- a/packages/qwik/src/core/use/use-core.ts
+++ b/packages/qwik/src/core/use/use-core.ts
@@ -17,7 +17,7 @@ import {
 } from '../util/markers';
 import { isPromise, safeCall } from '../util/promises';
 import { seal } from '../util/qdev';
-import { isArray, ValueOrPromise } from '../util/types';
+import { isArray, type ValueOrPromise } from '../util/types';
 import { setLocale } from './use-locale';
 import type { Subscriber } from '../state/common';
 import type { Signal } from '../state/signal';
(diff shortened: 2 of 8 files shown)


hbendev:
1. This might not only be related to `useResource` and `server` QRLs, but other combinations where one invokes other tasks.
2. Maybe we can/should fix this in `useResource` instead.

I'll try to gather a little more context and come back to the issue.

packages/qwik/src/core/use/use-core.ts
```diff
@@ -118,6 +119,26 @@ export function invoke<ARGS extends any[] = any[], RET = any>(
...
+  const previousContext = _context;
+  _context = context;
+  return safeCall(
+    () => fn.apply(this, args),
+    (value) => {
+      _context = previousContext;
+      return value;
+    },
+    (e) => {
+      throw e;
```
wmertens, line 137: Shouldn't we restore context here?

packages/qwik/src/core/qrl/qrl-class.ts
```diff
@@ -35,6 +36,7 @@ export interface QRLInternalMethods<TYPE> {
   getHash(): string;
   getCaptured(): any[] | null;
   getFn(
+    maybeAsync?: boolean,
```
wmertens, line 39: If you put this as the last argument, you don't have to pass it if it's false

wmertens (commented):
Is this PR still relevant? There were some changes to requestcontext IIRC.

hbendev:
Converting to draft, as the fix is not yet ready. So far, I've only extended the test suite with another similar related, and also unexpected functionality. I also added this test to a separate [qwik-city app](https://github.com/hbendev/qwik-svg-render-bug/commit/6211894c769b4c9b72982d3c92b7c22830989170), if you want to have a look. The source of this other issue is [a discord thread](https://discord.com/channels/842438759945601056/1160923192538439711/1160923192538439711).

packages/qwik/src/core/use/use-core.ts
```diff
@@ -123,6 +124,26 @@ export function invoke<ARGS extends any[] = any[], RET = any>(
   return returnValue;
 }
 
+export function asyncInvoke<ARGS extends any[] = any[], RET = any>(
```
wmertens, line 127: actually, this isn't possible in the browser environment. Concurrent invokes will mess up things unless you use async localstorage.

wmertens:
I think we're looking at the same problem, right? I'm trying to solve it in #5510 but got stalled.

hbendev:
I think so, yes!

packages/qwik/src/core/use/use-core.ts
PatrickJS: we can just have a check for now 
```js
                    if (maybeAsync && isServerPlatform())  {
                        return asyncInvoke.call(this, context, fn, ...args);
                    }
                    return invoke.call(this, context, f, ...args);
```

PatrickJS: I got all this working I had to change a few things

PatrickJS: @hbendev if you have time right now. Can update your PR I can apply my changes after and we can get this merged

hbendev: Hi @PatrickJS. Just this change you mean? I can surely add in a few minutes.

PatrickJS: @hbendev yeah fix the conflicts and I'll refactor your pr with my changes

hbendev: On it!

packages/qwik/src/core/use/use-core.ts
PatrickJS: I'll add this

packages/qwik/src/core/qrl/qrl-class.ts
PatrickJS: I'll refactor for this

PatrickJS:
@hbendev keep the tests there 😅

hbendev:
Sorry, I'll push with the merge resolutions soon.

PatrickJS:
yeah thats fine the ci can run the tests. did you accidentally format everything

hbendev:
@PatrickJS it's done. Fixed the formatting too (I had older node and npm package versions...)

PatrickJS:
@hbendev something is wrong with the formatting which is different

PatrickJS:
@hbendev this should be good enough I'll fix everything

PatrickJS:
@hbendev can you add me to your fork so I can push to it

hbendev:
Just invited you, @PatrickJS

PatrickJS:
we need better concurrent tests I don't think my fix will work

PatrickJS:
ok it seems like it doesn't work well with hmr but you don't have concurrent issues locally

wmertens:
@PatrickJS So this still can't work 100%. The problem is when different sync or async calls are happening at the same time with different invoke contexts. They will overwrite each other's context.

Since invoke is a sync API and both sync+async need to use tryGetInvokeContext, you can't even pause all invokes while an async invoke is running. So in the browser, it's impossible to make this work.

What could be done on the server: use https://nodejs.org/api/async_context.html#class-asynclocalstorage. It's supported by most runtimes, and it allows you to provide a store to a "thread" of async calls.

PatrickJS:
@wmertens can we do AsyncLocalStorage and/or workers per user request to fix the global issue. we should refactor tryGetInvokeContext out or have a better API

PatrickJS:
we're fixing this with this pr
https://github.com/BuilderIO/qwik/pull/5905

PR #4644, opened 2023-06-29
fix: csstypes dep

manucorporat:


hbendev:
Hi @manucorporat, can you share what kind of issue arised from csstype being declared as a devDependency? It only ever exports a .d.ts file (and a flow typing file). I really want to undesrtand what went wrong, and sorry for the extra work.

manucorporat:
Hey! it was my fault!

The issue is that this dependency becomes part of our public API, out JSX api. devDependencies will not be installed for our users, so their JSX is pointing to a dependency that does not exist.

It does not matter if it's .ts, .d.ts, the point is that it's not there for our final users!

Makes sense?

hbendev:
Thanks! Got it, and I should have known better. I was just confused because i had `@types/react` installed... I'm using `qwik-react` on my playground project. They do list `csstype` as a dependency, hence `qwik/jsx-runtime.dts` had reference to it in my case.

PR #4666, opened 2023-06-30
fix(eslint): useMethod rule for regular functions

shairez:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

When using regular functions in tasks and async await, eslint would throw an error

For example in the following code: 
```
useTask$(async function myTask(){
  await fetch()
});

useTask$(function myTask2(){
  // ESLINT WOULD SHOW AN ERROR HERE 👈
});

```

# Checklist:


manucorporat:
Amazing! also! great test!

PR #4668, opened 2023-07-01
fix: the property parameter can be symbol

wtlin1228:
# Overview

- Change the type for `prop` param of `getOwnPropertyDescriptor` method.
- Add return types for `getOwnPropertyDescriptor` and `has` methods.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

I think `getOwnPropertyDescriptor`'s prop parameter can also be symbol type since we're checking:

```ts
typeof prop === 'symbol'
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


shairez:
thanks for your contribution @wtlin1228 !

PR #4678, opened 2023-07-01
feat: improved Link prefetching behavior, API, and docs; make it similar to Next.js

jordanw66:
# Overview

This PR makes `prefetch` on `Link` similar to prefetch on `Next.js`:

- Prefetching occurs by default using `on:qvisible`.
- Opt-out completely with `prefetch={false}`.
- Or, optionally, only prefetch javascript bundles (and not data) to render the page with `prefetch="js"`.
- In dev mode, prefetch will occur with `mouseover` or `focus` instead of `qvisible`.
- If a user has their `data saver` setting set, prefetching is skipped.
- Optimized to not prefetch nor attach handlers on same-page `Link`. (hashes, navbar, etc)
- Added detailed jsdoc/API explanation of what `prefetch` does and how it works, along with options.
- Removed `aria-pressed` because this is for toggle buttons, not `<a>` tags. Screen readers already handle links, in fact adding and removing `aria-pressed` will cause confusion for these users as it's not normal link behavior.
- Improved the reliability of `pathname` comparison by normalizing path trailing slashes first, added unit tests for this.
- Added and modified unit tests for verifying decisions from prefetch checks.

Prefetching symbols (and/or data) to render a page before SPA navigation is a huge boost to user experience, especially on slow/mobile connections, preventing a burst of network activity during navigation.

If you think defaults or API should be different let me know.

References:
https://nextjs.org/docs/pages/api-reference/components/link#prefetch
https://nextjs.org/docs/messages/prefetch-true-deprecated
https://www.builder.io/blog/prefetching-nextjs-visual-guide
https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/pressed/#intended-use

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


wmertens:
@jordanw66 this is great! Could I implore you to rebase?

An e2e test would be very nice too, just not quite sure how to test prefetching.

wmertens:
@jordanw66 wow, I'm glad you're back now! Totally understand, if you could do a quick rebase that would be fine.

wmertens:
Closing in favor of #5480

PR #4688, opened 2023-07-02
fix(deno): origin resolution

manucorporat:


wmertens:
Seems outdated now, the deno starter uses `Deno.serve` which seems to do the same thing

PR #4727, opened 2023-07-05
feat(server-functions): routeAction$ validation based on RequestEvent

tzdesign:
Passed the request event to the builder functions to make request event based designed zod schemas.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

We have an e-commerce site with multiple shops that are powered by a single application. This has complex validation structures as a shop in Denmark has some fields as required type and for the international shop this type is optional. To determine the correct zodSchema we need the RequestEvent to work with. See the example below.

I have to admit that I am still in a relationship with ```yarn```, so I added the link option to it, if you don't mind.

```ts
export const useAddUser = routeAction$(
  async (user) => {
    // The "user" is still strongly typed, but firstname 
    // is now optional: { firstName?: string | undefined, lastName: string }
    const userID = await db.users.add({
      firstName: user.firstName,
      lastName: user.lastName,
    });
    return {
      success: true,
      userID,
    };
  },
  // Zod schema is used to validate that the FormData includes "firstName" and "lastName"
  zod$((z, ev) => {
    // The first name is optional if the url contains the query parameter "firstname=optional"
    const firstName =
      ev.url.searchParams.get("firstname") === "optional"
        ? z.string().optional()
        : z.string().nonempty();

    return z.object({
      firstName,
      lastName: z.string(),
    });
  })
);
```

# Use cases and why

- Return advanced schemas with superRefine based on RequestEventData
- Return dynamic zodSchemas based on routActions form data

# Questions

- How bad is it in terms of performance to move schema resolution to the validate function?
- Should this be another export like zodEvent$ with a simpler signature?

# Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,6 +182,7 @@ To use your build in your project, follow these steps:
    ```shell
    pnpm link.dist.npm
    ```
+
    or
 
    ```shell
@@ -201,6 +202,7 @@ To use your build in your project, follow these steps:
     pnpm install
     pnpm link --global @builder.io/qwik @builder.io/qwik-city
    ```
+
    or
 
    ```shell
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -182,6 +182,11 @@ To use your build in your project, follow these steps:
    ```shell
    pnpm link.dist.npm
    ```
+   or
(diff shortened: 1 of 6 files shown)


tzdesign:
@manucorporat here it is.

PR #4749, opened 2023-07-07
feat(qwik-core): expose isSignal helper function

octoper:
# Overview

I have found some cases when I'm creating custom hooks that I need to check if a given parameter is a signal, I have always checked if the value has the `value` or `untrackedValue` but I don't find that optimal because in hooks that can accept many forms of data these properties maybe are present but the value may not be a Signal.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This change just exposes the `isSignal` function.

# Use cases and why

```ts
import type { Signal } from '@builder.io/qwik'
import { useVisibleTask$, isSignal } from '@builder.io/qwik'

export function useTitle(
  title: Signal<string> | string,
) {
  const newTitle = useSignal(isSignal(title) ?  title.value : title)

  useVisibleTask$(({ track }) => {
    track(() => newTitle.value)
    document.title = newTitle.value
  }, {
    strategy: 'document-ready'
  })
}
```

# Checklist:


mhevery:
Thank you for your contribution.

packages/qwik/src/core/state/signal.ts
```diff
@@ -161,6 +161,13 @@ export class SignalWrapper<T extends Record<string, any>, P extends keyof T> ext
   }
 }
 
+/**
+ * Checks if a given object is a `Signal`.
+ * 
+ * @param obj - The object to check if `Signal`.
+ * @returns boolean - True if the object is a `Signal`.
```
gioboa, line 168:
```suggestion
 * @returns Boolean - True if the object is a `Signal`.
```

packages/qwik/src/core/state/signal.ts
```diff
@@ -161,6 +161,13 @@ export class SignalWrapper<T extends Record<string, any>, P extends keyof T> ext
   }
 }
 
+/**
+ * Checks if a given object is a `Signal`.
+ *
```
gioboa, line 166:
```suggestion
 *
```

gioboa:
thanks @octoper for your help.

PR #4798, opened 2023-07-15
refactor: extract group type

wtlin1228:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

While going over the Qwik source code, I noticed `SubscriberEffect | SubscriberHost | Node` type is used in many places that might be clearer if it is extracted into a standalone `Group` type.

I was wondering wether it would be better to derive `Group` type from `Subscriptions` type like the following because group can only be the host of subscription.

```ts
type Group = Subscriptions[1]; // SubscriberEffect | QwikElement | Text
```

But that will cause a problem in the `cleanupTree`. So I decided not to do so.

```
export const cleanupTree = (
  elm: Node | VirtualElement,
  // ...
) => {
  // ❌ Type 'Node' is not assignable to type 'SubscriberEffect | QwikElement | Text'.
  subsManager.$clearSub$(elm); 
  // ...
};
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/state/common.ts
+++ b/packages/qwik/src/core/state/common.ts
@@ -21,8 +21,8 @@ import type { Signal } from './signal';
 export interface SubscriptionManager {
   $groupToManagers$: GroupToManagersMap;
   $createManager$(map?: Subscriptions[]): LocalSubscriptionManager;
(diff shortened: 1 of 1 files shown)


wmertens:
Thanks!

PR #4799, opened 2023-07-15
Meta tag names

AL-Hareth:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Changed the type of the name property on the DocumentMeta interface from `string` to union types that includes all the values possible in a name attribute inside the meta tag

from: 
```ts
interface DocumentMeta {
    // ...
    readonly name?: string;
    // ...
}
```
to:
```ts
interface DocumentMeta {
    // ...
    readonly name?: "application-name" | "author" | "keyword" /* ...etc */;
    // ...
}
```

# Use cases and why

This gives better development experience where you get the correct name values if you're trying to add a meta tag data using qwik-city

# Checklist:


mhevery:
Are you still working on this change or is it abandoned?

AL-Hareth:
> Are you still working on this change or is it abandoned?

I would be happy to work on it, but I don't really understand the process that I should go with before making the pull request

mhevery:
I guess it just needs a rebase? Let me see if I can make it pass...

mhevery:
It looks like this change has already been merged https://github.com/BuilderIO/qwik/pull/4340

packages/qwik-city/runtime/src/api.md
```diff
@@ -167,6 +167,8 @@ export interface DocumentLink {
     type?: string;
 }
 
+type MetaNames =  "application-name" | "author" | "description" | "generator" | "keywords" | "referrer" | "theme-color" | "color-scheme" | "viewport" | "creator" | "googlebot" | "publisher" | "robots";
```
mhevery, line 170: I don't think this is an exhaustive list Where are these defined?

mhevery: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name

I don't think there is an exhaustive list, so such a change would make this breaking to our users.

mhevery:
So the web does not define a "standard" set of names for the `<meta>` tag see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name For this reason, I don't think we should change the type system to make it more restrictive as the user can really add any value they wish.

I am going to close this PR. Let me know if you disagree..

PR #4827, opened 2023-07-19
feat: add snooze for image size cls devtool

jessezhang91:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

https://github.com/BuilderIO/qwik/issues/4773

Adding a snooze button to hide the CLS image resize dev tool so we can disable when working on designs.

I followed the conventions set by the `clickToSource` option.

# Use cases and why

Currently, there's no way to disable this functionality and it makes the iterative dev cycle a little difficult before we figure out the exact sizes for the design. I imagine with this flag, we'll disable this when it gets cumbersome and re-enable after we've made our changes.

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/request-handler/cookie.ts
+++ b/packages/qwik-city/middleware/request-handler/cookie.ts
@@ -52,6 +52,14 @@ const createSetCookieValue = (cookieName: string, cookieValue: string, options: 
   return c.join('; ');
 };
 
+function tryDecodeUriComponent(str: string) {
+  try {
+    return decodeURIComponent(str);
+  } catch {
+    return str;
+  }
+}
+
 const parseCookieString = (cookieString: string | undefined | null) => {
   const cookie: Record<string, string> = {};
   if (typeof cookieString === 'string' && cookieString !== '') {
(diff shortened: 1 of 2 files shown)


zanettin:
thx a lot @jessezhang91 for submitting this PR 🙏 

we had last week a chat with the core team regarding that ropic and there will be a solution in the future to snooze this warning for a bit. the goal behind the message is to remind the devs that here is something on the code that needs your attention.

that's why we don't want to provide an option to disable the performance relevant notices at the moment since they will help you at the end to achieve better page metrics.

cc @mhevery / @manucorporat

mhevery:
I think snooze is the way to go, and I would like to encourage you in that direction.

jessezhang91:
@mhevery @zanettin i've added a "snooze for an hour" button to the image-warning element
image

i'm curious to get your thoughts on how the PerformanceObserver CLS should be handled since that's attached as an `::after` pseudoelement.
image

my thought is maybe refactoring these warnings to something similar to the image-warning element but i imagine layout shifts that move the item around would "break" the initial outline positioning (if it layout-shifted multiple times).

zanettin:
thanks a lot for the rewrite of this PR 🙏  amazing! and sorry for the late reply.

imo we could set an even higher value than just an hour. @manucorporat @mhevery what do you think? imo a value around 2d+ would be nice?

regarding the CLS warnings: my personal opinion would be to have a global snooze for all perf warnings on a single place. could be a flag for the dev server or an element on the UI. but here we would need some inputs from the core team 😄 thanks again for the made changes 🙏 🎉 

PS: CI is not happy atm because a `pnpm fmt` is required to format the code 👼

jessezhang91:
thanks for the response! i was thinking some "dev tools config screen" that is injected on the page that can control things like snoozing CLS errors or enabling/disabling the click-to-source functionality and whatever else is added in the future could be a good way to manage things more globally. but yeah, would definitely love some of the core team's opinions here

PR #4841, opened 2023-07-21
fix(render): empty string for boolean props

### LINKED ISSUE #4824: [🐞] inconsistent rendering between SSR and CSR

Qwik Runtime

### Describe the bug

SSR interprets `<button disabled={''}>` as disabled but CSR interprets it as NOT disabled. 

### Reproduction

https://stackblitz.com/edit/qwik-starter-u3vczw?file=README.md,src%2Froutes%2Findex.tsx

### Steps to reproduce

Click on the toggle button and observe that the target button disabled does not correctly follow from SSR.

### System Info

```shell
n/a
```

### Additional Information

_No response_

Varixo:
fix #4824

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Empty string value should render property

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/dom/operations.ts
+++ b/packages/qwik/src/core/render/dom/operations.ts
@@ -7,7 +7,7 @@ import { isElement, isNode, isQwikElement } from '../../util/element';
 import { logDebug, logError, logWarn } from '../../util/log';
 import { QSlot, QSlotRef, QStyle } from '../../util/markers';
 import { qDev } from '../../util/qdev';
-import { directGetAttribute, directSetAttribute } from '../fast-calls';
+import { directGetAttribute, directRemoveAttribute, directSetAttribute } from '../fast-calls';
 import type { RenderStaticContext } from '../types';
 import type { QwikElement, VirtualElement } from './virtual-element';
 import {
@@ -36,7 +36,7 @@ export const setAttribute = (
 
 const _setAttribute = (el: QwikElement, prop: string, value: any) => {
   if (value == null || value === false) {
-    el.removeAttribute(prop);
+    directRemoveAttribute(el, prop);
   } else {
     const str = value === true ? '' : String(value);
     directSetAttribute(el, prop, str);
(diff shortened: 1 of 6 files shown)


manucorporat:
Imo, this is a hack, and it should be considered a warning, or an error. It's changing the attribute twice.

manucorporat (changes requested):
imo, it's not a bug, but an invalid value. Maybe we should consider a warning instead

Varixo:
I don't know if this is still required as typescript should do the job. Feel free to close if changes are not desired :D

javascript-unsafe:
# Still broken in 1.20.0 — a page straight from SSR shows empty inputs

Hi! I hit a real user-visible bug that this PR is about, and wanted to share a simple reproduction.

**What happens:** write `<input value={false} />`. If the page comes from the server (SSR), the input shows up **empty**. If the exact same component is rendered on the client (CSR), it shows **"false"**. Same code, two different things on screen.

### Screenshot

SSR vs CSR

*Top: what the user sees on a server-rendered page — every input is empty. Bottom: the same component rendered on the client — inputs show "false" / "true".*

### How to reproduce (3 steps)

1. Make a tiny component:

   ```tsx
   const sig = useSignal(false);
   <input value={sig.value} />
   ```

2. Render it with SSR → the HTML is `<input>` (no `value` attribute at all), so the user sees an empty box.

3. Render the same component only on the client → the HTML is `<input value="false">`, so the user sees "false".

Same story with `value={true}`: SSR gives a bare `value` attribute (which the browser reads as `""`), CSR gives `value="true"`.

| What you write | SSR page shows | Client-rendered page shows |
|---|---|---|
| `value={false}` | *empty* ❌ | "false" |
| `value={true}` | *empty* ❌ | "true" |
| `value={null}` / `{undefined}` | *empty* ✅ (consistent) | *empty* |

(`null`/`undefined` are fine — it's only the booleans that disagree.)

### Why it matters

This isn't just a hydration warning — on a resumed SSR page the user's first paint shows an empty field where the component logic says there should be a value. It bit me with a filter dropdown built on `<option value={...}>` (same as reported in #4824).

### Note on the fix

From what I can see in 1.20.0, the SSR side already treats `false` as "remove the attribute", but the client-side render still writes `String(value)` → `"false"`. So the two sides just need to agree. If this PR (or a follow-up) normalizes the CSR side, the table above should go all-green — happy to re-test against a branch or nightly build.

PR #4901, opened 2023-08-01
fix: skip serializing functions returned by exported functions during pause

nynevi:
Overview:

Prevents serializing functions that return other functions which otherwise yields in error `Only primitive and object literals can be serialized`:

Fixes https://github.com/BuilderIO/qwik/issues/4371

# Checklist:

### DIFF
--- a/packages/qwik/src/core/container/pause.ts
+++ b/packages/qwik/src/core/container/pause.ts
@@ -25,7 +25,12 @@ import {
   QError_missingObjectId,
   QError_verifySerializable,
 } from '../error/error';
-import { isArray, isObject, isSerializableObject } from '../util/types';
+import {
+  isArray,
+  isFunction,
+  isObject,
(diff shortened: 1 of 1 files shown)


mhevery (changes requested):
Thank you for your contribution; unfortunately, I don't think this is the correct fix.

1. This fix silently ignores function values during serialization and deserializes them as `undefined` This will cause a surprise in the browser. The correct behavior is to tell the developer that they have made a mistake.
  - There should be a clear runtime error (instead of the current cryptic error)
  - The linter should be updated to underline the offending code correctly. 
2. We need to have tests with the change.

mees-van-wel:
What's the status on this?

gioboa:
@nynevi thanks for opening this PR.
I'm closing this for now because it's stale, feel free to create a new PR with your improvements. :muscle:

PR #4905, opened 2023-08-02
style(pnpm-lock.yaml): synchronize lockfile version with version of pnpm in use

jensmeindertsma:
# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This workspace is pinned to pnpm@8.6.1 which uses lockfileVersion 6.1 and not 6.0

# Checklist:


zanettin (commented):
LGTM 🙏 thx u very much @jensmeindertsma 🙏 

quickly checked on pnpm side (https://github.com/pnpm/pnpm/tree/v8.6.1) and the lockfile version was upgraded but just for pnpm `v8.6.1` and `v8.6.2`. after that it went down again to `v6.0`. 

@manucorporat what is the general plan with upgrading pnpm to latest?

jensmeindertsma:
Ah, yeah, you're right. I guess that means we need to decide whether we bump "packageManager". I think it may a good idea, just to stay up to date, also performance wise

mhevery:
Sorry, I don't understand this change. the `pnpm` generated this file. Do you have a different version of `pnpm`? If so there is more work to update the `pnpm` version than just lock file...

zanettin:
@mhevery the root [pkg.json](https://github.com/BuilderIO/qwik/blob/main/package.json#L96) sets the `packageManager` to `pnpm@8.6.1`. everybody having this pnpm version or `8.6.2` installed will set the `pnpm-lock.yaml` to `6.1`. new and older versions don't affect the version. i was able to reproduce this with that pnpm version but since re-upgraded to `8.6.22` it doesn't happen anymore.

so my question would be: do we upgrade the value to `pnpm@8.6.3` or higher, or should we update the lock file version to 6.1 and enforce that this pnpm version is used for all contributors. i am not yet familiar with [corepack](https://nodejs.org/docs/latest-v19.x/api/corepack.html) 👼

mhevery:
> @mhevery the root [pkg.json](https://github.com/BuilderIO/qwik/blob/main/package.json#L96) sets the `packageManager` to `pnpm@8.6.1`. everybody having this pnpm version or `8.6.2` installed will set the `pnpm-lock.yaml` to `6.1`. newer and older versions don't affect the version. i was able to reproduce this with that pnpm version but since re-upgraded to `8.6.22` it doesn't happen anymore.
> 
> so my question would be: do we upgrade the value to `pnpm@8.6.3` or higher, or should we update the lock file version to 6.1 and enforce that this pnpm version is used for all contributors. i am not yet familiar with [corepack](https://nodejs.org/docs/latest-v19.x/api/corepack.html) 👼
> 
> to ensure a min. version we could set the `engines.pnpm` key and set that to `>8.6.2` as well. imo they should be in sync with the version defined on `packageManager`.

I think the correct next step is to update `engines.pnpm` to force everyone on a new version and than use whatever lock file it generates.

Problem with updating just the lock file is that next time someone will npm install with old version they will downgrade the lock file again.

zanettin:
@mhevery agree 👍 any version u'd suggest? starting at  `8.6.3` we could pick all values and set enginge as well as the pkgManager values. latest is `8.6.22`.

mhevery:
> @mhevery agree 👍 any version u'd suggest? starting at `8.6.3` we could pick all versions and set enginge as well as the pkgManager values. latest is `8.6.22`.

I have no opinions...

mhevery (approved):

mhevery:
merge at will...

jensmeindertsma:
Thanks all, this seems like the right move, as more and more tools respect `engines` and `packageManager`

zanettin:
thank YOU for starting this concersation and your time 🙏

PR #4912, opened 2023-08-07
feat(qwik-city): add a way to get server-side env vars from `onStaticGenerate`

### LINKED ISSUE #4907: [✨] Dynamic SSG routes (`onStaticGenerate`) and server-side environment variables

Hi everyone :wave:

The [Dynamic SSG routes](https://qwik.builder.io/docs/guides/static-site-generation/#dynamic-ssg-routes) are a nice way to generates pages during the build, given some data that you fetch from somewhere.

However, if this "somewhere" is a private database or a private API, and the associated credentials are stored in environments variables (because we want to do things well, https://12factor.net/config), then there is no way to get those server-side env vars when being in `onStaticGenerate()` function.

Like it's done for `RequestHandler` and `RequestEventLoader`, I would like to inject a function in parameters a function `env.get(key: string)` that will give access to server-side env vars (by their name).

First, I would change the `StaticGenerateHandler` signature from:

https://github.com/BuilderIO/qwik/blob/2e96ae030a299e66191627352cb9252057064409/packages/qwik-city/runtime/src/types.ts#L324

to:
```ts
declare interface EnvGetter {
    get(key: string): string | undefined;
}

export type StaticGenerateHandler = ({ env }: { EnvGetter }) => Promise<StaticGenerate> | StaticGenerate;
```

Then change:   https://github.com/BuilderIO/qwik/blob/2e96ae030a299e66191627352cb9252057064409/packages/qwik-city/static/main-thread.ts#L186

For this: 
```ts
const staticGenerate = await pageModule.onStaticGenerate({
  env: {
    get(key) {
      return sys.getKey(key);
    }
  }
});
```

No

### Additional context

If this solution is acceptable, I can try to open a PR for that, since I definitely think this is a must-have feature.

Kocal:
# Overview

Add a way to access server-side env vars from `onStaticGenerate()`.

Fix #4907.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

The function `onStaticGenerate()` now takes an `EnvGetter`, allowing the end-user to generate SSG
routes parameters by fetching data from a service which is protected by credentials (e.g.: an
API or a database), and where those credentials are available as server-side env vars.

# Use cases and why

- Our Qwik app use a Strapi backend and needs to call its API to get articles data. This API need a token, which is stored inside a server-side env var (stored in a `.env.local`).

# Checklist:

- ~[ ] Added new tests to cover the fix / functionality~  I didn't find any tests around `onStaticGenerate()`.


zanettin:
Hi @Kocal 👋 
Thanks for creating this PR 🙏 may i ask u to run `pnpm api.update` to automatically update the api docs? the CI is complaining atm 😄

Kocal:
Oh, weird, I thought everything was correct before pushing 🤔 

I will double-check and push if needed.

mhevery (approved):

PR #4915, opened 2023-08-08
fix: fix cookie parsing to allow invalid uri encoding

jessezhang91:
# Overview

Cookies with invalid URI encoding throws a `URIError: URI malformed` error and bails out of processing.

Cookie values do not need to be URL encoded. Looking at the `cookie` library used by `cookie-parser` and many others, they do a [`tryDecode`](https://github.com/jshttp/cookie/blob/master/index.js#L86) and return the raw value on failure to decode.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

I updated the cookie parsing function to mimic this `tryDecode` functionality and added a test to validate that.

# Use cases and why

A cookie header with `foo=%bar` should have a cookie with key `foo` have value `%bar` and the rest of the app should continue to function. Currently, it gets a `URIError: URI malformed` and bails with a server error.

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/request-handler/cookie.ts
+++ b/packages/qwik-city/middleware/request-handler/cookie.ts
@@ -52,6 +52,17 @@ const createSetCookieValue = (cookieName: string, cookieValue: string, options: 
   return c.join('; ');
 };
 
+function tryDecodeUriComponent(str: string) {
+  if (str.indexOf('%') === -1) {
+    return str;
+  }
+  try {
+    return decodeURIComponent(str);
+  } catch {
+    return str;
+  }
+}
+
 const parseCookieString = (cookieString: string | undefined | null) => {
   const cookie: Record<string, string> = {};
   if (typeof cookieString === 'string' && cookieString !== '') {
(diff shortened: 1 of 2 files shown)


jessezhang91:
Bump on this PR. We're running into production issues with this due to what looks like ad tracking cookies. It's not _that_ high volume but it's not insignificant.

packages/qwik-city/middleware/request-handler/cookie.ts
```diff
@@ -52,15 +52,26 @@ const createSetCookieValue = (cookieName: string, cookieValue: string, options:
   return c.join('; ');
 };
 
+function tryDecodeUriComponent(str: string) {
+  if (str.indexOf('%') === -1) {
```
mhevery: I think this check is not-needed. If not `%`, then `decodeURIComponent` just returns a normal string.

mhevery (approved):

packages/qwik-city/middleware/request-handler/cookie.ts
jessezhang91: Yeah, agreed. I added that only because the `cookie` library had something like that (probably for optimization?).

PR #4928, opened 2023-08-11
docs: Add modal cookbook example.

### LINKED ISSUE #4575: [✨] Built-in Portal Support

There doesn't seem to be a good way to implement Portal functionality right now.  Applications can overcome this, but it may be difficult when creating an a11y-friendly component library with complex Overlay behavior

Providing `<Portal />` components or other qwikized ways

No

### Additional context

_No response_

mhevery:
fix #4575

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/demo/cookbook/modal/index.tsx
+++ b/packages/docs/src/routes/demo/cookbook/modal/index.tsx
@@ -1,0 +1,37 @@
+import { component$, useContext, useStylesScoped$ } from '@builder.io/qwik';
+import { PopupManagerContext } from './popup-manager';
+import PopupExampleCSS from './popup-example.css?inline';
+
+export default component$(() => {
+  // Retrieve the popup manager API
+  const popupManager = useContext(PopupManagerContext);
+  return (
+    <button
+      onClick$={async () => {
+        await popupManager.show(PopupExample, {
+          salutation: 'Hello',
+          name: 'World',
+        });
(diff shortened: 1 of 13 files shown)


brandonpittman:
@mhevery 

Thank you for this. My only worry is that it seems like you'd have to set up multiple `PopupManager` wrappers for each possible portal you'd want to have. I'm thinking of a scenario from an app we have at work:

1. Modal portal
2. Custom select portal (for the dropdown content)
3. Tooltip portal (for mouseover hints)

Would you need three wrapping components in this case?

mhevery:
> Thank you for this. My only worry is that it seems like you'd have to set up multiple `PopupManager` wrappers for each possible portal you'd want to have. I'm thinking of a scenario from an app we have at work:
> 
> 1. Modal portal
> 2. Custom select portal (for the list in a custom dropdown)
> 3. Tooltip portal (for mouseover hints)
> 
> Would you need three wrapping components in this case?
> 
> With a function like `createPortal()` you get to portal out dynamically and as many times as you want—you don't have to plan out your portal destinations ahead of time.

@brandonpittman  I provided a simple example that is easy to follow. One could easily design a more complex popupManager which could handle multiple popups as in your case.

PR #4935, opened 2023-08-14
fix(insights): add useIsAuthUser guard

gioboa:
Only auth users can access to /app/[...] routes
Removed unused Supabase API keys.


packages/insights/src/routes/app/layout.tsx
```diff
@@ -0,0 +1,14 @@
+import type { Session } from '@auth/core/types';
+import { Slot, component$ } from '@builder.io/qwik';
+import { routeLoader$ } from '@builder.io/qwik-city';
+
+export const useIsAuthUser = routeLoader$(({ sharedMap, redirect }) => {
+  const session = sharedMap.get('session') as Session | null;
+  if (!session) {
+    throw redirect(308, '/');
```
mhevery, line 8: I think it should be 307? As this is not a permanent redirect.

mhevery (changes requested):

mhevery (approved):

PR #4937, opened 2023-08-14
deps(qwik-auth): update @auth/core

ulic75:


ulic75:
Pfft, I don't know.  Hopefully someone can update the @auth/core version.

shairez:
what happened @ulic75 ?

ulic75:
> what happened @ulic75 ?

Oh, just the failed build stuff.  I thought it might be easy to slap the new version in there, but I guess not and don't have the time right now to sort it out.  And I'd rather not leave the dangling PR, there's enough danglers already :)

shairez:
OK cool, thanks for the info
is there an issue opened about it?

ulic75:
I don't think so. But @auth/core is still pre 1.0 so it's changing weekly.
Nothing that breaks qwik-auth so far though.

shairez:
OK Cool thanks!

PR #4957, opened 2023-08-17
cli(library): fix release script

fabiobiondi:
# What is it?

- [x] Bug

# Description

The `release` script in library starter is failing. 
To solve it I added the `--no-tests` to skip the test step.
By the way there is no testing library configured in the starter

# Steps to reproduce the issue

- Create a library project with Qwik CLI: `npm create qwik` -> Library
- Run the `release` script
- "Running test using npm" step fails

# Checklist:

### DIFF
--- a/starters/apps/library/package.json
+++ b/starters/apps/library/package.json
@@ -29,7 +29,7 @@
     "lint": "eslint \"src/**/*.ts*\"",
     "start": "vite --open --mode ssr",
     "qwik": "qwik",
-    "release": "np"
+    "release": "np --no-tests"
   },
   "devDependencies": {
     "@builder.io/qwik": "latest",
--- a/starters/apps/library/package.json
+++ b/starters/apps/library/package.json
@@ -27,9 +27,10 @@
     "fmt": "prettier --write .",
     "fmt.check": "prettier --check .",
     "lint": "eslint \"src/**/*.ts*\"",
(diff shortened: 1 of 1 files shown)


zanettin:
Hi @fabiobiondi 👋 
Thx for this change as well 🙏  just a quick question w/o having tested the issue on my end: wouldn't it be sufficient to specify a `test` script on the `package.json` like `"test": "echo \"No test specified\" && exit 0"`?
Imo that would be preferable if this works 👼  thx again

fabiobiondi:
Hi @zanettin, 
I have tried your solution right now and I confirm it perfectly works.
Honestly I like both solutions 😅

mhevery:
> `"test": "echo \"No test specified\" && exit 0"`?

I like this solution as it makes it clear where the test will go in the future. And it does not special case this, it just says we don't have tests.

zanettin:
thx for checking the proposal @fabiobiondi 🙏  thought the same as @mhevery just mentioned. as soon as a user makes use of any tests he wouldn't need to adjust the np script as well. let me know when i should do this change. thanks again for your time and help 🙏

fabiobiondi:
Thank you @mhevery and @zanettin!
@zanettin: I pushed (and tested again) your proposal right now on the same PR  so you can merge it ♥️
Thanks : )

zanettin (approved):
awesome job 🙏 thx a lot!

PR #4960, opened 2023-08-17
fix(auth): Cookies get updated if session returns Updated Cookies

aliyss:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The response from ``await Auth(new Request...) may return the headers 'set-cookie'. When using jwt with token rotation this get's called and subsequently the rotation errors out. This allows the token rotation to work without errors.

# Use cases and why

**Actual Behaviour**
``await Auth(...)`` gets called returns an updated cookie. Cookie is not set for ``req``.

**Expected Behaviour**
``await Auth(...)`` returns an updated cookie. Cookie is set for ``req``.

# Checklist:

- [-] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)

### DIFF
--- a/packages/qwik-auth/src/index.ts
+++ b/packages/qwik-auth/src/index.ts
@@ -12,7 +12,7 @@ import {
 import { isServer } from '@builder.io/qwik/build';
 import { parseString, splitCookiesString } from 'set-cookie-parser';
 
-export type GetSessionResult = Promise<Session | null>;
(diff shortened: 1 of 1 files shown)


zanettin:
Thx a lot @aliyss for creating this PR 🙏 
@ulic75 would you have a minute to test this? 🙏

mhevery (approved):

PR #4961, opened 2023-08-17
docs(qwikcity): html attributes docs page

bab2683:
added docs page to easily find how to add html attributes in a qwikcity app

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Added docs page to easily find how to add HTML attributes in Qwik City as currently is not as straightforward as it may seem. Added also a `vscode` extension to the recommendations to spot common typos.
While trying to setup the environment, `pnpm` required a node version of at least 18 so I updated that bit as well

# Checklist:

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -7,7 +7,8 @@
     "rust-lang.rust-analyzer",
     "ms-azuretools.vscode-docker",
     "manucorporat.vermoji",
-    "vadimcn.vscode-lldb"
+    "vadimcn.vscode-lldb",
+    "streetsidesoftware.code-spell-checker"
   ],
   "unwantedRecommendations": []
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,7 +120,7 @@ If you're not able to use the dev container, follow these instructions:
 
 1. Make sure [Rust](https://www.rust-lang.org/tools/install) is installed.
 2. Install [wasm-pack](https://rustwasm.github.io/wasm-pack/installer/) with `cargo install wasm-pack` .
-3. Node version >= `16.8.0`.
+3. Node version >= `18`.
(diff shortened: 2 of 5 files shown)


.vscode/extensions.json
```diff
@@ -7,7 +7,8 @@
     "rust-lang.rust-analyzer",
     "ms-azuretools.vscode-docker",
     "manucorporat.vermoji",
-    "vadimcn.vscode-lldb"
+    "vadimcn.vscode-lldb",
+    "streetsidesoftware.code-spell-checker"
```
wmertens, line 11: Is this intentional?

CONTRIBUTING.md
```diff
@@ -120,7 +120,7 @@ If you're not able to use the dev container, follow these instructions:
 
 1. Make sure [Rust](https://www.rust-lang.org/tools/install) is installed.
 2. Install [wasm-pack](https://rustwasm.github.io/wasm-pack/installer/) with `cargo install wasm-pack` .
-3. Node version >= `16.8.0`.
+3. Node version >= `18`.
```
wmertens, line 123: The main package.json still allows v16.8

.vscode/extensions.json
bab2683: yes, I wrote it in the description of the PR, this extension spot grammatical errors, and it even supports a range of cases so it's useful even when writing actual code and not  just documentation

CONTRIBUTING.md
bab2683: I changed that section because that was the one I was using, following step by step, when running `pnpm install`, it threw an error saying I needed 18.
I installed `pnpm` with `brew install pnpm`, so I guess it installs the latest. If we need a specific version of `pnpm` that actually supports `Node version >= 16.8.0`, I guess it would be a good idea to specify it as well

PR #4967, opened 2023-08-18
fix(qwik-city): cleanup matchRouteRequest

gioboa:
# Overview

Polish matchRouteRequest method and fix typo in a comment.

# Checklist:

### DIFF
--- a/packages/qwik-city/buildtime/vite/dev-server.ts
+++ b/packages/qwik-city/buildtime/vite/dev-server.ts
@@ -35,20 +35,11 @@ import { matchRoute } from 'packages/qwik-city/runtime/src/route-matcher';
 
 export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
   const matchRouteRequest = (pathname: string) => {
+    pathname += ctx.opts.trailingSlash && !pathname.endsWith('/') ? '/' : '';
     for (const route of ctx.routes) {
       const params = matchRoute(route.pathname, pathname);
       if (params) {
         return { route, params };
-      }
-    }
-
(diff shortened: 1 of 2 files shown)


packages/qwik-city/buildtime/vite/dev-server.ts
```diff
@@ -35,23 +35,14 @@ import { matchRoute } from 'packages/qwik-city/runtime/src/route-matcher';
 
 export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
   const matchRouteRequest = (pathname: string) => {
+    pathname += ctx.opts.trailingSlash && !pathname.endsWith('/') ? '/' : '';
```
mhevery, line 38: I am not convinced that this is the same behavior. 
I think the old one checked it twite as it checked the URLs both with and without the trailing slash. The new one does one or the other.

gioboa: You are right, if the behavior is to check with or without then this code is not correct. Anyway [trailingSlash](https://qwik.builder.io/docs/advanced/vite/#trailingslash) is a global vite option, do we need to check both cases?

gioboa: Btw Array.find() is another improvement

mhevery: I would leave the original behavior. The cost of checking the URLs is not that much.

gioboa: With this code I removed one for loop and the behaviour remains the same

mhevery (approved):

PR #4970, opened 2023-08-18
docs: Improved media, fixed height thumbnails

the-r3aper7:
# Overview

Before:
image
After:
image

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/(ecosystem)/media/index.tsx
+++ b/packages/docs/src/routes/(ecosystem)/media/index.tsx
@@ -45,7 +45,6 @@ export const MEDIA = mediaObj({
     youtube('Qwik en 2 horas', 'FF3D4tppyag', {
       language: 'es',
       promoted: true,
-      country: '🇪🇸',
       author: 'Leifer Mendez',
     }),
     {
@@ -53,7 +52,6 @@ export const MEDIA = mediaObj({
       imgSrc:
         'https://import.cdn.thinkific.com/643563/courses/2285840/hD0Nof2mTK2cVOWvRJg2_QWIK-COVER-CURSO.jpg',
       title: 'Fernando Herrera + DevTalles',
-      country: '🇪🇸',
       language: 'es',
       author: 'Dev Talles',
       promoted: true,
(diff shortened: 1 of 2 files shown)


mhevery (approved):

packages/docs/src/routes/(ecosystem)/media/index.tsx
```diff
@@ -391,22 +517,44 @@ export const MEDIA = mediaObj({
 export interface MediaEntry {
   title: string;
   href: string;
+  author?: string;
   language?: string;
+  country?: string;
```
zanettin, line 522: i am not sure if a country has an impact when a user wants to view a video. what was the idea behind this?

zanettin (commented):
thx for the improvement @the-r3aper7 🙏 looks nice! just left a question regarding the country you've added. thx again 🙏

packages/docs/src/routes/(ecosystem)/media/index.tsx
the-r3aper7: i don't know which variable name should i put because some of the video title was prefixed with country flag so i assumed it would be better and it shows the language's country

zanettin: thx a lot for the quick reply 🙏 imo just showing the language would be sufficient. the goal of the flag in front of the title is just to indicate in which language smth was recorded 👼  would you mind to remove it again?

the-r3aper7: should i remove the country the type from the MediaObj Interface, put the flag in front of title and i will put the language only.

zanettin: imo just having the language badge would be sufficient. guess the flag could be left out then 🙏 thx again

zanettin (approved):
love it ❤️  thx a lot 🙏

PR #4971, opened 2023-08-19
feat(insights): manage routes and timeline metrics

gioboa:
# Overview

### current route

This PR read the container `q:route` attribute and send the information to the database.
Insights will receive this information and will save them in the routes table.

### timeline

We are collecting `timeline` info for every symbol.
This is the number of ms between the `q:route` attribute change and the `qsymbol` custom event is delivered.
    
MOTIVATION:
The motivation for knowing the current route is for the Insights
to be able to collect it so that we can have a better understanding of
which symbols are needed for each route.

# What is it?

- [X] Feature / enhancement

# Checklist:

### Questions

- is `timeline` a good name for this metric?
- is `timeline` value correct for what we are trying to measure?


packages/insights/src/db/query.ts
```diff
@@ -167,3 +167,40 @@ export async function updateEdge(
     await db.insert(edgeTable).values(edgeRow).run();
   }
 }
+
+export async function updateRoutes(
```
mhevery, line 171: look at `updateEdge` that function is much more efficient as it only needs a single SQL to achieve the goal rather than your function requiring two.

packages/insights/src/db/schema.ts
```diff
@@ -194,3 +194,24 @@ export const edgeTable = sqliteTable(
...
+export const routesTable = sqliteTable(
+  'routes',
+  {
+    id: integer('id').primaryKey(),
+    publicApiKey: text('public_api_key').references(() => applicationTable.publicApiKey),
+    manifestHash: text('manifest_hash').notNull(),
+    route: text('route').notNull(),
+    symbol: text('symbol').notNull(),
+    interactions: integer('interactions').notNull(),
+    timeline: integer('timeline').notNull(),
```
mhevery, line 207: This `timeline` needs to be expanded to `timeline00`, ... `timeline49`

packages/qwik-labs/src/insights/index.tsx
```diff
@@ -164,13 +170,26 @@ function symbolTracker(
...
+  let qRouteChangeTime = new Date().getTime();
+  const qRouteEl = document.querySelector('[q\\:route]');
+  if (qRouteEl) {
+    const observer = new MutationObserver((mutations) => {
+      const mutation = mutations.find((m) => m.attributeName === 'q:route');
+      if (mutation) {
+        qRouteChangeTime = new Date().getTime();
+      }
+    });
+    observer.observe(qRouteEl, { attributes: true });
```
mhevery, line 182: So this logic is missing something. When `q:route` changes we must flush the data, because the `route` is shared on `payload` and if we don't flush we will attribute the symbols to the wrong route. 

So a `q:route` change must flush the current symbols and start a new queue.

gioboa: I will add the route in the symbol so we can avoid extra logic

packages/insights/src/db/query.ts
gioboa: 🤔 I looked at the updateEdge implementation but in this case I need to update interactions based on what is in the database.

packages/qwik-labs/src/insights/index.tsx
gioboa: Fixed with [2ed9ae3](https://github.com/BuilderIO/qwik/pull/4971/commits/2ed9ae3f265344979d96200848c36e7c98ae7043)

packages/insights/src/db/schema.ts
gioboa: Fixed [2c1794e](https://github.com/BuilderIO/qwik/pull/4971/commits/2c1794e964d4a0127e88140005e89b3293e2f161)

packages/insights/src/db/query.ts
```diff
@@ -167,3 +167,40 @@ export async function updateEdge(
...
+  db: AppDatabase,
+  row: {
+    publicApiKey: string;
+    manifestHash: string;
+    route: string;
+    symbol: string;
+    timeline: number;
```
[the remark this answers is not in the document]
gioboa: I created 50 cols, which scale do you think is better for our purpose?

packages/insights/src/db/query.ts
gioboa: FIxed [e739e8f](https://github.com/BuilderIO/qwik/pull/4971/commits/e739e8f10d15eb6a0a8763d5025210182f07b4a5)

packages/insights/src/db/query-helpers.ts
```diff
@@ -1,5 +1,5 @@
 import { sql, type InferModel } from 'drizzle-orm';
-import { edgeTable } from './schema';
+import { edgeTable, routesTable } from './schema';
```
mhevery, line 2:
```suggestion
import { edgeTable, type routesTable } from './schema';
```

packages/insights/src/stats/vector.ts
```diff
@@ -8,29 +8,39 @@ export const NUMBER_OF_BUCKETS = 50;
  * result in the last bucket.
  */
 export const MAX_VALUE = 10000;
+export const TIMELINE_MAX_VALUE = 60000 * 15;
```
mhevery, line 11: can you change this to make it easier to read please?

```suggestion
export const TIMELINE_MAX_VALUE = 15 * 60 * 1000;
```

mhevery (approved):

PR #4985, opened 2023-08-21
fix: only show qwik eslint errors in dev mode

jessezhang91:
# Overview

See https://github.com/BuilderIO/qwik/issues/4981 for more details.

The dev-mode linter is reporting warnings for missing rules due to ESLint ignore comment directives.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The implemented solution is to only report ESLint rules with rule id beginning with `qwik/` aka just the `eslint-qwik-plugin` rules.

From my understanding of the original issue https://github.com/BuilderIO/qwik/issues/4858, the goal of PR https://github.com/BuilderIO/qwik/pull/4873 is to show the `qwik/valid-lexical-scope` errors more visibly during dev mode to assist in debugging and prevent confusion. Given this, I think it makes sense to only run and show the `qwik` rules. (Technically the code right now only runs the `qwik` rules as those are the only ones defined in the `createLinter` function. It just is noisy with the warnings about missing rules.)

# Use cases and why

See https://github.com/BuilderIO/qwik/issues/4981

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/eslint-plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/eslint-plugin.ts
@@ -51,6 +51,9 @@ export async function createLinter(
 
         report.forEach((file) => {
           for (const message of file.messages) {
+            if (message.ruleId != null && !message.ruleId.startsWith('qwik/')) {
+              continue;
+            }
             const err = createRollupError(file.filePath, message);
             ctx.warn(err);
           }


packages/qwik/src/optimizer/src/plugins/eslint-plugin.ts
```diff
@@ -51,6 +51,9 @@ export async function createLinter(
 
         report.forEach((file) => {
           for (const message of file.messages) {
+            if (message.ruleId != null && !message.ruleId.startsWith('qwik/')) {
```
jessezhang91, line 54: It seems `ruleId` can be `null` when it's some eslint configuration error. I assume we still want to show those.

PR #5062, opened 2023-08-30
fix: loader type

wtlin1228:
# Overview

Fix loader type for the case: `loaderQrl` returns `fail(...)` without any validator.

Related issue: https://github.com/BuilderIO/qwik/issues/3979

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

I only do `StrictUnion` when the `loaderQrl` returns `fail(...)` to keep the most simple case's type the same.

```ts
// Loader<number>
export const useLoader1 = routeLoader$(() => 123); 

// Loader<{
//     foo: number;
// }>
export const useLoader2 = routeLoader$(() => ({ foo: 123 })); 
```

And for those returning `fail(...)`:

```ts
// Loader<{
//     loaderFailedReason?: string | undefined;
//     failed?: true | undefined;
//     pet?: undefined;
// } | {
//     pet?: string | undefined;
//     failed?: undefined;
//     loaderFailedReason?: undefined;
// }>
export const useRandomFailedLoader = routeLoader$(({ fail }) => {
  if (Math.random() > 0.5) {
    return fail(500, {
      loaderFailedReason: "Reach Limit",
    });
  }

  return {
    pet: "guinea pig",
  };
});
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/types.ts
+++ b/packages/qwik-city/runtime/src/types.ts
@@ -622,7 +622,7 @@ export interface LoaderConstructor {
   <OBJ>(
     loaderFn: (event: RequestEventLoader) => ValueOrPromise<OBJ>,
     options?: LoaderOptions
-  ): Loader<OBJ>;
+  ): Loader<[Extract<OBJ, Failed>] extends [never] ? OBJ : StrictUnion<OBJ>>;
 
   // With validation
   <OBJ extends Record<string, any> | void | null, REST extends readonly DataValidator[]>(
@@ -729,12 +729,14 @@ export interface ActionStore<RETURN, INPUT, OPTIONAL extends boolean = true> {
   >;
 }
 
+type Failed = {
+  failed: true;
+};
+
 /**
  * @public
  */
-export type FailReturn<T> = T & {
-  failed: true;
-};
+export type FailReturn<T> = T & Failed;
 
 /**
  * @public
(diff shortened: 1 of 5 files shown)


packages/qwik-city/runtime/src/types.ts
```diff
@@ -729,12 +729,14 @@ export interface ActionStore<RETURN, INPUT, OPTIONAL extends boolean = true> {
   >;
 }
 
+type Failed = {
+  failed: true;
+};
+
```
wtlin1228, line 735: For reviewers: 

Should this type be exported?

mhevery (commented):
Thank you for the contribution!

I see you have an issue with reproduction, but I don't see a test. Did you forget to check it in?

wtlin1228:
> Thank you for the contribution!
> 
> I see you have an issue with reproduction, but I don't see a test. Did you forget to check it in?

Hi @mhevery, test is `starters/apps/qwikcity-test/src/routes/(common)/loaders/issue3979/index.tsx`.

NexonTreeHouse (approved):

mhevery (approved):
Thank you for your contribution. ❤️

packages/qwik-city/runtime/src/types.ts
mhevery: I think it is fine not being exported.

PR #5072, opened 2023-08-31
fix(core): Missing Object ID for captured object

### LINKED ISSUE #4979: [🐞] "Missing Object ID for captured object" error when using useComputed$ with mutable props

Qwik Runtime

### Describe the bug

I have an `<Icon>` component that modifies the `props` in a `useComputed$` and spreads the result in an `<i>` element. I expect this to work (it was working in v1.2.6) but I get a "QWIK ERROR Error: Code(27): Missing Object ID for captured object" instead.

I'm not sure what the exact problem is but I've managed to replicate it in a simpler form that doesn't represent the actual real-world use case. From what I can tell the replication requires: (1) mapping over an array, (2) accessing the entire `props` object (i.e. not just a field of props) in `useComputed$`, and (3) mutating a value that gets put into the component's `props`.

### Reproduction

https://stackblitz.com/edit/qwik-starter-t3qhf8?file=package.json,src%2Froutes%2Findex.tsx

### Steps to reproduce

1. npm install && npm run dev

It immediately fails but if it doesn't I've noticed refreshing the page will cause the error to be thrown.

### System Info

```shell
System:
    OS: macOS 13.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.29 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
    npm: 8.19.3 - ~/.nvm/versions/node/v18.13.0/bin/npm
  Browsers:
    Chrome: 116.0.5845.96
    Firefox Developer Edition: 116.0
    Safari: 16.6
  npmPackages:
    @builder.io/qwik: ^1.2.7 => 1.2.7
    @builder.io/qwik-city: ^1.2.7 => 1.2.7
    undici: ^5.23.0 => 5.23.0
    vite: 4.4.9 => 4.4.9
```

### Additional Information

I believe this is related to https://github.com/BuilderIO/qwik/pull/4872. My code was working in v1.2.6.

mhevery:
Fix #4979

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/container/pause.ts
+++ b/packages/qwik/src/core/container/pause.ts
@@ -400,6 +400,14 @@ export const _pauseFromContexts = async (
     return key;
   };
 
+  const getOrCreateObjId = (obj: any): string => {
+    let key = getObjId(obj);
+    if (key === null) {
+      objToId.set(obj, (key = String(objToId.size)));
+    }
+    return key;
+  };
+
   // Compute subscriptions
   const subsMap = new Map<any, (Subscriptions | number)[]>();
   for (const obj of objs) {
@@ -535,7 +543,7 @@ export const _pauseFromContexts = async (
       if (elementCaptured) {
         assertDefined(renderQrl, 'renderQrl must be defined');
         const propsId = getObjId(props);
-        metaValue.h = mustGetObjId(renderQrl) + (propsId ? ' ' + propsId : '');
+        metaValue.h = getOrCreateObjId(renderQrl) + (propsId ? ' ' + propsId : '');
         add = true;
       } else {
         const propsId = getObjId(props);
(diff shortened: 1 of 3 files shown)


manucorporat:
Seems like the app got into a bad state in the first place, no? looking into it more

mhevery:
This is not the right fix.
- tries to fix regression introduced by https://github.com/BuilderIO/qwik/pull/4872, which in turn tried to fix https://github.com/BuilderIO/qwik/pull/4872
- We could revert the regression but given that it was so long ego, I think it is better just to fix this correctly. So leaving for @manucorporat to have a look.

PatrickJS:
it looks like this is already fixed

Issue #5065, opened 2023-08-31
[✨] Better either/or typings for route action values


kfranqueiro:
### Is your feature request related to a problem?

I've been trying out Qwik and really like some of its ideas, but have run into a significant developer ergonomics issue when using `routeAction$`.

It seems that when working with an action's resolved value, the return typing is forcibly converted to a partial, even when the callback would otherwise produce an object with required properties.

For example, if I write the following:

```ts
export const useFoo = routeAction$(async () => ({ foo: 42 }));
```

I get a typing like `{ foo?: number | undefined }` even though foo is definitely always defined as returned by my function.

In the case where a zod schema is also passed, I receive a similarly-partial typing:

```ts
{
  foo?: number | undefined;
  formErrors?: undefined;
  fieldErrors?: undefined;
  failed?: undefined;
} | {
  foo?: undefined;
  formErrors?: string[] | undefined;
  fieldErrors: { ... } | undefined;
  failed?: true | undefined;
}
```

This makes it onerous to pick out individual properties on successful values, requiring type assertions or redundant conditions for every individual property accessed in order to appease tsc, even though IIUC we should know which side of the union we're on as soon as we note the presence/absence of any one of our properties (or `failed`).

### Describe the solution you'd like

In the no-schema case, I would expect exactly the type my function returns: `{ foo: number }`.

In the case of passing a zod schema, I would expect either the type my function returns, or a type with only the failure-related properties, ideally definitely defined (though I'm unsure if there are implementation details that would preclude that):

```ts
{
  foo: number
} | {
  formErrors: string[];
  fieldErrors: { ... };
  failed: true;
}
```

This should make it trivial to access any number of properties on the return value within a single `if (!value.failed)` check, with no need for type assertions, or redundant conditions.

### Describe alternatives you've considered

As mentioned above, currently I'd need to work around this by adding type assertions or redundant conditions. Alternatively, I'm otherwise contemplating switching to using `server$` for my particular use cases at the moment, but I'd lose out on built-in zod support.

### Additional context

The oops-all-partials typings appear to originate in `StrictUnion`, which uses `Prettify` and `StrictUnionHelper`, both of which seem to involve use of `Partial`. I'm not sure what the goal was in building `StrictUnion` this way, or if there's some reason/limitation for the current typings conundrum.

`StrictUnion` also appears to be used in `routeLoader$` typings specifically when multiple arguments are passed, which I'm not sure is even documented?

wtlin1228:
[...]

PR #5079, opened 2023-08-31
fix(qwik-city): Improve routeAction types

mhevery:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/types.ts
+++ b/packages/qwik-city/runtime/src/types.ts
@@ -383,7 +383,7 @@ type StrictUnionHelper<T, TAll> = T extends any
 type StrictUnion<T> = Prettify<StrictUnionHelper<T, T>>;
 
 type Prettify<T> = {} & {
-  [K in keyof T]?: T[K];
+  [K in keyof T]: T[K];
 };
 
 /**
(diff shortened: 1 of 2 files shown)


mhevery:
@manucorporat This change breaks a lot of existing code, so I don't think we can merge it. But I agree with #5065 that the current situation is not good, and I don't know how to fix it in a non-breaking way.

kfranqueiro:
After taking a whack at this myself in #5113 then stepping away for a while, it occured to me that if you make just the change of removing the `?` within `Prettify`'s index type, with no other change to typings, I think this will get close enough to what I was hoping for in #5065 without breaking things. It would still accomplish ensuring that `failed` is always truthy on failure, and the result type remains intact on success.

With this suggestion you would also need to change the `if ("failed" in zodValue)` in the new test file to `if (zodValue.failed)` - which I'm guessing is more in line with existing usage patterns anyway and related to the build passing.

mhevery:
> d also need to change the `if ("failed" in zodValue)` in the new test file to `if (zodValue.failed)` - which I'm guessing is more in line with existing usage patterns anyway and related to the build passing.
> 
> Thanks for working on this so soon after I opened it!

I don't think that is correct either. So, at this point, I think this change is dead as it is very breaking... So I am going to close it.

Issue #5087, opened 2023-09-02
[✨] Allow binding `this` to the store in useStore


wmertens:
### Is your feature request related to a problem?

It would be nice to have have functions in stores be bound to the store object.

### Describe the solution you'd like

I think a `bind` option to useStore would be nice. It could even be `true` by default because the majority use case for functions in the store is using the store object itself.

This can be used orthogonally with `deep` and `reactive`.

```tsx
import { component$, useStore, $, useTask$ } from '@builder.io/qwik';

export default component$(() => {
  const store = useStore({
    count: 1,
    inc: $(function () {console.log('I have a this:', !!this);this.count++})
  }, {bind: true})

  return <p>Count: {store.count} <button onClick$={store.inc}>one up</button></p>;
});
```

### Describe alternatives you've considered

letting the deserializer do it, but then the function is only bound when it's deserialized.

### Additional context

This allows making zustand-like APIs.

justoverclockl:
pr opened, hope it helps:

https://github.com/BuilderIO/qwik/pull/5088

mhevery:
I assume this workaround is too wordy

```typescript
import { component$, useStore, $, useTask$ } from '@builder.io/qwik';

export default component$(() => {
  const store = useStore({
    count: 1,
    inc: null!
  })
  store.inc = $(function () {store.count++})

  return <p>Count: {store.count} <button onClick$={store.inc}>one up</button></p>;
});
```

Wondering if something like this would work:

```typescript
import { component$, useStore, $, useTask$ } from '@builder.io/qwik';

export default component$(() => {
  const store = useStore({
    count: 1,
    inc: storeMethod$(function () {store.count++})
  })

  return <p>Count: {store.count} <button onClick$={store.inc}>one up</button></p>;
});
```

The semantics of `method$` would be such that it would ensure that `store.inc` would pre-bind it to the `store`.

wmertens:
I found how to do it!

https://qwik.builder.io/playground/#v=1.1.5&buildMode=production&f=Q0o0xobG2BKNjgIQwWpdilMQJDjS8vOBoVENbhlYgYuuWqAcUFQP0liwhVuoAdIE0gYs9q0UDHXAvMy8ZCsFkKkgQ%2BG69MCKtLVrNYGKwATEsmqwuA5IVy3QZLh6LpQE7QyxAaK4FhiY4FYOonKvBmm3A%2FoHWFnC20CjaX8op30A

```tsx
[...]

PR #5088, opened 2023-09-02
enhancement: Allow binding this to the store in useStore

justoverclockl:
# What is it?
Allow binding this to the store in useStore
https://github.com/BuilderIO/qwik/issues/5087

- [x ] Feature / enhancement
- [ ] Bug

# Description

In this modified code, I've added a bind option to the opts parameter, and by default, it's set to true. When the bind option is true, the bind method is called on the newStore object to ensure that the this context is bound to the store object.

You can now use the bind option to control whether the this context should be bound when using the useStore function, and it can be used in combination with the deep and reactive options as needed.

# Checklist:

- [x ] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [x ] I have performed a self-review of my own code

### DIFF
--- a/packages/qwik/src/core/use/use-store.public.ts
+++ b/packages/qwik/src/core/use/use-store.public.ts
@@ -110,18 +110,18 @@ export const useStore = <STATE extends object>(
     const newStore = getOrCreateProxy(value, containerState, flags);
 
     const bindFunctionsRecursively = (obj: any) => {
-      for (const key in obj) {
-        if (typeof obj[key] === 'function') {
-          obj[key] = obj[key].bind(newStore);
-        } else if (typeof obj[key] === 'object' && obj[key] !== null) {
-          bindFunctionsRecursively(obj[key]);
+      for (const [key, value] of Object.entries(obj)) {
+        if (typeof value === 'function') {
+          obj[key] = value.bind(newStore);
+        } else if (typeof value === 'object' && value !== null) {
+          bindFunctionsRecursively(value);
         }
       }
     };
 
     bindFunctionsRecursively(newStore);
     set(newStore);
-    
+
     if (opts?.bind !== false) {
       return newStore as STATE;
     }
--- a/packages/qwik/src/core/use/use-store.public.ts
+++ b/packages/qwik/src/core/use/use-store.public.ts
@@ -100,15 +100,16 @@ export const useStore = <STATE extends object>(
     return get;
   }
   const value = isFunction(initialState) ? invoke(undefined, initialState) : initialState;
-  if (opts?.reactive === false) {
-    set(value);
-    return value;
-  } else {
-    const containerState = iCtx.$renderCtx$.$static$.$containerState$;
-    const recursive = opts?.deep ?? true;
-    const flags = recursive ? QObjectRecursive : 0;
-    const newStore = getOrCreateProxy(value, containerState, flags);
 
+  if (!opts?.reactive || !opts?.bind) {
+    return value as STATE;
+  }
+
+  const containerState = iCtx.$renderCtx$.$static$.$containerState$;
+  const flags = opts?.deep ? QObjectRecursive : 0;
+  const newStore = getOrCreateProxy(value, containerState, flags);
+
+  if (opts?.bind) {
     const bindFunctionsRecursively = (obj: any) => {
       for (const [key, value] of Object.entries(obj)) {
         if (typeof value === 'function') {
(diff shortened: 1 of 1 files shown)


wmertens:
Hi, this doesn't actually implement the binding of functions when the getter for them fires.

justoverclockl:
sorry, forgot to push

wmertens:
Hmm that just seems to implement the current behavior in a slightly different way.

But it looks like the code was already supposed to work. Perhaps you can start by adding a failing test? In my example code, store.inc should be able to be copied to a var and calling it that way should still make `this` be the store object.

wmertens:
Actually it looks like you're close but it doesn't work when reactive is false, and it always binds even when bind is false, and I'm not sure if doing it like that won't cause it to register listeners. Perhaps this should be done in the proxy wrapper

justoverclockl:
> Actually it looks like you're close but it doesn't work when reactive is false, and it always binds even when bind is false, and I'm not sure if doing it like that won't cause it to register listeners. Perhaps this should be done in the proxy wrapper

you mean something like

```javascript
if (bind) {
    const proxiedStore = new Proxy(newStore, {
      get(target, prop, receiver) {
        const value = Reflect.get(target, prop, receiver);
        if (typeof value === 'function') {
          return value.bind(target);
        }
        return value;
      },
    });
    set(proxiedStore);
    return proxiedStore as STATE;
  } else if (reactive) {
    set(newStore);
  }
```

devcaeg:
It would be great if this functionality could be implemented. But I don't see the need for it to be activated through an option, be it "bind" or other.

packages/qwik/src/core/use/use-store.public.ts
```diff
@@ -93,15 +100,29 @@ export const useStore = <STATE extends object>(
...
-  } else {
-    const containerState = iCtx.$renderCtx$.$static$.$containerState$;
-    const recursive = opts?.deep ?? true;
-    const flags = recursive ? QObjectRecursive : 0;
-    const newStore = getOrCreateProxy(value, containerState, flags);
-    set(newStore);
-    return newStore;
+
+  if (!opts?.reactive || !opts?.bind) {
+    return value as STATE;
```
wmertens, line 105: you're not binding here

packages/qwik/src/core/use/use-store.public.ts
```diff
@@ -93,15 +100,29 @@ export const useStore = <STATE extends object>(
...
+      for (const [key, value] of Object.entries(obj)) {
+        if (typeof value === 'function') {
+          obj[key] = value.bind(newStore);
+        } else if (typeof value === 'object' && value !== null) {
+          bindFunctionsRecursively(value);
+        }
+      }
+    };
+
+    bindFunctionsRecursively(newStore);
```
wmertens, line 123: By calling this here, you're triggering all the listeners. You should implement this binding in `getOrCreateProxy`

packages/qwik/src/core/use/use-store.public.ts
```diff
@@ -19,6 +19,13 @@ export interface UseStoreOptions {
    * Default is `true`.
    */
   reactive?: boolean;
+
+  /**
+   * If `true` then bind method is called on the `newStore` object to ensure
+   * that the `this` context is bound to the store object.
+   * Default is `true`.
+   */
+  bind?: boolean;
```
mhevery, line 28: I don't think this is the right approach:

1. Store should not be aware of binding
2. Even if Store does the bind, it will not survive serialization, since serialization does not understand binding.

mhevery (changes requested):

mhevery:
I don't think this is the right approach:

1 `useStore` should not be aware of `binding` of functions. Functions should be `bindable` independently of store. That is how it works in regular JS. When you tear of a function you need to understand that `this` is not coming along. In #5087 when `<button onClick$={store.inc}>` the `inc` function is torn from `store` hence no `this`.
2. `useStore` can't be the right thing as people can move a function between stores, and then they will really be surprised by the `this` behavior. 
3. The way this is solved in JS is that we use `.bind(this)`. We should follow the same pattern here. The problem is that Qwik would need to be able to serialize the `bind(exp)` and that is not something that is currently supported. A good thing to support, but a bit of o head scratcher on how to implement it. For now we have been ignoring this, but if @justoverclockl would like to explore this approach happy to coach you through it. But expect that this would be very much an exploratory approach so it may involve some dead ends as we work through it.
  - Of the top of my head, I think the QRLs would have to understand bindings and serialize them along with other closed-over expressions. Then when the function is invoked some of the closed-over variables would need to be sent to bind to the function. ..

wmertens:
We could intercept .bind() and mark the target so we can serialize it.

wmertens:
Here's how to do it, I think the optimizer could do it automatically: https://github.com/BuilderIO/qwik/issues/5087#issuecomment-1707185010

wmertens:
I'm going to close it since the correct solution involves changes to the optimizer, not the wrapping code, and the workaround is known.

PR #5089, opened 2023-09-02
feat(optimizer): add qwikVite configuration option optimizerOptions.inlineStylesUpToBytes

### LINKED ISSUE #5069: Optimizer inlines large blobs of styles
The optimizer inlines css below a certain length/size, which is awesome for "small" stuff. I use tailwind and it adds quite a bit, but apparently not enough :)

https://github.com/BuilderIO/qwik/blob/86a9452be1572c84868a5ebee3ad417cf304b70f/packages/qwik/src/optimizer/src/plugins/vite.ts#L463

Changing the setting from 20000 to 200 makes my global.css a separate file again.
-rw-r--r--@ 1 roel  wheel  26671 Aug 31 12:17 /tmp/2e2.html
-rw-r--r--@ 1 roel  wheel  40947 Aug 31 12:17 /tmp/2e4.html

I personally could try to not use global.css/tailwind, maybe use pandacss instead. Maybe make the 20000 configurable?

Any other opinions?

jaroel:
Closes #5069

# What is it?

- [X] Feature / enhancement
- [ ] Bug

# Description

The Optimizer inlines styles up to 2e4 bytes. This PR makes that value configurable using:
```
qwikVite({
  optimizerOptions: {
    inlineStylesUpToBytes: 400,
  },
}),
```

Feedback needed:
1. is `inlineStylesUpToBytes` a good enough name?
2. is placing it inside `optimizerOptions` the correct scope?

# Use cases and why

- 1. Inlining 20k bytes of css might be invalid for certain use cases.

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/plugin.ts
@@ -267,7 +267,7 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
     }
     opts.csr = !!updatedOpts.csr;
 
-    opts.inlineStylesUpToBytes = optimizerOptions.inlineStylesUpToBytes ?? 0;
+    opts.inlineStylesUpToBytes = optimizerOptions.inlineStylesUpToBytes ?? 20000;
     if (typeof opts.inlineStylesUpToBytes !== 'number' || opts.inlineStylesUpToBytes < 0) {
       opts.inlineStylesUpToBytes = 0;
     }
@@ -902,6 +902,11 @@ export interface QwikPluginOptions {
     | ((transformedModules: TransformModule[]) => Promise<void> | void)
     | null;
   devTools?: QwikPluginDevTools;
+  /**
+   * Inline styles up to a certain size (in bytes) instead of using a separate file.
+   *
+   * Default: 20kb (20,000bytes)
+   */
   inlineStylesUpToBytes?: number;
 }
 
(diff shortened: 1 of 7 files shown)


jaroel:
This works. Could use some docs and some tests.

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -266,6 +267,11 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
     }
     opts.csr = !!updatedOpts.csr;
 
+    opts.inlineStylesUpToBytes = optimizerOptions.inlineStylesUpToBytes ?? 0;
```
jaroel, line 270: We should add a default value here.

jaroel: Maybe having the _option_ to inline is worth more than inlining by default. Especially when an external .css file can be hashed and cached forever.

mhevery: I think we should keep the previous behavior unless it is overwritten...

mhevery (changes requested):

mhevery (approved):
Added my changes

mhevery:
Thank you for your contribution ❤️

jaroel:
Thanks for the fixup, nice touch with the docstring!

PR #5103, opened 2023-09-04
chore: release qwik-auth 0.1.1

ulic75:
There have been a few changes to qwik-auth that are included in the latest release (0.1.0).  I'm not completely sure if merging this would cause a new package release or not.  If not some guidance or something about releasing a new version would be nice.


ulic75:
> Let me know when it it is ready and I will cut a release...

I think it should be ready.  I just wasn't sure the process.

mhevery:
Published as v0.1.1

Issue #4835, opened 2023-07-20
[✨] qwik-city: routes should allow aliases for i18n


wmertens:
### Is your feature request related to a problem?

When doing i18n, SEO people like it when URLs are also translated. This is the same as aliasing paths.

### Describe the solution you'd like

Allow comma separated names as a route path, for example `/[...locale]/blog,blogue,blog/[id]`

### Describe alternatives you've considered

`|` is also a good candidate but it can be problematic on the command line. Normally, people don't put commas in paths so that should be safe.

claudioshiver:
I added route rewriting in this PR, hope it will be merged soon: #5122

mhevery:
Resolved by #5122

PR #5122, opened 2023-09-07
feat: Add route rewriting

claudioshiver:
# What is it?

- [X] Feature / enhancement
- [ ] Bug

# Description

Now you can rewrite your urls using the same page component, layout and middlewares. 
You can use a prefix to localize your routes or you can simply rewrite urls without prefix.

# Use cases and why

- 1. **Translate localized urls with prefix**: for localization purposes you may want to translate your routes from `/products` to `/it/prodotti` or to `/fr/produits` and `/products/product-name` to `/it/prodotti/nome-prodotto` or `/fr/produits/nom-du-produit` without having multiple routes files for each locale but reusing the same page component, layouts, middlewares and so on. parameters name will not be changed so if the routePath is `/products/[slug]` you will receive `product-name`, `nome-prodotto` or `nom-du-produit` as the values for the same `slug` path parameter.
- 2. **Rewrite urls without prefix**: it's rare but you may want to have aliases for the same path as stated in the issue below. for example you may want both `/docs` and `/documents` to be rendered from the same page component.

# Checklist:

# Issues

This should close the issue #4835

### DIFF
--- a/packages/qwik-city/buildtime/build.ts
+++ b/packages/qwik-city/buildtime/build.ts
@@ -31,6 +31,7 @@ export async function updateBuildContext(ctx: BuildContext) {
         })
         .then((sourceFiles) => {
           const resolved = resolveSourceFiles(ctx.opts, sourceFiles);
+          rewriteRoutes(ctx, resolved);
           ctx.layouts = resolved.layouts;
           ctx.routes = resolved.routes;
           ctx.entries = resolved.entries;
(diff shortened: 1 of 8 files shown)


packages/qwik-city/buildtime/build.ts
```diff
@@ -46,6 +47,70 @@ export async function updateBuildContext(ctx: BuildContext) {
   return ctx.activeBuild;
 }
 
+function rewriteRoutes(ctx: BuildContext, resolved: ReturnType<typeof resolveSourceFiles>) {
+  if (ctx.opts.rewriteRoutes) {
+    for (let rewriteIndex = 0; rewriteIndex < ctx.opts.rewriteRoutes.length; rewriteIndex++) {
```
gioboa, line 52: I prefer forEach because it's more readable, anyway it looks good to me.

claudioshiver: I can change it if you prefer

mhevery:
Yes, thank you for this awesome work! My initial reaction is that I like it. I am going to sit on it for a few hours to better internalize it.

mhevery (approved):

PR #5130, opened 2023-09-08
bug(vite): populate rollup cache after transform

wmertens:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

I think this might fix what was wrong with the playground.

The Qwik optimizer transform will create new QRL imports and stores them in memory until they are loaded by rollup/vite.

If a subsequent run tries to load a QRL, it won't be in the transform cache.

This change hopefully loads the code into the cache, but I wasn't able to actually test this.

I also wonder how rollup knows that a QRL  has changed when the source file has changed.

# Checklist:


mhevery (approved):
You rock!

PR #5168, opened 2023-09-15
fix(qwik-city): Fix rewrite home route

claudioshiver:
# What is it?

- [ ] Feature / enhancement
- [X] Bug

# Description

The home route is not translated if rewriteRoutes config is provided. 
If there is a prefix in a rewriteRoute item the home page as well needs to be translated.

# Use cases and why

- Actual: I need to translate the homepage to `/it-IT/` but the corresponding route is not created.
- Expected: If i add a rewriteRoute with the prefix `it-IT` i want the `/it-IT/` route to be created.

# Checklist:


mhevery (approved):
Thank you for being part of OSS and contributing to Qwik. ❤️

PR #5172, opened 2023-09-15
fix(qwik-city): do not write into a destroyed stream

tuurbo:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Check if the response is already destroyed similar to checking if the response is closed.

PR https://github.com/BuilderIO/qwik/pull/5074 fixed some of the `Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed` errors but sometimes `closed` is undefined and `destroyed` is true.

If checking for `destroyed`, I don't know if that makes checking for `closed` redundant so I kept the `closed` check.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/node/http.ts
+++ b/packages/qwik-city/middleware/node/http.ts
@@ -96,8 +96,8 @@ export async function fromNodeHttp(
       }
       return new WritableStream<Uint8Array>({
         write(chunk) {
-          if (res.closed) {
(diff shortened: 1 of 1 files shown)


tuurbo:
Not sure if this fix is enough. I've been testing this PR in production and now i'm seeing error
```
```

mhevery (approved):

Issue #5175, opened 2023-09-15
[🐞] Qwik serializer could fail on binary string


genki:
### Which component is affected?

Qwik Runtime

### Describe the bug

I am writing an app that receives a random binary string generated in server side.
The random string is generated like this:
```ts
const getRandomString = () => {
  const random = new Uint8Array(16);
  crypto.getRandomValues(random);
  return String.fromCharCode.apply(null, random);
};
```
It is usually works fine, but sometimes, I found the Qwik serializer fails on the serialization of that binary string.
Here is the error log:
```
Uncaught (in promise) SyntaxError: Cannot convert ;,ëÆ¹L¥yçRÂ«8âc;E©vÚõüx* to a BigInt (at serializers.js:344:16)
    at BigInt (<anonymous>)
    at Object.$prepare$ (serializers.js:344:16)
    at Object.prepare (serializers.js:463:37)
    at _deserializeData (resume.js:65:72)
    at serverQrl_stuff_wOIPfiQ04l4 (index.qwik.mjs:1248:29)
    at async qrl (qrl-class.js:23:24)
    at async routes_component__Fragment_button_onClick_ahXys9A4F8E (index.tsx:19:36)
    at async dispatch (?qfunc=fYK4WiVJmoo:96:29)
    at async HTMLDocument.processDocumentEvent (?qfunc=fYK4WiVJmoo:113:17)
```

I guess the serializer has a bug on the treatment of the binary string.

### Reproduction

https://github.com/genki/qwik-test/tree/bad_serialize

### Steps to reproduce

I made a reproduction app at the link above.
Please clone the repo and `pnpm run dev` then click the button to reproduce this issue.

### System Info

```shell
System:
    OS: macOS 13.5.2
    CPU: (8) arm64 Apple M2
    Memory: 91.97 MB / 24.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.5.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.8.1 - /opt/homebrew/bin/npm
    pnpm: 8.7.5 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 116.0.5845.187
    Safari: 16.6
  npmPackages:
    @builder.io/qwik: 1.2.11 => 1.2.11 
    @builder.io/qwik-city: latest => 1.2.11 
[...]

PR #5176, opened 2023-09-16
fix(core): Added StringSerializer for strings that start with the prefixes of serializers.

genki:
In order to solve the issue #5175 
added `NOSERIALIZE_PREFIX` to mark the binary strings that are not serialized

# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The Qwik serializer had treated all strings as possibly serialized and try to resume them even for binary strings that start with prefixes for serializer.
So I introduced the `NOSERIALIZE_PREFIX` to mark the binary strings so that the Qwik serializer can distinguish they don't need to be deserialized.

# Use cases and why

We can use binary strings correctly.

# Checklist:


mhevery (changes requested):
If I understand this PR correctly all strings will now get a special prefix.... I think this will bloat the output....

I think a better approach is to create a new serializer that will trigger on strings that start with a special character and only then on the serialization prefix. This way we are not polluting all of the existing strings in the serialization.

Do you think you could take on that?

genki:
@mhevery I see, I will try :)

genki:
I introduced the `StringSerializer` that serializes strings of which start with the prefixes of serializers.
Other normal strings are stay being themselves.

mhevery:
Thanks for the contribution. I did some "minor" cleanup...

mhevery (approved):

genki:
@mhevery Thanks a lot :)

genki:
Failed on E2E Tests? Hmm...

genki:
I have confirmed the E2E test `pnpm run test.e2e.chromium` at my local environment.

At 5d85187970d0e0d5b611f73a942c7509af2e0dc0 Passed
At d25968e638aa4b0612893254ac89986f96e9a9b4 Passed
At 4c90bf0fab95133d225bb837bfa943b27bb303c4 Failed

@mhevery Could you confirm it?

genki:
Found a issue in the `serializeObjects` so fixed it.
Now the StringSerializer tests the `UNDEFINED_PREFIX` as it must be serialized as special character.

genki:
Now passed the E2E test :)
But it still fails on the treeshaking of the `core.min.mjs`

genki:
Now all tests have passed.
@mhevery Could you confirm?

mhevery (approved):

PR #5178, opened 2023-09-16
feat: bootstrap integration adapter

mugan86:
# What is it?

- [X] Feature / enhancement
- [ ] Bug

# Description

In this functionality I want to facilitate the process to start working with Bootstrap in Qwik using the Bootstrap library.

An entire section has been added with a main page and several component examples, where they have been implemented based on the content provided by its documentation, adapting it to Qwik's features.

Everything that is needed, the basics in the configurations, is already implemented there.

If you want to know more about what each file will do, in the documentation section, an explanation of what each section does has been added.

Captura de pantalla 2023-09-16 a las 19 17 10

To execute this integration, we must have the project prepared and the latest changes compiled

Complete Compilation:
`pnpm build.full`

Create the symbolic link with the new feature (if you haven't done it before) :
`pnpm link.dist`

And creating a new Qwik project we install those dependencies of what we just added, to be able to run the script with `npm run qwik add` to show the entire list or directly execute the proposed integration with `npm run qwik add boostrap`.

When executed, we will have the following:

Captura de pantalla 2023-09-16 a las 19 21 30

Select `Yes` and result if process is correct:

Captura de pantalla 2023-09-16 a las 19 22 11

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


starters/features/bootstrap/src/components/bootstrap/navbar.tsx
```diff
@@ -0,0 +1,47 @@
+import { component$, useVisibleTask$ } from "@builder.io/qwik";
+import { Link } from "@builder.io/qwik-city";
+export const Navbar = component$(() => {
+  useVisibleTask$(async () => {
```
hamatoyogi, line 4: I think this might be less than ideal. Not sure what's the right course of action, but it seems it might be a good idea to import the entire library instance for this to work.

Thought here @gioboa @mhevery ?

anartzdev: Hi (I'm mugan86 from another account),

What do you mean by: "I'm not sure what the correct course of action is, but it seems like it might be a good idea to import the entire library instance to make this work."

If you are referring to the "bootstrap" import, this is used so that we already have the Javascript functionalities, in this case so that if we have our application in responsive, the toggle menu of that navbar can be opened or not. Without this import element, Boostrap could only be used at the style level, but not at the functionality level.

I was thinking about putting it in the src/bootstrap/layout.tsx file but I am waiting for your suggestions for improvement and accordingly.

Perhaps I have not understood you well, I am waiting for you to tell me what would be the most correct way to implement these details.

I hope you see the rest as correct and if not, obviously let me know so I can leave it as you recommend your team.

I have already been improving many details for the procedure thanks to the suggestions that @gioboa made in a previous PR ;)

Thank you so much

hamatoyogi: Hi @anartzdev !
First off, thanks for the PR! (forgot to say that before 😅 )

I was chatting with @gioboa and we are still mulling it over. What we're currently thinking is that it might be better to avoid using `useVisableTask` here and loading the bootstrap script via an HTML script tag, via the `root.tsx`.

hamatoyogi: @gioboa mentioned another option:
something like this --> https://github.com/BuilderIO/qwik/blob/b330651d33bfa9665f1619e29c831e69286c11a0/starters/apps/basic/src/components/router-head/router-head.tsx#L32
```
export const RouterHead = component$(() => {
  [...]
  return (
    <>
      [...]
      {head.scripts.map((s) => (
        <script key={s.key} {...s.props} dangerouslySetInnerHTML={s.script} />
      ))}
    </>
  );
});
```

But we both think `root.tsx` is fine.

mhevery: So `useVisibleTask$` causes code to execute eagerly, which is why we try to push back (especially in libraries) on its use...

The other point about `useVisibleTask$` is that it causes the FW to be loaded on event....

1. So the best option would be to do nothing, which sounds like is not possible in this case as we do need to load the `bootstrap`
2. If it would be used everywhere, I would say put it in global, but in this case, it is only specific to the component, so visible is appropriate. So I would use `useOn('qvisible', $(...))` as this code will nat cause the framework to load when the component becomes visible. 
3. The option of last resort is `useVisibleTask$`, which we are trying to avoid because it will cause the FW to be loaded as well...

So I think the correct thing here is to switch to `useOn('qvisible', $(...))`

PS: Thank you for the hard work of creating this PR.

starters/features/bootstrap/src/components/bootstrap/navbar.tsx
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * It is required for Javascript functions within Bootstrap components to work.
+     * For example, if we adapt to a mobile screen, so that the toggle menu button works.
+     * Without this it will NOT work (try it if you want)
```
mhevery, line 8: (maybe remove this line, as it feels a bit snarky )

starters/features/bootstrap/src/components/bootstrap/navbar.tsx
mugan86: Thank you very much for your answers and feedback. All the points you are presenting to me are really interesting and serve as a learning experience for me. The truth is that by creating some things to contribute I am learning more than by diving into the documentation.

So, before resorting to useVisibleTask$ (which should already be the "desperate" situation when there is no other option) we are going to resort to useOn('qvisible', $(...)).

Very interesting, I make that change and when I have it, I'll let you know.

Thank you very much team, to @hamatoyogi, @gioboa and @mhevery. A pleasure to contribute to this great project and I hope to continue doing so! I have some ideas in mind, which I will propose, as I learn from you :)

starters/features/bootstrap/src/components/bootstrap/navbar.tsx
mugan86: No problem @mhevery, I'll delete it. If in the future I have to make any indication, I understand that I have to indicate it in the documentation so that it is cleaner. Thank you so much! .

anartzdev:
I am having some problems, I did not realize that there was a new version and many problems have arisen, among which some code has disappeared, so I am going to choose to recreate it and as I see that there are quite a few things to stir up, the better I'm going to create a new version from what was uploaded in 1.2.12 to avoid any problems.

Today I don't think I can do it, but I think tomorrow I could dedicate enough time to do it correctly.

I've already tried `useOn` as suggested and it works perfectly.

Thank you very much and sorry for the inconvenience

anartzdev:
By the way, where is the information regarding the 'qvisible' event? I'm interested in knowing more and although I'm looking for it, I can't find anything and I'm curious to know more. Thank you

gioboa:
At the moment there is no documentation for `qvisible` event.

anartzdev:
> At the moment there is no documentation for `qvisible` event.

Well thank you very much. Good to Know. What I'm missing is unlocking the status of the conflicts. Maybe close this PR and open a new one? I already have the branch well created with version 1.2.12 but when submitting the changes, it does not detect the CI flow and is stuck in "Revolved conflicts"

anartzdev:
Do any changes need to be made? Not knowing for days, I would like to know since it may not be to accept it

mhevery:
Hey, sorry about the delay; been traveling...

I am trying to review this now, and want to push a minor fix, but it seems like you have locked me out from pushing changes to this branch. Usually the PR shows 
image
but not in your case.  I am assuming you have changed a setting someplace in your repo configuration which prevents me from sending a fix. Do you think you could open that up?

mhevery:
I have moved the PR to https://github.com/BuilderIO/qwik/pull/5222

anartzdev:
Hello good morning,

You don't have to apologize, don't worry ;) she was just asking about the status of this, nothing more.

Well, the truth is, I don't know what I touched that prevented you from making changes, apologies on my part.

I will fork the repository again and start from 0, in case I want to contribute something new, since I more or less already have some ideas. Little by little.

Thank you very much and apologies again.

Greetings!

PR #5179, opened 2023-09-16
docs: added copy button

the-r3aper7:
# Overview

# What is it?

image

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/root.tsx
+++ b/packages/docs/src/root.tsx
@@ -1,4 +1,4 @@
-import { component$, useContextProvider, useStore } from '@builder.io/qwik';
+import { $, component$, useContextProvider, useStore, useVisibleTask$ } from '@builder.io/qwik';
 import { QwikCityProvider, RouterOutlet, ServiceWorkerRegister } from '@builder.io/qwik-city';
 import RealMetricsOptimization from './components/real-metrics-optimization/real-metrics-optimization';
 import { RouterHead } from './components/router-head/router-head';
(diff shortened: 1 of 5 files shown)


the-r3aper7:
I have created a eventlistener which is listening to whole body which i think is not optimal is there any other solution to this?
however when i am trying to do `document.getElementsByName` or `document.body.getElementsByClass` none of them is showing up? 

What should i do?

packages/docs/src/root.tsx
```diff
@@ -16,6 +16,21 @@ export default component$(() => {
 
   useContextProvider(GlobalStore, store);
 
+  useVisibleTask$(({ cleanup }) => {
```
mhevery, line 19: Could we solve this without `useVisbelTask$` which forces code to run eagerly?

mhevery (changes requested):

packages/docs/src/root.tsx
the-r3aper7: i'll look into it

mhevery: I think you want `useOnDocument('click', $(...));` https://qwik.builder.io/tutorial/hooks/use-on/

gioboa:
@the-r3aper7 are you still on this PR?

the-r3aper7:
yeah i will work on it busy with college actually

gioboa:
@the-r3aper7 thanks for the fast feedback. The best solution is close this one because there is also a conflict to solve.
You can create a new PR when you are free/ready, more then welcome to merge the future PR. 💪

PR #5180, opened 2023-09-17
fix(qwik-auth): don't overwrite response headers in qwik auth

Aslemammad:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

In Authjs, we cannot return a custom error currently, the only way is to redirect to a certain URL with the error being injected in the search params. So when we try setting `location` header, it can be missed since when Authjs gives the error it overwrites that `location` header with the CallbackRouteError URL which is not useful for a product or a website.

In this PR, we try checking if we're overwriting what's in the response or no. So we skip it if the header is already there which gives control over the redirecting ability of qwik instead of it being handled by Authjs.
# Use cases and why

- 1. Making Redirecting by the `location` property possible in Qwik in the `authorize` function of the Credentials provider of Authjs

# Checklist:

### DIFF
--- a/packages/qwik-auth/src/index.ts
+++ b/packages/qwik-auth/src/index.ts
@@ -141,7 +141,7 @@ async function authAction(
     skipCSRFCheck,
   });
   res.headers.forEach((value, key) => {
-    if (!req.headers.has(key)){
+    if (!req.headers.has(key)) {
       req.headers.set(key, value);
     }
   });
--- a/packages/qwik-auth/src/index.ts
+++ b/packages/qwik-auth/src/index.ts
@@ -141,7 +141,9 @@ async function authAction(
     skipCSRFCheck,
   });
   res.headers.forEach((value, key) => {
-    req.headers.set(key, value);
+    if (!req.headers.has(key)){
+      req.headers.set(key, value);
+    }
   });
   fixCookies(req);
 


mhevery (approved):

ulic75:
I haven't had time to investigate yet, but this appears to break oauth authentication with PKCE code verifier errors

ulic75:
@Aslemammad can you may provide your authjs config so I can understand what you're trying to accomplish?

mhevery:
Let me know if this needs to be reverted...

ulic75:
It is a zero dot release, folks can pin it to 0.1.1 I suppose until we sort through it.  Upon some further research, it only appears to impact things when trying to login directly to a provider bypassing the auto-generated login page.

Aslemammad:
@ulic75 Sure, here you go:

```ts
class AuthError extends Error {}

declare module '@auth/core/types' {
  export interface Session {
    user: {
      id: number
      name: string | null
      email: string | null
      image: string | null
    }
  }
}

export const { onRequest, useAuthSession, useAuthSignin, useAuthSignout } =
  serverAuth$((req: RequestEventCommon) => ({
    callbacks: {
      session({ user, session, token}) {
        return {...session, user: {...session.user, id: token.id}}
      },
      async jwt({ token }) {
        const { db } = await createClient(req);
        const selectedUser = (
          await db
            .select()
            .from(user)
            .where(eq(user.email, token.email!))
            .execute()
        )[0];

        if (!selectedUser) {
          (
            await db
              .insert(user)
              .values({
                name: token.name!,
                email: token.email!,
                avatar: token.picture,
              })
              .returning()
              .execute()
          )[0];
        }
        return { ...token, id: selectedUser.id};
      },
    },
    secret: req.env.get("AUTH_SECRET"),
    trustHost: true,
    providers: [
      GitHub({
        id: "github",
        clientId: req.env.get("GITHUB_ID")!,
        clientSecret: req.env.get("GITHUB_SECRET")!,
      }),
      Google({
        id: "google",
        clientId: req.env.get("GOOGLE_CLIENT_ID")!,
        clientSecret: req.env.get("GOOGLE_CLIENT_SECRET")!,
      }),
      Credentials({
        id: "credentials",
        name: "Credentials",
        credentials: {
          // name is defined only in Signup
          name: { label: "Name", type: "text" },
          email: { label: "Email", type: "text" },
          password: { label: "Password", type: "password" },
        },
        async authorize(_credentials) {
          const isSignup = !!_credentials.name;
          try {
            const credentials = _credentials as {
              name?: string;
              email: string;
              password: string;
            };
            const { db } = await createClient(req);
            let selectedUser = (
              await db
                .select()
                .from(user)
                .where(eq(user.email, credentials.email))
                .execute()
            )[0];

            let justCreated = false;
            if (!selectedUser) {
              if (isSignup) {
                // sign up
                justCreated = true;
                selectedUser = (
                  await db
                    .insert(user)
                    .values({
                      name: credentials.name!,
                      email: credentials.email,
                      password: await pbkdf2(credentials.password),
                    })
                    .returning()
                    .execute()
                )[0];
              } else {
                throw new AuthError("User does not exist.");
              }
            }
            if (!selectedUser.password) {
              throw new AuthError(
                "You have been using OAuth instead of password, use the methods below like Google or Github.",
              );
            }
            if (
              !justCreated &&
              !(await pbkdf2Verify(
                selectedUser.password!,
                credentials.password,
              ))
            ) {
              return null;
            }

            return {
              id: `${selectedUser.id}`,
              email: selectedUser.email,
              name: selectedUser.name,
              image: selectedUser.avatar,
            };
          } catch (err: unknown) {
            const isAuthError = err instanceof AuthError;
            if (err instanceof Error) {
              const url = new URL(req.url);
              for (const k of url.searchParams.keys()) {
                url.searchParams.delete(k);
              }
              url.searchParams.set(
                isSignup ? "sign-up" : "sign-in",
                isAuthError
                  ? err.message
                  : "An error occured, Feel free to contact us",
              );
              req.headers.set("location", url.toString());
            }
            return null;
          }
        },
      }),
    ] as Provider[],
  }));

// session for server use
export const getSession = (req: RequestEventBase) => {
  const session: Session | null = req.sharedMap.get("session");
  return session;
};

```
Let me know if you need anything more.

mhevery:
OK, let me know how things need to change, and happy to cut another dot release.  PRs are welcome 😁

PR #5194, opened 2023-09-19
fix(qwik-auth): allow multiple set-cookie headers

ulic75:
# Overview

In PKCE OAuth authentication @auth/core sets two cookies (callback url and pkce code_verifier), but the second one was failing because of the condition added in #5180, which checked to see if the header already existed. This may be fine for most headers, but not for 'set-cookie' headers, in which multiple are allowed.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Checklist:

### DIFF
--- a/packages/qwik-auth/src/index.ts
+++ b/packages/qwik-auth/src/index.ts
@@ -141,7 +141,7 @@ async function authAction(
     skipCSRFCheck,
   });
   res.headers.forEach((value, key) => {
-    if (!req.headers.has(key)) {
+    if (!req.headers.has(key) || key === 'set-cookie') {
       req.headers.set(key, value);
     }
   });
--- a/packages/qwik-auth/src/index.ts
+++ b/packages/qwik-auth/src/index.ts
@@ -141,6 +141,10 @@ async function authAction(
     skipCSRFCheck,
   });
   res.headers.forEach((value, key) => {
+    /**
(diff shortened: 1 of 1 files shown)


mhevery (approved):

packages/qwik-auth/src/index.ts
```diff
@@ -141,7 +141,7 @@ async function authAction(
     skipCSRFCheck,
   });
   res.headers.forEach((value, key) => {
-    if (!req.headers.has(key)) {
+    if (!req.headers.has(key) || key === 'set-cookie') {
```
mhevery, line 144: Can we add a comment explaining the reasoning for this....

ulic75: Added a comment, if it doesn't include the details you feel is important feel free to suggest another :).

mhevery (approved):

mhevery:
This this need a new NPM release to be cut?

ulic75:
> This this need a new NPM release to be cut?

Probably, since the other PR introduced a bug that this resolves.

ksjogo:
Heya, this PR still seems bugged.
As multiple cookies won't be handled by fetch, only the last one will apply.
I tried fixing this there: https://github.com/BuilderIO/qwik/pull/5194

PR #5238, opened 2023-09-28
feat: When closing over values in `server$` the values should be marked as readonly

gitstart:
### What does this PR do?
This PR ensures that values serialized by server$() are immutable and triggers a clear error message if any mutation is attempted

### Issue reference
https://github.com/BuilderIO/qwik/issues/5019

### Demo video/scrennshot: 
https://www.loom.com/share/abc8b587001d4845bc55748000b0412f?sid=945d461e-79cc-4cd3-a80f-0e99bb0aa065
Happy hacking! 🎉

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.


mhevery:
Love this change. Let me know when it is ready.

packages/qwik-city/runtime/src/server-functions.ts
```diff
@@ -301,52 +312,57 @@ export const serverQrl: ServerConstructorQRL = (qrl: QRL<(...args: any[]) => any
         const requestEvent = useQwikCityEnv()?.ev ?? this ?? _getContextEvent();
         return qrl.apply(requestEvent, args);
```
wmertens, line 313: This is the only place where you should deep freeze the args. This is the code that runs on the server and the args are the ones sent by the client.

wmertens (changes requested):
Can you also add an e2e spec that tries to change an object on the server? Something like

```js
server$(() => {
  try {
    myObject.deeper.foo = 'test'
    return false
  } catch (e) {
    if (/* some test that checks it errored because of freeze */)
      return true
    return false
  }
})
```

gitstart:
> Hi @gitstart thanks for your help, are you still on this PR or is it abandoned?

Sorry for the delay. we are still working on the issue

wmertens:
@gitstart could you rebase?

packages/qwik-city/runtime/src/server-functions.ts
```diff
@@ -276,22 +289,17 @@ export const serverQrl: ServerConstructorQRL = (qrl: QRL<(...args: any[]) => any
...
         const ctxElm = _getContextElement();
         const filtered = args.map((arg) => {
           if (arg instanceof SubmitEvent && arg.target instanceof HTMLFormElement) {
             return new FormData(arg.target);
           } else if (arg instanceof Event) {
-            return null;
+            throw new Error('Cannot serialize instances of Event.');
           } else if (arg instanceof Node) {
-            return null;
+            throw new Error('Cannot serialize instances of Node.');
```
wmertens, line 302: Is this a good idea? Seems to me they are deliberately being ignored

wmertens:
@gitstart still interested in finishing this?

PatrickJS:
@wmertens @mhevery I'm changing this to only freeze in dev

PatrickJS:
in the future technically we can pass the current client value to the server then update the client signals from the server response

PR #5258, opened 2023-10-02
feat(core): Allow useContext in same elem as provider

wmertens:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This allows `useContext()` within the same component that already ran `useContextProvider()`.

# Use cases and why

This is useful for terseness, code reuse, and it eliminates an element of surprise.

# Checklist:

### DIFF
--- a/packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
+++ b/packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
@@ -1001,6 +1001,15 @@ renderSSRSuite('component useContextProvider()', async () => {
   );
 });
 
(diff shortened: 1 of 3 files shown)


wmertens:
So this turns out to be a breaking change for that one e2e test: It was reading the context after the useContextProvider hook and expecting it to hold the old context.

IMHO this is incorrect usage and in any case undocumented behavior, so it doesn't warrant a major version change?

mhevery (approved):

mhevery:
Thank you. ( I think the change to test is fine.)

Issue #5257, opened 2023-10-02
[✨] document or remove useContextBoundary


wmertens:
### Is your feature request related to a problem?

There's an undocumented but public function called [useContextBoundary](https://github.com/BuilderIO/qwik/blob/716812e3997f8a0c5ab64507ec3791608754ac63/packages/qwik/src/core/use/use-context.ts#L220).

It takes context ids and copies their context into the current element. It does a shallow copy, so deep changes change the context, and proxy get/set still runs.

It also adds an entry `_` to the contexts, which when encountered during context search, stops the context search.

This means two things:
- only the copied contexts will be visible as contexts
- a context named `_` will not work, this is not documented

It is not clear what this is useful for, and it's not used in the repo at all.

### Describe the solution you'd like

Either remove it or document it

### Describe alternatives you've considered

N/A

mhevery:
Sweet, any chance to send a PR to remove it?

wmertens:
@mhevery done in #5274

PR #5263, opened 2023-10-03
No slot parent

wmertens:
# Overview

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

This removes $slotParent$ from the internal contexts and makes the context search more correct and slightly more efficient.

The bug was that it's supposed to stop searching context when hitting `_` but it didn't when following the dom. (not that I think that's a good feature, see #5257)

# Checklist:


gioboa:
Thanks @wmertens for your useful comments and for this nice refactor

mhevery (approved):

PR #5268, opened 2023-10-04
fix(starter): localize container attribute

tzdesign:
I changed the lang of the container attributes to serverData.locale with fallback en-us

# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Just added the container attribute for locale

# Use cases and why

# Checklist:

### DIFF
--- a/starters/features/localize/src/entry.ssr.tsx
+++ b/starters/features/localize/src/entry.ssr.tsx
@@ -25,7 +25,7 @@ export default function (opts: RenderToStreamOptions) {
     base: extractBase, // determine the base URL for the client code
     // Use container attributes to set attributes on the html tag.
     containerAttributes: {
-      lang: "en-us",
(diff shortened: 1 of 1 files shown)


tzdesign:
@gioboa thanks for fixing the localize so fast. I'm not 100% sure if locale is the correct ISO value, but now it should be the locale qwik picks up.

@mhevery you probably know if this is the correct location in serverData.

Thank you guys for being that fast with the starters.

wmertens (approved):

gioboa:
Thanks, which problems did you find with the locale?

tzdesign:
@gioboa we just had a meeting with Miško on Monday and we had the old starter installed and where confused that client side rendering was not getting the right translations. So your change is actually what was missing.

gioboa (approved):
Thanks for your help 👍

PR #5269, opened 2023-10-04
docs: Update SSG documentation to run build instead of preview

ibousfiha:
# Overview
This PR aims to correct the documentation for SSG, specifically to suggest running build instead of preview.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

This PR updates the SSG documentation to replace the recommendation of using the preview command with the build command for generating static pages. This change is important to highlight because the preview command currently does not generate any HTML static files, whereas the build command does. This ensures developers are fully aware of what each command will yield.

# Use cases and why

- Clarity: The documentation should accurately guide the developer in generating static pages, which is achieved by using build.
- Avoiding Confusion: Mentioning that preview does not generate HTML static files clarifies its limitations.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/guides/static-site-generation/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/guides/static-site-generation/index.mdx
@@ -45,10 +45,10 @@ Running the above command will make the following changes to your project:
 
 Your build files will be generated into the `dist` folder.
 
-You can build and preview your static site using:
+You can build your static site using:
 
 ```shell
-npm run preview
+npm run build
 ```
 
 ### SSG Config


wmertens:
Would it not be better to have preview generate the SSG pages instead?

wmertens:
Would it not be better to have preview generate the SSG pages instead?

ibousfiha:
I'm on the same page with you; having the `preview` command generate SSG pages would be ideal. I've already brought up this issue on Discord, and @mhevery suggested updating the documentation for now ([Discord conversation link](https://discord.com/channels/842438759945601056/1156589798761451521/1158848137117565038)).

Additionally, when I attempted to remove the `--ssr` flag from the `preview` command in `package.json`, it resulted in the following error:

```bash
Could not resolve entry module "src/entry.preview.tsx/index.html".
error during build:
RollupError: Could not resolve entry module "src/entry.preview.tsx/index.html".
    at error (file:///Users/issambousfiha/Projects/qwik-app/node_modules/rollup/dist/es/shared/node-entry.js:2287:30)
    at ModuleLoader.loadEntryModule (file:///Users/issambousfiha/Projects/qwik-app/node_modules/rollup/dist/es/shared/node-entry.js:24881:20)
    at async Promise.all (index 0)
```

I can open a separate issue to address this problem. Would that be a good next step?

wmertens (approved):

wmertens:
Thanks! Yes, an issue for fixing preview would be good.

PR #5326, opened 2023-10-17
fix: @builder.io/qwik/build should not SSR render

### LINKED ISSUE #5325: [🐞] library starter throws an error when trying to load @builder.io/qwik/build

Qwik Rollup / Vite plugin

### Describe the bug

I initially found this bug following the steps to reproduce, which are very straightforward as you can repro on the component library starter. The bug only seems to be present when QwikCity is not used so it affects both the default vite dev server on component library projects as well as storybook, if it is used.

After some digging, I found the issue seems to be the logic introduced as part of #5234 as the call to load `@builder.io/qwik/build` has an `Accepts: */*` http header and no extension, so under the new logic it triggers an ssr render which breaks it.

I think the solution would be to add another special case to `shouldSsrRender` like the ones for vite-ping and such, so that the path `@builder.io/qwik/build` does not ssr render. I will open a PR for this as the fix works for me locally, but feel free to go a different way if there's a better solution.

### Reproduction

https://github.com/BuilderIO/qwik/tree/main/starters/apps/library

### Steps to reproduce

```
yarn create qwik
# select component library option (this bug does not seem to be present when qwikcity vite plugin is used)
yarn dev
```

When you click on the counter button, it will throw a client-side error as it tries to load `@builder.io/qwik/build` and fails 

### System Info

```shell
Qwik v1.2.13 (only, bug was introduced in this version)
```

### Additional Information

_No response_

mpeg:
# Overview

As per bug #5325 `@builder.io/qwik/build` is a special url that should not ssr render. #5234 introduced new logic on `shouldSsrRender` that causes it to ssr render, which breaks things.

fix #5325 

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This PR adds a new conditional to `shouldSsrRender` similar to the existing ones for `vite-ping`, `html-proxy` and `__open-in-editor` for `@builder.io/qwik/build` that fixes the issue.

# Use cases and why

As per #5325, the library starter template throws an error when interacting with any client side code on v1.2.13

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/vite-server.ts
+++ b/packages/qwik/src/optimizer/src/plugins/vite-server.ts
@@ -309,6 +309,9 @@ const shouldSsrRender = (req: IncomingMessage, url: URL) => {
   if (InternalPrefixRE.test(url.pathname)) {
     return false;
   }
+  if (pathname.includes('@builder.io/qwik/build')) {
+    return false;
+  }
   const acceptHeader = req.headers.accept || '';
   const accepts = acceptHeader.split(',').map((accept) => accept.split(';')[0]);
   if (accepts.length == 1 && accepts.includes('*/*')) {


wmertens:
Thanks!

PR #5334, opened 2023-10-19
fix(cli): set args on interactive cli

mpeg:
# Overview

When using the interactive CLI mode, the build preview option was in the help commands list but wasn't working, selecting it would return you back to the interactive help.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The `build preview` option wasn't working as the command runner expects the task to be named `build` with preview being an extra argument.

Initially considered adding it as a special case to the `runCommand` switch statement, but this fix is actually better as it will work more consistently if other two word commands are added, and it prevents unexpected behaviour. 

For instance, without the fix, if you run `qwik preview` (unrecognized command) and when the interactive help shows up you select the production build option it would still run a preview build as the args are inherited from your initial commands.

# Checklist:

### DIFF
--- a/packages/qwik-city/adapters/vercel-edge/vite/index.ts
+++ b/packages/qwik-city/adapters/vercel-edge/vite/index.ts
@@ -134,8 +134,8 @@ export interface VercelEdgeAdapterOptions extends ServerAdapterOptions {
   /**
    * Define the `target` proeprty in the `ssr` property in the `vite.config.ts` file.
    *
-   * Defaults to `webworker` for not having a breaking change. But `node` will become the default
-   * in an upcoming release.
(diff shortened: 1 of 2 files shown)


zanettin:
thx a lot @mpeg for your contribution 🙏 may i aks you to run `pnpm fmt` on this branch so that the CI is happy? thx again!

mpeg:
I did see the lint error, but didn't correct it as the issue it complains about has nothing to do with my PR so I didn't want to make things confusing, the prettier issue is in `qwik-city/adapters/vercel-edge/vite/index.ts`

I've fixed it in my branch anyway as it might cause more confusion in other PRs if goes unfixed

gioboa (approved):
Thanks for your help 👍

Issue #5333, opened 2023-10-19
[🐞] Sparse array cannot be serialized.


genki:
### Which component is affected?

Qwik Runtime

### Describe the bug

I have tried to use an array that is sparsely initialized in the `useTask$` then the serialization failed.

This is an example:
```ts
const a = [];
a[1] = 1;
useTask$(() => {
  console.log(a); // failed to serialize
})
```

I think this should be serialized as `[undefined, 1]`

### Reproduction

https://stackblitz.com/edit/qwik-starter-1w77uc?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

Please see the link above.

### System Info

```shell
System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M2
    Memory: 124.56 MB / 24.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.5.0 - /opt/homebrew/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.8.1 - /opt/homebrew/bin/npm
    pnpm: 8.8.0 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 118.0.5993.88
    Safari: 17.0
  npmPackages:
    @builder.io/partytown: ^0.8.0 => 0.8.0 
    undici: 5.22.1 => 5.22.1 
    vite: 4.4.7 => 4.4.7
```

mhevery:
Ohh. Interesting. Sparse is not the same as undefined. But I think in this case this is a good approximation. 

Any chance you would be up for creating a fix? I would be happy to guide you along.

genki:
okey, I will try.

gioboa:
Thanks @genki 💪

genki:
I have recognized this is not an issue but the specification that we shouldn't use any sparse arrays.
So I close this issue.
If you want to use the sparse array or have to use libraries that use it, you can fill the sparse array by yourself to avoid the serialization error. Gook luck :)

PR #5336, opened 2023-10-20
fix(core): Fill the holes of the sparse array with `undefined`

genki:
# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

To fix the issue #5333 fill the holes of the sparse array with `undefined` in order to serialize any arrays.

# Use cases and why

Please see the issue #5333 

# Checklist:

### DIFF
--- a/packages/qwik/src/core/state/common.ts
+++ b/packages/qwik/src/core/state/common.ts
@@ -1,5 +1,4 @@
 import { assertFail, assertTrue } from '../error/assert';
-import { qError, QError_verifySerializable } from '../error/error';
 import { isNode } from '../util/element';
 import { seal } from '../util/qdev';
 import { isArray, isFunction, isObject, isSerializableObject } from '../util/types';
@@ -64,7 +63,10 @@ const _verifySerializable = <T>(value: T, seen: Set<any>, ctx: string, preMessag
           // Make sure the array has no holes
           unwrapped.forEach((v, i) => {
             if (i !== expectIndex) {
-              throw qError(QError_verifySerializable, unwrapped);
+              const array = value as any[];
+              for (let j = expectIndex; j < i; j++) {
+                array[j] = undefined;
+              }
             }
             _verifySerializable(v, seen, ctx + '[' + i + ']');
             expectIndex = i + 1;


mhevery (changes requested):
- Could we have a test for this?
- I noticed fuel on above. There is a comment that specifically says to check for a raise with holes, but then we fill it in, so I think the common needs to change.

genki:
@mhevery I couldn't find the tests that are checking to raise error when with holes.
It's okay that simply remove the comment line and add a test to check if the sparse array doesn't raise the error?

wmertens:
This feels pretty hacky IMHO. Wouldn't it be better to support sparse arrays in the serializer?

genki:
@wmertens I think so too, but the `JSON.stringify()` does equivalent thing. This may be the JavaScript's way.

wmertens:
There's two problems:
- (minor) we change a data structure given to us
- (major) we allow the dev to write `a[100000]=1` and then we're invisibly serializing 100000x `undefined`

I prefer having the dev consider their options over this approach.

mhevery:
Thinking more about this, I think arrays with holes are very bad for performance anyway and should be discouraged. Why can't the developer just fix their code?

genki:
I think there is no one to use the sparse array intentionally but it could be mixed in if using outer APIs that use it.
Even though, we application engineer can fill the holes in advance. And Qwik can do as well.
The problem is which one do it.

genki:
Qwik has been checking all arrays if they have holes. Filling them at that time can be taken with minimal cost.
If we have applications done it, it introduced extra cost to check the arrays if they have holes twice.

mhevery:
I agree that it is minimal effort, but it is not the same thing. Arguably filling in the array changes the meaning of the application....

```
> a = [];
[]
> a[1] = 123
123
> Array.prototype[0] = 'WEIRD'
'WEIRD'
> a[0]
'WEIRD'
```

So I don't think we should be doing it.

I think the correct thing is to either:
1. Error (as we do now)
2. Serialize the array with holes and then re-create the holes.


Arguably holy arrays are just bad practice, so I think Erroring is probably a better solution...

genki:
Haha, I have never think there could be someone who changes the `Array.prototype[0]` :)
Okey, I will choose to fill the holes by myself. It might be the best for people who believe there may be sparse array but no one change the `Array.prototype`
Thank you for conversation about it that brings me forward :)

PR #5339, opened 2023-10-21
fix(cli): parseTemplatePath doesn't work in windows

yuyidegit:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

in windows path split with \,which makes parseTemplatePath does not work.

# Use cases and why

qwik new does not work in windows

# Checklist:

### DIFF
--- a/packages/qwik/src/cli/utils/templates.ts
+++ b/packages/qwik/src/cli/utils/templates.ts
@@ -1,5 +1,5 @@
 import fs from 'node:fs';
-import { join } from 'node:path';
+import { join, sep } from 'node:path';
 import type { TemplateSet } from '../types';
 import { getFilesDeep } from './utils';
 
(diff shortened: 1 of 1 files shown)


packages/qwik/src/cli/utils/templates.ts
```diff
@@ -58,7 +58,10 @@ export async function readTemplates(rootDir: string) {
 }
 
 function parseTemplatePath(path: string, type: string) {
-  const parts = path.split(`/${type}/`);
+  let parts = path.split(`/${type}/`);
+  if (parts.length == 1) {
```
mhevery, line 62: I don't think this is the best way to fix this. Can we just use file system separator?

```typescript
const path = require('path');
path.splipt(path.sep + type + path.sep);
```

mhevery (changes requested):
Could I ask for this change please?

packages/qwik/src/cli/utils/templates.ts
yuyidegit: Of course

mhevery (approved):

zanettin:
thx for the fix @yuyidegit 🙏

PR #5394, opened 2023-11-03
qwik-city: Link - Handle option clicks

ptz0n:
# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Escape client side routing for modifier/option click. Ctrl on Windows, CMD on macOS.

# Use cases and why

1. When users click on `<a>` in the `<Link>` component with a modifier key active, we want to allow user preferred behaviour. Usually open in new tab etc. Currently, we always prevent default which results in client render in the same window/frame.

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/link-component.tsx
+++ b/packages/qwik-city/runtime/src/link-component.tsx
@@ -19,7 +19,12 @@ export const Link = component$<LinkProps>((props) => {
           prefetchLinkResources(elm as HTMLAnchorElement, ev.type === 'qvisible')
         )
       : undefined;
-  const handleClick = event$(async (_: any, elm: HTMLAnchorElement) => {
+  const handleClick = event$(async (ev: any, elm: HTMLAnchorElement) => {
+    if (ev.ctrlKey || ev.metaKey) {
+      // Do not enter the nav pipeline if any option key is pressed.
+      return;
+    }
+
     if (!elm.hasAttribute('preventdefault:click')) {
       // Do not enter the nav pipeline if this is not a clientNavPath.
       return;


wmertens:
Seems like a good idea. Nitpick is that ideally you'd limit the modifier to the correct one for the platform, but devs can just add their own onClick handler.

mhevery:
Wil this actually fix the issue of being able to alt click to open in a new window? I think the issue is that `prevent:default` prevents the browser from doing anything.

I think the correct solution is to be able to run code synchronously on `prevent:default` to determine if it should run or not. 

I had a write up on this, but Qwik already serializes some functions as sync in `<script q:func="qwik/json">` It would be greater if we could have a function there for determining if the `prevent:default` should run or not. @wmertens any chance we could attack that one?

wmertens:
Linking this to #4496 - as soon as that's implemented this can be implemented (with e2e test)

packages/qwik-city/runtime/src/link-component.tsx
```diff
@@ -19,7 +19,12 @@ export const Link = component$<LinkProps>((props) => {
           prefetchLinkResources(elm as HTMLAnchorElement, ev.type === 'qvisible')
         )
       : undefined;
-  const handleClick = event$(async (_: any, elm: HTMLAnchorElement) => {
+  const handleClick = event$(async (ev: any, elm: HTMLAnchorElement) => {
+    if (ev.ctrlKey || ev.metaKey) {
+      // Do not enter the nav pipeline if any option key is pressed.
+      return;
```
mhevery, line 25: I don't think this will work because events are handled async in Qwik.
I am working on sync events and once that lands we can include this.

gioboa:
I close this PR b/c this behaviour is covered by the new [`sync$` API](https://qwik.builder.io/docs/cookbook/sync-events/)

PR #5423, opened 2023-11-11
fix(qwik-city): fix rendered svg?jsx component closing tag

riccardoperra:
# Overview

# What is it?

- [ ] Feature / enhancement
- [X] Bug

# Description

Qwik currently optimize svg components imported through the ?jsx param, but the given result includes an extra `<` character.
This is noticeable while dragging links which are includings an optimized svg.

That bug can be reproduced also in the docs page using also the inspection tool.

image

Since the custom plugin is modifying the `<svg>` tag into a `<g>`, we have to slice the content with 4 character instead of 3 (we are removing `</g>`).

I moved out the function in order to create a unit test file just to test if everything works but if there are better approaches please give me feedback

# Checklist:


wmertens (approved):
LGTM! 👍 
Thanks for adding the test too!

hbendev:
Thanks. I wanted to fix this too, but went one step further in my PR: https://github.com/BuilderIO/qwik/pull/5407. Added another unit test just now :)

Issue #5419, opened 2023-11-10
[✨] Automatically add `px` to CSS properties provided under `style` prop


sidmohanty11:
### Is your feature request related to a problem?

It would be awesome if we could support the auto-addition of `px` or metrics when a user provides us with a number for a CSS property under the style props. This will make us more synonymous with React where this support is already present.

### Describe the solution you'd like

```jsx
<h1 style={{ fontSize: 50 }}> Hello World </h1>
```
should resolve `font-size: 50px`

### Describe alternatives you've considered

- We could use native CSS files (.css) and write pure CSS, but this is limiting if we want conditional CSS to be applied
- use strings everywhere- '50px'

PR #5426, opened 2023-11-11
feat(core): auto px addition to css properties for unitless numbers

sidmohanty11:
# Overview
This PR adds support for auto-addition of 'px' to CSS properties so that unitless numbers can also be passed to CSS properties under style prop.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Ref: #5419 

With this change, we will automatically add 'px' to unitless numbers provided under the style prop.

```jsx
<h1 style={{ fontSize: 50 }}> Hello World </h1>
```

this resolves to `font-size: 50px`.

I've referred and taken inspiration from React-DOM's implementation, reference attached below:
- https://github.com/facebook/react/blob/432b9f1d9729aaea010730d546bda89b9842eaa1/packages/react-dom-bindings/src/client/CSSPropertyOperations.js#L90
- Referred to [this](https://github.com/facebook/react/blob/main/packages/react-dom-bindings/src/shared/isUnitlessNumber.js) for getting all CSS properties that support numbers and don't require units. I've trimmed it down to the ones that we support.

# Use cases and why

1. As of now if I add a style prop to a JSX component in Qwik and pass in let's say a fontSize with a number (`style={{ fontSize: 20 }}`) this will not throw any error and will be picked up by CSS types but this actually breaks in our page because no units are mentioned.
2. This will make React to Qwik migrations very easy as React/Next.js support this by default

# Checklist:

### DIFF
--- a/packages/qwik/src/core/util/unitless_number.ts
+++ b/packages/qwik/src/core/util/unitless_number.ts
@@ -1,0 +1,51 @@
+/** CSS properties which accept numbers but are not in units of "px". */
+const unitlessNumbers = new Set([
+  'animationIterationCount',
+  'aspectRatio',
+  'borderImageOutset',
+  'borderImageSlice',
+  'borderImageWidth',
+  'boxFlex',
+  'boxFlexGroup',
+  'boxOrdinalGroup',
+  'columnCount',
+  'columns',
+  'flex',
+  'flexGrow',
+  'flexShrink',
+  'gridArea',
+  'gridRow',
+  'gridRowEnd',
+  'gridRowStart',
+  'gridColumn',
+  'gridColumnEnd',
+  'gridColumnStart',
+  'fontWeight',
+  'lineClamp',
+  'lineHeight',
+  'opacity',
+  'order',
+  'orphans',
+  'scale',
+  'tabSize',
+  'widows',
+  'zIndex',
+  'zoom',
+  'MozAnimationIterationCount', // Known Prefixed Properties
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/util/unitless_number.ts
```diff
@@ -0,0 +1,51 @@
+/** CSS properties which accept numbers but are not in units of "px". */
```
gioboa, line 1: Thanks for your help. Where did you find this list? is there an official site with this list?

sidmohanty11: Hey @gioboa, unfortunately I couldn't find any official site that mentioned this but this list is referred from React's official codebase here: https://github.com/facebook/react/blob/main/packages/react-dom-bindings/src/shared/isUnitlessNumber.js

mhevery (changes requested):
1. Can we have a test for this? Specifically there is a different code path for SSR vs CSR, so. we should have a tests which executes both to show that it is OK. 

NOTE: Here is an example on how to write it: https://github.com/BuilderIO/qwik/commit/60a031aa2b847a61412014a73f4b09341d36db7d#diff-a6d89ff9ed0038f46c59edbb3b027bf1daf98d3027f2e0509a920154441656a5

sidmohanty11:
Thanks @mhevery for sharing this! I've added tests for this feature for both CSR and SSR rendering, can you please take a look?

mhevery (approved):

Issue #5282, opened 2023-10-07
[🐞] qwik new CamelCaseComponent exported component becomes Camelcasecomponent


chug2k:
### Which component is affected?

Starters / CLI

### Describe the bug

I created a new component using `qwik new` and I lost camelcase in the Props and export declaration.

### Reproduction

n/a

### Steps to reproduce

just use qwik new. 

### System Info

```shell
System:
    OS: macOS 12.6.6
    CPU: (8) arm64 Apple M1 Pro
    Memory: 79.23 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
  Browsers:
    Chrome: 117.0.5938.149
    Safari: 16.5
  npmPackages:
    @builder.io/qwik: ^1.2.13 => 1.2.13 
    @builder.io/qwik-city: ^1.2.13 => 1.2.13 
    undici: 5.22.1 => 5.22.1 
    vite: 4.4.7 => 4.4.7
```

### Additional Information

Sorry to drop such a trivial bug. I am learning qwik for the first time, so submitting the issue with a goal of submitting a PR for it shortly as soon as I get my bearings.

wmertens:
Can you provide more details? How are you creating CamelCaseComponent and how do you import it, where do you notice the problem?

PR #5430, opened 2023-11-12
fix(cli): casing for component and mdx route creation

sidmohanty11:
# Overview

This fixes the `qwik new` command to:
- preserve `CamelCased` components
- wrong `/route/newRoute` console.success
- mdx route creation provided with `.mdx` created wrong route

# What is it?

- [x] Bug

# Description

Improves/fixes:

1. Enhanced the `toPascal` and `toSlug` functions to better accommodate varying casings. For instance, if we input `new-component` as the name, it will correctly transform it into `NewComponent`. However, if we provide `NewComponent`, the default behavior (or no change) is expected, but currently, it converts it to `Newcomponent`. (Ref: #5282)
2. When generating an MDX route and supplying it with a `.mdx` extension, the resulting route appears as `/route/pagex`. The issue arises as `!(nameAnswer as string).startsWith('/')` is false if we already provide it starting `/`. This fails to differentiate between .md and .mdx extensions, leading to unexpected routes as it blindly replaces `.md` in the file, this leaves the `x` behind.
3. When logging our success message, it used `name` which transformed kebab casing to pascal casing which in turn outputted wrong route names. (For example, if we create route `/some/new-route` this outputs, `Route "/some/newRoute" created!`.

Fixes: #5282

# Checklist:

### DIFF
--- a/packages/qwik/src/cli/new/run-new-command.ts
+++ b/packages/qwik/src/cli/new/run-new-command.ts
@@ -95,11 +95,7 @@ export async function runNewCommand(app: AppCommand) {
 
     const fileOutput = await writeToFile(name, slug, template, outDir);
 
-    if (typeArg === 'markdown') {
-      log.success(`${green(`Markdown route "${name}" created!`)}`);
-    } else {
-      log.success(`${green(`${toPascal([typeArg])} "${name}" created!`)}`);
-    }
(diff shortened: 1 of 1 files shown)


gioboa (approved):
It looks good to me. Thanks @sidmohanty11 for your help

PR #5431, opened 2023-11-12
fix: 3rd party imports of libs during build

shairez:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

When trying to build an app that uses Qwik UI, the build failed when it reached a 3rd party import (floating-ui)

This fixes the vite plugin to support this

# Checklist:


wmertens:
@shairez Very nice 🎉!

I think we could use a comment explaining why [1] works?

Also, nitpick, any reason why those imports are moved around?

shairez:
Thanks @wmertens !

So what I found in my research is that only direct imports have an `origPath`, but when it comes down to 3rd party imports that are imported from another 3rd party import,  the `transformedOutput[1]` hold the original `importer` path, which is what needed to resolve from the correct place.

I think it's the missing part of your commit from last week (hopefully 😅)

gioboa:
Thanks @shairez 😎

PR #5440, opened 2023-11-14
fix: Wrap function in cleanup function instead of returning it

thasmo:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Updates the tutorial example for synchronous events with using the cleanup function instead of returning a function.


PatrickJS (approved):

gioboa (approved):
Thank you @thasmo nice catch

PR #5450, opened 2023-11-17
fix(docs): remove `inline-code` and `link` in tag `<a />`

JuerGenie:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

I removed the `inline-code` and `link` syntax in the `<a />` tag to fix the error caused by nesting `<a />` tags during rendering.

Sorry for making this error in the `.card-grid`’s render. 😢


gioboa (approved):
Thanks for fixing it @JuerGenie

Issue #5371, opened 2023-10-31
[🐞] Playground is broken in 1.2.15


wmertens:
### Which component is affected?

Qwik Playground

### Describe the bug

Error is

```
ReferenceError: require is not defined at globalThis.qwikOptimizer.module.exports.exports (eval at g (https://qwik.builder.io/repl/repl-sw.js:1:341), <anonymous>:2434:25) at eval (eval at g (https://qwik.builder.io/repl/repl-sw.js:1:341), <anonymous>:3561:2) at g (https://qwik.builder.io/repl/repl-sw.js:1:377) at async N (https://qwik.builder.io/repl/repl-sw.js:1:1086) at async K (https://qwik.builder.io/repl/repl-sw.js:12:5471)
```

which might mean a bundling issue.

The commits involved are https://github.com/BuilderIO/qwik/compare/v1.2.14...v1.2.15 which don't immediately have anything suspect.

### Reproduction

https://qwik.builder.io/playground/#v=1.2.15&f=G000o4mG5EQDAA

### System Info

```shell
website
```

wmertens:
paging @bluwy, your commit https://github.com/BuilderIO/qwik/commit/ec53ef7834dd76af0fca1f90c903d93f6cb802f6 is the only one that touched the optimizer, I'm assuming the build pulls in the new vitefu package that somehow uses require

bluwy:
Hmm seems like it. `packages/qwik/dist/optimizer.cjs` `require`s some modules now because I thought it's safe to run in node. The esbuild script also bundles for `platform: 'node'`. 

I suppose the fix is to import it [here](https://github.com/BuilderIO/qwik/blob/ab67f91f707ad3baa6e15499d11753e84f7b9a9b/packages/qwik/src/optimizer/src/plugins/vite.ts#L689) but I'm not really sure how that can be done with dependencies bundled, or maybe `vitefu` cannot be bundled. Not familiar with how the project is bundled 😅 

At the meantime, I don't mind if we revert that PR, and then figure it out the fix for Vite 5 later.

mhevery:
Reverting: https://github.com/BuilderIO/qwik/pull/5379

PR #5451, opened 2023-11-17
chore: Vite 5

wmertens:
This should be straightforward thanks to @bluwy's work to remove the resolvePackageData API, but that caused the playground to stop working (#5371).

Maybe Vite 5 magically makes it work again? We need to get a build in a way that the playground can consume it.

Things to do:
- [x] Change qwik repo to build with Vite 5
- [x] Make sure playground doesn't break
- [x] Make sure all qwik vite plugins work with Vite 4 & 5
- [ ] Make sure all starters work with Vite 5
- [ ] When creating a new Qwik project, use Vite 5 but make it clear that the user should downgrade to 4 if they have problems with some vite plugin they want to use


wmertens:
Should be available at github:BuilderIO/qwik-build#build/vite-5 and github:BuilderIO/qwik-city-build#build/vite-5 soonish

PatrickJS (approved):

gioboa:
Thanks for this great PR

Issue #5240, opened 2023-09-29
[🐞] Link prefetch does not prefetch on mobile after SPA navigation


maiieul:
### Which component is affected?

Qwik Runtime

### Describe the bug

`<Link prefetch>`'s current implementation is to: (**1**) prefetch on hover; and (**2**), if the viewport is smaller than 520px, prefetch on:qvisible. 

I'm not 100% sure why, but no `<Link prefetch>` present on the page, either in the layout.tsx or in index.tsx will prefetch the links available on the viewport after SPA navigation.

I think the problem is that the current implementation relies on the qvisible event, and this event is only fired on first page load when the element appears in the viewport for the first time, but it is not fired again after SPA navigation. Either the qvisible event should fire as well after SPA navigation, or Link prefetch should rely on another event. What do you think @mhevery?

I'd love to work on a PR for that, but I will need your feedback first. I think it would be preferable if Link prefetch relied on another event if qvisible is also used for other purposes and can't be changed.

### Reproduction

https://github.com/maiieul/qwik-prefetch-on-mobile-repro

### Steps to reproduce

- npm install
- npm run dev

- Open dev tools (ctrl + shift + i)
- Toggle device toolbar (ctrl + shift + M) 
- Select a real phone (not responsive design)

![image](https://github.com/BuilderIO/qwik/assets/45822175/eeec3d01-7a7b-420d-b59f-e74afb139c69)

- Click on `go to route` buttons.

Notice that after first page load, the next route will load automatically because it was prefetched properly. But then the route coming after that (with no page refresh in-between) will take 1s to load because it has to wait for the routeLoader$ to finish executing and it wasn't prefetched properly.

### System Info

```shell
System:
    OS: Linux 6.5 Fedora Linux 38 (Workstation Edition)
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i5-1340P
    Memory: 21.80 GB / 31.04 GB
    Container: Yes
    Shell: 3.6.1 - /usr/bin/fish
  Binaries:
    Node: 18.16.1 - /usr/local/bin/node
    npm: 9.7.2 - /usr/local/bin/npm
    pnpm: 8.7.4 - ~/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 117.0.5938.92
  npmPackages:
    @builder.io/qwik: ^1.2.12 => 1.2.12 
    @builder.io/qwik-city: ^1.2.12 => 1.2.12 
[...]

PR #5480, opened 2023-11-26
feat(link): prefetch symbols

wmertens:
rebase of #4678

@jordanw66 does this look ok?


wmertens:
@mhevery any idea what this error could be? https://app.netlify.com/sites/qwik-insights/deploys/65630257f2395100085905ed#L346

packages/qwik-city/runtime/src/link-component.tsx
```diff
@@ -1,75 +1,111 @@
...
       {...linkProps}
       onClick$={[onClick$, handleClick]}
-      data-prefetch={prefetchDataset}
-      onMouseOver$={onPrefetch}
-      onFocus$={onPrefetch}
-      onQVisible$={onPrefetch}
+      data-prefetch={prefetchData}
+      onMouseOver$={isDev ? handlePrefetch : undefined}
+      onFocus$={isDev ? handlePrefetch : undefined}
+      onQVisible$={!isDev ? handlePrefetch : undefined}
```
wmertens, line 80: @jordanw66 actually, shouldn't these include the given props if any (if so, that's an existing bug)?

And I don't understand why the handling happens only when isDev?

packages/qwik-city/runtime/src/client-navigate.ts
```diff
@@ -40,8 +40,12 @@ export const newScrollState = (): ScrollState => {
   };
 };
 
-export const dispatchPrefetchEvent = (prefetchData: QPrefetchData) => {
+export const prefetchSymbols = (path: string) => {
   if (isBrowser) {
-    document.dispatchEvent(new CustomEvent('qprefetch', { detail: prefetchData }));
+    path = path.endsWith('/') ? path : path + '/';
```
mhevery, line 45: You are assuming that all paths end with `/` This is the default but it is possible to configure Qwik without trailing slash.

maiieul:
This should fix #5240 (a.k.a SPA prefetch on mobile), right?

PR #5494, opened 2023-11-28
fix(localization-starter): update docs and add script

tzdesign:
I changed the file pathes in the docs to match the real ones and also added a prei18n-extract script to make sure the build is running before the extract. I guess a lot of users will miss that one.

# Checklist:


tzdesign:
@gioboa what do you think about having `vite build` running as `prei18n-extract`?

I changed the docs for auto-translation. Looks like you renamed from dash to point in the filename.

See ya

wmertens:
@tzdesign take a look at https://www.npmjs.com/package/compiled-i18n, it automates everything and adds plural support. Feedback very welcome

mhevery (approved):

Issue #5400, opened 2023-11-04
[🐞] from create qwik latest, I can't make cloudflare-pages work since 1.2.14


tidiview:
### Which component is affected?

Qwik Runtime

### Describe the bug

I am using Windows, Visual Code, from there create qwik latest, pnpm install, then cloudflare-pages adapter, pnpm build, then start. In local, no problem but when transfer to [github repo](https://github.com/tidiview/worktop-qwikcity), the cloudflare building do not work at all.
The site builds on cloudflare, but once it has build, the [site](https://66751608.worktop-qwikcity.pages.dev/) appears completely blank since 1.2.15, then 1.2.16 and now, 1.2.17.
I thought that other user would report it and that it would be solved, but it seemed that I am the only one to have that problem therefore may be I have a specific problem, so I decided to put it here.

### Reproduction

https://66751608.worktop-qwikcity.pages.dev/

### Steps to reproduce

pnpm create latest, pnpm install, then cloudflare-pages adapter, pnpm build, then start.

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (4) x64 Intel(R) Core(TM) i3-4160 CPU @ 3.60GHz
    Memory: 1.75 GB / 7.86 GB
  Binaries:
    Node: 18.12.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.8.0 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Spartan (44.19041.3570.0), Chromium (119.0.2151.44)
    Internet Explorer: 11.0.19041.3570
  npmPackages:
    @builder.io/qwik: ^1.2.17 => 1.2.17
    @builder.io/qwik-city: ^1.2.17 => 1.2.17
    undici: ^5.26.0 => 5.27.2
    vite: ^4.4.11 => 4.5.0
```

gioboa:
I bumped up Qwik to 1.2.16 in one project and I had the same problem.
@tidiview Can you bisect the commits to figure out the problematic commit please?
[Here is the guide](https://github.com/BuilderIO/qwik/blob/main/CONTINUOUS_BUILD.md#bisect-for-regression)

fun fact, the Qwik docs site is on Cloudflare and it's working 🤔

tidiview:
[...]

PR #5507, opened 2023-11-30
fix: cf pages polyfill only if needed

mpeg:
# Overview

As discussed in #5400, the previous patch to remove the CF Pages Stream polyfill was causing trouble for people with older pages apps as Cloudflare does not make new APIs available by default, they instead preserve backwards compatibility by making it opt-in – so it was only working correctly for apps less than a year old.

This patch checks whether the TextEncoderStream class is available before replacing it, preferring the native version if so, but keeps the polyfill for backwards compatibility

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

# Checklist:


wmertens (approved):
LGTM

PR #5508, opened 2023-11-30
feat(qwik): Experimental support for synchronous QRLs

### LINKED ISSUE #4496: [✨] prevent default functions

### LINKED ISSUE #5322: [✨] stopPropagation:{eventName}

I want to implement a card and like button positioned absolute on top of it. When a user clicks on the card body, a details drawer appears on the screen. But if the user clicks the button, a backend request is called.

https://qwik.builder.io/playground/#v=1.2.13&f=7VY9C8IwEN39FUGH0q1FBEnQxVVwcBSHlgZbtDSm9QOk%2F913%2BWg7CNJRkA4lcORyd%2B%2FeewPQLOLlaNBMbHlgHaNgKDFQVV1QazkxFhbnLgXpjgJZcJYSmQrwZdbknM2jSD0F1KA45U137NmcsxnoDB9S2UTWe3zIlaRgSeyYYNpeFmtZ4q4K8aU9mfeOQLmbDWxPL%2FEvGzV0FsFhk2gIPAUc2U6B3TMJOrzULNPJA4gNW%2B8qfJ%2FafmamoC8JtsVZutJ9nj2gajpFfy2vN6hbEGJWtFGrqQ2e%2BryDtrVruq1zcQ4Q5O0cJv6L%2FJOL%2FAY

I want to have stopPropagation:{eventName} attribute similar to preventDefault:{eventName}.

1. useVisibleTask$ - Doesn't work :( Why? https://qwik.builder.io/playground/#v=1.2.13&f=7ZbBSsNAEIbvPsVQDxuhhIqgkGAv0psgeBUPKTu1wZYuaa2C5N39ZybJrlHEq%2BAW2qTZzOwMM9%2F8SdFcXQ7hpEUzjdNXL1O8%2FVhQJxY6iKTTDeG7sNvXkvZCaIamOnIpMykAJAUtBbQlWOoP64IuZrPwVmJS1E%2Frw3AbSV%2FQKVCHD1yZI9Ml3%2FiqliAo%2Bq%2BkxoydN7yFrR32b%2B1Oz%2FubDjBX4Aca%2Bp5XcBS1yWLD0tXzDKkl7B3PgmiFqF6hHHsrOVjywlaKskYP8sr7xRGWZXQxMJs5ZaybUsbyf2JXVqp73MNt%2FcxdbgzNj6J1AkACnaVgc3Lebqm9fLRBA9Ilba6%2F%2BJbrL5UMkRgF0btF%2FPk8N1UDOWQHuQuYhZ4xPDZ78k31iv4%2Ba3sN1ldOGyvcwlCZOCSpVYpcT%2BzhpH87KYd2LkkYlGvXBKJnuz74h9efhNcH
2. [Event target](https://qwik.builder.io/docs/components/events/#event-target) - Current solution https://qwik.builder.io/playground/#v=1.2.13&f=7VZLa8JAEL73V0ztIS1IsRSkZEkvvfRQ6KG9iYdoVg2%2B4hofIPnvfjOTlw%2FEq2ACG5bs7MzufPN9UwNNu%2F1xDjRNuoybBz0hiEdEDKf0kvky5tv1mbRQO2trWHoS8IVPPeZTA8qM0pFP761WsjUQhHg4SstpReg%2BPYHR8MKVOtL244yvsAeiRJkZcrrZm7NT7DXH%2BqnOJN5rgK6u%2BqGLRNu%2FoQbMJwFhkV1jcZP6K%2FDsLP0XlSwNieIBPR%2F8o8cgIDF6VUlVzPGjKDAyy2SUod64eJ0vRKGk2qXfBOIRWbDtZEmRCzcoCCQS1jye4AotW9We7I6PkxXNTpG7rIKSXHLNVq7mMK6feGzzbBTh%2FaGAJHn8dXaxguZ6L4AP13nQ0MWNwm0tk9kn71b2ljlMuePMkXqnl5uklz0

### Additional context

_No response_

mhevery:
# Overview

feat(qwik): Experimental support for synchronous QRL `$sync()`.

It is often desirable to have call API on events synchronously. For example, `event.preventDefault()` and `event.stopPropagation()` must be called synchronously in order to have a helpful effect.

```TypeScript
  <button onClick$={[
    $sync(event => event.preventDefault()),
    $(() => {
      // normal behavior.
    })
  ]}>Click Me</button>
```

The idea is that `$sync()` will be inlined into HTML. For this reason, the function, `$sync()`, must not only be a pure function but also can not depend on any external code, such as imports.

The best way to think about it is that the function inside `$sync()` must be able to survive `fn.toString()` and then be reconstructed from the string into a proper function.

In practice, this means that the function must be a simple function which can't:
- Close over any state.
- Close over any imports.

Fix #5322
Fix #4496

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -49,7 +49,7 @@
       "internalConsoleOptions": "neverOpen",
       "program": "${workspaceFolder}/./node_modules/vitest/vitest.mjs",
       "cwd": "${workspaceFolder}",
-      "args": ["--threads=false", "packages/qwik/src/core/container/render.unit.tsx"]
+      "args": ["--threads=false", "${file}"]
     }
   ]
 }
--- a/packages/docs/src/routes/api/qwik-testing/api.json
+++ b/packages/docs/src/routes/api/qwik-testing/api.json
@@ -12,7 +12,7 @@
         }
       ],
       "kind": "Variable",
-      "content": "CreatePlatform and CreateDocument\n\n\n```typescript\ncreateDOM: () => Promise<{\n    render: (jsxElement: JSXNode) => Promise<import(\"@builder.io/qwik\").RenderResult>;\n    screen: HTMLElement;\n    userEvent: (queryOrElement: string | Element | keyof HTMLElementTagNameMap | null, eventNameCamel: string | keyof WindowEventMap, eventPayload?: any) => Promise<void>;\n}>\n```",
+      "content": "CreatePlatform and CreateDocument\n\n\n```typescript\ncreateDOM: ({ html }?: {\n    html?: string | undefined;\n}) => Promise<{\n    render: (jsxElement: JSXNode) => Promise<import(\"@builder.io/qwik\").RenderResult>;\n    screen: HTMLElement;\n    userEvent: (queryOrElement: string | Element | keyof HTMLElementTagNameMap | null, eventNameCamel: string | keyof WindowEventMap, eventPayload?: any) => Promise<void>;\n}>\n```",
       "editUrl": "https://github.com/BuilderIO/qwik/tree/main/packages/qwik/src/testing/library.ts",
       "mdFile": "qwik.createdom.md"
     }
--- a/packages/docs/src/routes/api/qwik-testing/index.md
+++ b/packages/docs/src/routes/api/qwik-testing/index.md
@@ -9,7 +9,7 @@ title: \@builder.io/qwik/testing API Reference
 CreatePlatform and CreateDocument
 
 ```typescript
-createDOM: () =>
+createDOM: ({ html }?: { html?: string | undefined }) =>
   Promise<{
     render: (
       jsxElement: JSXNode,
--- a/packages/docs/src/routes/api/qwik/api.json
+++ b/packages/docs/src/routes/api/qwik/api.json
@@ -3,6 +3,20 @@
   "package": "@builder.io/qwik",
   "members": [
     {
+      "name": "_qrlSync",
+      "id": "_qrlsync",
+      "hierarchy": [
+        {
+          "name": "_qrlSync",
+          "id": "_qrlsync"
+        }
+      ],
+      "kind": "Variable",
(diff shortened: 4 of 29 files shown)


maiieul:
I like $$ :+1:  Sounds like it's more expensive.

PatrickJS (approved):

maiieul:
Wouldn't it be a better API if Qwik abstracted `$sync` so that the developer doesn't even have to worry about it?

So instead of

```tsx
    <button
      onClick$={[
        $sync((event) => event.preventDefault()),
        $(() => {
          // normal behavior.
        })
      ]}
    >
      Click Me
    </button>
  ```
  
  We could simply have
  
  ```tsx
    <button
      onClick$={(event) => {
        event.preventDefault();
        // normal behavior.
      }}
    >
      Click Me
    </button>
  ```
  
  But the optimizer then transforms it into 
  
  ```tsx
    <button
      onClick$={[
        $sync((event) => event.preventDefault()),
        $(() => {
          // normal behavior.
        }),
      ]}
    >
      Click Me
    </button>
  ```
  
  Or even skips this step and transforms the code into what this new $sync API will be transformed into.

maiieul:
I'm not sure it would be too much magic. Could be similar to what's being done in `props_destructuring.rs`. 

Why create a new `$sync` API for the optimizer to transform the code if `event.preventDefault` doesn't work? The optimizer could as well detect `event.preventDefault` and do the required transformations. 

Only reason I can see is if `$sync` allows us to do more useful stuff that I haven't been able to think about. If it's only a matter of API syntax, I prefer the magic :star2: .

mhevery:
Closing in favor of https://github.com/BuilderIO/qwik/pull/5545

PR #5509, opened 2023-12-01
fix(qwik): handle cypress dev server path

dmitry-stepanenko:
# Overview
Adding a cypress dev server prefix to the regexp of `shouldSsrRender` function

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description
Cypress tries to load vite assets using a slightly different path than we usually have: it adds  `/__cypress/src` prefix so we end up having paths like `/__cypress/src/@vite/client` . This is not caught by the VITE_PUBLIC_PATH regex part here https://github.com/BuilderIO/qwik/blob/main/packages/qwik/src/optimizer/src/plugins/vite-server.ts#L291 because it checks for the /@vite/ part to be at the beginning of the url's pathname. 

# Use cases and why

# Checklist:


gioboa (approved):
@dmitry-stepanenko you are amazing 👏 
Thanks for fixing it.

PR #5510, opened 2023-12-01
fix: WIP UNTESTED add invoke context to qwikcity

wmertens:
Co-authored-by: Giorgio Boa <gioboa@users.noreply.github.com>

This hopefully makes getLocale work in server calls

NOTE: I got stuck trying to share async localstorage between qwik and qwik-city, but now I realized that qwik-city can probably keep it on its side only. Will revisit at some point.


wmertens:
Fixed in #5905

Issue #5514, opened 2023-12-02
[🐞] <Link> Fallback to Hard Navigation when Page's JS file is not found.


iamriajul:
### Which component is affected?

Qwik City (routing)

### Describe the bug

Expect a scenario where You've deployed the application initially and your user has visited your home page.
The user was sitting down a while after opening the homepage, In the meanwhile you changed some code to your Login page, and now what happens is that your js bundle's filename for the login page components has changed to something else (because of hash-calculated based on code). 
At this moment when the user tries to navigate to the Login page via the Link component (using Soft Navigation), the navigation just hangs! The user doesn't understand what is happening and why the website is taking so long, but we know if the user waits infinitely the page won't load because the js filename has changed, to reflect the change user must hard reload.
This is the scenario we want to fix by falling back to the old good reliable MPA / Hard Navigation.

### Reproduction

https://github.com/iamriajul/repro-fallback-to-hard-navigation-when-page-js-not-available

### Steps to reproduce

- `pnpm install`
- `pnpm build`
- Now delete the file: `/build/q-ef638c7f.js` (to simulate you have pushed a new build with code changes, which in turn changes the filename.)
- `pnpm serve`
- Visit: `http://localhost:3000`
- Scroll Down and Click [Todo App](http://localhost:3000/demo/todolist) Link. Now see what happens! nothing happens the application just silently crashes.

### System Info

```shell
System:
    OS: macOS 13.5.1
    CPU: (16) x64 8-Core Processor             
    Memory: 988.09 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.20.2 - ~/.nvm/versions/node/v16.20.2/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.20.2/bin/yarn
    npm: 8.19.4 - ~/.nvm/versions/node/v16.20.2/bin/npm
    pnpm: 8.9.2 - ~/.nvm/versions/node/v16.20.2/bin/pnpm
  Browsers:
    Chrome: 119.0.6045.105
    Edge: 119.0.2151.97
    Safari: 16.6
```

gioboa:
Thanks for your feedback. Do you have a possibile solution on top of your head?

iamriajul:
I've been able to fix the issue like this:
![image](https://github.com/BuilderIO/qwik/assets/24244379/ff096123-7d52-472e-840b-54ca95622325)

knpwrs:
[...]

PR #5521, opened 2023-12-04
fix(routing): fallback to hard navigation when soft navigation fails.

iamriajul:
Issue: #5514

# Overview

This just adds a try/catch for soft navigation, so when soft navigation fails (may be due to js file does not exist) it will fallback to hard navigation using browser's `window.location.href`.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This issue is related to: https://github.com/BuilderIO/qwik/issues/5514

# Checklist:

- [ ] I have made corresponding changes to the documentation (not needed)


wmertens:
I like it, but I wonder:

- is setting the href the best way? What does it do for navigation history?
- can this be surprising for devs that keep a lot of state in their SPA?

For the second question, I'm thinking that it's probably ok because the user can reload the page at all times.

iamriajul:
I'm concerned about the other cases where `onClick$` listener's code has changed! 
As per opinion: There should some kind advance control should be provided.
For example:
The developer can decide what to do when a module (might be route module, click listener module, etc) loading is failed.
The options can be like:
- do nothing - silently crash
- save the scroll state now + refresh the current page and scroll back to the place where it was so the user can click again.
- (for Route only): fallback to hard navigation.

wmertens:
@mhevery what are your thoughts on this?

wmertens (approved):

wmertens:
@iamriajul Could you rebase? The tests failed back then and I can't re-run them in this state, the rebase will trigger them.

wmertens:
Thanks!

PR #5531, opened 2023-12-06
refactor(types): more precision

wmertens:
Specifics:
- piggyback on TS HTML types as much as possible (now automatically supports new props like popover)
- Remove QwikEvents and SyntheticEvents, they are really just DOM events
- improve QRL and QRLInternal types for better matching
- useOn* now can infer type and the passed element is also typed
- move the QRL AbortSignal to server$, the only place where it's supported
- generic parameter to useDocumentHead for the frontmatter
- DOM ref={} functions get the type of the element

Generic:
- add some tests for the types
- use `unknown` or actual type instead of `any` where possible
  - if not used as part of an `extends` test, it's probably ok to change `any` to `unknown`. It will surface type mismatches.
- fix function signatures so they can match extends ( `(...args: any) => any` instead of `(...args: any[]) => any`
- change some interfaces to types if they're not meant to be merge-extended and they're not extending (interfaces have a performance advantage there)
- add jsdoc comments

Actual code changes:
- In use-task.ts, allow `track(signal)` inside a task
- in docs, remove useless filtering from class array
- changed some hardcoded strings to constants


mhevery (approved):

gioboa:
🤯 what a great work. you rock

PR #5550, opened 2023-12-09
refactor(types): optimize, QRL props, PropsOf, track(Signal) type, ...

wmertens:
fix(tasks): type of track(Signal)
fix(types): typing intensifies

Type changes only, no real code changes.

- inside functions, handlers are QRLs only
- handler type inferral works through $()
- PropsOf now works on strings and inline components too, e.g. `PropsOf<'div'>`
- also type SVG elements and fix the incorrect SVG examples
- provide JSX.ElementType (supported since TS 5.1)
- examples now use the element argument of onInput$
- some more Qwik events are documented
- improve popover types
- fix api.ts: don't remove `{};` from api docs

### DIFF
--- a/packages/docs/src/routes/(ecosystem)/ecosystem/menu-items.tsx
+++ b/packages/docs/src/routes/(ecosystem)/ecosystem/menu-items.tsx
@@ -193,10 +193,10 @@ export const MenuItems = () => {
             xmlns="http://www.w3.org/2000/svg"
             aria-hidden="true"
           >
-            <g fill="none" stroke="currentColor" strokeLinejoin="round" strokeWidth="4">
+            <g fill="none" stroke="currentColor" stroke-linejoin="round" stroke-width="4">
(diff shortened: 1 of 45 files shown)


mhevery:
Love these changes, but they may need to go in into a minor bump. So please don't merge them into `main` Let's merge them into `build/minor` for now and when we are ready we can get it out.

mhevery:
Also, can you give this PR a title which will be mere useful when reading the git log months from now.

wmertens:
@mhevery I think the tracker commit needs to go in ASAP, and PropsOf is very nice dx.

What changes are you worried about for the minor? The fact that some incorrect SVG props will now become visible?

My reasoning is that we're very close to the 1.3.0 release so it's ok to make these changes in 1.3.1. There was at least one report of a typing issue in 1.3.0 that this fixed.

wmertens:
Discussed with Misko and will merge now.

PR #5558, opened 2023-12-11
feat(util): add a utility function for combining RequestHandlers

erikras:
# Overview

`combineRequestHandlers()` takes any number of `RequestHandler`s and combines them such that they will run in the correct order.

# What is it?

- [X] Feature / enhancement
- [ ] Bug

# Description

When making endpoints with Qwik City, I found myself writing several different middlewares, e.g. to create a database connection, to create a Stripe object, to instantiate a Slackbot, and I found it frustrating that you can only export a single `onRequest` value.

So I came up with a way to combine multiple `RequestHandler` functions whilst maintaining the order that they are called.

# Use cases and why

- 1. If you have one middleware to connect to a database, and another to load the current user record, the latter depends on the former, so they must be called in order.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
@@ -11,7 +11,7 @@ contributors:
   - mrhoodz
   - harishkrishnan24
   - erikras
-updated_at: '2023-09-18T19:04:44Z'
+updated_at: '2023-12-11T09:55:00Z'
 created_at: '2023-06-13T06:46:09Z'
 ---
 
--- a/packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
@@ -10,6 +10,7 @@ contributors:
   - gparlakov
   - mrhoodz
   - harishkrishnan24
+  - erikras
 updated_at: '2023-09-18T19:04:44Z'
 created_at: '2023-06-13T06:46:09Z'
 ---
@@ -51,6 +52,14 @@ export const onDelete: RequestHandler = async (requestEvent) => { ... }
 
 Each middleware function is passed a [`RequestEvent`](#requestevent) object which allows the middleware to control the response.
 
+If you want to combine multiple `RequestHandler`s together, there is a utility function called `combineRequestHandlers()` that will allow you to chain mulitple `RequestHandler`s together.
+```typescript
+import type { RequestHandler } from '@builder.io/qwik-city';
+import { combineRequestHandlers } from "@builder.io/qwik-city";
+
+export const onRequest: RequestHandler = combineRequestHandlers(handler1, handler2, handler3);
+```
+They will be called in the order they are given, so `handler2` can depend on something `handler1` has placed into the `sharedMap`, for example.
 
 ## Order of invocation
 
--- a/packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
@@ -59,7 +59,7 @@ import { combineRequestHandlers } from "@builder.io/qwik-city";
 
 export const onRequest: RequestHandler = combineRequestHandlers(handler1, handler2, handler3);
 ```
-They will be called in the order they are given, so `handler2` can depend on something `handler1` has placed into the `sharedMap`, for example.
+They will be called in the order they are given, so `handler2` can depend on something `handler1` has placed into the `sharedMap`. For example, if you have one middleware to connect to a database, and another to load the current user record, the latter depends on the former, so they must be called in order.
 
 ## Order of invocation
 
--- a/packages/docs/src/routes/api/qwik-city/api.json
+++ b/packages/docs/src/routes/api/qwik-city/api.json
@@ -59,6 +59,20 @@
       "mdFile": "qwik-city.actionstore.md"
     },
     {
+      "name": "combineRequestHandlers",
+      "id": "combinerequesthandlers",
+      "hierarchy": [
+        {
+          "name": "combineRequestHandlers",
+          "id": "combinerequesthandlers"
+        }
+      ],
+      "kind": "Variable",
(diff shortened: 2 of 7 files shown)


packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
```diff
@@ -10,6 +10,7 @@ contributors:
   - gparlakov
   - mrhoodz
   - harishkrishnan24
+  - erikras
 updated_at: '2023-09-18T19:04:44Z'
```
erikras, line 14: Was I supposed to manually update this `updated_at`?

packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
```diff
@@ -51,6 +52,14 @@ export const onDelete: RequestHandler = async (requestEvent) => { ... }
...
 Each middleware function is passed a [`RequestEvent`](#requestevent) object which allows the middleware to control the response.
 
+If you want to combine multiple `RequestHandler`s together, there is a utility function called `combineRequestHandlers()` that will allow you to chain mulitple `RequestHandler`s together.
+```typescript
+import type { RequestHandler } from '@builder.io/qwik-city';
+import { combineRequestHandlers } from "@builder.io/qwik-city";
+
+export const onRequest: RequestHandler = combineRequestHandlers(handler1, handler2, handler3);
+```
+They will be called in the order they are given, so `handler2` can depend on something `handler1` has placed into the `sharedMap`, for example.
```
gioboa, line 63: I read your PR description and the use case is clear to me.
Can you add that description here too pls?

wmertens:
@erikras you're making all the handlers async, can't you use something like maybeThen from qwik to keep everything sync if possible?

packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
erikras: Done.

erikras:
> can't you use something like maybeThen from qwik to keep everything sync if possible?

Maybe, but with all the looping, I'm not sure how to switch to using a callback. 🤷‍♂️

packages/docs/src/routes/docs/(qwikcity)/middleware/index.mdx
gioboa: Thanks

wmertens:
Looks like it supports sync returns https://github.com/BuilderIO/qwik/blob/609d070156794aba1a245841b572465eb001d277/packages/qwik-city/middleware/request-handler/request-event.ts#L74-L81

erikras:
> Looks like it supports sync returns

Oh well. This function is good enough for me in my project(s). I thought I'd offer it to the open source gods. But I can't bring myself to care enough to figure out how to make it work sync. 🤷‍♂️ Up to y'all...

I understand – and _deeply_ appreciate – how perf-driven the Qwik team is, so if this is insufficient, perhaps the existence of this PR and its Google-ability will be enough to help the next person that comes against this problem.

erikras:
btw, @wmertens, I've missed you since the ye olde Redux days. 👋 💚

wmertens:
@erikras actually, if you don't await before returning, doesn't your function boil down to this, and shouldn't it just work?

```tsx
export const combineRequestHandlers =
  (...handlers: RequestHandler[]): RequestHandler =>
  (originalContext) => {
    let lastNext = originalContext.next;
    for (let i = handlers.length - 1; i >= 0; i--) {
      const currentHandler = handlers[i];
      const nextInChain = lastNext;
      lastNext = () => currentHandler({ ...originalContext, next: nextInChain });
    }

    return lastNext();
  };
```

erikras:
The tests pass, but...
Screenshot 2023-12-12 at 6 54 35 PM

Adding `Promise.resolve()` appeases the TS gods, but then isn't it back to being async?
```ts
export const combineRequestHandlers =
  (...handlers: RequestHandler[]): RequestHandler =>
  (originalContext) => {
    let lastNext = originalContext.next;
    for (let i = handlers.length - 1; i >= 0; i--) {
      const currentHandler = handlers[i];
      const nextInChain = lastNext;
      lastNext = () => Promise.resolve(currentHandler({ ...originalContext, next: nextInChain }));
    }

    return lastNext();
  };
```

wmertens:
@mhevery ping, what's your take on this? The types say `next` returns a promise but the reality seems to say that returns a `ValueOrPromise`.
- should we adjust the type of `next`?
- does this nice combiner have a place in the API?

erikras:
In general, I think APIs should have a minimal possible surface area. However, this little code snippet feels very useful. Perhaps for a first-party `@builder.io/qwik-city-utils` library? 🤷‍♂️ Up to Miško.

👋 Hey, Miško! 😃

mhevery:
> In general, I think APIs should have a minimal possible surface area. However, this little code snippet feels very useful. Perhaps for a first-party `@builder.io/qwik-city-utils` library? 🤷‍♂️ Up to Miško.
> 
> 👋 Hey, Miško! 😃

Hey @erikras, Thank you for your PR. Your contribution means a lot.

1. I don't think this is something that we should be adding to our codebase. I think it comes down to the philosophy of the FW, and in this case, the philosophy is that the FW should do the least amount of API surface. Instead, we should encourage the community to create an ecosystem around Qwik, and I think this kind a function would be a good example. In general, if things can be done in the user space, they should, and the FW should only do things that are FW only, and can't be done in user space.
2. I think it limits the ability to create new APIs in the future. For example, One thing I could imagine is that we change our code to take `onRequest` or `onRequest_foo`. This way you can easily compose the API without creating a helper function. If the helper function exists, it would mean that adding `onRequest_foo` would create two different ways to do the same thing.
3. In general, I am extremely cautious (and have a high bar) about expanding the API surface area of FW. Because it means more stuff to document, maintain and ensure that all of the surface area has the same "feel". So the requirements for new API is that a) it can only be done in FW and b) it has a strong need. 


So I don't think we should merge this. Perhaps a new 3rd party lib? OR maybe we discuss the idea of `onRequest_foo`?

gioboa:
we can add it to the [cookbook section](https://qwik.builder.io/docs/cookbook/) b/c I think it's a nice piece of code.

Issue #5460, opened 2023-11-19
[✨] Remove this warning: `routeLoader$()` can only be declared in `layout.tsx`, `index.tsx` and `plugin.tsx` inside the src/routes directory


Nefcanto:
### Is your feature request related to a problem?

Yes,

we want to reuse some `routeLoader$` functions across multiple routes, and in many projects.

This means that we create a file called `Loader.jsx` and have a `routeLoader$` in it. Yet it's not used directly.
It exports the function, and then we import it in another place:

For example, this is our `BlogLoader.jsx` file:

```
import { routeLoader$ } from '@builder.io/qwik-city'
import {
    getFromCacheOrApi,
    useAsync
} from 'Base'
import { getLayout } from 'Contents'
import { getForm } from 'Forms'
import { getGlobalization } from 'Globalization'

const loadPost = routeLoader$(async (props) => {
    const {
        params,
        status,
        query
    } = props
    const { slug } = params || {}
    const locale = query?.get('locale')
    const [
        data,
        layout,
        globalization,
        form
    ] = await useAsync([
        getFromCacheOrApi(`/blogPost/get?slug=${slug}&locale=${locale}`),
        getLayout('post', props),
        getGlobalization(props),
        getForm(`comment`, props)
    ])
    if (data.error?.code === "404") {
        return status(404)
    }
    if (data.statusCode) {
        return status(data.statusCode)
    }
    const { content, ...contentLessForm } = form
    return {
        ...data,
        ...layout,
        ...globalization,
        ...contentLessForm
    }
})

export default loadPost
```

This loader can be used in `blog/index.jsx` or `artciles/index.js` or `news/index.jsx`. For example, this is our `blog/index.jsx` in our Blog module:

```
import { component$ } from '@builder.io/qwik'
import { NotFound } from 'Base'
import { useSeo } from 'Seo'
import {
    Layout,
    loadBlog,
} from 'Blog'
import { loadBlog as runnableLoader } from 'Loaders'
import { Layout as RunnableLayout } from 'BlogParts'

const Blog = component$(() => {

    const data = loadBlog().value
    const extraData = runnableLoader().value
    if (data === 404) {
        return <NotFound />
    }


    const {
        categories,
        latest,
        mostViewed,
        popular,
        posts,
        seo,
        tags,
    } = data || {}

    return RunnableLayout
        ?
        <RunnableLayout {...data} />
        :
        <Layout {...data} />
})

export default Blog

export { loadBlog }

[...]

PR #5563, opened 2023-12-11
fix: use existing local .eslintrc configuration

### LINKED ISSUE #5555: [🐞] console "[vite] warning :..." messages still show up even when eslint rule is disabled

Qwik Rollup / Vite plugin

### Describe the bug

Some Eslint rules appear in the dev console as "[vite] warning : ..." messages. This has gotten especially annoying with the new "qwik/no-use-visible-task" eslint rule.

Ex:

> 10:27:45 PM [vite] warning: Qwik tries very hard to defer client code execution for as long as possible.
This is done to give the end-user the best possible user experience.
Running code eagerly blocks the main thread, which prevents the user from interacting until the task is finished.
"useVisibleTask$" is provided as an escape hatch.

> When in doubt, instead of "useVisibleTask$()" use:
> - useTask$ -> perform code execution in SSR mode.
> - useOn() -> listen to events on the root element of the current component.
> - useOnWindow() -> listen to events on the window object.
> - useOnDocument() -> listen to events on the document object.

> Sometimes it is the only way to achieve the result.
In that case, add:
// eslint-disable-next-line qwik/no-use-visible-task
to the line before "useVisibleTask$" to acknowledge you understand.

  > Plugin: vite-plugin-qwik
  > File: /home/maieul/dev/vite-warnings-repro/src/routes/index.tsx:5:3
  3  |  
  4  |  export default component$(() => {
  5  |    useVisibleTask$(() => {
     |     ^
  6  |      console.log("Hello World!");
  7  |    });
^C⏎          

### Reproduction

https://github.com/maiieul/qwik-vite-warnings-repro

### Steps to reproduce

- pnpm dev
- load the page to see the warning

### System Info

```shell
System:
    OS: Linux 6.6 Fedora Linux 38 (Workstation Edition)
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i5-1340P
    Memory: 16.89 GB / 31.06 GB
    Container: Yes
    Shell: 3.6.1 - /usr/bin/fish
  Binaries:
    Node: 20.9.0 - /usr/local/bin/node
    npm: 10.1.0 - /usr/local/bin/npm
    pnpm: 8.10.5 - ~/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 120.0.6099.71
  npmPackages:
    @builder.io/qwik: ^1.3.0 => 1.3.0 
    @builder.io/qwik-city: ^1.3.0 => 1.3.0 
    undici: ^5.26.0 => 5.28.2 
    vite: ^5.0.6 => 5.0.7
```

### Additional Information

I think those kinds of vite warnings shouldn't appear in the dev console at all since this should be the job of eslint.

gioboa:
Closed #5555 #5460

With this PR, priority is given to the configuration that eslint can load from the project.
if it is invalid or does not exist `eslintConfig.parser === null` then we fallback to the one we have always used.

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/eslint-plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/eslint-plugin.ts
@@ -12,31 +12,37 @@ export async function createLinter(
   tsconfigFileNames: string[]
 ): Promise<QwikLinter> {
   const module: typeof import('eslint') = await sys.dynamicImport('eslint');
-  const options: ESLint.Options = {
-    cache: true,
-    useEslintrc: false,
-    overrideConfig: {
-      root: true,
-      env: {
-        browser: true,
-        es2021: true,
-        node: true,
-      },
 
-      extends: ['plugin:qwik/recommended'],
-      parser: '@typescript-eslint/parser',
-      parserOptions: {
-        tsconfigRootDir: rootDir,
-        project: tsconfigFileNames,
-        ecmaVersion: 2021,
-        sourceType: 'module',
-        ecmaFeatures: {
-          jsx: true,
+  let eslint = new module.ESLint({ cache: true }) as ESLint;
+  const eslintConfig = await eslint.calculateConfigForFile('no-real-file.tsx');
+  const invalidEslintConfig = eslintConfig.parser === null;
+
+  if (invalidEslintConfig) {
+    const options: ESLint.Options = {
+      cache: true,
+      useEslintrc: false,
+      overrideConfig: {
+        root: true,
+        env: {
+          browser: true,
+          es2021: true,
+          node: true,
+        },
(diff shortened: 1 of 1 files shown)


PatrickJS (approved):

PatrickJS:
can we tell the dev if their config is invalid

maiieul (approved):
Tested on the 𝕏 stack. Looks like it's working :+1:

gioboa:
> can we tell the dev if their config is invalid (but only if its invalid and exists)

eslint config is invalid
- if there is no parser object in the config
- if there is no eslint config file
- other cases that I didn't explore

so we need to test all the cases to print a good message in the terminal.

This is an internal linter, I don't think we need to bug the developer for that.

PR #5601, opened 2023-12-16
fix: Bun integration env.get is undefined

phyrog:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Fixes this error when using `req.env` with Bun (e.g. for qwik-auth):

```
TypeError: req.env.get is not a function. (In 'req.env.get("AUTH_SECRET")', 'req.env.get' is undefined)
```

# Checklist:

### DIFF
--- a/packages/qwik-city/middleware/bun/index.ts
+++ b/packages/qwik-city/middleware/bun/index.ts
@@ -75,7 +75,11 @@ export function createQwikCity(opts: QwikCityBunOptions) {
         mode: 'server',
         locale: undefined,
         url,
-        env: Bun.env,
+        env: {
+          get(key) {
+            return Bun.env[key];
+          },
+        },
         request,
         getWritableStream: (status, headers, cookies, resolve) => {
           const { readable, writable } = new TransformStream<Uint8Array>();


packages/qwik-city/middleware/bun/index.ts
```diff
@@ -75,7 +75,11 @@ export function createQwikCity(opts: QwikCityBunOptions) {
         mode: 'server',
         locale: undefined,
         url,
-        env: Bun.env,
+        env: {
+          get(key) {
```
Mouvedia, line 79: question

Shouldn't you `...Bun.env` before?
Shouldn't you use `Object.defineProperty()` so that `writable` can be set to `false` for `get`?

phyrog: Don't know about setting writable to false, but this is exactly how it is implemented for node too: https://github.com/BuilderIO/qwik/blob/main/packages%2Fqwik-city%2Fmiddleware%2Fnode%2Fhttp.ts#L85-L89

I don't think we need to add `...Bun.env` either, `env.get` is the only documented way to read environment variables.

gioboa:
Thanks @phyrog 💪 
This fix looks great. did you try this implementation locally?

phyrog:
> Thanks @phyrog 💪 
> This fix looks great. did you try this implementation locally?

Yes I did.

phyrog:
@gioboa Really? I'm just getting an empty page when following your instructions (works when running in dev, though).


Here's my system specs:
```
  System:
    OS: macOS 13.3.1
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 17.72 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.4.0 - /usr/local/bin/node
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 8.12.1 - ~/Library/pnpm/pnpm
    bun: 1.0.18 - ~/.bun/bin/bun
  Browsers:
    Chrome: 120.0.6099.109
    Safari: 16.4
  npmPackages:
    @builder.io/qwik: ^1.3.1 => 1.3.1
    @builder.io/qwik-city: ^1.3.1 => 1.3.1
    undici: ^5.26.0 => 5.28.2
    vite: ^5.0.6 => 5.0.10
```

phyrog:
Ok so I debugged this a bit more:

The current version works for `routeLoader$` (with the above mentioned issue that it actually does not read the variable from env at runtime), but neither for `routeAction$`, nor `globalAction$` (which is used by qwik-auth).

phyrog:
@gioboa I also looked more into why it works for `routeLoader$`, turns out when you `add bun`, it automatically configures SSG, which explains why the env variable only changes after a rebuild. It is actually built into `dist/q-data.json`. When you disable SSG in `adapters/bun/vite.config.ts`, it will show the same behavior as `routeAction$` and `globalAction$`, i.e. `env.get is not a function. (In 'env.get("TEST")', 'env.get' is undefined)`.

gioboa:
Thanks @phyrog for your commitment. Your change is working fine.

---
I have this other error in the terminal but it's not related to your change.
Do you have the same?
Screenshot 2023-12-18 at 21 18 28

phyrog:
@gioboa No I don't think I have seen this one

PR #5602, opened 2023-12-17
fix: multiple cookies in qwik-auth

ksjogo:
# Overview

https://github.com/BuilderIO/qwik/pull/5194 introduced functionality to set multiple cookies.
With fetch setting multiple cookies via the set-cookie key is not supported though.
The testing was limited to the PCKE cookie, which passed as that cookie is last in the list coming from auth.js.
Other cookies like next-auth.callback-url would come earlier and would have been discarded by the later cookie.
 

I am not sure, if we should handle pre-existing values for set-cookies?

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

Instead of setting cookies directly, instead collect them and then create a semicolon seperated string.

# Checklist:

### DIFF
--- a/packages/qwik-auth/src/index.ts
+++ b/packages/qwik-auth/src/index.ts
@@ -140,15 +140,21 @@ async function authAction(
     ...authOptions,
     skipCSRFCheck,
   });
+
+  const cookies: string[] = [];
   res.headers.forEach((value, key) => {
-    /**
-     * Do not set the header if already set accept in the case of set-cookie which is allowed
-     * https://httpwg.org/specs/rfc6265.html#rfc.section.3
-     */
-    if (!req.headers.has(key) || key === 'set-cookie') {
+    if (key === 'set-cookie') {
+      // while browsers would support setting multiple cookies, the fetch implementation does not, so we join them later.
+      cookies.push(value);
+    } else if (!req.headers.has(key)) {
       req.headers.set(key, value);
     }
   });
+
+  if (cookies.length > 0) {
+    req.headers.set('set-cookie', cookies.join(', '));
+  }
+
   fixCookies(req);
 
   try {


Lemon2311 (approved):
Overview:
The pull request intends to consolidate multiple Set-Cookie headers into a single header due to limitations in the fetch implementation. The loop correctly identifies Set-Cookie headers and accumulates them in the cookies array. Non-cookie headers are managed appropriately, avoiding potential duplication. Combining cookies into a single header is implemented as intended. This approach could be necessary given the mentioned limitation but requires careful consideration of how combined cookies are parsed on the client side.
Potential Issues:     
Cookie Delimiter: Using a comma to separate multiple cookies may cause issues if cookie values themselves contain commas.
    Standards Deviation: The standard way to handle multiple cookies is with multiple Set-Cookie headers. This deviation should be documented, explaining the limitation that led to this design choice.
Improvements:
 Additional comments may be beneficial to explain the implications of the chosen workaround for future maintenance.
Addition comments:
Some testing and validation are recommended to ensure compatibility and correctness.

ulic75:
I think this is an issue with the Fetch API.  Perhaps their addition of the [getSetCookie()](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) method will resolve your use case?

ksjogo:
> I think this is an issue with the Fetch API. Perhaps their addition of the [getSetCookie()](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) method will resolve your use case?

Not sure if it is specifically my use case. It seems to be broken with any sitatuation of having a csrf cookie + anything.

Do you know if the getSetCookie array is writable? The doc doesn't seem to say that.

Similar discussion on next: https://github.com/vercel/next.js/issues/54033

ksjogo:
next-auth itself seems to be using this to handle multiple: https://github.com/nextauthjs/next-auth/blob/6c958d7e10edabb2915e101096e1dba5e22fff6b/packages/core/src/lib/utils/web.ts#L94

ksjogo:
Who is leading qwik-auth at the moment/whom could I have a chat with to get this completed? Would prefer to path to be in mainline so it's easier to handle on our end.

wmertens:
@ksjogo could you rebase? The tests are hanging and I can't restart them.

ksjogo:
> @ksjogo could you rebase? The tests are hanging and I can't restart them.

Done.

wmertens:
Thanks!

PR #5618, opened 2023-12-21
feat: qwik-prefetch-worker

mhevery:
# New implementation of SW (Service Worker)

Issues with old implementation:
- Part of Qwik-city therefore it can be used in qwik-only solutions such as  Astro
- Only one SW can be per page. If MFE (micro frontends) than we can't prefetch each MFE

# How to use

1. Include a single `<PrefetchServiceWorker />` somewhere on the page. This is usually in `root.tsx` but can be a different place, especially if MFE.
2. For each MFE include `<PrefetchGraph />`

Example of what a simple application may look like:
```html
      <head>
        ...
        <PrefetchServiceWorker />
        <PrefetchGraph />
      </head>
```
That is it! Happy coding.


thejackshelton:
> as opposed to the simple application example, do you have an example of how to integrate this when you have MFEs and different languages? For example we currently have an Angular app hosting our Qwik MFE and we plan to introduce additional Qwik MFEs before we change the host later:
> 
> ```
> <head>
>   <title>Angular app</title>
> </head>
> <body>
>   <my-angular-app />
>   
>   
>   <iframe src="qwik-app1.com">
>     <head>
>       <title>Qwik App 1</title>
>       <PrefetchServiceWorker />
>       <PrefetchGraph />
>     </head>
>     ...
>   </iframe>
> 
>   
>   <iframe src="qwik-app2.com">
>     <head>
>       <title>Qwik App 2</title>
>       <PrefetchServiceWorker />
>       <PrefetchGraph />
>     </head>
>     ...
>   </iframe>
> </body>
> ```
> 
> Or should the PrefetchServiceWorker exist in the host only, and the iframes only have the PrefetchGraph? What about host apps that aren't JSX like Angular

There should only be one prefetch service worker on the page, and then the prefetch graphs in each MFE. As for apps that aren't JSX I'm not sure 🤔 , maybe taking the js of the prefetch graph and adding that to the angular parts? 😅

PR #5632, opened 2023-12-23
feat(route-matcher): optional param

samyung0:
# Overview

Adds optional param for route matching

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Draft Sytax: `[param.]`

This is my ever first pull request and I apologize if I messed anything up.

Currently `[...param]` matches zero or more than 1 segments while [param] matches exactly one segment. This might not cover all the use cases. Suppose I am integration i18n into my app and I put all routes under `[...lang]`. Then `/`, `/en-US/`, etc are all valid routes and works perfectly. However, visiting an invalid route will also be rendered by `[...lang]/index.tsx` instead of showing the 404 page.

One of the solution is to use `[lang]` instead of [...lang] and redirect `/` to a default locale but redirecting home page seems a bit sketchy to me (maybe implications towards SEO?).

Second solution is to analyze the path in `[...lang]/index.tsx`, if the first segment is not a supported locale, then throw a redirect or error. I haven't tested this yet but it seems like a workaround instead of a good solution.

 The current draft for the optional param syntax is `[lang.]` and it matches exactly one or zero path segment. With optional param, any invalid route with more than 1 segment will render the 404 page without having to manually check the path.

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/route-matcher.ts
+++ b/packages/qwik-city/runtime/src/route-matcher.ts
@@ -7,10 +7,7 @@ import type { PathParams } from './types';
  * @param path Actual path to match
  * @returns Returns PathParams or null if did not match.
  */
-export function matchRoute(
-  route: string,
-  path: string,
-): PathParams | null {
+export function matchRoute(route: string, path: string): PathParams | null {
   const routeIdx: number = startIdxSkipSlash(route);
   const routeLength = lengthNoTrailingSlash(route);
   const pathIdx: number = startIdxSkipSlash(path);
@@ -24,7 +21,7 @@ function matchRoutePart(
   routeLength: number,
   path: string,
   pathIdx: number,
-  pathLength: number,
+  pathLength: number
 ): PathParams | null {
   let params: PathParams | null = null;
   while (routeIdx < routeLength) {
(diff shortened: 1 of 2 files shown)


wmertens:
I suppose this could be useful although the language usecase is not as hard as you make it out to be https://github.com/wmertens/compiled-i18n/blob/HEAD/qwik.md#route-based-locale-selection

In any case I'm not sure of the syntax, because `[...params]` means that `params` is a variable name, and `[lang.]` means the word "lang" might be in there.

samyung0:
Sorry for the confusion, [lang.] means that a param named lang might be there, I can change the name to [param.] and the name will change to param.
I do agree that i11n is not the best example here since mostly people make the [lang] part required instead of optional anyway, but I hope this can be useful in some other ways :)

wmertens:
Oh hmm, so it's a wildcard match on anything, and if there's not enough path to do the match, it is left out?

That seems like it might cause subtle bugs, no?

I'd like to postpone this until we have a real usecase.

Also, how about this syntax: `<string or [param]>[0,1]` ?

samyung0:
I agree to leave it open until there is a real use case that requires it. Currently the matching route fn will try to match the route _without_ the [param.] first before going back to matching the route with the param. It also takes precedence over catchall routes. It could introduce some subtle bugs but unlikely, but I am totally fine with leaving it open at the moment.

wmertens:
Paging @mhevery to weigh in on the API implications

wmertens (changes requested):
marking this as Change Requested because the syntax isn't ok right now and we're not sure if we want to add it.

wmertens:
I'm going to close this for now to keep the PR list relevant.

If a usecase pops up, it should be findable in the closed PR list.

PR #5633, opened 2023-12-23
docs: pnpm contributors

AmirSa12:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/community/groups/index.mdx
+++ b/packages/docs/src/routes/community/groups/index.mdx
@@ -14,7 +14,8 @@ contributors:
   - apherio
   - salamaashoush
   - adamdbradley
-updated_at: '2023-10-15T10:09:17Z'
+  - AmirSa12
+updated_at: '2023-12-07T19:48:02Z'
 created_at: '2023-02-03T20:09:34Z'
 ---
 import ImgDiscord from '~/media/ecosystem/discord.svg?jsx';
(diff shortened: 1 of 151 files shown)


AmirSa12:
this script should get executed manually as it's not automated yet. 
So, the maintainers/contributors should run this script to keep the frontmatter of the docs updated. we are finding a way to make this script automatically executed.

gioboa (changes requested):
Thanks @AmirSa12 
I have one concern, now you are in all the files even if you didn't change them all I guess
Can you verify this behaviour?

AmirSa12:
> Thanks @AmirSa12 
> 
> I have one concern, now you are in all the files even if you didn't change them all I guess
> 
> Can you verify this behaviour?

Yes I didn't change all of them, but I think it's because of the rebase, I'm in all the old doc's files. Not the new docs which were recently added after the rebase. I'm still not sure if it's really because the rebase, but I think it's the reason the script adds me.

gioboa:
I would like to have a review from a core team member b/c change these file over and over with a PR is not a good thing to me. A specific GitHub action for this work seems better.

AmirSa12:
> I would like to have a review from a core team member b/c change these file over and over with a PR is not a good thing to me. A specific GitHub action for this work seems better.

I totally agree, I've told @mhevery about this case and  the GHA idea, and a friend and I are still working on it.

gioboa:
Ok, so I think better thing to do is close this one and wait for the GitHub action one.

PatrickJS:
closing this because of this https://github.com/BuilderIO/qwik/pull/5633#issuecomment-1869442122 comment. If you think this should be open feel free to open the PR again

PR #5455, opened 2023-11-18
feat: add qwik/no-use-visible-task eslint rule

gioboa:
We did a pair programming PR

Co-authored-by: Matteo Pietro Dazzi <ilteoood@users.noreply.github.com>
Co-authored-by: Pasquale De Lucia <pasquale.delucia96@gmail.com>
Co-authored-by: Gloria Giannascoli <gloriag-dev@users.noreply.github.com>
Co-authored-by: pietrodev07 <pietrodev07@users.noreply.github.com>

# Overview

# What is it?

- [X] Feature / enhancement

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

- [ ] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] Added new tests to cover the fix / functionality

PR #5638, opened 2023-12-26
feat(eslint): rule for sync$()

Lemon2311:
@gioboa send me as an example merge #5455, but since then  @mhevery in #5572 refactored the code as all the ruleTesters were just placed inside qwik.unit.ts. I made this changes, implemented the noSyncClosure rule, but I can`t seem to find where he moved the ruleTester implemetations so that I can include my test for the noSyncClosure rule.


packages/eslint-plugin-qwik/src/noSyncClosure.ts
```diff
@@ -0,0 +1,48 @@
+const { Rule } = require('eslint');
+
+const isBrowserAPI = (variable, context) => {
+  const browserAPIs = new Set([
+    'window', 'document', 'navigator', 'location', 'history', 'localStorage',
+    'sessionStorage', 'alert', 'fetch', // ... other known browser APIs
+  ]);
+
+  return context.getScope().through.some(scopeVar => 
+    scopeVar.identifier.name === variable && browserAPIs.has(variable)
```
wmertens, line 10: Won't this run into trouble when you do `{ const document = "foo"; sync$(() => document)}`?

Since all these are global and globals aren added to the scope array ([see the bundles here](https://qwik.dev/playground/#f=Q0405gbYEo2OAijzUjHtgCrMxLwAYOGhYlsNNhuoCqgoBZor9BxDQoI8nUJDXOP9%2FF1cNWtHE9swSWwA)), would it instead be possible to just check if it's using only globals?

You could even check if they're browser globals but since it would crash immediately for undefined globals I think that's not necessary.

gioboa:
Hi @Lemon2311 are you still working on this or is it abandoned?

Lemon2311:
Hello @gioboa, I haven\`t looked into this more since this draft. I saw u assigned @ahmaddonishyar to the issue so I suppose not for now, if he doesn\`t fix it I can give it another go then.

gioboa:
@Lemon2311do you need an help to finalise this PR?

Lemon2311:
@gioboa if you want you can continue from where I left off, if not i can look into it in a few days.

gioboa:
Unfortunately I don't have the possibility to look at it in the next days, I'm so sorry

PR #5646, opened 2023-12-28
docs: fix entry file name

AustinGil:
# Overview

Simple change to the docs based on behavior I found in the CLI

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Docs currently mention that `createQwikCity` exists in `/src/entry.[server].tsx`, but actual behavior is in `/src/entry.preview.tsx`

I'm not sure about some of the other references in the rest of the page.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/deployments/node/index.mdx
+++ b/packages/docs/src/routes/docs/deployments/node/index.mdx
@@ -78,7 +78,7 @@ By default, all Qwik City applications are protected against [CSRF attacks](http
 
 This protection is enabled by default and it's the reason why you need to set the `ORIGIN` environment variable when deploying your application for production.
 
-If you want to disable CSRF protection, you can set `checkOrigin: false` in the `createQwikCity()` options in `src/entry.preview.tsx`:
+If you want to disable CSRF protection, you can set `checkOrigin: false` in the `createQwikCity()` options in `src/entry.preview.tsx` or `src/entry.[server].tsx`:
 
 ```tsx {6} /checkOrigin/ title="entry.preview.tsx"
 // ...
--- a/packages/docs/src/routes/docs/deployments/node/index.mdx
+++ b/packages/docs/src/routes/docs/deployments/node/index.mdx
@@ -78,9 +78,9 @@ By default, all Qwik City applications are protected against [CSRF attacks](http
 
 This protection is enabled by default and it's the reason why you need to set the `ORIGIN` environment variable when deploying your application for production.
 
-If you want to disable CSRF protection, you can set `checkOrigin: false` in the `createQwikCity()` options in `src/entry.express.tsx`:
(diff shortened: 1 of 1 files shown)


gioboa (changes requested):
Thanks for your help @AustinGil 
If we are using node deployment, `entry.[server].tsx` comes in to play during production build so the docs is correct. A good improvement in this page is add `deno` and `node-server` in this phrase.
```
In order to deploy the server, you need to run the server/entry.[server].js file in the server of your choice, where [server] can be [express](https://expressjs.com/) or [fastify](https://fastify.dev/) **[<here>]**
```
and change `entry.express` with `entry.[server]` everywhere

AustinGil:
@gioboa I understand that the production build setup produces a `entry.[server].tsx`, but that's actually where I want to suggest a fix. This PR is specifically concerning the CSRF protection. The docs say that if you want to modify the `createQwikCity` options, you need to modify the `src/entry.experess.tsx` file (for example). What I'm saying is that `createQwikCity` does not exist in that file at all. In it's **current** form, the CLI generates a `/src/entry.preview.tsx` file that contains the `createQwikCity`. Therefore, that's the file that should be referenced in the docs, and I don't think it matters which server you actually choose.

nelsonprsousa:
> @gioboa I understand that the production build setup produces a `entry.[server].tsx`, but that's actually where I want to suggest a fix. This PR is specifically concerning the CSRF protection. The docs say that if you want to modify the `createQwikCity` options, you need to modify the `src/entry.experess.tsx` file (for example). What I'm saying is that `createQwikCity` does not exist in that file at all. In it's **current** form, the CLI generates a `/src/entry.preview.tsx` file that contains the `createQwikCity`. Therefore, that's the file that should be referenced in the docs, and I don't think it matters which server you actually choose.

Actually, if you follow that page, you will end up with `src/entry.express.tsx` or `src/entry.fastify.tsx` files.

So, for someone actually reading and following this page from the start, it makes sense to have `src/entry.[server].tsx`, IMO. We can instead add a note regarding the `/src/entry.preview.ts` file. WDYT?

AustinGil:
Yeah, I get that the CLI generates a `/src/entry.fastify.tsx` file, but this PR is around the documentation for the CSRF protection (https://qwik.builder.io/docs/deployments/node/#csrf-protection) which says you have to change the options passed to `createQwikCity` in the `/src/entry.express.tsx` file, but this function call no longer exists in this file when you use the CLI. The function is used in `/src/entry.preview.tsx`. So the `/src/entry.[server].tsx` files are not really relevant in this context. In other words, the current state of the docs referecing the Express file is incorrect, but changing it to [server].tsx would also be incorrect.

Anyway, this was from my experience generating from the CLI at the time of opening the PR, but I think it used to be the other way.

PR #5174, opened 2023-09-15
docs: added Package manager tabs

the-r3aper7:
# Overview

# What is it?

image
image

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests / types / typos

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

- [ ] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] Added new tests to cover the fix / functionality

PR #5649, opened 2023-12-29
docs: added package manager tabs

the-r3aper7:
# Overview

#5174 

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


the-r3aper7:
should i add `@qwik-ui/headless` to vite optimization dependencies. @gioboa

gioboa:
No, the latest `@qwik-ui/headless` solves the big bundle size issue.

the-r3aper7:
some CI tests failed how can i make it right?

packages/docs/src/components/package-manager-tabs/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+      cookie = cookie.substring(1, cookie.length);
+    }
+    if (cookie.indexOf(cookieName) == 0) {
+      return cookie.substring(cookieName.length, cookie.length);
+    }
+  }
+  return null;
+};
+
+const setCookie = (name: string, value: string, options: CookieOptions = {}) => {
```
gioboa, line 28: Can you read cookie inside routeLoader$ pls?
--
We already have a setCookie method here `packages/docs/src/routes/(shop)/utils.ts`.
Can you create a common method to manage both cases pls?

the-r3aper7: i will make those changes by tomorrow.

gioboa:
I pushed a small change. Thanks for your commitment @the-r3aper7

the-r3aper7:
> I pushed a small change. Thanks for your commitment @the-r3aper7 

thanks i like that icon change.

the-r3aper7:
@gioboa i have make changes regarding to get the cookie but i am able to think of getting `setCookie` correct as we need to pass `Cookie` object to setCookie and  should i the `setCookie` function in `shop/utils` WDYT?

the-r3aper7:
@gioboa i have tried basic `document.cookie` in `setCookie` it does not work with the shop as it is not able to set `cartId` in cookies.

wmertens:
@the-r3aper7 So you're having trouble with the setCookie function? Can you explain more, and rebase?

the-r3aper7:
> I'm confused about your setCookie function that takes 3 arguments, the third called options, but you're passing a string

i am directly adding cookie to the browser. am i misinterpreting something?

packages/docs/src/components/package-manager-tabs/index.tsx
wmertens: Still relevant.

wmertens (commented):
I don't see you reading the package manager cookie to set the context for SSR of the package manager?

gioboa:
Closed by #6636 6636

PR #5651, opened 2023-12-30
feat(qwik-react): directive aliases for Astro and other package conflicts

thejackshelton:
This PR adds aliases for the directives so that they can be used in `.astro` files.

But, it can also be used in other places where packages might have `client:*` directives that conflict.

### DIFF
--- a/packages/qwik-react/src/react/types.ts
+++ b/packages/qwik-react/src/react/types.ts
@@ -14,7 +14,7 @@ export interface QwikifyBase {
    * **Use case:** Immediately-visible UI elements that need to be interactive as soon as possible.
    */
   'client:load'?: boolean;
-  'qwik-client:load'?: boolean;
+  'qwik:load'?: boolean;
 
   /**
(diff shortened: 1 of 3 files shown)


gioboa (approved):
👏 @thejackshelton nice change 🚀

packages/qwik-react/src/react/types.ts
```diff
@@ -14,6 +14,7 @@ export interface QwikifyBase {
    * **Use case:** Immediately-visible UI elements that need to be interactive as soon as possible.
    */
   'client:load'?: boolean;
+  'qwik-client:load'?: boolean;
```
mhevery, line 17: why `qwik-client` Why no just `qwik`? 🤷‍♂️

thejackshelton: I thought it would be less clear that it is a hydration directive, but after getting some feedback changing it to just `qwik`

PR #5669, opened 2024-01-06
fix(starter): rename tailwind postcss.config file

iamriajul:
The new vite 5 project is module type by default.

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

As we know we've changed our default vite version to vite 5 and vite 5 is mostly should be "type": "module", and postcss require's it's configuration to be exposed via common js, so that's why we need to explicitly declare the postcss.config file's type to common js by using .cjs extension.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


gioboa (approved):
Thanks @iamriajul you rock

PR #5671, opened 2024-01-07
feat: Slot default value

hassanzohdy:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Now `Slot` component can receive default value if the component did not receive any children, this work with named slots as well.

Example of usage:

```tsx
import { Slot, component$ } from "@builder.io/qwik";

const SaveButton= component$(() => {
  return (
    <button>
      <Slot>Save</Slot>
    </button>
  );
});

  return (
    // <button>Save</button>
    <SaveButton /> 
    // <button>Update</button>
    <SaveButton>Update</SaveButton>
  )
});
```

Works as well with named slots:

```tsx

import { Slot, component$ } from "@builder.io/qwik";

const SaveButtonWithIcons = component$(() => {
  return (
    <button>
      <Slot name="left">
        <svg width="24" height="24" viewBox="0 0 24 24">
          <path
            fill="currentColor"
            d="M21 3H3a2 2 0 00-2 2v14a2 2 0 002 2h18a2 2 0 002-2V5a2 2 0 00-2-2zM3 5h18v14H3V5zm9 4v6h6V9h-6z"
          ></path>
        </svg>
      </Slot>
      <Slot>Save</Slot>
      <Slot name="right">
        <svg width="24" height="24" viewBox="0 0 24 24">
          <path
            fill="currentColor"
            d="M21 3H3a2 2 0 00-2 2v14a2 2 0 002 2h18a2 2 0 002-2V5a2 2 0 00-2-2zM3 5h18v14H3V5zm9 4v6h6V9h-6z"
          ></path>
        </svg>
      </Slot>
    </button>
  );
});

  return (
    // <button>Save</button>
    <SaveButtonWithIcons /> 
    // <button>Update</button>
    <SaveButtonWithIcons >
       <span>Update</span>
       <div q:slot="left">Left Arrow</div>
</SaveButtonWithIcons >
  )
});
```

# Use cases and why

A good use case for this feature for package authors, for example when creating a Carousel component, we can allow the developer to define custom  left and right arrows, otherwise we can use the default value.

# Checklist:


wmertens (changes requested):
This approach needs to be different:
- render fallbacks in templates
- add support in the DOM renderer, not just in SSR

Note that this also means that if you have a list of 1000 elements with slot fallbacks that aren't used, you still have those fallbacks in the DOM.

So perhaps it's better to re-render the slot parent if the child goes away, but that can already be achieved by using Inline Components to render the fallback content if they don't get children.

wmertens:
I'm going to close this because the current approach won't work and we'll have it in v2 which is coming along nicely.

PR #5702, opened 2024-01-12
chore: Remove undici, set minimum node version to 18.17

wmertens:
`sharp` requires "^18.17.0 || ^20.3.0 || >=21.0.0", and that's the cause of installation issues.

Furthermore, global fetch is in node since v18, so we don't need undici any more.

Bun and Deno also include fetch.


devcaeg:
@wmertens 

When will this change be merged?

wmertens:
@manucorporat the macro plugin keeps erroring on insights/src/root.tsx, even though it hasn't changed. It says the code is unreachable. Any idea what could cause this?
This single commit PR must be the reason because I keep rebasing on main which has no error.

PatrickJS:
can we change the min version to node 20? 😁 20 is a lot faster

gioboa:
is this PR still valid or is it abandoned ?

devcaeg:
> is this PR still valid or is it abandoned ?

If I am not mistaken they are waiting for version 2.0.

PR #5709, opened 2024-01-12
Fix: losing this context when multiple server functions are calls in one useResource$ hook

### LINKED ISSUE #4417: [🐞] `this` is undefined in server$ function when called consequentially.

hassanzohdy:
# Overview

So, this PR fixes losing `this` context in server functions if they are called inside `useResource$` hook (#4417) and also introduce a new helper function to the passed object to the resource callback function called `call`.

The idea is to keep tracking of current context of the resource when initially called and use it in any inner function calls that uses promises to perform some actions (i.e calling multiple server functions sequentially).

### Usage
```tsx
import { server$ } from "@builder.io/qwik-city";
import {
  $,
  Resource,
  component$,
  useResource$,
} from "@builder.io/qwik";

const sleep = async (ms: number) =>
  new Promise((resolve) => setTimeout(resolve, ms));

const serverFunctionA = server$(async function SererA() {
  await sleep(4000);

  return "A";
});

const serverFunctionB = server$(async function ServerB(name: string) {
  await sleep(3000);

  return name;
});

const serverFunctionC = server$(async function ServerC() {
  await sleep(2000);

  return "C";
});

export const ResCaller= component$(() => {
  const resource = useResource$(
    $(async ({ call }) => {
    /**
     * The call function is used to keep track of the current context it accepts any function that
     * will be executed and rest of params are passed as arguments to the function using `call`
     * function is only needed when there functions (i.e server functions) that run sequentially and
     * need to be executed in the same context
     */
      const a = await call(serverFunctionA);
      const b = await call(serverFunctionB, "Hello");
      const c = await call(serverFunctionC);

      return { a, b, c };
    }),
  );

  return (
    <Resource
      value={resource}
      onResolved={(d) => {
        console.log("Resolved!", d);

        return <div>Done</div>;
      }}
    />
  );
});
```

I've tested it in multiple concurrent requests and it works fine.

# What is it?

- [ x ] Feature / enhancement
- [ x ] Bug

# Description

This PR fixes #4417 when server functions are called in sequence in useResource$ hook.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


gioboa:
Thanks @hassanzohdy 
Did you try with concurrent request?
This kind of implementation with a global variable is not "request safe". multiple requests at the same time can recreate the problem because the global variable mess us things.

hassanzohdy:
@gioboa You're right, the previous solution failed with concurrent requests, i fixed it by introducing a new helper function to the resource object called `call` which will be mainly needed when sequential functions are needed to be called, it keeps the resource `context`.

If everything is okay, i'll update the docs to highlight the `call` function in the `useResource$` docs.

wmertens:
@hassanzohdy unfortunately invoke isn't magic either, as soon as the called function goes async things are broken again.

The solution is using async localstorage and I started a PR but didn't finish it.

hassanzohdy:
@wmertens the idea here is simple, each time an async function will be called in sequence, it should be wrapped with `call` helper function, it will use the same context by calling the async function and will not lose the context.

wmertens:
whenever a server function awaits, another server function can be called and it will change the global context that invokeFn maintains. You're just moving the problem.

If invokeFn worked, it would already work.

See https://github.com/BuilderIO/qwik/pull/5510

wmertens (changes requested):
As discussed, cannot be merged with the current implementation

PR #5741, opened 2024-01-18
fix(ssr): dirty tasks

### LINKED ISSUE #5738: [🐞] computed signals do not get serialized with refs

Qwik Runtime

### Describe the bug

It seems that Qwik does not register when the computed signal should be serialized.

According to wout, this might be fixed in v2?

image

### Reproduction

https://qwik.builder.io/playground/#v=1.3.5&f=Q0o0xhbG2BKNDqL2BTPhiVlHgfIEBQmd3Mqg1DRg8CCqeY8QXx8ncDvDNScVlFXsNIDuRaiHigJVQ7QhchjCbHggg43XA%2BbX0lSQGbWaXHCTMoDFfE4qpP4GmoSqHbn9gGwlxChNuFGwXAFpGAG5wEYM2M5aRNugGtmmWjswreCbCm9NjeagoZqDAA

### Steps to reproduce

Load playground
Click on the button

### System Info

```shell
N / A
```

### Additional Information

_No response_

wmertens:
Fixes #5738.

The problem is that dirty tasks never run again, whereas on the client they are run on the next tick.

This PR makes executeComponent rerun tasks during SSR until they are no longer dirty.

### DIFF
--- a/packages/qwik/src/core/render/dom/notify-render.ts
+++ b/packages/qwik/src/core/render/dom/notify-render.ts
@@ -297,6 +297,54 @@ const executeTasksBefore = async (containerState: ContainerState, rCtx: RenderCo
   }
 };
 
+/** Execute tasks that are dirty during SSR render */
+export const executeSSRTasks = (containerState: ContainerState, rCtx: RenderContext) => {
+  const containerEl = containerState.$containerEl$;
+  const staging = containerState.$taskStaging$;
+  if (!staging.size) {
+    return;
+  }
+  const taskPromises: ValueOrPromise<SubscriberEffect>[] = [];
+
+  let tries = 20;
+  const runTasks = () => {
+    // SSR dirty tasks are in taskStaging
+    staging.forEach((task) => {
+      console.error('task', task.$qrl$.$symbol$);
+      if (isTask(task)) {
+        taskPromises.push(maybeThen(task.$qrl$.$resolveLazy$(containerEl), () => task));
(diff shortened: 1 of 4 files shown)


mhevery:
So my concern is that this does not fully solve it. For example re-running task can now change a signal, and that signal will not update any already streamed components. So in same cases this makes things better, but it only moves the goal post. 

I don't know what the right fix is.

wmertens:
@mhevery right, it doesn't solve the problem completely, but it does solve more of the problem at a low cost. Interactions between tasks of the same component are handled. There are several issues opened about the dirty tasks, and these should all be resolved with this PR.

Right now you can make code like the test case that is very straightforward, works fine with CSR and yet breaks in SSR.

It is possible to write code that changes output that was already emitted, but at that point that's on the developer. We can't update sent HTML, but we can try a bit harder before sending outdated HTML, and that's this PR.

mhevery:
OK, lets merge it than.

jessezhang91:
Hey @wmertens, is this `console.error` intentionally left in or should it be removed?
https://github.com/BuilderIO/qwik/pull/5741/files#diff-d4d1f7e177c9f3a0e6ea553b1f4c16f5546f754eb3bbc31ed8d327229953bdcfR313

I was creating a calculator that did some computations asynchronously which dirtied some input signals and saw this logging in my terminal. Took me a bit to find it but finally traced it back to here.

PR #5745, opened 2024-01-21
feat(vite-plugin): set assets path in build to /assets

wmertens:
Right now, if `rollupOptions.output.assetFilenames` is not defined, assets will be output as `build/q-[hash].ext`.

This PR changes that default setting to `assets/[hash].[ext]`.

The reasoning is that for post-processing like `$localize` that copies the build files into `/build/[locale]/`, it makes no sense to have to copy `.css` files and others, especially since all css files will be automatically added to the head and so you'll get a per-language copy of the css file.

This is a slightly breaking change, in case some custom scripting is expecting all files to live under `/build`. (Seems unlikely though)

### DIFF
--- a/.changeset/sour-papayas-travel.md
+++ b/.changeset/sour-papayas-travel.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': minor
+---
+
+The default asset filenames in the build have changed. Now they are under `assets/hash-name.ext`, so they are clearly separated from code.
--- a/packages/qwik/src/optimizer/src/plugins/rollup.ts
+++ b/packages/qwik/src/optimizer/src/plugins/rollup.ts
@@ -182,35 +182,23 @@ export function normalizeRollupOutputOptionsObject(
 ): Rollup.OutputOptions {
   const outputOpts: Rollup.OutputOptions = { ...rollupOutputOptsObj };
 
-  if (!outputOpts.assetFileNames) {
-    outputOpts.assetFileNames = 'build/q-[hash].[ext]';
-  }
   if (opts.target === 'client') {
     // client output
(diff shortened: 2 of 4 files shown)


PatrickJS:
oh I actually need this change haha

wmertens:
@PatrickJS you can configure this for yourself, it just changes the default

PatrickJS:
yeah I see that now 😂 I'll update on my end. plus I also want to keep the names for the assets the same for SEO. It's good to fix this as the default

PatrickJS:
@wmertens I want to move all assets to another folder. so dist/client and dist/assets and dist/server

the idea is that I may want to handle assets differently or uploaded somewhere else. I can configure dist/client and dist/server

PatrickJS:
for SEO reasons you want to keep the name of the file so I changed it to this for my app
```js
assetFileNames: 'assets/[hash]/[name][extname]'
```
it also helps to see which fonts are loading in

PatrickJS:
for images we want to keep the image name for seo so `assets/[hash]/[name][extname]` is a better way to include the hash for caching. we also need to make sure the headers are set to immutable for /assets like we do or /build

wmertens:
> for images we want to keep the image name for seo so assets/[hash]/[name][extname]

how about `assets/[hash]-[name][extname]` so that we don't create a zillion directories?

PatrickJS:
@wmertens yup that works

PatrickJS:
I updated the pr

PatrickJS:
for v1 can we do 
`assetFileNames: 'build/assets/[hash]-[name][extname]'`

shairez (approved):

PR #5748, opened 2024-01-23
fix(qwik): ensure that SW cache is properly opened

mhevery:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


packages/docs/src/repl/repl.tsx
```diff
@@ -77,27 +76,19 @@ export const Repl = component$((props: ReplProps) => {
     }
   });
```
hassanzohdy, line 78: Missing `  // eslint-disable-next-line qwik/no-use-visible-task` :)

wmertens: The eslint rules don't work in the Qwik codebase because eslint can't load local plugins :(. I tried with workspace:* but got in trouble somehow.

wmertens:
I'm confused about the repl revert, that is already on main. Rebase and remove?

wmertens (approved):
LGTM

PR #5752, opened 2024-01-24
feat(vite): new file filter option

thejackshelton:
This PR adds the ability to filter out files from the optimizer with a predicate function. It can be useful if you're using multiple frameworks alongside Qwik, and gives the ability to customize what files get passed into `qwikVite`.

Without this PR, other JSX frameworks do not work with Qwik, this is something we thought previously was not necessary, but it is 😅 

I am currently using the build artifact in the Astro integration so that `@astrojs/react` can work alongside Qwik.


gioboa:
@thejackshelton 👏 you rock

thejackshelton:
> @thejackshelton 👏 you rock

Haha this is actually code misko and I wrote a while back. The missing piece being some integration code that was what we needed. Figured I'd go ahead and merge since we both looked at it together at the time.

PR #5754, opened 2024-01-24
fix: insights error handling and qwik-city json imports

wmertens:


tscpp:
This made my Vercel edge deployment fail :(

I think Vercel fails because it does not support importing absolute paths. I can confirm it worked on v1.4.0.

```
Error: The Edge Function "_qwik-city" is referencing unsupported modules:
	- __vc__ns__/0/entry.vercel-edge.js: /vercel/path0/.vercel/output/functions/_qwik-city.func/@qwik-city-not-found-paths.js, /vercel/path0/.vercel/output/functions/_qwik-city.func/@qwik-city-static-paths.js
```

wmertens:
:( ok I'll revert this import, perhaps I can make it work with relative imports

PR #5778, opened 2024-01-26
feat: allow non-object contexts + parameter in errors

hassanzohdy:
# Overview

This PR enhances the displayed message when a context is being `used` without declaring it's value in an ancestor component.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Besides enhancing the displayed message either in terminal on the browser error, the current state when a context value is missing is to stop the entire application, which is a little annoying in terms of just providing the context provider value so we don't have to start the development server again.

It also enhances the error messages generated by `qError` to display the stack at last because the passed options (in our case the context name would be lost among a very large stack).

# Changes done in this PR

- Removed context id transformation to `kebab-case`, it was originated for historical reasons and it's not needed anymore.
- Enhanced the error message when a context is being used without declaring it's value in an ancestor component.
- If thee is no context value, the error will be displayed but the server won't crash (this happens at the moment).
- If an element renders children with `dangerouslySetInnerHTML` the error will be thrown but the server will still be running.
- Updated unit tests based on the changes above.
- Added more unit tests to cover the children with html set combination (there was no test for this case).
- Enhanced the error message if there are params sent with it to be displayed then the stack, otherwise the current behavior will be kept.
- Enhanced types for `useContext` and `createAndLogError` function.
- Allowed `useContext` to use different types like string instead of just objects see [#5793](https://github.com/BuilderIO/qwik/issues/5793).

# Checklist:


packages/qwik/src/core/use/use-context.ts
```diff
@@ -134,6 +134,7 @@ export const createContextId = <STATE = unknown>(name: string): ContextId<STATE>
   assertTrue(/^[\w/.-]+$/.test(name), 'Context name must only contain A-Z,a-z,0-9, _', name);
   return /*#__PURE__*/ Object.freeze({
     id: fromCamelToKebabCase(name),
+    name, // keep the original name to be used when context is being used but not yet provided in one of the parents
```
wmertens, line 137: Can't we just use the id? I know it's not much, but it's still data that we keep just for the dev

packages/qwik/src/core/use/use-context.ts
```diff
@@ -268,27 +269,34 @@ export interface UseContext {
 // </docs>
 export const useContext: UseContext = <STATE extends object>(
   context: ContextId<STATE>,
-  defaultValue?: any
+  defaultValue?: STATE | ((value: STATE | undefined) => STATE), // type of default value should be the same as the context
```
wmertens, line 272: 👍

packages/qwik/src/core/util/log.ts
```diff
@@ -81,8 +81,12 @@ const printElement = (el: Element) => {
 
 const createAndLogError = (asyncThrow: boolean, message?: any, ...optionalParams: any[]) => {
   const err = message instanceof Error ? message : new Error(message);
-  const messageStr = err.stack || err.message;
-  console.error('%cQWIK ERROR', STYLE, messageStr, ...printParams(optionalParams));
+  
+  // display the error message first, then the optional params, and finally the stack trace
+  // the stack needs to be displayed last because the given params will be lost among large stack traces so it will
```
wmertens, line 86: 👍

wmertens (changes requested):
Good changes!

packages/qwik/src/core/use/use-context.ts
hassanzohdy: I was initially using it but displaying the original context id will make it more appropriate because displaying `user-profile` instead of `UserProfile` (the original context id) will be somehow a misleading, that's why i added the name to not make any confusion.

wmertens:
Actually I wonder why the id needs to be kebab case, I can't think of a reason. Thoughts @mhevery ? Can we just leave the context id as-is?

hassanzohdy:
> Actually I wonder why the id needs to be kebab case, I can't think of a reason. Thoughts @mhevery ? Can we just leave the context id as-is?

I asked him same exact question 😅.

I was thinking of removing this transformation but didn't do it because i don't know if it would make break in the app.

hassanzohdy:
@wmertens I went back to Miško and he told me it was for a historical reason, so i guess we no longer need it, updated the code accordingly.

wmertens:
@hassanzohdy you need to run `pnpm fmt` so the lint test can pass

wmertens:
And there's a failed unit test

wmertens:
looks like the unit test still fails?

hassanzohdy:
All good now, i updated the PR with all changes in this PR.

PatrickJS:
@hassanzohdy can you fix the build

wmertens:
@hassanzohdy kleur is not added to the package JSON

hassanzohdy:
@wmertens I tried but still failing in build, it is already in `@builder.io/qwik` package.json file, i tried to add it to the root package and still the same.

@PatrickJS Still trying to figure out why it is upset with it :)

wmertens:
@hassanzohdy actually, kleur shouldn't be in qwik core. Right now it is only being used in the CLI and it should stay that way. No reason to increase the package size so devs have colors in the console.

hassanzohdy:
I removed it and just used a simpler function for output, later we can enhance this in V2.

hassanzohdy:
I guess we are finally done the checks, except E2E tests, it went timeout, is this something related to something that i can do @wmertens ?

packages/qwik/src/core/render/dom/render.unit.tsx
```diff
@@ -67,9 +67,7 @@ test('should serialize events correctly', async () => {
       <div
         on-document:load=""
         on-document:thing=""
-        on-document:-thing=""
         on-window:scroll=""
-        on-window:-scroll=""
```
wmertens, line 72: I don't understand why this is here

packages/qwik/src/core/render/dom/render.unit.tsx
```diff
@@ -1113,3 +1126,15 @@ test('should render value="" on option', async () => {
   assert.equal(option.getAttribute('value'), '');
   assert.equal(option.outerHTML, '<option value="">Empty</option>');
 });
+
+async function throws<T>(fn: () => T, expected?: string): Promise<void> {
```
wmertens, line 1130: vitest's expect() already has https://vitest.dev/api/expect.html#tothrowerror, use that instead

packages/qwik/src/core/render/jsx/jsx-runtime.ts
```diff
@@ -285,7 +285,9 @@ const validateJSXNode = (node: JSXNode) => {
             `The JSX element <${type}> can not have both 'dangerouslySetInnerHTML' and children.`,
             node
           );
-          logError(err);
+
+          // just throw it as an error because we do not need to crash the app
```
wmertens, line 289: So logError crashes the app but throwing doesn't? If so, fix logError instead, no?

packages/qwik/src/core/error/error.ts
```diff
@@ -82,3 +101,23 @@ For more information see: https://qwik.builder.io/docs/components/tasks/#use-met
...
+ * Get an error when context is being used but not provided in a parent This function displays the
+ * error message in the console in more readable way than qError() Also, it does not stop the app,
+ * so when the developer fixes the issue, the app will reload automatically without the need to
+ * start the development server again.
+ */
+export const missingContextValueError = (contextId: string) => {
+  if (qDev) {
+    const error = `${red('Qwik ERROR Error')}: Code(${cyan(
+      '13'
+    )}) Actual value for useContext() '${magenta(
```
wmertens, line 115: Please use codeToText, like the other errors do. Add a parameter to it that will replace a placeholder like `{{info}}`.

packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
```diff
@@ -1494,13 +1494,15 @@ test('null component', async () => {
...
 
 test('cleanse attribute name', async () => {
-  const o = {
-    '"><script>alert("ಠ~ಠ")</script>': 'xss',
-  };
-  await testSSR(
-    <body {...o}></body>,
-    '<html q:container="paused" q:version="dev" q:render="ssr-dev" q:manifest-hash="test"><body></body></html>'
-  );
+  throws(async () => {
```
wmertens, line 1497: So you changed how qwik renders, possibly breaking existing applications

packages/qwik/src/core/use/use-context.ts
```diff
@@ -130,11 +129,11 @@ export interface ContextId<STATE> {
  * @public
  */
 // </docs>
-export const createContextId = <STATE = unknown>(name: string): ContextId<STATE> => {
+export const createContextId = <STATE = unknown>(name: string) => {
```
wmertens, line 132: Please keep the annotation here, so that later code changes don't accidentally change the return signature

packages/qwik/src/core/use/use-context.ts
```diff
@@ -263,32 +267,39 @@ export interface UseContext {
...
   if (value !== undefined) {
     return set(value);
   }
+
   if (defaultValue !== undefined) {
     return set(defaultValue);
   }
-  throw qError(QError_notFoundContext, context.id);
+
+  throw missingContextValueError(context.id);
```
wmertens, line 302: Please inline the function here, no need to make a short function that is used only from one place.

packages/qwik/src/core/util/case.unit.ts
```diff
@@ -2,7 +2,7 @@ import { assert, test } from 'vitest';
 import { fromCamelToKebabCase } from './case';
 
 test('should convert to kebab', () => {
-  assert.equal(fromCamelToKebabCase('HelloWorld'), '-hello-world');
+  assert.equal(fromCamelToKebabCase('HelloWorld'), 'hello-world');
   assert.equal(fromCamelToKebabCase('on:ClicK'), 'on:-clic-k');
```
wmertens, line 6: Shouldn't you handle this case then too? `.replace(/(^|:)-/, '')`?

All in all though, we're doing extra work for no real benefit. I think it's better to leave the kebabcase as-is.
There might even be someone that sends two different events `myEvent` and `MyEvent`, and this change would break that.

packages/qwik/src/core/util/log.ts
```diff
@@ -79,10 +87,29 @@ const printElement = (el: Element) => {
...
+  message?: Error | string,
+  ...optionalParams: any[]
+) => {
+  const hasOptions = optionalParams.length > 0;
+  // if three are params sent, then do not add the message (if string) in the Error object.
+  const err = message instanceof Error ? message : new Error(hasOptions ? '' : message);
+
+  if (hasOptions) {
+    // // Display the error message first, then the optional params, and finally the stack trace.
+    // // This gives a better developer experience so the printed params do not get lost in the stack
```
wmertens, line 101: double comments

packages/qwik/src/core/util/log.ts
```diff
@@ -8,7 +8,15 @@ const STYLE = qDev
   : '';
 
 export const logError = (message?: any, ...optionalParams: any[]) => {
-  return createAndLogError(true, message, ...optionalParams);
+  const error = createAndLogError(true, message, ...optionalParams);
+
+  // make sure to throw the error in test mode so that the test fails
+  // TODO: uncomment the following to fix any unit test that MUST fail and catch it
```
wmertens, line 14: To do this correctly, you need to have async localstorage, and that's not supported in the browser environment.

I propose you revert this change?

packages/qwik/src/core/util/case.ts
```diff
@@ -1,5 +1,8 @@
 export const fromCamelToKebabCase = (text: string): string => {
-  return text.replace(/([A-Z])/g, '-$1').toLowerCase();
+  return text
+    .replace(/([A-Z])/g, '-$1')
+    .replace(/^-/, '') // remove leading dash, this would be present if the first letter was uppercase
```
wmertens, line 4: see comment below. I think this should be reverted.

packages/qwik/src/core/use/use-context.ts
```diff
@@ -191,29 +190,34 @@ export const createContextId = <STATE = unknown>(name: string): ContextId<STATE>
...
+// default value callback function type
+type DefaultStateFn<STATE> = (value: STATE | undefined) => STATE;
+
 export interface UseContext {
-  <STATE extends object, T>(context: ContextId<STATE>, transformer: (value: STATE) => T): T;
-  <STATE extends object, T>(context: ContextId<STATE>, defaultValue: T): STATE | T;
-  <STATE extends object>(context: ContextId<STATE>): STATE;
+  <STATE, T>(context: ContextId<STATE>, transformer: (value: STATE) => T): T;
+  <STATE, T>(context: ContextId<STATE>, defaultValue: T): STATE | T;
+  <STATE>(context: ContextId<STATE>): STATE;
```
wmertens, line 220: You should add a test for number, string, etc context to make sure it works and keeps working.

wmertens (changes requested):
This PR has several changes, I'd like it to only add the contextId to the error, and to allow non-objects for context.

packages/qwik/src/core/render/jsx/jsx-runtime.ts
hassanzohdy: I'd have done this but because `logError` is being used in multiple places, and idk if it should be intentional to halt the app, that's why i just made it like this (for now) until we decide later what should halt the dev server and what should not.

wmertens: well, nothing should halt the devserver ever. But invalid qwik code should definitely break SSR.
For tests, logError should always throw, except that it doesn't have to log when the test is expecting an error.
So maybe we could have something like `invokeFn` called `quietErrorsInvoke()`, that is used around tests that expecte errors?

wmertens:
@hassanzohdy so the context scalars and improved logging are in main now, but the children throwing isn't yet.

Calling logError throws async, but that can probably be turned off provided someone checks all the uses to see they don't rely on actually throwing.

The kebab case event naming is actually more involved, and in v2 it's getting an overhaul, so I propose to leave it as-is.

I propose you make new PRs for each feature you'd still like to add?

PR #5779, opened 2024-01-26
feat(QwikCityMockProvider): add optional goto prop

alexismch:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

While trying to run a Cypress e2e test based on a storybook, I've challenged an issue due to the thrown error of the `goto` function not implemented in `QwikCityMockProvider`.
In order to tackle, I've moved the error to a warning. On top of that I've added an optional prop that allows users to pass a custom goto function.

# Checklist:

### DIFF
--- a/packages/docs/src/routes/api/qwik-city/api.json
+++ b/packages/docs/src/routes/api/qwik-city/api.json
@@ -477,7 +477,7 @@
         }
       ],
       "kind": "Interface",
-      "content": "```typescript\nexport interface QwikCityMockProps \n```\n\n\n|  Property | Modifiers | Type | Description |\n|  --- | --- | --- | --- |\n|  [params?](#) |  | Record&lt;string, string&gt; | _(Optional)_ |\n|  [url?](#) |  | string | _(Optional)_ |",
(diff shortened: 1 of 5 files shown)


alexismch:
> Nice, could you document an example of how to use it?

I've made an update, Is it what you expect?

wmertens:
> I've made an update, is it what you expect?

Well, I was hoping for something that people new to Qwik would be able to easily understand, showing how to use it in their tests.

alexismch:
> > I've made an update, is it what you expect?
> 
> Well, I was hoping for something that people new to Qwik would be able to easily understand, showing how to use it in their tests.

Here it is!

wmertens:
@alexismch great, thanks! I do notice you replaced the error with a console log, shouldn't it remain an error?

Also, the log doesn't say where the log comes from, whereas the Error would show the stack, so even if logging is preferred it would still need a change.

alexismch:
> @alexismch great, thanks! I do notice you replaced the error with a console log, shouldn't it remain an error?
> 
> 
> 
> Also, the log doesn't say where the log comes from, whereas the Error would show the stack, so even if logging is preferred it would still need a change.

I effectively moved to a warn in order to avoid blocking the code after the `navigate` to be executed. And so not restrict people to always define a `goto` fn. What do you think?

mhevery (approved):

wmertens:
I extended the log message a little. Looks good, thanks!

PR #5817, opened 2024-02-04
fix(link): Prefetch redirects to error pages. Fixes #5474

### LINKED ISSUE #5474: [🐞] Prefetch redirects to error pages.

Qwik City (routing)

### Describe the bug

`<Link prefetch={true} href="/this-page-doesnt-exist" />`

If the route on the href is not found on prefetch (such as when the user hovers over the element), the page is immediately redirected to the 404 page. This shouldn't happen until the user clicks or navigates to the page.

### Reproduction

https://github.com/eglove/examples/blob/main/apps/qwik-example/src/routes/index.tsx

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Windows 11 10.0.22621
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i9-12900K
    Memory: 5.58 GB / 31.76 GB
  Binaries:
    Node: 20.9.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.10.5 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (119.0.2151.58)
    Internet Explorer: 11.0.22621.1
```

### Additional Information

_No response_

iamriajul:
# Overview

Fixes: https://github.com/BuilderIO/qwik/issues/5474

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This PR fixes the bug by adding a flag to the `loadClientData` function, and update `handlePrefetch` in link-component to use the flag to indicate that the call is intended to just prefetch not actual navigation, So in-case the route does occur unexpected error or doesn't exist it shouldn't redirect to the page.

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/link-component.tsx
+++ b/packages/qwik-city/runtime/src/link-component.tsx
@@ -48,7 +48,10 @@ export const Link = component$<LinkProps>((props) => {
           prefetchSymbols(url.pathname);
 
           if (elm.hasAttribute('data-prefetch')) {
-            loadClientData(url, elm, { prefetchSymbols: false });
+            loadClientData(url, elm, {
+              prefetchSymbols: false,
+              isPrefetch: true,
+            });
           }
         }
       })
--- a/packages/qwik-city/runtime/src/use-endpoint.ts
+++ b/packages/qwik-city/runtime/src/use-endpoint.ts
@@ -11,6 +11,7 @@ export const loadClientData = async (
     action?: RouteActionValue;
     clearCache?: boolean;
     prefetchSymbols?: boolean;
+    isPrefetch?: boolean;
   }
 ) => {
   const pagePathname = url.pathname;
@@ -61,7 +62,9 @@ export const loadClientData = async (
           return clientData;
         });
       } else {
-        location.href = url.href;
+        if (opts?.isPrefetch !== true) {
+          location.href = url.href;
+        }
         return undefined;
       }
     });


packages/qwik-city/runtime/src/link-component.tsx
```diff
@@ -48,7 +48,10 @@ export const Link = component$<LinkProps>((props) => {
           prefetchSymbols(url.pathname);
 
           if (elm.hasAttribute('data-prefetch')) {
-            loadClientData(url, elm, { prefetchSymbols: false });
+            loadClientData(url, elm, {
+              prefetchSymbols: false,
```
iamriajul, line 52: I'm a little bit confused about `prefetchSymbols` being `false`!

wmertens: Isn't this specifically to fetch the data separately? The symbols are being fetched elsewhere.

wmertens (approved):
I was a little worried that there might be other places that prefetch data that would need this flag, but it's indeed only in this place. Nice concise solution!

LGTM

packages/qwik-city/runtime/src/link-component.tsx
iamriajul: > Isn't this specifically to fetch the data separately?

I don't know really. Currently the prefetch only data. and symbols are prefetched via the `service-worker` maybe.

wmertens: Actually it's done two lines above this call

PR #5846, opened 2024-02-12
feat(qwik-core): Uint8Array serializer

### LINKED ISSUE #4416: [✨] Serialize Uint8Array & other TypedArray

I'm using webauthn to manage credentials. It using a lot of Uint8Array. It would be a great devX improvement to support serializing TypedArray out-of-the-box.
```jsx
// Generate random Uint8Array challenge on the server
export const useChallenge = routeLoader$(() => new Uint8Array([...]));

export const verify = server$((credentials: AuthenticatorResponse) => {
  // verify credentials from the client
});

export default component$(async () => {
  const challenge = useChallenge();
  const signup = event$(() => {
    const credentials = await navigator.credential.create({
      publicKey: {
        challenge,
        ...
      }
    });
    await verify(credentials.response);
  });
  return <button onClick$={signup}>Signup</button>
});
```

Ideally Qwik would serialize/deserialize the native TypedArray. As they are supported by all browsers, node, deno and web workers, it would greatly reduce the difficulty of doing it manually at every step.

Doing it manually

### Additional context

_No response_

genki:
# Overview
Serializer for the `Uint8Array`.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

As the `Uint8Array` can be used for large data blob, its way is compactness conscious. 
It encodes the `Uint8Array` into the UTF-16 code in 2 bytes-wise.
To achieve the compactness, there ended up being two kinds of `Uint8Array`, even and odd length ones.
So there are two serializers `Uint8ArrayESerializer` and `Uint8ArrayOSerializer` to distinguish them.

# Use cases and why

The `Uint8Array` is widely used for data buffer, credencial data and so on compared to other typed arrays.
Fixes #4416

# Checklist:


wmertens (changes requested):
I like it, but:

- it's missing tests
  - especially tests for edge cases involving non-printable characters, XSS result strings like `<script`, unicode modifiers etc.
  - You could look at https://github.com/minimaxir/big-list-of-naughty-strings for inspiration of what shouldn't be allowed as a result
- Most likely you'll need an escape mechanism?
- the code can be deduplicated by moving the functions out and making them into factories depending on odd/even

:+1:

genki:
@wmertens I see, this has some problems.

genki:
@wmertens Thank you. It's just my careless.
I found this code is needing large amount of testing on various environment to check if the unmatched surrogate pairs in the encoded strings are replaced with `0xFFDF`.
JavaScript permits the existence of those invalid UTF-16 strings, but no guarantees on its treatment.
This behaviour is depending on browsers or intermediate text processing libraries.
Unfortunately, currently I have no time to do such testing.

genki:
I have fixed the previous implementation to serialize `Uint8Array` into valid UTF-16 string without no unmatched surrogate pairs.
Now the two serializers are into one.
I have tested locally more than 1,000,000 times over random buffers and the estimated redundancy  is less than 7% in bytes.
But I am not being familiar with how place the serialization tests to be in the core package.
Where I have to see?
I have thought so called the sanitization is out of the responsibility of the serializer module.

wmertens:
Wow that's a lot of handling needed :) are you sure that it's impossible for an even sequence to look like an odd one?

You can add any test files you like.

It's a bit heavy to review for me right now, I'll get back to this later. Notice the lint error btw.

genki:
@wmertens
Yep, I am not hurrying :)
The odd sequence has a marker at the end like`0xFFFD, 0x00XX`, where the `XX` is the last byte of the `Uint8Array`.
I know it is possible to make more space efficient algorithm if we use the bitwise encoding.
But it introduces more computational cost in JavaScript. I think the bytewise approach is neat balance.

wmertens:
Right but I meant, what if the even array has the same bytes at the end? Or is that sequence reserved by the escaping?

genki:
@wmertens
That is why the escape character `0xFFFD` is put in front of the last byte. Every escape character themselves are encoded into doubled as `0xFFFD 0xFFFD`, so the single `0xFFFD` and  sucessing `0x00XX` means always the last byte of the odd array.
There are 3 kind of use case of the escape character.
* `0xFFFD 0xD800`, `0xFFFD 0xD801` for making a fake surrogate pair for unmatched low surrogates.
* `0xFFFD 0xDC00`, `0xFFFD 0xDC01` for making a fake surrogate pair for unmatched high surrogates.
* `0xFFFD 0x00XX` for meaning the array is odd and holding its the last byte as `XX`.

genki:
When encoding the `Uint8Array` into UTF-16 string, there are 4 parts.
1. Normal code points.  96.875% of the bytes. They can be used as is.
2. Unmatched high surrogate `0xDC00-0xDFFF`. They need to be paired with the low surrogate. 
3. Unmatched low surrogate `0xD800-0xDBFF`. They need to be paired with the high surrogate.
4. The last byte of the odd length bytes.

genki:
Added tests and moved utility functions into a separate file.

genki:
Fixed the unit test because the `TextEncoder` drops the BOM character even if that is a valid UTF-16 character.

genki:
I came up with the BOM also should be escaped to protect from unexpecting replacement or vanishing by text processors.
So I have changed implementation slightly to do it.
Added special sequence `0xFFFD 0xD801 0xDC02` that means the escaped BOM.

As well as I know, there's no characters in the UTF-16 having undefined treatment other than unmatched surrogates and the BOM.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,144 @@
...
+      const prev = dbytes[i - 1];
+      const [hi, lo] = prev === 0xd800 ? [0xd801, 0xdc01] : [prev, 0xdc00];
+      // put the 0xFFFD and the fake surrogate pair to make it valid
+      code += String.fromCharCode(0xfffd, hi, lo);
+      surrogate = false; // reset surrogate
+    }
+    // escape the BOM
+    if (c === 0xfeff) {
+      // BOM
+      code += String.fromCharCode(0xfffd, 0xd801, 0xdc02);
```
wmertens, line 65: this could be `FFFD 1000`, but is it even needed? The serialization will never start with the BOM.

wmertens (changes requested):
I like how the code is structured, I'm still unsure about embedding UTF-16 directly in UTF-8 HTML (maybe that's fine?).

Since this code ships with every Qwik, it should be as small as possible, so any tricks are nice.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,144 @@
...
+//  * 0xD800 for fake high surrogate to be with unmatched low surrogate
+//  * 0xDC00 for fake low surrogate to be with unmatched high surrogate
+//
+// if the unmatched high surrogate is 0xD800, it is collided with the fake
+// high surrogate, so use [0xD801, 0xDC01] as the fake surrogate pair
+// representing the 0xD800.
+//
+// If the length of the bytes is odd, the last byte is put after the escape
+// character. As the bytes after the escape character are in 0xD800 to 0xDFFF,
+// we can distingwish the last byte by its high byte being 0x00.
```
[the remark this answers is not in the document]
genki: Oh, I will fix

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,144 @@
+'./string.unit.ts';
+// pack bytes into valid UTF-16 string
+//
+// strategy:
+//
+// * using 0xFFFD as the escape character
```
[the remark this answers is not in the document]
genki: Actually, any other character can be the escape character other than surrogates and the BOM.
It's just I couldn't came up with any other something special character in the unicode.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,144 @@
+'./string.unit.ts';
+// pack bytes into valid UTF-16 string
```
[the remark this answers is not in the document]
genki: I don't know about the Qwik 2.0, as far as I can see at least at existing code, the serialized contents are written as JSON string in the script element of the type `qwik/json`.
This is an example of the serialized Uint8Array at there.
```js 
"\u001c씂絯흲뿰ꔊ閧ዕ曤⾉솨拖梖꽗㔄䓭�Ä"
```
I think all control characters are escaped by the stringify process.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,144 @@
...
+//
+// strategy:
+//
+// * using 0xFFFD as the escape character
+//  * if there is 0xFFFD in the bytes, double it
+// * if there is unmatched surrogate pair, mark it by the escape character
```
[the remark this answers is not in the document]
genki: Yes.
In the valid UTF-16 string, every high surrogates `0xD800-0xDBFF` must be followed by low surrogates `0xDC00-0xDFFF`. If not, they might be replaced with the `0xFFFD`.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,144 @@
...
+// * and put a fake surrogate pair to make it valid
+//  * 0xD800 for fake high surrogate to be with unmatched low surrogate
+//  * 0xDC00 for fake low surrogate to be with unmatched high surrogate
```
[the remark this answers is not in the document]
genki: It sounds good. I will try to change the implementation.

packages/qwik/src/core/util/string.ts
genki: Sorry, I couldn't understand what could be `FFFD 1000`?
This code is escaping the `0xfe 0xff` in the Uint8Array.
This 2 bytes is same to the BOM character in the UTF-16, so some environments meddlingly delete them.
So we have to save them.

wmertens: I meant that FFFD 1000 is still available for escaping after you handle the unmatched pairs, but if you do it with FFFD F999 or so, so still only escaping is needed for FFFD F...

packages/qwik/src/core/util/string.ts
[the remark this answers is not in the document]
genki: Did it. Now the code is simpler and more efficient :)

genki:
Inspired by @wmertens's proposal, I have changed the algorithm to make unmatched surrogates escape to be normal code points by omitting the MSB of the character.

packages/qwik/src/core/util/string.ts
genki: Sorry, as of the change of the algorithm, this code is outdated.

wmertens:
@genki another improvement: When converting to JSON string, most control characters become 7 bytes. I propose you also escape all characters 0-31 by adding 0xf200 to them? Then they become 6 bytes in UTF-8 :).

```
> JSON.stringify(String.fromCharCode(0x00))
'"\\u0000"'
> JSON.stringify(String.fromCharCode(0xfffd)+String.fromCharCode(0xf200))
'"�"'
```

genki:
@wmertens It is depending on the encoding of the HTML document. You are right if it is the `utf-8`, but we have to think other cases. 
I think this serialization prefix is not only for the `Uint8Serializer` but relating to all serializers.
So the improvement of it may be better to be an another issue.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,118 @@
...
+// unpack encoded valid UTF-16 string into Uint8Array
+export const unpackUint8Array = (code: string) => {
+  const dbytes = new Uint16Array(code.length);
+  let j = 0;
+  let escaped = false;
+  for (let i = 0; i < code.length; i++) {
+    const c = code.charCodeAt(i);
+    // check the replacement character
+    if (c === 0xfffd) {
+      if (escaped) {
```
wmertens, line 89: If you move this if outside, then you can skip several ifs. While scanning through the string, you're only interested in the escape character.
Also, when you get the escape character, you're only interested in the next character, so you can immediately check it without looping through again.

Also, if you change the encoding a little bit so that all escape sequences are 0xfffd 0xf..., then you still have plenty of escape namespace but you need to escape slightly less.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,118 @@
...
+    const c = dbytes[i];
+    // test high surrogate
+    if (c >= 0xd800 && c <= 0xdbff) {
+      if (surrogate) {
+        // unmatched high surrogate
+        // omit the MSB to make it a normal codepoint
+        const prev = dbytes[i - 1];
+        code += String.fromCharCode(0xfffd, prev & 0x7fff);
+      }
+      surrogate = true;
```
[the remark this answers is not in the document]
genki: The parser comes here (L29) when the second high surrogate (uncertain if matching or unmatched) comes after the first unmatched high surrogate.
And the parser outputs the escaped *previous* first unmatched surrogate.
The state of the parser is still watching the second uncertain high surrogate until the next iteration, so the `surrogate` keeps being `true`.

packages/qwik/src/core/util/string.ts
genki: I see. I will change it.

packages/qwik/src/core/util/string.ts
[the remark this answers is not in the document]
wmertens: Maybe we should pick one that looks like the Qwik logo ;-)

packages/qwik/src/core/util/string.ts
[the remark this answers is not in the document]
wmertens: I meant, could you add this explanation to the comments so that later readers understand the code faster?

wmertens (approved):
Looks great!

wmertens:
@mhevery is this good to merge? Should we use tricks to reduce size like moving String.fromCharCode into a separate function?

wmertens:
> @wmertens I think this serialization prefix is not only for the `Uint8Serializer` but relating to all serializers. So the improvement of it may be better to be an another issue.

I was just referring to when there's 00 or low values inside a uint8array. That's likely to happen often.

Ideally we'd have an encoding that uses minimal bytes. Are there cases where utf-8 encodes non-ascii using only two bytes instead of 3?

genki:
@wmertens Indeed, I understood. Surely there is a room to improve the space efficiency.
Before to think about it, IMHO, I wonder if the current choice of the serialization prefixes are the best as they are going to be stringified with significant inflation in size.
And I think it is better the anti stringify inflation encoding/decoding you said can be applied to before the stringify process and after the deserialization not only for the Uint8Array serializer but for all serializers.
It will be generic improvement of the `JSON.stringify` and `JSON.parse`.

wmertens:
You're right that the current prefixes aren't good and we should fix those, but for the rest the data is mostly within ASCII range. For byte arrays however it's very likely to encounter low codes since they are used for calculations.

So that's why I think it's better to spend more effort on encoding these efficiently, using as little code as possible.

genki:
@wmertens
I see. I was expecting the Uint8Array to be serialized is not very small often.
The inflation in size during the stringify happens in small part of the UTF-16 space.
So If the size of the data is large and random, the count of control characters is usually ignorable compared with 65536.
The inflation ratio is converged to the ratio of the count of control characters/65536, it is less than 1%.
So my worry was on mainly other smaller data to be serialized.

genki:
I added the extra compaction for the control characters (0-31).

packages/qwik/src/core/util/string.ts
```diff
@@ -57,6 +57,11 @@ export const packUint8Array = (bytes: Uint8Array) => {
       code += String.fromCharCode(0xfffd, 0xffff);
       continue;
     }
+    // extra comaction againt the strinfigy of the control characters
+    if (c <= 31) {
+      code += String.fromCharCode(0xfffd, c + 0xf200);
```
wmertens, line 62: I just realized that because we don't limit ourselves to 0xfffd f..., you can just add 32 and it will be stringified smaller.

wmertens:
I looked up utf-8 encoding and I think it would be better to use 0x007f as escape code. It's only one byte after stringify. All the other escaped values should be kept as low as possible but > 31.

genki:
@wmertens The stringified `0x007f` is 4 characters `\x7F` in UTF-8.
```js
> JSON.stringify(String.fromCharCode(0x007f))
'"\x7F"'
```
I think the escape character can be any of normal code points but it must become 1 character after being stringified.
For example, how about the `0x0071` = `q` for qwik to be the escape character? `q` is 1 byte in the UTF-8 document and stringified as is `q`.

wmertens:
@genki acually that's just the REPL showing the character escaped. When you embed it in a UTF-8 document, it's only one byte. You can check this in Node with `Buffer.from(JSON.stringify(String.fromCharCode(0x007f)))`

UTF-8 encodes ASCII in one byte, and higher codes grow to more bytes. Emoji are 4 to 8 bytes :scream:.

How about:
- 0x007f: escape, because it's rare but still only one utf-8 byte. To escape itself, use 0x007f 0x08ff (four bytes utf-8)
- 0x0000->0x001f: converted to esc + 0x0020->0x003f (two bytes utf-8)
- unmatched pairs: converted to esc + (code-0xd800+0x0040) 0x0040->0x083f (two-four bytes utf-8)
- BOM: esc + 0x08fe (four bytes utf-8)

The current encoding uses 6 bytes per escaped item. so it's all profit.

We can assume that the tests run in Node, so you could calculate the after-JSON size by checking `Buffer.from(JSON.stringify(encoded)).length - 2` (doesn't include the quotes)

genki:
@wmertens 
I agree with you.
I will try to change the implementation.

genki:
Completed :)

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,128 @@
+// pack bytes into valid UTF-16 string
+//
+// strategy:
+//
+// * using esc as the escape character
+//  * if there is esc in the bytes, double it
```
wmertens, line 6: This is no longer correct

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,128 @@
...
+// BOM: esc + 0x08fe (four bytes utf-8)
+//
+// If the length of the bytes is odd, the last byte XX is put after the escape
+// character as 0xFFXX.
+//
+const ESC = 0x007f;
+const SURROGATE_OFFSET = 0xd800 - 0x0040;
+export const packUint8Array = (bytes: Uint8Array) => {
+  const odd = bytes.length % 2 === 1;
+  const dbytes = new Uint16Array(bytes.buffer, 0, bytes.length >> 1);
```
wmertens, line 23: Actually, I think you have to use the bytes offset of the u8 array here?

Could be problematic when it starts at an odd location :-/

Definitely worth adding some tests.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+  if (surrogate) {
+    // ended with unmatched high surrogate
+    // subtract the offset to make it a normal codepoint
+    const c = dbytes[dbytes.length - 1];
+    code += String.fromCharCode(ESC, c - SURROGATE_OFFSET);
+  }
+  if (odd) {
+    // put the last byte
+    code += String.fromCharCode(ESC, bytes[bytes.length - 1] + 0x0900);
```
wmertens, line 85: Since this can happen often, perhaps we should just start the output with O or E? Only one byte.

wmertens (changes requested):
One more thing: endianness. What happens if you convert uint8 on arm and unpack on Intel?

genki:
@wmertens As this serialization process is not across the network but closed in the local machine, I think the endianness is out of the scope. The serializer and deserializer are run at the same endian.

wmertens:
Perhaps we can sidestep the endianness and offset issue by not using a uint16 array but by combining bytes ourselves. Byte0*256+byte1

genki:
To deal with the endinan, the `DataView` is useful.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView
Is it possible to use it in Qwik?
If we can use it, we can fix the endian to the little endian both server side and client side.

packages/qwik/src/core/util/string.ts
genki: The UTF-16 string, JavaScript using internally, must be even length in bytes, so if the Uint8Array is odd length, always one extra byte padding is needed.
Current approach is add ESC(1 byte) + 1 character for odd ones and none for even ones.
If we put the O or E, it costs `O` (1 byte) for odd ones and `E` + `padding` (2 bytes) for even ones.
So I think they are mostly same efficiency.

wmertens:
@genki but I think we can just go by byte pair using only the uint8, that means no endianness and no worry about offset

packages/qwik/src/core/util/string.ts
genki: In addition, I think the length of the buffers tend to be even byte length.

wmertens: Yes but it turns a single byte into 4 bytes...

wmertens: Hmm you're right, most likely they're even.

wmertens: So the algorithm would be, go through the uint8 two bytes at a time, and if odd encode the last byte as 0x0000+byte and add esc.

On decode just check if the final byte is esc, then the last charcode does not add the high byte.

genki: 0x0000+byte means 0x00XX where XX is the byte?
If so, that isn't the swap the position of the ESC and the byte, of current algorithm?

genki:
@wmertens To deal with the endianness manually, we have to detect the endianness by a little trick using Uint8Array and Uint16Array, and the computational cost in JavaScript will be doubled.

For example, we have to add an utility function like this and save its result somewhere to be referred.
```js
function checkEndian() {
    const arrayBuffer = new ArrayBuffer(2);
    const uint8Array = new Uint8Array(arrayBuffer);
    const uint16Array = new Uint16Array(arrayBuffer);

    uint8Array[0] = 0xAA;
    uint8Array[1] = 0xBB;

    if (uint16Array[0] === 0xBBAA) {
        return 'little endian';
    } else if (uint16Array[0] === 0xAABB) {
        return 'big endian';
    } else {
        return 'unknown';
    }
}
```

genki:
As of today, the DataView is widely available on almost all browsers and the requirement for the environment may be lighter than the the current of Qwik's.

If we use only the Uint8Array, the result of the string is changed depending on the endian because the JavaScript environment is internally using `UTF-16LE` or `UTF-16BE` string.
If the endians of the server side and client side are not matched, the encoded strings are swapped its high/low bytes and the resulting character will be larger in bytes.
As long as we use the Uint16Array for the view, we ignore the difference of them because they are in same endian.

packages/qwik/src/core/util/string.ts
genki: Changed the implementation to use `0x00XX ESC` for the last byte.
It was needed to take a special treatment for `0x007f 0x007f` coming at the end to avoid the collision.

genki:
Bye the way, I have no idea for testing the endians.
Is it possible to provide testing environments over varied endianness at the CI?

packages/qwik/src/core/util/string.ts
genki: Found an issue. I am fixing....

genki:
@wmertens I have fixed the collision issue of when the `0x007f 0x007f` comes at the end.
Now the `0x007f 0x08fd` is for escaped `0xd83f` that was to be `0x007f 0x007f`
And I have used only the Uint8Array to solve the endianness issue.

wmertens:
Nice! There's no endianness issues now because we walk the uint8 manually and put 16 bit values in the String, without relying on how those 16 bits are stored.

I'm surprised you need a special case for esc esc at the end, the signal for odd is the start of a new escape sequence at the end of the string. So esc esc should not cause that.

genki:
@wmertens You are right for odd arrays. But the even arrays can have the ESC ESC at the end if `0xd83f` comes as the last character.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,149 @@
...
+  return code;
+};
+
+// unpack encoded valid UTF-16 string into Uint8Array
+export const unpackUint8Array = (code: string) => {
+  const odd = code.charCodeAt(code.length - 1) === ESC;
+  const bytes = new Uint8Array(code.length * 2);
+  let j = 0;
+  for (let i = 0; i < code.length; i++) {
+    const c = code.charCodeAt(i);
```
wmertens, line 100: change this to `for (const s of code) { const c = s.charCodeAt(0);`, it's [vastly faster](https://jsben.ch/Ng3YQ).

That does mean that we now need to maintain the `escaped` state, which actually makes the code easier to understand.

wmertens:
Actually, when iterating through the uint8 array, using for..of is also [vastly faster](https://jsben.ch/XcDca).
So then the lookahead wouldn't work for the stringify, and the code would like like (rough pseudocode):
```ts
let code, surrogate, low = true
// walk uint8
for (const b of arr) {
  if (low) {
    code = b
    low = false
    continue
  }
  code |= b<<8
  low = true
  if (surrogate) {
    if (is_low(code)) {
      output+=surrogate
    } else {
      output+= escape_surrogate(surrogate)
    }
    surrogate = undefined
  }
  if (is_high(code)) {
    surrogate = code
    continue
  }
  output += maybe_escape(code)
}
if (low) {
  // odd array
  output += maybe_escape(code) + ESC
} else if (surrogate) {
  output += escape_surrogate(surrogate)
}
```

genki:
@wmertens 
What did you mean at `if (true) {` at L4?

genki:
BTW, if we think about the performance, isn't it to be better the unpacking to use the `for...of` as well?

wmertens:
Sorry fixed the true. Yes about for of, see my other comments

genki:
@wmertens I see.
The pseudo code seemed lacking treatment of the unmatched low surrogate coming while the surrogate is undefined.
Anyway I will change the implementation after some work.

genki:
Isn't it better the handling of the unmatched low surrogate as the `else if` clause following the `if (surrogate) {`?

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,149 @@
...
+      // restore the unmatched surrogates
+      const x = e + SURROGATE_OFFSET;
+      bytes[j++] = x & 0xff;
+      bytes[j++] = x >>> 8;
+    }
+  }
+  if (odd) {
+    j--; // omit the last ESC
+  }
+  return bytes.subarray(0, j);
```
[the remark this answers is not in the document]
genki: Even the even length arrays, the length of the bytes would be reduced by unescaping.
So the result should be cut by length `j`

genki:
I have recalled the strange behaviour of `for...of` for string.
The JavaScript runtime makes the surrogate pair into single code point while iterating by `for...of`.
So the iteration count may be less than the length of the string if it including surrogate pairs.
To avoid this, we have to split the string in advance like this.
```ts
  for (const s of code.split('')) {
    const c = s.charCodeAt(0);
    if (!escaped) {
      if (c === esc) {
```
Is it acceptable?

genki:
Or, it is needed to do like this.  
```ts
  for (const s of code) {
    const c = s.charCodeAt(0);
    if (!escaped) {
      if (c === esc) {
        escaped = true;
      } else {
        // normal codepoint
        bytes[j++] = c & 0xff;
        bytes[j++] = c >>> 8;
        if (c >= 0xD800 && c <= 0xDBFF) {
          const d = s.charCodeAt(1);
          if (d !== undefined) {
            bytes[j++] = d & 0xff;
            bytes[j++] = d >>> 8;
          }
        }
      }
      continue;
    }
```
At any rate, it draws extra computing cost. Does this still have gain over the original implementation?

genki:
Updated implementation to use `for...of` and fixed things.
I think the cost of checking high surrogate is relatively trivial, I have adopted it.

genki:
As the treatment of the ESC ESC may feel strange, let me explain in advance.
* For even length array that is ended with `0xD83F`, its last 2 bytes are encoded to ESC ESC as unmatched high surrogate.
* For odd length array that is ended with `0x7F`, its last byte encoded to ESC ESC as the last byte and ESC.
* There's no way to distinguish them as is, it is need to encode in different way for the ending`0xD83F` or `0x7f`.
So I choosed the way to change the encoding of the `0xD83F` at the end from ESC ESC to ESC `0x08FD`.
And then, now the `0x007f` at the end means always the odd length array.

wmertens:
Well no, not quite. You can't know if an array is odd or even by looking at the end. You have to walk the array, and if you end with `escaped` true, then you know that it is odd.

For that you don't need any special esc handling.

genki:
@wmertens 
Even if we walk the array, we can't distinguish the final 0x007f 0x007f because that comes both while `escape = false`.
For simplest example:

1. `0x007f 0x007f`: encoded even array of 2 bytes `0xD8 0x3F`.
2. `0x007f 0x007f`: encoded odd array of 1 byte `0x7f`.

How to distinguish them?

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,140 @@
...
+    // normal codepoint
+    code += String.fromCharCode(c);
+  }
+  if (surrogate) {
+    // ended with unmatched high surrogate
+    code += String.fromCharCode(ESC, surrogate - SURROGATE_OFFSET);
+  }
+  if (!low && bytes.length > 0) {
+    // put the last byte
+    code += String.fromCharCode(c === ESC ? 0x100 : c, ESC);
```
wmertens, line 82: I really think you don't need this now

wmertens (commented):
Almost there :sunglasses:

wmertens:
BTW, you can see how much this PR adds to the qwik bundle by looking at this line inside the Qwik CI check: 
https://github.com/BuilderIO/qwik/actions/runs/8056685930/job/22006261963#step:7:38
(you can also see this when you run `pnpm build.core`)

For [main](https://github.com/BuilderIO/qwik/actions/runs/8040830046/job/21959374763#step:7:38) it's `core.min.mjs: { original: '67.6kb', brotli: '23.2kb' }`, so currently we're adding 1kb of code, 400 bytes minified. If we reduce repetition by using internal helper functions that can probably go down a little still.

This code will always be shipped, so every byte helps as is multiplied by millions of eventual views :)

wmertens:
> Even if we walk the array, we can't distinguish the final 0x007f 0x007f because that both come while escaped = false.
> For simplest example:

> 0x007f 0x007f: encoded even array of 2 bytes 0xD8 0x3F.
> 0x007f 0x007f: encoded odd array of 1 byte 0x7f.
> How to distinguish them?

Your second example is wrong, encoding 0x7f should be `0x007f 0x08fe 0x007f`.

packages/qwik/src/core/util/string.unit.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+});
+
+test('space efficient', () => {
+  const a = new Uint8Array(65536);
+  crypto.getRandomValues(a);
+  const packed = packUint8Array(a);
+  // 0xD800-0xDFFF = 2048. 2048/65536 = 0.03125
+  // These are doubled in length, so inflating ratio is about 1.0625
+  expect((packed.length * 2) / a.length).toBeLessThan(1.07);
```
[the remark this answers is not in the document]
genki: It depends on the encoding of the HTML document.
If it uses UTF-16, our encoder/decoder reaches the best inflation ratio 106%.
But it uses UTF-8, it drops to about 147%.

in UTF-8:
* 1 byte: 0x00-0x7F: IR = 50%, share is 256/65536
* 2 bytes: 0x80-0x7FF:  IR = 100%, share is 1920/65536
* 3 bytes: 0x800-0xFFFF: IR = 150%, share is (63488 - 2048)/65536 = 61440/65536
* 4 bytes: 0x10000-0x10FFFF (all surrogate pairs in UTF-16) IR = 100%, 2048/65536
So totally,  IR is about 147%

If we can omit the support other than the UTF-8, the base64 may be better choice because of the UTF-8 is latin1 friendly encoding. The inflation ratio (IR) of the base64 encoding is 4/3, it's about 133%.
It may be the best way to choose the encoder/decoder depending on the HTML document.

The UTF-8 is the standard today, but if you think about efficiency on serializing random bytes, choosing UTF-16 is better.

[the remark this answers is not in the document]
genki: It depends on the encoding of the HTML document.
If it uses UTF-16, our encoder/decoder reaches the best inflation ratio (IR) 106%.
But it uses UTF-8, it drops to about 147%.

in UTF-8:
1 byte: 0x00-0x7F: IR = 50%, share is 256/65536
2 bytes: 0x80-0x7FF: IR = 100%, share is 1920/65536
3 bytes: 0x800-0xFFFF: IR = 150%, share is (63488 - 2048)/65536 = 61440/65536
4 bytes: 0x10000-0x10FFFF (all surrogate pairs in UTF-16) IR = 100%, 2048/65536
So totally, IR is about 147%

If Qwik omits the support other than the UTF-8, the base64 may be better choice because of the UTF-8 is latin1 friendly encoding. The IR of the base64 encoding is 4/3, it's about 133%.
It may be the best way to choose the encoder/decoder depending on the encoding of HTML document, for example, if it is UTF-8 or latin1, uses base64, otherwize this serializer.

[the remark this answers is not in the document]
genki: This result comes from the UTF-8 is not space efficient other than latin1 characters.
In spite of the UTF-8 is common for HTML, using UTF-16 is better choice if the document contains many characters over the 0x7f code points because most of 3 bytes in UTF-8 are 2 bytes in UTF-16.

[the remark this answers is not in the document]
genki: In particular the Qwik, as of using many serialization that produces special characters, it may be good to set the UTF-16 as default encoding for better performance.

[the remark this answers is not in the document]
wmertens: I think Qwik only supports utf-8 at the moment, so then base64 would be a better choice :-/.

However, we have 96 safe-ish characters at our disposal that encode to one byte, so we could implement our own base-96. However, [here's a very cool package](https://github.com/qntm/base65536?tab=readme-ov-file#:~:text=Note%20that%20in%20UTF%2D16%2C%20Base32768%20significantly%20outperforms%20either%20choice%2C%20and%20in%20UTF%2D8%2C%20Base64%20remains%20the%20preferred%20choice) that does UTF-32 encoding and it talks about other encodings, saying that for UTF-8, base-64 is still preferred :-/.

wmertens:
Very interesting: https://blog.kevinalbs.com/base122

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,149 @@
...
+  let c = 0;
+  for (const b of bytes) {
+    if (low) {
+      c = b;
+      low = false;
+      continue;
+    }
+    c |= b << 8;
+    low = true;
+    if (surrogate !== undefined) {
```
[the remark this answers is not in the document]
genki: I see.

wmertens:
Ok so the base-122 encoding assumes embedding into HTML directly, and it is smaller, but after gzip encoding, [it is larger](https://blog.kevinalbs.com/base122#:~:text=Unfortunately%2C%20base%2D64%20seems%20to%20compress%20better%20than%20base%2D122) :-(

So it looks like we should just use base64 encode/decode. Too bad because this was fun code, but the boring way seems to be the most efficient.

packages/qwik/src/core/util/string.ts
```diff
@@ -0,0 +1,149 @@
...
+    j -= 2;
+    bytes[j++] = ESC;
+  }
+  // if ended while escaped, the length is odd
+  if (escaped) {
+    // Array is odd-length, remove last byte
+    return bytes.subarray(0, j - 1);
+  } else {
+    return bytes.subarray(0, j);
+  }
```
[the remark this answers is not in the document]
genki: Yes, this was wrong. I will fix.

As of my custom, it is preferable not to use function in the tight inner loop. But it may be old fashioned.
You think the optimizer works well?

genki:
Okey, so I will change the implementation to use base64.
Don't worry the code we have made is still usable for serialization to localStorage or cookie.

wmertens:
Yeah and w3 [discourages using UTF-16](https://www.w3.org/International/questions/qa-html-encoding-declarations#utf16) for webpages.

genki:
Completed :)
Now I know the base64 is not so bad in UTF-8.

packages/qwik/src/core/container/serializers.ts
```diff
@@ -442,6 +442,22 @@ const BigIntSerializer = /*#__PURE__*/ serializer<bigint>({
...
+  $prefix$: '\u001c',
+  $test$: (v) => v instanceof Uint8Array,
+  $serialize$: (v) => btoa(decoder.decode(v)),
+  $prepare$: (data) => {
+    const buf = atob(data);
+    const array = new Uint8Array(buf.length);
+    for (let i = 0; i < buf.length; i++) {
+      array[i] = buf.charCodeAt(i);
+    }
+    return array;
```
wmertens, line 456: Why don't you use TextEncoder here?

genki: Sorry, I recalled the TextEncoder and TextDecoder is not usable for this purpose.
The TextDecoder can decode only the encoded array of text.
It is need to use the generic base64 encoder for any binary data.

genki:
Fixed. Now uses `for...of` for performance.

packages/qwik/src/core/container/serializers.ts
```diff
@@ -442,6 +442,28 @@ const BigIntSerializer = /*#__PURE__*/ serializer<bigint>({
...
 
+const Uint8ArraySerializer = /*#__PURE__*/ serializer<Uint8Array>({
+  $prefix$: '\u001c',
+  $test$: (v) => v instanceof Uint8Array,
+  $serialize$: (v) => {
+    let buf = '';
+    for (const c of v) {
+      buf += String.fromCharCode(c);
+    }
+    return btoa(buf);
```
wmertens, line 453: You can make this a tiny bit smaller by doing `btoa(buf).replace(/=+$/, '')`

genki: Indeed. I fixed

wmertens (approved):
LGTM

genki:
@wmertens 
BTW, I had put the final result here. I was misunderstand the treatment of the last byte of the odd length arrays. It was fixed.
https://gist.github.com/genki/e86e4907d0f5ed04340ab0ec55250499

PR #5849, opened 2024-02-14
fix(create-qwik): update engine version in package.json for generated projects

AlexanderShushunov:
create-qwik

This PR is fixing a minor issue

I created a new project using `npx create-qwik@latest` and tried to run `npm run dev`. Although I had Node.js v16 and the `package.json` in the generated project specified that the appropriate version is `>=15.0.0`, I encountered an error because `sharp` does not support this version.


gioboa (approved):
Thank you @AlexanderShushunov 
Appreciated 💪

PR #5851, opened 2024-02-14
test(use-visible-task): add more use visible task tests

Varixo:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This PR adds more use visible task tests with fixes for v2

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/v2/client/vnode-diff.ts
+++ b/packages/qwik/src/core/v2/client/vnode-diff.ts
@@ -15,6 +15,7 @@ import { ELEMENT_KEY, ELEMENT_PROPS, OnRenderProp, QSlot, QSlotParent } from '..
 import { isPromise } from '../../util/promises';
 import type { ValueOrPromise } from '../../util/types';
 import type { QElement2, fixMeAny } from '../shared/types';
+import { isQrlSSREvent } from '../ssr/ssr-render';
 import type { SsrAttrs } from '../ssr/types';
 import type { DomContainer } from './dom-container';
 import {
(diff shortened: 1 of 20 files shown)


gioboa:
@Varixo 👏 great

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -702,7 +721,7 @@ export const vnode_applyJournal = (journal: VNodeJournalEntry[]) => {
         while (typeof (key = journal[idx] as string | null) === 'string') {
           idx++;
           const value = journal[idx++] as string | null;
-          if (key.startsWith(':')) {
+          if (key.startsWith(':') || key.startsWith('document:') || key.startsWith('window:')) {
```
mhevery, line 724: So this is not right. The code needs to stay as is, which means the code above needs to change.

The way it works now is that any property that starts with `:` is not serializable. 

For example `:onClick$` So i think we need to have `:document:onClick$` or something like that.

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -567,9 +573,22 @@ export const vnode_diff = (container: ClientContainer, jsxNode: JSXOutput, vStar
       const element = vnode_getNode(vnode) as QElement2;
       if (!element.qDispatchEvent) {
         element.qDispatchEvent = (event: Event) => {
-          const eventName = event.type;
+          let eventName = event.type;
```
mhevery, line 576: I don't think this is right. See comment bellow about it having to start with `:`

mhevery (changes requested):
Great work!  the only issue I seed is the `:` in the `vnode_getProp` function. 

Alos since you are adding `document` and `window` prefix, can we add `v2/use-on.unit.tsx` tests?


Otherwise this looks great!

PR #5869, opened 2024-02-20
fix: read q:base to get the script path

jcfauchet:
# Overview

The initial issue revolves around the SPA Shim Script, which fails to utilize the correct base for loading the JavaScript script when a custom base is configured in `entry.ssr.ts`. After discussing with Miško, he guided me on how to make the script read the path from the `q:base` attribute, mirroring its behavior in dev mode.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

The proposed fix involves removing the `isDev` condition and adopting the same approach for production mode.

# Checklist:


packages/qwik-city/runtime/src/spa-shim.ts
```diff
@@ -33,18 +31,15 @@ const shim = async (path: string, symbol: string) => {
...
-      const base = new URL(container.getAttribute('q:base')!, document.baseURI);
-      const url = new URL(path, base);
-
-      // Bypass dev import hijack. (not going to work here)
-      // eslint-disable-next-line no-new-func
-      const imp = new Function('url', 'return import(url)');
-      (await imp(url.href))[symbol](currentScript);
-    }
+
+    // Importing @qwik-city-plan here explodes dev, get basePathname manually.
```
mhevery, line 35: Not sure this comment is relevant.

packages/qwik-city/runtime/src/spa-shim.ts
```diff
@@ -33,18 +31,15 @@ const shim = async (path: string, symbol: string) => {
...
+    const container = currentScript.closest('[q\\:container]')!;
+    const base = new URL(container.getAttribute('q:base')!, document.baseURI);
+    const url = new URL(path, base);
+
+    // Bypass dev import hijack. (not going to work here)
+    // eslint-disable-next-line no-new-func
+    const imp = new Function('url', 'return import(url)');
```
mhevery, line 42: seems like this hack is needed for dev mode only. So I think we still in `if (isDev) {}`

mhevery (changes requested):

mhevery (approved):

PR #5882, opened 2024-02-23
feat(qwik-city): upgrade to mdxjs v3

wmertens:


wmertens:
I'll merge this, and push out a dev release, we can always revert.

PR #5894, opened 2024-02-26
feat: drizzle orm integration

LazyClicks:
# Overview

# What is it?

- [x ] Feature / enhancement
- [ ] Bug
- [ x ] Docs / tests / types / typos

# Description

Drizzle Orm integration with qwik, based this integration on the existing prisma integration

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

- [ x ] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [x ] I have made corresponding changes to the documentation


starters/features/drizzle/package.json
```diff
@@ -0,0 +1,43 @@
...
+      ]
+    }
+  },
+  "devDependencies": {
+    "drizzle-kit": "^0.20.14"
+  },
+  "dependencies": {
+    "drizzle-orm": "^0.29.1",
+    "@neondatabase/serverless": "^0.8.1",
+    "pg": "^8.11.3"
```
wmertens, line 35: Is it not possible to install a lighter DB by default? Perhaps there's something builtin without any deps?

wmertens: If not, how about sqlite? That would allow it to work out-of-the-box

wmertens:
Can you also run `pnpm fmt`?

LazyClicks:
> Can you also run `pnpm fmt`?

i wasnt able to install the pnpm dependencies, i keep getting wrong node version errors, i tried following the contribution guide by installing the rust thingy and running pnpm i but that diddnt work either, tried changing my own version of node and it didnt work as well

starters/features/drizzle/package.json
LazyClicks: i had thought drizzle doesnt support schemas with sqlite as their docs says but it seems it does, just not multiple schemas? im still confused by that, checking with their support, but i did switch to better-sqlite3 as it is the documented one in their docs

LazyClicks:
> Can you also run `pnpm fmt`?

my test included using my implementation in a new qwik project
i also ran pnpm fmt that comes with qwik projects inside them before pasting inside the integrations files as i couldnt install with pnpm i

wmertens (approved):
LGTM

gioboa (approved):
Thanks @LazyClicks 😉

LazyClicks:
@gioboa @wmertens for some reason the db folder inside the drizzle directory didnt get pushed into the commit? idk if its due to it being empty, but without it the drizzle migrate command which is required for creating the database wont run
is there anything i can do to modify it here or should i make a new PR?

LazyClicks:
also the integration is not visible in the integrations docs, did i mess something up?
its accessible by link tho?
https://qwik.dev/docs/integrations/drizzle/

wmertens:
Empty dirs don't exist in git so they need a `.empty` file in them (or whatever).

As for the integration TOC, not sure :-/

PR #5903, opened 2024-02-28
chore(docs): change indicator side from left to the right

LazyClicks:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

switched the updated indicator to be on the right of the a tag text, as on desktop it is not visible due to the small sidebar width
after change
vs
before change

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


gioboa (approved):
That's awesome 😎 @LazyClicks 
Nice catch

PR #5919, opened 2024-03-01
test(vnode-data): tests for vnode data refids

Varixo:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This PR fixes vnodes ref ids and adds basic tests

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/v2/ssr/vnode-data.unit.tsx
+++ b/packages/qwik/src/core/v2/ssr/vnode-data.unit.tsx
@@ -1,10 +1,13 @@
 import { describe, expect, it } from 'vitest';
-import { componentQrl } from '../../component/component.public';
+import { component$, componentQrl } from '../../component/component.public';
 import { inlinedQrl } from '../../qrl/qrl';
 import { useSignal } from '../../use/use-signal';
 import { ssrRenderToDom } from '../rendering.unit-util';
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/v2/ssr/vnode-data.unit.tsx
```diff
@@ -0,0 +1,215 @@
...
+            </div>
+          );
+        }, 's_parent')
+      );
+
+      const { container } = await ssrRenderToDom(<Parent />, { debug });
+      const state = container.$rawStateData$;
+      const vnodeData = getVNodeDataFromQwikState(state);
+
+      expect(vnodeData).toEqual(['3A', '4A', '4B']);
```
mhevery, line 65: I don't have an answer for you, but I would like to figure out something better here.

The issue is that:
- You have an integration test, but really, it asserts an implementation detail. If the implementation details change, it will be hard to know what the test should assert.
- Ideally this integration tests, should be integration all the way through. Could you assert that a correct VNode can be retrieved?

I know I don't have answers here, but was hoping that maybe you could come up with something.

mhevery (approved):
Would love to see a better test, but if you can't think of anything, I guess this is OK.

mhevery (approved):

PR #5942, opened 2024-03-04
fix: Use types from `@types/bun` package

octet-stream:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This PR removes typings for Bun adapter in favor of the official `@types/bun` package. This will further simplify the setup for this adapter. I've also updated `starters/adapters/bun/src/entry.bun.ts` file to rely on that package.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


octet-stream:
Two things in addition to this PR:
1. I wasn't sure what `@types/bun` at the project's root is used for so I kept it there, but I think it can be removed too, because `@types/bun` (the package) was installed at the root of the monorepo, so it should be available across the project.
2. I wasn't able to build and test my changes, because neither `pnpm build` nor `pnpm build.full` can finish successfully. I was running these two commands within docker container via terminal attached to the container. Is this the right way to run these commands or I misinterpreted the contributing guides?

wmertens (approved):
LGTM

wmertens:
@octet-stream you need to run `pnpm api.update`

octet-stream:
Yeah, I see. Did the update.

wmertens:
Oh and you should be able to run `pnpm build.local`. If not, what is the error?

octet-stream:
Was able to successfully run `pnpm build.local`

wmertens:
I'm really surprised about Loader_2 though

octet-stream:
Yeah, I have no idea why is it there.

I did in the last commit:
1. Switched to main branch
2. Pulled changes from the upstream (with --rebase, it's my default)
3. Switched to bun-types
4. Merged main into my branch
5. Run `pnpm api.update` inside a docker container

wmertens:
ok, thanks!

gioboa (approved):

PR #5954, opened 2024-03-05
fix(qwik-city): spa-shim.ts import scripts from wrong path

blueagler:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

spa-shim.ts does not respect q:base attribute when importing script

In a subpath, for example http://localhost:3000/blog/
It will import the script from http://localhost:3000/blog/eg.js
Not from http://localhost:3000/build/eg.js

CleanShot 2024-03-06 at 00 47 30@2x

# Use cases and why
Just a mistype maybe

- 1. One use case
- 2. Another use case

# Checklist:


gioboa (approved):
Thanks @blueagler

PR #5967, opened 2024-03-07
fix: A bug fix

### LINKED ISSUE #5708: [🐞] Qwik City app breaks in dev mode with legacy decorators

Qwik Rollup / Vite plugin

### Describe the bug

I've been playing around with Qwik and Qwik City building simple demos. Pretty great framework so far, by the way! But when I decided to try it with [Mikro ORM](https://mikro-orm.io/) I stumbled upon following issue: Qwik breaks with TypeScript legacy decorators and Mikro ORM relies heavily on decorators for database schema definition, so I need those to work properly to be able to use Mikro ORM with Qwik app. The issue appears only in in dev mode.

### Reproduction

https://github.com/octet-stream/qwik-legacy-decorators-issue

### Steps to reproduce

1. Clone this repo
2. Install dependencies via `pnpm i`
3. Run `pnpm start` and you'll see following error

```
[vite] Pre-transform error: Expression expected
[vite] Pre-transform error: Unexpected token `@`. Expected identifier, string literal, numeric literal or [ for the computed key
[vite] Pre-transform error: Unexpected token `@`. Expected identifier, string literal, numeric literal or [ for the computed key (x2)
```

This error does not appear in preview mode. Run `pnpm preview` to verify it. Vite will open your browser and you'll see a todo list app. You will be able to add and remove todos.

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 95.73 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.3.0 - ~/Library/Caches/fnm_multishells/9803_1704038600457/bin/node
    npm: 10.2.4 - ~/Library/Caches/fnm_multishells/9803_1704038600457/bin/npm
    pnpm: 8.13.1 - /opt/homebrew/bin/pnpm
    bun: 1.0.21 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 120.0.6099.216
    Safari: 17.2.1
  npmPackages:
    @builder.io/qwik: ^1.3.5 => 1.3.5
    @builder.io/qwik-city: ^1.3.5 => 1.3.5
    undici: * => 6.3.0
    vite: ^5.0.11 => 5.0.11
```

### Additional Information

I saw [similar issue](https://github.com/BuilderIO/qwik/issues/2951), but the error is still there and happens only in dev mode, so I believe there's a way to make it work for dev mode as well without bringing some additional plugins. Also, from what I can tell reading the comments the solution with babel plugin doesn't work.

JerryWu1234:
# Overview
fix #5708

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


wmertens:
This is not the right way to tackle this. The reason that in dev mode the decorators don't work is that the qwik plugin (which understands them) isn't running.

So the options are:
- making qwik plugin work during dev mode (might be difficult with HMR but worth trying)
- add very light-weight support for decorators (this excludes babel plugins) during dev only

JerryWu1234:
> This is not the right way to tackle this. The reason that in dev mode the decorators don't work is that the qwik plugin (which understands them) isn't running.
> 
> So the options are:
> 
> * making qwik plugin work during dev mode (might be difficult with HMR but worth trying)
> * add very light-weight support for decorators (this excludes babel plugins) during dev only

thank for your tips. it's useful for me, especially as a new one.
let me dig into it deeply,

wmertens:
I'd like to add that it's fine to use this as a standalone vite plugin, but for qwik it needs to be very light weight because not many people use decorators in qwik projects.

JerryWu1234:
> I'd like to add that it's fine to use this as a standalone vite plugin, 

Could you clarify the conception about `qwik plugin`?
image
such as isn't `qwikCity` a Vite plugin? So which is `qwik plugin`?

>but for inclusion in qwik core it needs to be very light weight because not many people use decorators in qwik projects.

I totally a agree, babel is such heavy, I'll find out another

wmertens:
Actually I've just tried to figure out why the decorator transform doesn't run in Feb mode.

Here's the rust plugin enabling it: https://github.com/BuilderIO/qwik/blob/6b6b3acfa698b449a6debe932d2a8b9dd33e0f81/packages/qwik/src/optimizer/core/src/parse.rs#L522

And when I look at the qwik optimizer plugin.ts, it only changes `mode` to be `"dev"` but it calls the optimizer in the same way in the `transform()` hook.

So maybe something else is erroring in dev mode, running before the Qwik optimizer?

JerryWu1234:
> Actually I've just tried to figure out why the decorator transform doesn't run in dev mode.
> 
> Here's the rust plugin enabling it:
> 
> https://github.com/BuilderIO/qwik/blob/6b6b3acfa698b449a6debe932d2a8b9dd33e0f81/packages/qwik/src/optimizer/core/src/parse.rs#L522
> 
> And when I look at the qwik optimizer plugin.ts, it only changes `mode` to be `"dev"` but it calls the optimizer in the same way in the `transform()` hook.
> 
> So maybe something else is erroring in dev mode, running before the Qwik optimizer?


it;

PR #5985, opened 2024-03-09
fix: prevent SW exception

gioboa:
This fix will solve this edge case error. 
With:
- empty browser cache
- no active service worker

sometimes the variable `$cache$` is null when that code is evaluated.
 
Pasted image 1
Pasted image

This code could be the problem https://github.com/BuilderIO/qwik/blob/157f7b01df22448638f7cc55546a051aa4eea5f0/packages/qwik/src/prefetch-service-worker/setup.ts#L42

### DIFF
--- a/packages/qwik/src/prefetch-service-worker/direct-fetch.ts
+++ b/packages/qwik/src/prefetch-service-worker/direct-fetch.ts
@@ -79,8 +79,8 @@ function taskTick(swState: SWState) {
     if (task.$isFetching$) {
       outstandingRequests++;
     } else if (
-      outstandingRequests < swState.$maxPrefetchRequests$ ||
-      task.$priority$ >= DIRECT_PRIORITY
+      swState.$getCache$() &&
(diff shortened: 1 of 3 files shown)


packages/qwik/src/prefetch-service-worker/direct-fetch.ts
```diff
@@ -79,8 +79,8 @@ function taskTick(swState: SWState) {
     if (task.$isFetching$) {
       outstandingRequests++;
     } else if (
-      outstandingRequests < swState.$maxPrefetchRequests$ ||
-      task.$priority$ >= DIRECT_PRIORITY
+      swState.$getCache$() &&
```
gioboa, line 82: This prevents $put$ without cache in place

wmertens: Shouldn't you await that?

gioboa: Yes, nice catch

PR #5986, opened 2024-03-09
fix: delay write cookie headers to first write

PatrickJS:
# Overview
fixes https://github.com/BuilderIO/qwik/issues/5951

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

set headers once but allow for dev to add cookie values


packages/qwik-city/middleware/node/http.ts
```diff
@@ -89,13 +89,18 @@ export async function fromNodeHttp(
     },
     getWritableStream: (status, headers, cookies) => {
       res.statusCode = status;
-      headers.forEach((value, key) => res.setHeader(key, value));
-      const cookieHeaders = cookies.headers();
-      if (cookieHeaders.length > 0) {
-        res.setHeader('Set-Cookie', cookieHeaders);
-      }
+      let once = false;
```
gioboa, line 92: This is an oversight I guess

gioboa (changes requested):

packages/qwik-city/middleware/node/http.ts
PatrickJS: I want to write once since write can happen multiple times (unless it's renderToString)

gioboa: I see. I think this should be true to execute the next if once

PatrickJS: oh yeah I meant to set the other one to true

gioboa (approved):

PatrickJS:
@wmertens we need to delay writing to headers until we first "write" to the client. that way any app code can run and add cookie headers before we stream.

PatrickJS:
yes this applies to all headers. the e2e tests hang with this fix but when I run it locally it works correctly in preview-only. so there is something going on. @wmertens is there a better way to hook into this?

wmertens:
I don't think there's a better way because each adapter had their own way of sending cookies.

I think the e2e problem may be when there's no write and the response just closes without sending the cookies? Odd but maybe.

In any case you should handle `once` still being true on close.

wmertens:
@PatrickJS notice that the e2e tests fail on this

wmertens (approved):
Looks good!

wmertens:
...but sadly the e2e fails. Try removing the once=false in close? Not sure that's needed or even desired

PatrickJS:
@wmertens I'm thinking e2e is waiting for headers which is why it fails during e2e but not when running locally. maybe there's another way

PatrickJS:
I think it's better having devs use `plugin@<name>.ts` and middleware for this

PR #6005, opened 2024-03-13
chore(rust): update all deps + fixups

wmertens:
Napi v1 doesn't work on ARM macs. So we need to upgrade to v2 as well, or use the wasm optimizer on ARM macs.

UPDATE: after upgrading to napi v2, everything builds and runs, but `cargo test` no longer works, breaking on qwik-napi. Seems like a cosmetic issue.

UPDATE: worked around by only building what's needed for cargo test.


shairez:
@wmertens do you need help with this one?
running into the same issues of when trying to build the repo

shairez (approved):

shairez:
thanks @wmertens !
Great job!

PR #6013, opened 2024-03-14
fix(insights): gracefully handle no manifests and missing symbols

intellix:
# Overview

Closes https://github.com/BuilderIO/qwik/issues/6012

When fetching insights, it throws a 500 error each time for my projects when there are no manifests

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

. After pulling it locally and running it on a brand new app, I've noticed there were 2x errors:
 - Cannot use inArray without at least 1 item (so I've added a condition to not use it when it's empty)
 - Cannot read name of undefined:

```
[vite] Internal server error: Cannot read properties of undefined (reading 'name')
  File: /Projects/qwik/packages/insights/src/stats/edges.ts:211:4
  209|    clusters.forEach((cluster) => {
  210|      const symbols = cluster.map((id) => symbolVectors.symbols[id]);
  211|      const symbolNames = symbols.map((s) => s.name);
```

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


wmertens (approved):
LGTM

PR #6017, opened 2024-03-15
fix(qwik-city): support vite.base in dev mode

### LINKED ISSUE #5984: [🐞] qwik-city: vite base doesn't work in dev mode

Qwik City (routing)

### Describe the bug

When you set `config.base` of vite, the build correctly uses it but dev mode breaks, getting module files without the base path.

### Reproduction

https://stackblitz.com/edit/qwik-starter-wvwual?file=vite.config.ts

### Steps to reproduce

try clicking Home

### System Info

```shell
stackblitz, 1.5.1
```

### Additional Information

_No response_

JerryWu1234:
Fixes #5984


packages/qwik/src/qwikloader.ts
```diff
@@ -74,7 +74,7 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
       const container = element.closest('[q\\:container]')!;
       const base = new URL(container[getAttribute]('q:base')!, doc.baseURI);
       for (const qrl of attrValue.split('\n')) {
-        const url = new URL(qrl, base);
+        const url = new URL(`${container[getAttribute]('q:route')}${qrl}`, base);
```
wmertens, line 77: what is this for? Aren't all qrls hosted in the same place?

wmertens, line 77: what is this for? Aren't all qrls hosted in the same place?

JerryWu1234:
@wmertens didn't finish yet.
I create PR just for convenient sync code. lol

JerryWu1234:
@wmertens I'm sorry to bother you, I know where problem is, but I was stuck on a crucial problem.

please take a look at [this line ](https://github.com/BuilderIO/qwik/blob/9959b1875ac2b5f70de28e3f97baf86bda0a56c6/packages/qwik/src/server/utils.ts#L30)

Why does it set a default `'/build/'`,
and what's between `q:route` and `q:base`.

wmertens:
/build is where the scripts are stored by default. The result of this function should then get the vite base added.

JerryWu1234:
image
image

@wmertens 
I have almost done.
because it's different between pro and dev.
pro `q:base="/newbase/build/"`
dev `q:base="build/"`

I want to check the pro env, but code is compressed by the tool. when I run `pnpm preview`
Could you give me some tips? 
thanks

wmertens:
Btw, you can set minify:false in the build to see prod as-is

JerryWu1234:
yR6clyXv70
pro

mLx3DdTwLv
dev

everything is right

JerryWu1234:
I also tested env and pro without base added

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -134,6 +135,13 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
 
     opts.debug = !!updatedOpts.debug;
 
+    opts.base =
+      !updatedOpts?.base || updatedOpts.base.length === 1
```
wmertens, line 139: What if base is 'a'? Also, you default it to '/' and then replace it with ''

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -957,6 +965,7 @@ export interface QwikPluginOptions {
    * large projects. Defaults to `true`
    */
   lint?: boolean;
+  base?: string;
```
wmertens, line 968: Perhaps you should rename this to baseNoSlash because the vite base requires a slash.

Or else make it include the slash so it's the same and then handle the slash when using it

packages/qwik/src/optimizer/src/plugins/vite-server.ts
```diff
@@ -118,7 +118,7 @@ export async function configureDevServer(
                   location: 'head',
                   attributes: {
                     rel: 'stylesheet',
-                    href: url,
+                    href: `${opts.base}${url}`,
```
wmertens, line 121: Isn't it possible that done of these urls already include the base?

Better to change them at the source, no?

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -155,6 +155,7 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
         devTools: qwikViteOpts.devTools,
         sourcemap: !!viteConfig.build?.sourcemap,
         lint: qwikViteOpts.lint,
+        base: viteConfig.base,
```
wmertens, line 158: This one includes a slash

packages/qwik/src/qwikloader.ts
```diff
@@ -72,9 +72,11 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
     const attrValue = element[getAttribute](attrName);
     if (attrValue) {
       const container = element.closest('[q\\:container]')!;
-      const base = new URL(container[getAttribute]('q:base')!, doc.baseURI);
+      const baseURI = new URL(doc.baseURI);
+      const base = container?.getAttribute('q:base') ?? '/';
       for (const qrl of attrValue.split('\n')) {
-        const url = new URL(qrl, base);
+        const pathUrl = (base + qrl).replace(/\/+/g, '/');
```
wmertens, line 78: Same comments as above

wmertens (commented):
Great, but we need some unit tests and more consistent `/` handling

JerryWu1234:
> Great, but we need some unit tests and more consistent `/` handling

I'll add test after you reviewed.

wmertens:
@JerryWu1234 how's it going with the tests? Need help?

JerryWu1234:
> @JerryWu1234 how's it going with the tests? Need help?



> @JerryWu1234 how's it going with the tests? Need help?

I haven't tried to add tests yet because I’m waiting for your reply on the problem you mentioned

wmertens:
@JerryWu1234 which one specifically? I don't have any more information to add right now?

JerryWu1234:
> @JerryWu1234 which one specifically? I don't have any more information to add right now?

https://github.com/BuilderIO/qwik/pull/6017#discussion_r1533292949

https://github.com/BuilderIO/qwik/pull/6017#discussion_r1533286154

Here

wmertens:
So I meant that we should have unit tests, and that you shouldn't have to replace the double slashes if you don't put them in.

So you need to be careful about base, don't call replace, and have unit tests that prove it doesn't add double slashes for various test cases

JerryWu1234:
> So I meant that we should have unit tests, and that you shouldn't have to replace the double slashes if you don't put them in.
> 
> So you need to be careful about base, don't call replace, and have unit tests that prove it doesn't add double slashes for various test cases



> So I meant that we should have unit tests, and that you shouldn't have to replace the double slashes if you don't put them in.
> 
> So you need to be careful about base, don't call replace, and have unit tests that prove it doesn't add double slashes for various test cases

Let me try to add some tests

JerryWu1234:
> @JerryWu1234 how's it going with the tests? Need help?

@wmertens 

I really need your help.

To test the base attribute, we need to rely on both **build** and **start**. However, the current e2e only has build.

If I modify it, I'm afraid it will break the project's testing structure.

wmertens:
@JerryWu1234 🤔 actually the tests in v2 are a lot better. Take a look at the build/V2 branch, do you think you can incorporate your test there?

And actually, maybe that's not needed. It could be enough to make unit tests, not e2e

JerryWu1234:
> @JerryWu1234 🤔 actually the tests in v2 are a lot better. Take a look at the build/V2 branch, do you think you can incorporate your test there?
> 
> And actually, maybe that's not needed. It could be enough to make unit tests, not e2e

https://github.com/BuilderIO/qwik/blob/87f0b9e5acf89f03204cc8e1cddaca8f54ad8ee9/starters/playwright.config.ts#L28
https://github.com/BuilderIO/qwik/blob/87f0b9e5acf89f03204cc8e1cddaca8f54ad8ee9/starters/dev-server.ts#L63
The build/v2 branch does not seem to have many changes compared to the master branch. It also seems that it is not possible to dynamically pass in the base parameter.

Here is my idea:

I referenced the [Astro](https://github.com/withastro/astro/blob/57959e607e81ec3b13bc55db6daa253a63485a97/packages/astro/test/test-utils.js#L97) test code, and I think I can dynamically pass in the **base** parameter in both **build** and **start** modes. **Vitest** can cover most scenarios, but I don't know how to test the **qwikLoader** function in Vitest, , so I want to use **e2e** tests.

wmertens:
> I think I can dynamically pass in the base parameter in both build and start modes

sounds good, give it a shot!

wmertens:
@JerryWu1234 there's a TS error https://github.com/BuilderIO/qwik/actions/runs/8508365123/job/23414373840?pr=6017#step:7:14

JerryWu1234:
@wmertens 

1. Could you please evaluate the feasibility of my test plan? 
2. I can't write a complete test yet because there is an error. The error is a bit complicated. I hope you can help me take a look.
image

wmertens (changes requested):
I propose you leave the tests out, on the v2 branch it will be clearer how to test e2e

wmertens:
@JerryWu1234 actually the e2e tests don't test qwik-city yet in v2, so we'll have to wait for that to add a test

JerryWu1234:
> actually the e2e tests don't test qwik-city yet in v2, so we'll have to wait for that to add a test

ok . add tests late

JerryWu1234:
last check before merge 

dev
has /newbase/
image


no /newbase/
image

pro
has /newbase/
image

no /newbase/
image

PatrickJS:
I do need this but I do set `base` to a cloudfront url so the regexp may not work

JerryWu1234:
> I do need this but I do set `base` to a cloudfront url so the regexp may not work

good idea
I will cover this function

JerryWu1234:
image
image
@PatrickJS 
it has already covered it

JerryWu1234:
@wmertens When will **build/v2** branch finish the test enhancement ?

JerryWu1234:
@wmertens is there anything else I need to do in this PR? 
It will encourage me to keep going if this PR is merged.
thanks

packages/qwik/src/core/platform/platform.ts
```diff
@@ -65,9 +65,10 @@ export const createPlatform = (): CorePlatform => {
  * @returns Fully qualified URL.
  */
 export const toUrl = (doc: Document, containerEl: QwikElement, url: string | URL): URL => {
-  const baseURI = doc.baseURI;
+  const baseURI = new URL(doc.baseURI);
   const base = new URL(containerEl.getAttribute('q:base') ?? baseURI, baseURI);
-  return new URL(url, base);
+  const pathUrl = (base.pathname + url).replace(/\/+/g, '/');
```
wmertens, line 70: `url` can be anything, including query string and hash, so you shouldn't run replace on it.
For example `toUrl(doc, el, "/hello/foo?stuff=///meep#123")` should work.
Just check if url isn't empty and starts with /

packages/qwik/src/core/platform/platform.ts
```diff
@@ -65,9 +65,10 @@ export const createPlatform = (): CorePlatform => {
  * @returns Fully qualified URL.
  */
 export const toUrl = (doc: Document, containerEl: QwikElement, url: string | URL): URL => {
-  const baseURI = doc.baseURI;
+  const baseURI = new URL(doc.baseURI);
```
wmertens, line 68: not sure why this is needed?

packages/qwik/src/core/platform/platform.ts
```diff
@@ -65,9 +65,10 @@ export const createPlatform = (): CorePlatform => {
...
   const base = new URL(containerEl.getAttribute('q:base') ?? baseURI, baseURI);
-  return new URL(url, base);
+  const pathUrl = (base.pathname + url).replace(/\/+/g, '/');
+  return new URL(pathUrl, base.origin);
```
wmertens, line 71: .origin can be `null`, for example when using Capacitor which uses a base of `capacitor://`

wmertens (changes requested):
I propose that you make sure that everywhere, `base` ends with a `/`. If you need to have it without `/` somewhere, slice it.

Also keep in mind that urls can be any url with any schema.

JerryWu1234:
I done

JerryWu1234:
@wmertens

PatrickJS:
@JerryWu1234 can you fix the conflicts

JerryWu1234:
> @JerryWu1234 can you fix the conflicts

Will fix it today

wmertens:
wait, actually, it's only pnpm-lock.yaml, so you can `git checkout main -- pnpm-lock.yaml && pnpm i`

JerryWu1234:
image

@wmertens I done this

JerryWu1234:
but still changed yaml file

JerryWu1234:
I copied code directly from the main branch

starters/apps/basic/src/root.tsx
```diff
@@ -20,7 +21,7 @@ export default component$(() => {
     <QwikCityProvider>
       <head>
         <meta charSet="utf-8" />
-        <link rel="manifest" href="/manifest.json" />
+        <link rel="manifest" href={`${BASE_URL}manifest.json`} />
```
PatrickJS, line 24: does BASE_URL end with a /?

PatrickJS (changes requested):

starters/apps/basic/src/root.tsx
PatrickJS: @JerryWu1234 one question then I can merge

PatrickJS: I'm going to merge and we can figure out later (at least we have time before we cut a release)

PatrickJS:
e2e tests are failing 🤔

JerryWu1234:
image
image

it's weird, I didn't modify entry.ssr.ts. however, I have to dive into

PatrickJS:
yeah very weird

PatrickJS:
oh maybe its erroring since the e2e tests are running the app in a browser so if the files can't load. thats the only thing i can think of

PatrickJS:
idk about this I think we need to make sure /build/ is added for scripts even if base was set

wmertens:
I think we need to take a step back here. `q:base` is the absolute path or full URL of where Qwik is supposed to get the dist files, and vite `base` is the prefix path to the url of the application.

- We need to make sure we use the right thing at the right time.
- q:base needs to be absolute or a full URL and end with `/`. It needs to include the vite base if it's not explicitly set.
- vite base needs to be absolute and end with `/`

I see a bunch of duplication around / handling, we can't have that. It needs to be simple throughout, with a helper function if needed. We know that the bases will always end with `/`.

Furthermore, it's up to the dev to use the vite base in their URLs, we can't add it for them, because there might be links between microfrontends.

JerryWu1234:
> I think we need to take a step back here. `q:base` is the absolute path or full URL of where Qwik is supposed to get the dist files, and vite `base` is the prefix path to the url of the application.
> 
> * We need to make sure we use the right thing at the right time.
> * q:base needs to be absolute or a full URL and end with `/`. It needs to include the vite base if it's not explicitly set.
> * vite base needs to be absolute and end with `/`
> 
> I see a bunch of duplication around / handling, we can't have that. It needs to be simple throughout, with a helper function if needed. We know that the bases will always end with `/`.
> 
> Furthermore, it's up to the dev to use the vite base in their URLs, we can't add it for them, because there might be links between microfrontends.

just want to confirm I understand or not.
such as if user set a Vite base: '/newbase/'
so that the `q:base` is `http://localhost:3000/newbase/` in dev 
and the `q:base` is `http://localhost:3000/newbase/build/` in pro 

@PatrickJS @wmertens 

I think we can try to easy way to fix this bug and waiting for built/v2.
because it's very difficult to guarantee whether it's  right or not without tests

PatrickJS:
q:base is for assets so it's saying this is where the /build/ files are. I think it makes sense right now to always make sure /build/ is in the url

starters/apps/basic/src/root.tsx
sreeisalso: It will be better to add / here instead of assuming BASE_URL ends with /, because even if it has a // will be treated as /

wmertens:
Actually I wonder if https://github.com/QwikDev/qwik/commit/b381b228531badce0253a6341c27d65f1a8023c0 fixed this - can you check if 1.5.5 still needs this PR @JerryWu1234 ?

JerryWu1234:
image
image
still has a problem

wmertens:
Thank you for checking @JerryWu1234, can you figure out exactly what is wrong? Is the q:base wrong, or is it requesting specific files wrongly etc

JerryWu1234:
> Thank you for checking @JerryWu1234, can you figure out exactly what is wrong? Is the q:base wrong, or is it requesting specific files wrongly etc

Just check where wrong is? or fix it in this PR?

wmertens:
@JerryWu1234 yes just where. Is it by any chance when it's trying to load `/src/something..._hashhashhash.js`? Then I know where to fix it (and I'm replacing that anyway)

PatrickJS:
I'll fix the pr

JuanmaVargas:
In production, is there any workaround to fix this? I'm trying to create a multi-zone in Next.js for a Qwik application, and for that, I need the vite.base set in the configuration to work for all the assets

PatrickJS:
I'll revert my changes except for the important ones and update the pr

PatrickJS:
@JuanmaVargas can you use [this](https://github.com/QwikDev/qwik/pull/6017#issuecomment-2155490844) to test out the changes

wmertens:
@JerryWu1234 thank you for your efforts, but I think it's fixed now in 1.7.1 (in a different way) so I'll close this.

PR #6022, opened 2024-03-17
fix(qwik-city): add explicit magic attr to select Link for SPA resume

jordanw66:
# Overview
Simple/partial fix for this issue: https://github.com/BuilderIO/qwik/issues/5983

Resuming SPA after context loss with refresh/reload requires a Link on the page or it falls back to location.href on popstate, it initially relied on a Qwik-internal key attached to Links. My impression of upcoming v2 changes is that this might disappear.

Either way this PR fixes this potential problem simply by adding a magic attr explicitly and selecting on that instead.

I also removed the erroneous `link:app` attribute, this was added in the `sync$` commit, as a check whether the Link was same-app or foreign, but this check got replaced in another PR with a better impl to just not attach the sync$ handler at all if it wasn't. There was no reason for it to exist at this point. Its presence in the API also seemed to be a workaround for linting complaining about it being on `linkProps` without being documented, this API didn't actually do anything at all, needless to say this was removed as well.


wmertens (approved):
LGTM, thanks!

PR #6043, opened 2024-03-22
chore: improve mutable/immutable split by optimizer

wmertens:
very wip but the snapshots should give an idea


wmertens:
this was merged

PR #6046, opened 2024-03-23
fix(v2): optimizer test fixes

Varixo:
# Overview

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/state/common.ts
+++ b/packages/qwik/src/core/state/common.ts
@@ -489,11 +489,20 @@ export class LocalSubscriptionManager {
           }
         } else {
           const signal = sub[SubscriptionProp.SIGNAL];
-          scheduler.$scheduleNodeDiff$(
-            host as fixMeAny,
-            sub[SubscriptionProp.ELEMENT] as fixMeAny,
-            untrack(() => signal.value)
-          );
+          if (type == SubscriptionType.PROP_IMMUTABLE || type == SubscriptionType.PROP_MUTABLE) {
+            scheduler.$scheduleNodeProp$(
+              host as fixMeAny,
+              sub[SubscriptionProp.ELEMENT] as fixMeAny,
+              sub[SubscriptionProp.ELEMENT_PROP],
+              signal
+            );
+          } else {
+            scheduler.$scheduleNodeDiff$(
+              host as fixMeAny,
+              sub[SubscriptionProp.ELEMENT] as fixMeAny,
+              untrack(() => signal.value)
+            );
+          }
         }
       } else {
         notifyChange(sub, this.$containerState$);
--- a/packages/qwik/src/core/state/common.ts
+++ b/packages/qwik/src/core/state/common.ts
@@ -489,20 +489,11 @@ export class LocalSubscriptionManager {
           }
         } else {
           const signal = sub[SubscriptionProp.SIGNAL];
-          if (type == SubscriptionType.PROP_IMMUTABLE || type == SubscriptionType.PROP_MUTABLE) {
-            scheduler.$scheduleNodeProp$(
-              host as fixMeAny,
-              sub[SubscriptionProp.ELEMENT] as fixMeAny,
-              sub[SubscriptionProp.ELEMENT_PROP],
-              signal
-            );
-          } else {
-            scheduler.$scheduleNodeDiff$(
-              host as fixMeAny,
-              sub[SubscriptionProp.ELEMENT] as fixMeAny,
-              untrack(() => signal.value)
-            );
-          }
+          scheduler.$scheduleNodeDiff$(
+            host as fixMeAny,
+            sub[SubscriptionProp.ELEMENT] as fixMeAny,
+            untrack(() => signal.value)
+          );
         }
       } else {
         notifyChange(sub, this.$containerState$);
--- a/packages/qwik/src/core/state/common.ts
+++ b/packages/qwik/src/core/state/common.ts
@@ -25,15 +25,11 @@ import type { QRL } from '../qrl/qrl.public';
 import { ELEMENT_PROPS, OnRenderProp } from '../util/markers';
 import { JSX_LOCAL } from '../v2/shared/component-execution';
 import { untrack } from '../use/use-core';
-import {
-  ElementVNodeProps,
-  type ClientContainer,
-  type VirtualVNode,
-  type VNode,
-} from '../v2/client/types';
+import { ElementVNodeProps, type VirtualVNode, type VNode } from '../v2/client/types';
 import { serializeClassWithHost2, stringifyStyle } from '../render/execute-component';
-import { type VNodeJournal, vnode_setAttr, VNodeJournalOpCode } from '../v2/client/vnode';
+import { vnode_setAttr, VNodeJournalOpCode } from '../v2/client/vnode';
 import type { ClassList } from '../render/jsx/types/jsx-qwik-attributes';
+import type { DomContainer } from '../v2/client/dom-container';
 
 /**
  * Top level manager of subscriptions (singleton, attached to DOM Container).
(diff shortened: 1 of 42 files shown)


packages/qwik/src/core/v2/ssr/ssr-render-jsx.ts
```diff
@@ -228,6 +228,21 @@ export function toSsrAttrs(
       continue;
     }
 
+    if (isSignal(value)) {
+      const signalNode = ssrContainer.getLastNode();
+      // TODO(hack): last node is previous node, not current,
```
mhevery, line 233: Yeah, so this hack will not work as we may clobber the parent ID. So let's not do that.

So the issue is that we need to get a hold of the current element. But the current element has not been created, because we are in the process of creating the attributes for the current element. Hence, we have a circular dependency that we need to solve somehow. 🤔 

Here is my suggestion: 
1. In `toSsrAttrs` don't try to create a subscription, just put the signal directly into the `SsrAttrs`
2. Change the `SsrAttrs` so that it is `export type SsrAttrs = Array<string | Signal<any> | null>;`
3. in `writeAttrs`, check `isSignal(value)`, and if it is a signal, then do the code which you have here. Inside `writeAttrs` a call to `getLastNode()` will return a correct value. 

What do you think? I think that would be much cleaner.

packages/qwik/src/core/state/common.ts
```diff
@@ -489,11 +489,20 @@ export class LocalSubscriptionManager {
           }
         } else {
           const signal = sub[SubscriptionProp.SIGNAL];
-          scheduler.$scheduleNodeDiff$(
-            host as fixMeAny,
-            sub[SubscriptionProp.ELEMENT] as fixMeAny,
-            untrack(() => signal.value)
-          );
+          if (type == SubscriptionType.PROP_IMMUTABLE || type == SubscriptionType.PROP_MUTABLE) {
+            scheduler.$scheduleNodeProp$(
```
mhevery, line 493: I don't think a call to `scheduler` is necessary. Check my thinking if it makes sense.

The scheduler is needed because we want to ensure that the side-effects of calling components functions (`useTask` etc.) are called in depth-first order. (Because they have `use*` hooks)

But updating a prop/attr (or text) does not need to go through the scheduler because it does not have any side effects (ie no `useTask` or hooks will be invoked). (Not considering DOM update as a side-effect we care about. More specifically, the order of DOM updates does not matter.)

Because no hooks are involved, the order of updating the DOM does not matter. (as long as we do all of the updates in a single micro-task, which we do because of journal) 

So the scheduler here is just overhead, and I think it is safe to skip it, as that will gain us speed.

packages/qwik/src/core/state/common.ts
```diff
@@ -489,11 +489,20 @@ export class LocalSubscriptionManager {
...
+              host as fixMeAny,
+              sub[SubscriptionProp.ELEMENT] as fixMeAny,
+              sub[SubscriptionProp.ELEMENT_PROP],
+              signal
+            );
+          } else {
+            scheduler.$scheduleNodeDiff$(
```
mhevery, line 500: FYI: This must go through scheduler because hooks may be involved.

packages/qwik/src/core/v2/shared/scheduler.ts
```diff
@@ -76,6 +76,7 @@ export const createScheduler = (container: Container2, scheduleDrain: () => void
     $scheduleTask$: scheduleTask,
     $scheduleComputed$: scheduleComputed,
     $scheduleNodeDiff$: scheduleNodeDiff,
+    $scheduleNodeProp$: scheduleNodeProp,
```
mhevery, line 79: I think this whole file can be reverted. See comment about scheduler not needed.

mhevery (changes requested):
WOW! Very nice work sir! Most impressed. You are absolutely killing it. 

I left some comments about the hack and the scheduler. Check me if they make sense to you, and if they do please update the PR.

packages/qwik/src/core/v2/ssr/ssr-render-jsx.ts
Varixo: While this works there is a problem with this.
I need to import these things inside ssr-container and this is not allowed:

```
import { trackSignal } from '../core/use/use-core';
import { SubscriptionType } from '../core/state/common';
```

packages/qwik/src/core/state/common.ts
Varixo: While this works I need to import few things that are not allowed to import inside the ssr-container.
```ts
import { trackSignal } from '../core/use/use-core';
import { SubscriptionType } from '../core/state/common';
```

packages/qwik/src/core/v2/shared/jsx-props.ts
```diff
@@ -0,0 +1,34 @@
+import type { JSXNode } from '@builder.io/qwik/jsx-runtime';
+import { _IMMUTABLE } from '../../internal';
+
+export function fixJsxProps(jsx: JSXNode<unknown>) {
```
mhevery, line 4: So this function scans both objects and does a whole bunch of megamorphic reads.

Can you explain why is it needed? Are there other faster ways to do this?

packages/qwik/src/core/state/common.ts
```diff
@@ -490,11 +499,15 @@ export class LocalSubscriptionManager {
...
           const signal = sub[SubscriptionProp.SIGNAL];
           if (type == SubscriptionType.PROP_IMMUTABLE || type == SubscriptionType.PROP_MUTABLE) {
-            scheduler.$scheduleNodeProp$(
-              host as fixMeAny,
-              sub[SubscriptionProp.ELEMENT] as fixMeAny,
-              sub[SubscriptionProp.ELEMENT_PROP],
-              signal
+            const target = sub[SubscriptionProp.ELEMENT] as fixMeAny as VirtualVNode;
+            const propKey = sub[SubscriptionProp.ELEMENT_PROP];
+            updateNodeProp(
```
mhevery, line 504: I would expect a call to `container.scheduleRender()` so that the journal gets flushed.

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -603,8 +600,8 @@ export const vnode_diff = (container: ClientContainer, jsxNode: JSXOutput, vStar
     const srcLength = srcAttrs.length;
     let dstIdx = ElementVNodeProps.PROPS_OFFSET;
     let dstLength = dstAttrs.length;
-    let srcKey: string | null = srcIdx < srcLength ? srcAttrs[srcIdx++] : null;
-    let dstKey: string | null = dstIdx < dstLength ? dstAttrs[dstIdx++] : null;
+    let srcKey = srcIdx < srcLength ? (srcAttrs[srcIdx++] as SsrAttrKey) : null;
```
mhevery, line 603: I don't think this change is correct. 

`SsrAttrKey` is SSR only type. `vnode_diff` is Client only. So I think it is wrong to have SSR type here. Maybe create a new type?

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -38,7 +38,7 @@ import {
 import { addPrefixForScopedStyleIdsString, isClassAttr } from '../shared/scoped-styles';
 import type { QElement2, fixMeAny } from '../shared/types';
 import { DEBUG_TYPE, VirtualType } from '../shared/types';
-import type { SsrAttrs } from '../ssr/ssr-types';
+import type { SsrAttrKey, SsrAttrs } from '../ssr/ssr-types';
```
mhevery, line 41: We should clean this up. `vnode_diff` is a client only and should not have any references to SSR. We may need to create new types.

mhevery (approved):
Few more nit-picks.

packages/qwik/src/core/v2/shared/jsx-props.ts
Varixo: so basically sometimes mutable props are immutable and we need fix that
 the code is from here
 https://github.com/BuilderIO/qwik/blob/cf86daa15f62b429b5b25be83ccda9b40a4d3d7c/packages/qwik/src/core/render/ssr/render-ssr.ts#L602

packages/qwik/src/core/v2/client/vnode-diff.ts
Varixo: right, this is client only

mhevery (approved):

wmertens (approved):

PR #6054, opened 2024-03-26
refactor(v2): varProps and constProps and PropsProxy

wmertens:
- refactors mutable + immutable props to varProps and constProps
- adds a proxy that merges var+const props
- removes _jsxS


packages/qwik/src/optimizer/core/src/transform.rs
```diff
@@ -838,27 +838,16 @@ impl<'a> QwikTransform<'a> {
...
+		let (jsx_func, mut args) = if is_fn || has_spread_props {
 			(
 				self.ensure_core_import(&_JSX_C),
-				vec![node_type, mutable_props, flags, key],
-			)
-		} else if dynamic_props {
-			(
-				self.ensure_core_import(&_JSX_S),
-				vec![node_type, mutable_props, immutable_props, flags, key],
+				vec![node_type, var_props, const_props, flags, key],
```
mhevery, line 845: potential difference

mhevery, line 845: potential difference

wmertens: Not sure what you mean, this will definitely give different results than before

PR #6059, opened 2024-03-27
fix: Build breaking after adding bun adapter 

Twiggeh:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description
[Issue](https://github.com/BuilderIO/qwik/issues/6058)
When running the adapter for adding bun, it misses adding the types to `tsconfig.json` (this is also not detailed in the documentation for bun). When these types are missing the build fails.

I used lodash to do a deep merge over the jsons, I am sorry if there was another merging mechanism in place. I only found `@typescript-eslint/utils` which had a deepMerge, but that balooned the output of the cli.js to over 110k lines, so I opted to use `lodash` instead, which already is a peer dependency. 

# Use cases and why

- I don't expect to check which files the cli didn't modify after adding the adapter

# Checklist:

- [-] I have made corresponding changes to the documentation
- [-] Added new tests to cover the fix / functionality


starters/adapters/bun/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "compilerOptions": { "types": ["bun-types"] }
```
gioboa, line 2: To prevent the extra dependency we can define all the `"types"` (dafault + bun ) in the bun tsconfig.json
This revert some of your changes as well

gioboa (changes requested):
Thanks @Twiggeh for your help

starters/adapters/bun/tsconfig.json
octet-stream: Or types can be moved into an `.d.ts` file instead, like in Nuxt or Remix and then referenced in `includes` field. This way typescript will be able to pick types from `$PWD/node_modules/@types`. Also, this way TypeScript should pick bun types from `@types/bun` package without any additional configuration

See the documentation: https://www.typescriptlang.org/tsconfig#types

If `types` field specified, then TS won't pick anything that is not included in the list:

> This tsconfig.json file will only include ./node_modules/@types/node, ./node_modules/@types/jest and ./node_modules/@types/express. Other packages under node_modules/@types/* will not be included.

So suggestion is to create a file, let's call it `qwik.env.d.ts` for example, where we reference `vite/client` types, and then add it to the `includes`.

The template would look like this:

In `qwik.env.d.ts`:

```ts
/// <reference types="vite/client" />
```

Then in `tsconfig.json`:

```json5
{
  "include": ["qwik.env.d.ts", "**/*.ts", "**/*.tsx"]
}
```

Then for those who want to keep using `bun-types` (as they might be updated more frequently, aligned with the runtime releases) instead of `@types/bun`, they can just remove `@types/bun` after they installed the adapter and reference bun-types same way in `qwik.env.d.ts`

package.json
```diff
@@ -97,6 +98,7 @@
     "execa": "8.0.1",
     "express": "4.18.3",
     "install": "^0.13.0",
+    "lodash": "^4.17.21",
```
wmertens, line 101: Is there no smaller dependency? Or just few lines of helper code?

At the very least you should use lodash-es

packages/qwik/src/cli/add/update-files.ts
```diff
@@ -86,11 +87,15 @@ async function mergeJsons(fileUpdates: FsUpdates, srcPath: string, destPath: str
   try {
     const srcPkgJson = JSON.parse(srcContent);
     const destPkgJson = JSON.parse(await fs.promises.readFile(destPath, 'utf-8'));
-    Object.assign(srcPkgJson, destPkgJson);
+    const mergedJson = mergeWith(srcPkgJson, destPkgJson, (value, srcValue) => {
+      if (Array.isArray(value)) {
```
wmertens, line 91: This seems dangerous to do for unknown fields

PR #6083, opened 2024-04-01
fix(v2): fix attributes e2e tests

Varixo:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/starters/apps/e2e/src/components/attributes/attributes.tsx
+++ b/starters/apps/e2e/src/components/attributes/attributes.tsx
@@ -103,9 +103,9 @@ export const AttributesChild = component$<{ v: number }>(({ v }) => {
       <div>
         {hide.value ? (
           <>
-            <label id="label" key="abc" />
-            <input id="input" key="1" />
-            <svg id="svg" key="cba">
+            <label id="label" />
+            <input id="input" />
+            <svg id="svg">
               <feGaussianBlur></feGaussianBlur>
               <foreignObject>
                 <foreignObject></foreignObject>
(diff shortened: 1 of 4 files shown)


starters/e2e/e2e.attributes.spec.ts
```diff
@@ -62,12 +62,12 @@ test.describe("attributes", () => {
       await stuffBtn.click();
       await expect(inputCopy).toHaveJSProperty("value", "Hello");
       await expect(inputValue).toHaveText("Hello");
-      await expect(renders).toHaveText("2");
+      await expect(renders).toHaveText("1");
```
wmertens, line 65: Why the change here?

wmertens (commented):
Nice!

starters/e2e/e2e.attributes.spec.ts
Varixo: this additional rerender was caused by removed console.log. I don't know if this is needed

wmertens (approved):

starters/apps/e2e/src/components/attributes/attributes.tsx
```diff
@@ -104,9 +103,9 @@ export const AttributesChild = component$<{ v: number }>(({ v }) => {
       <div>
         {hide.value ? (
           <>
-            <label id="label" />
-            <input id="input" />
-            <svg id="svg">
+            <label id="label" key="abc" />
```
mhevery, line 106: This does not seem right. Why do we need to add the key?

mhevery (changes requested):

Issue #6084, opened 2024-04-02
[🐞] Qwik + Astro, useVisibleTask$ does not work when using VT


ITakayuki:
### Which component is affected?

Qwik Runtime

### Describe the bug

When using useVisibleTask$ in a Qwik component, if <ViewTransition/> is set, useVisibleTask$ is not executed on page transitions in qwikdev/astro.
track, etc. will not work as well.
Please let me know if there is any workaround🙏
Thanks.

## code
```tsx
import { component$, useVisibleTask$ } from "@builder.io/qwik";

export default component$(() => {

  // Not executed when transitioned by ViewTransition
  useVisibleTask$(() => {
    console.log("useVisibleTask$");
  });
  return /* .... */
});
```

## packages
```json
"dependencies": {
  "@astrojs/check": "0.5.10",
  "@builder.io/qwik": "1.5.1",
  "@qwikdev/astro": "0.5.13",
  "astro": "4.5.12",
  "typescript": "5.4.3"
}
```

## Issues
- [ ] document events run once per container, even when a container is added later
- [ ] we add a page change event for SPA. It also fires on first load together with document idle
- [ ] we fix visibility event with transition
- [ ] we fix not working `track` and `cleanup`

### Reproduction

https://codesandbox.io/p/devbox/tvj6y4?file=%2Fsrc%2Fcomponents%2FQwikComponent.tsx%3A4%2C26

### Steps to reproduce

1. Add `<ViewTransition /> in `.astro`  
2. Transition to a page using Qwik Component.
3. `useVisibleTask$` is not executed

### System Info

```shell
System:
    OS: macOS 13.6.3
    CPU: (8) arm64 Apple M2
    Memory: 53.94 MB / 24.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 18.14.1 - ~/.volta/tools/image/node/18.14.1/bin/node
    npm: 9.3.1 - ~/.volta/tools/image/node/18.14.1/bin/npm
  Browsers:
    Chrome: 123.0.6312.87
  npmPackages:
npmPackages:
    @builder.io/qwik: ^1.5.1 => 1.5.1 
    @qwikdev/astro: ^0.5.13 => 0.5.13
```

maiieul:
Hey @ITakayuki, please could you re-create your issue in the [qwik/astro repository](https://github.com/QwikDev/astro/issues) 🙏? 

[...]

PR #6087, opened 2024-04-03
fix: make the event listener on the window to use the capture option

Varixo:
# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


wmertens:
Does this also solve #6084 @ITakayuki?

ITakayuki:
@wmertens   
No, it doesn't seem to have been resolved yet.
The problem with #6084   is that the contents of `processReadyStateChange` are not executed unless `hasInitialized` is set back to 0, so I think that needs to be fixed as well.

PR #6123, opened 2024-04-16
fix(qwikloader): stopPropagation for any async handlers

PatrickJS:
sync$ may not be async function and e.stopPropagation() won't work unless it's fired immediately

fixes https://github.com/QwikDev/qwik/issues/5322

used this repo with the fix in a custom qwik loader
https://github.com/PatrickJS/bug-qwik-loader-stopPropagation

### DIFF
--- a/packages/qwik/src/qwikloader.ts
+++ b/packages/qwik/src/qwikloader.ts
@@ -28,7 +28,7 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
     return doc.querySelectorAll(query);
   };
 
-  const isPromise = (promise) => promise && typeof promise.then === 'function';
+  const isPromise = (promise: Promise<any>) => promise && typeof promise.then === 'function';
 
   const broadcast = (infix: string, ev: Event, type = ev.type) => {
     querySelectorAll('[on' + infix + '\\:' + type + ']')[forEach]((el) =>
@@ -67,7 +67,7 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
     if (relevantListeners && relevantListeners.length > 0) {
       for (const listener of relevantListeners) {
         // listener[1] holds the QRL
-        const results = listener[1].getFn([element, ev], () => element[isConnected])(ev, element);
+        const result = listener[1].getFn([element, ev], () => element[isConnected])(ev, element);
         let cancelBubble = ev.cancelBubble;
         if (isPromise(result)) {
           await result;
@@ -142,14 +142,12 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
     while (element && element[getAttribute]) {
       const results = dispatch(element, '', ev, type);
       let cancelBubble = ev.cancelBubble;
-      if (isPromise(results) {
+      if (isPromise(results)) {
         await results;
       }
       // if another async handler stopPropagation
       cancelBubble =
-          cancelBubble ||
-          ev.cancelBubble ||
-          element.hasAttribute("stoppropagation:" + ev.type);
+        cancelBubble || ev.cancelBubble || element.hasAttribute('stoppropagation:' + ev.type);
       element = ev.bubbles && ev.cancelBubble !== true ? element.parentElement : null;
     }
   };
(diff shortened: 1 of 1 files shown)


packages/qwik/src/qwikloader.ts
```diff
@@ -99,7 +99,11 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
                 element: element,
                 reqTime,
               });
-            await handler(ev, element);
+            const results = handler(ev, element);
+            // sync$ may not be async function and e.stopPropagation() won't work unless it's fired immediately
+            if (results && typeof results.then === 'function') {
```
PatrickJS, line 104: I'm trying to avoid `instanceof Promise` here in favor for thenable but maybe it's less code just using instanceof Promise

PatrickJS:
ok so this really should fix this problem if you see here only the first sync fn with e.stopPropagation() will make it work correctly

https://qwik.dev/playground/#f=7VdNS8NAEP0rSw7tBtvQeJLgIh4F8SB6kh7UxNpLLdKDEPa%2F%2B2Znd5LdtEURtEKDtelkdna%2B9r1Jr2lOy7Nepbr6TxSdXnzhL7BvyHqWVilLI795qxtE3m8nkCuEMOs5BPld3TMD2R0JwuzhA%2FH8TCp84chj%2F0qsFQ75jAEd09YXKivXHwoYuqzVE8gqU1VftHhvmlU2EXPrx7pG8StVzoLQWn9D%2FD8lq1Paw7R067YLCjIKtA9i0CVP60YiCxd4sCkwdqyvwDn1EuHjcGP2YbBBebao7tMYUJ8GddCmHW2TZKJmeS4rbd6Fzp4yVmvUhweA3X7vc6Y%2Fe7GdAp%2FLDU4W5sdGj%2BNcjvPUQGgJLUmONH4r2L%2BP9KAqe%2FjBzvksBtwbnlGWPNM8B7xXo5EYiR4UGOK189KliXGIr3OHafSIIcAK9pg23FmJ1LThznZwDPjN2SF%2Br%2FJ4jE%2BMnndApwF6biD8GnqS5v6IfhgL1XfoNTFP7LQ45l%2F5vGVoBUbR3IhUKeCzB2Q37VfKIbHbJhiQkI2CdddGXTMnTOBscpPJE36NMKngRJWkGG1FOYRm5B7ReXDRZ7dSAfK7ExW01XXzEvS3r4nXDdamBhIj8%2F4T2%2F0Quci2%2BHa7XLzucC5d735jaOhNgq5BKUWmpf%2FfaR%2FfOem4huY5DpL%2Fb5D8BA

PatrickJS:
this fix wasn't the fix I'm looking into qwikloader

PatrickJS:
this might be a chrome bug
```js
// Add a click event listener to the document body
document.body.addEventListener('click', function(e) {
  // Stop propagation
  e.stopPropagation();
  
  // Log cancelBubble value immediately after stopping propagation
  console.log('Immediate cancelBubble:', e.cancelBubble); 

  // Set a timeout
  setTimeout(function() {
    // Log cancelBubble value after the timeout
    console.log('After setTimeout cancelBubble:', e.cancelBubble);
    // e.cancelBubble <--- should be true but it's false 
  }, 1000);
});

```

thejackshelton:
LGTM 🙌 . Merging

packages/qwik/src/qwikloader.ts
```diff
@@ -60,15 +62,20 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
...
         // listener[1] holds the QRL
-        await listener[1].getFn([element, ev], () => element[isConnected])(ev, element);
+        const results = listener[1].getFn([element, ev], () => element[isConnected])(ev, element);
+        const cancelBubble = ev.cancelBubble;
+        if (isPromise(results)) {
+          await results;
+        }
+        // forcing async with await resets ev.cancelBubble to false
+        if (cancelBubble) {
+          ev.stopPropagation();
```
wmertens, line 77: Shouldn't this be called before awaiting?

thejackshelton: > Shouldn't this be called before awaiting?

Ah yes, I don't think we want any possible delay here, so something like:

```tsx
const cancelBubble = ev.cancelBubble;
if (cancelBubble) {
  ev.stopPropagation();
}
if (isPromise(results)) {
  await results;
}
```

does that look right @wmertens ?

PR #6126, opened 2024-04-16
fix(v2): styles scoped id fixes

Varixo:
This PR fixes the styles scoped id rendering


packages/qwik/src/core/render/jsx/jsx-runtime.ts
```diff
@@ -142,6 +142,22 @@ export class JSXNodeImpl<T> implements JSXNode<T> {
   }
 }
 
+function retrieveStyleScopedIdFromInvokeContext() {
+  let styleScopedId: string | null = null;
+  const ctx = tryGetInvokeContext();
+  if (ctx && ctx.$container2$) {
+    let scopedStyleIds: Set<string> | undefined;
+    // calculations of styleScopedId for dom container are done inside the vnode-diff, because of optimizations (number of runs)
```
Varixo, line 150: I don't like this. We can do setting the stylesScopedId for DomContainer here too, but this will be less efficient than the current logic (more `getObjectById` calls). 

This will be confusing, but I don't have better idea. Maybe the cost of calling `getObjectById` function for every JSXNode is not that big and we can do this here?

PR #6140, opened 2024-04-21
docs(deprecated-features): routeAction$

PatrickJS:
routeAction$


gioboa (approved):
Nice catch @PatrickJS

wmertens:
@gioboa it doesn't support base yet, I don't think this was ready to merge :(

Anyway @PatrickJS you can use `import.meta.env.BASE_URL` I think.

wmertens:
oops I meant #6141 - sorry for the noise

Issue #5227, opened 2023-09-27
[🐞] callbackUrl does not work with Auth.js integration


k-schneider:
### Which component is affected?

Qwik Runtime

### Describe the bug

When specifying either a relative or absolute url (where the protocol and origin match hosted site) the callback url does not work as expected and you are always redirected to the root of the site.

### Reproduction

https://stackblitz.com/edit/qwik-starter-ug5wfp

### Steps to reproduce

- Open .env and replace the `GITHUB_ID` and `GITHUB_SECRET` with valid values
- Run `npm install` followed by `npm run dev`
- Click the login button I added to the scaffolded header component and login via GitHub

Expected:
Because I hardcoded a callbackUrl of `http://localhost:5173/demo/flower` you should end up on the flower demo after signing in.

Actual:
You end up on the homepage.

**Note: This plugin doesn't seem to like stackblitz so you'll have to do this locally**

### System Info

```shell
System:
  OS: Windows 10 10.0.19045
  CPU: (12) x64 AMD Ryzen 5 7600X 6-Core Processor
  Memory: 40.40 GB / 63.12 GB
Binaries:
  Node: 18.17.1 - C:\Program Files\nodejs\node.EXE
  npm: 9.6.7 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Edge: Spartan (44.19041.1266.0), Chromium (117.0.2045.43)
  Internet Explorer: 11.0.19041.1566
npmPackages:
  @builder.io/qwik: ^1.2.11 => 1.2.11
  @builder.io/qwik-auth: 0.1.1 => 0.1.1
  @builder.io/qwik-city: ^1.2.11 => 1.2.11
  undici: 5.22.1 => 5.22.1
  vite: 4.4.7 => 4.4.7
```

### Additional Information

I didn't include this in the example but I noticed that the `redirect` callback seems to be invoked on every request.  The documentation states that it should only be called anytime the user is redirected to a callback URL so this seems incorrect.

In `src/routes/plugin@auth.ts` you can add this callbacks section to see how often the callback is invoked:

```
import { serverAuth$ } from "@builder.io/qwik-auth";
import GitHub from "@auth/core/providers/github";
import type { Provider } from "@auth/core/providers";

export const { onRequest, useAuthSession, useAuthSignin, useAuthSignout } =
  serverAuth$(({ env }) => ({
    secret: env.get("AUTH_SECRET"),
    trustHost: true,
    providers: [
      GitHub({
[...]

PR #6156, opened 2024-04-24
fix: #5227

lieranderl:
- fix for #5227


PatrickJS:
@lieranderl can you install @dev version and confirm that callback is correct

PR #6161, opened 2024-04-26
feat(image-jsx): automatic compression and transcoding for AVIF, WebP…

xhcy1314:
Automatic compression and transcoding for AVIF, WebP, and original formats.

Provides smaller image formats while ensuring proper display on low-end devices.

# Overview

This commit enhances the image processing functionality. Now, the image tools automatically compress and transcode AVIF, WebP, and original formats, providing smaller image formats while ensuring proper display on low-end devices.
# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

This commit enhances the image processing functionality. Now, the image tools automatically compress and transcode AVIF, WebP, and original formats. It provides smaller image formats while ensuring proper display on low-end devices. This improvement contributes to enhancing user experience, particularly for users with lower-performance devices.

# Use cases and why

- 1. A single configuration allows selecting either AVIF, WebP, or PNG image formats. Choosing AVIF format will result in images not being displayed on devices that do not support AVIF.
- 2. Similarly, selecting WebP format will result in images not being displayed on devices that do not support WebP.

# Checklist:


wmertens (changes requested):
The idea is good but then you also need to change the jsx output so that it uses `<picture>` instead of ``, because srcset doesn't let you tell the browser what the type is

https://stackoverflow.com/a/67698595/124416

xhcy1314:
> The idea is good but then you also need to change the jsx output so that it uses `<picture>` instead of ``, because srcset doesn't let you tell the browser what the type is
> 
> https://stackoverflow.com/a/67698595/124416

I'll work on implementing it and testing its effectiveness as soon as possible. Thanks  for the reminder!

xhcy1314:
> 这个想法很好，但是您还需要更改 jsx 输出，以便它使用`<picture>`而不是``，因为 srcset 不允许您告诉浏览器类型是什么很好，但是您需要更改 jsx 输出，以便它使用 <picture> 而不是 ，因为 srcset 没有告诉浏览器类型是什么
> 
> [https://stackoverflow.com/a/67698595/124416https://stackoverflow.com/a/67698595/124416](https://stackoverflow.com/a/67698595/124416)

Is there anything else I need to do ？

PatrickJS:
@wmertens is this ready?

wmertens (commented):
This looks good, but:
- it is slightly breaking in terms of CSS, no? I'm not sure if this should wait for v2 or if a new major is enough. @PatrickJS ?
- it would be great if we had some unit tests of the output, just provide fake imagedata and show what the result should be.

PatrickJS:
if we have unit/e2e tests we should be fine

wmertens:
I think we should make this configurable, and in V2 we enable it by default. 

What do you think @xhcy1314 ?

xhcy1314:
> I think we should make this configurable, and in V2 we enable it by default.
> 
> So right now, by default keep the img output as-is, but allow avif with picture via a setting.
> 
> What do you think @xhcy1314 ?

I think this approach is reasonable. By making this feature configurable, we can introduce the new format without disrupting the current workflow. Enabling it by default in V2 will reduce the cognitive load on users, allowing them to focus on their business code.

However, i need your suggestion on what to name this configuration key.

wmertens:
@xhcy1314 how about this: 

- make the default format webp again
- add jsdoc to the format parameter to explain that you can set multiple formats separated by `;`
- if there are multiple formats, generate a picture, otherwise an img

xhcy1314:
> @xhcy1314这个怎么样：
> 
> * 重新设置默认格式 webp
> * 在格式参数中添加 jsdoc，说明可以设置多个格式，用`;`
> * 如果有多种格式，则生成一张图片，否则生成一个 img

Apologies for the delay. I was deeply immersed in business tasks until now, but I have now completed the code modifications as per our previous agreement. Could you please confirm if everything now meets the requirements?

packages/qwik-city/buildtime/vite/image-jsx.ts
```diff
@@ -84,17 +134,25 @@ export function imagePlugin(userOpts?: QwikCityVitePluginOptions): PluginOption[
...
               `
   import { _jsxQ } from '@builder.io/qwik';
-  const PROPS = {srcSet, width, height};
+  const PROPS = {...img};
   export default function (props, key, _, dev) {
-    return _jsxQ('img', {...{decoding: 'async', loading: 'lazy'}, ...props}, PROPS, undefined, 3, key, dev);
+    if(!sources?.length) return null;
+    if(sources.length < 2){
+      // Single format directly returns the <img />
+      return _jsxQ('img', {...{decoding: 'async', loading: 'lazy'}, ...props}, sources[0], undefined, 3, key, dev);
```
wmertens, line 150: Since 1.8.1 we can use jsx inside qwik-city. This will also make it work with v2. So could you rename this file to .tsx and then just use `<img decoding="async"...`? There's also some merge conflicts so perhaps best to squash and rebase.

wmertens (changes requested):
:raised_hands: looks almost right to merge, we just need to fix the merge conflicts and switch to `.tsx`.

Issue #6177, opened 2024-04-29
[🐞] lib mode do not respect custom lib.entry in vite config


diecodev:
### Which component is affected?

Qwik Rollup / Vite plugin

### Describe the bug

When you try to overwrite the entrypoints in vite.config.ts file in lib mode, the optimizer just replace your specified entrypoint for the default, disrespecting the one you specified. I take a look to the optimizer.js file and the error occurs in here `line 1870`:
<img width="1408" alt="image" src="https://github.com/QwikDev/qwik/assets/51871681/fbb6b604-3c9e-4dcc-93b2-07d2f013421c">
 

### Reproduction

https://github.com/diecodev/qwik-date/blob/main/vite.config.ts

### Steps to reproduce

- create a qwik component library project
- edit vite config file and add a custom lib.entry object or array ~ you can do this in vite, is in the [docs](https://vitejs.dev/config/build-options.html#build-lib) ~ (only if you pass a custom string as entry point this is gonna be the entry, otherwise always is gonna be replaced by 'src/index.ts'

### System Info

```shell
System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M2
    Memory: 67.77 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2 - ~/Library/Caches/fnm_multishells/28998_1714366621848/bin/node
    Yarn: 1.22.21 - ~/Library/Caches/fnm_multishells/28998_1714366621848/bin/yarn
    npm: 10.5.0 - ~/Library/Caches/fnm_multishells/28998_1714366621848/bin/npm
    pnpm: 9.0.6 - ~/Library/Caches/fnm_multishells/28998_1714366621848/bin/pnpm
    bun: 1.1.6 - ~/.bun/bin/bun
  Browsers:
    Chrome: 124.0.6367.91
    Safari: 17.4.1
  npmPackages:
    @builder.io/qwik: 1.5.2 => 1.5.2
    undici: * => 6.14.1
    vite: ^5.1.6 => 5.2.10
```

### Additional Information

Fix this is very easy, you only need to check when the lib mode is activate if the `inputOpts.input` parameter in the `normalizeOptions` function is type Object or Array and make the proper handling for each case.

PatrickJS:
fixed by https://github.com/QwikDev/qwik/pull/6178

diecodev:
I just forgot to closr this issue. Thank you.

PR #6178, opened 2024-04-29
fix(plugin/optimizer): 🐛 Enhance support for custom entry points in Vite config

diecodev:
# Overview

This PR aims to enhance the support for custom entry points in the Vite configuration in library mode, enabling developers to specify custom entry points and facilitating the creation of libraries with custom submodules. By extending the capability of Vite configuration, this enhancement provides flexibility and empowers developers to tailor their projects more effectively.

# What is it?

- [X] Feature / enhancement
- [X] Bug Fix

# Description

This PR modifies the code to allow for custom entry points in the Vite configuration (library mode). It includes updates to handle custom entry points provided in the Vite config file. This enhancement aims to address the need for developers to create libraries with custom submodules by providing a more flexible configuration option.

# Use cases and why
The primary use cases for this enhancement include:

1. **Custom Entry Points:** Developers can now specify custom entry points in the Vite config file, allowing for more granular control over project structure and module loading.
2. **Library Creation:** This enhancement facilitates the creation of libraries with custom submodules, enabling developers to structure their libraries in a way that best suits their needs.

# Checklist:

# Issues Related:
This PR solve my own issue: #6177

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/plugin.ts
+++ b/packages/qwik/src/optimizer/src/plugins/plugin.ts
@@ -215,22 +215,37 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
           // client input default
           opts.input = [path.resolve(srcDir, 'root.tsx')];
         } else if (opts.target === 'lib') {
-          // lib input default
-          opts.input = [path.resolve(srcDir, 'index.ts')];
+          if (typeof updatedOpts.input === 'object') {
+            for (const key in updatedOpts.input) {
+              const resolvedPaths: { [key: string]: string } = {};
+              if (Object.hasOwnProperty.call(updatedOpts.input, key)) {
+                const relativePath = updatedOpts.input[key];
+                const absolutePath = path.resolve(opts.rootDir, relativePath);
+                resolvedPaths[key] = absolutePath;
+              }
+
+              opts.input = { ...opts.input, ...resolvedPaths };
+            }
+          } else {
+            // lib input default
+            opts.input = [path.resolve(srcDir, 'index.ts')];
+          }
         } else {
           opts.input = [];
         }
       }
-      opts.input = opts.input.reduce((inputs, i) => {
-        let input = i;
-        if (!i.startsWith('@') && !i.startsWith('~')) {
-          input = normalizePath(path.resolve(opts.rootDir, i));
-        }
-        if (!inputs.includes(input)) {
-          inputs.push(input);
-        }
-        return inputs;
-      }, [] as string[]);
+      opts.input = Array.isArray(opts.input)
+        ? opts.input.reduce((inputs, i) => {
+            let input = i;
+            if (!i.startsWith('@') && !i.startsWith('~')) {
+              input = normalizePath(path.resolve(opts.rootDir, i));
+            }
+            if (!inputs.includes(input)) {
+              inputs.push(input);
+            }
+            return inputs;
+          }, [] as string[])
+        : opts.input;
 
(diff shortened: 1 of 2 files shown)


thejackshelton:
On first glance everything looks good. It takes a custom path which is resolved, and allows for object inputs. If anyone else could review before merging.

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -215,22 +215,37 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
...
-        }
-        if (!inputs.includes(input)) {
-          inputs.push(input);
-        }
-        return inputs;
-      }, [] as string[]);
+      opts.input = Array.isArray(opts.input)
+        ? opts.input.reduce((inputs, i) => {
+            let input = i;
+            if (!i.startsWith('@') && !i.startsWith('~')) {
```
PatrickJS, line 240: can you also include `#` since thats the official alias now

diecodev: Done!😁

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -299,8 +314,7 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
         if (typeof opts.srcDir === 'string' && !fs.existsSync(opts.srcDir)) {
           throw new Error(`Qwik srcDir "${opts.srcDir}" not found.`);
         }
-        for (const alias in opts.input) {
-          const input = opts.input[alias];
+        for (const [_, input] of Object.entries(opts.input)) {
```
wmertens, line 317: Use `Object.values`, put it in a const and then you can use it instead of firstInput too.

packages/qwik/src/optimizer/src/plugins/vite-server.ts
```diff
@@ -84,7 +84,13 @@ export async function configureDevServer(
           return;
         }
 
-        const ssrModule = await server.ssrLoadModule(opts.input[0]);
+        let firstInput: string;
+        if (Array.isArray(opts.input)) {
+          firstInput = opts.input[0];
+        } else {
+          firstInput = Object.values(opts.input)[0];
```
wmertens, line 91: Just use this always

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -535,12 +555,18 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
       debug(`load()`, 'Found', id);
 
       let code = transformedModule[0].code;
+      let firstInput: string;
+      if (Array.isArray(opts.input)) {
+        firstInput = opts.input[0];
+      } else {
+        firstInput = Object.values(opts.input)[0];
```
wmertens, line 562: Just use this always

wmertens (changes requested):
Tiny nitpick, other than that it looks great!

packages/qwik/src/optimizer/src/plugins/plugin.ts
diecodev: I can do that in a global variable (let instead of const), but I need to make the Array.isArray verification and the assignment in every case, because the inputs are used in three different functions.

PatrickJS: yeah, I think this is fine. any more changes it seems like a larger refactor

PR #6181, opened 2024-04-29
feat(v2): SSRBlock and SSRStream implementation

Varixo:
This PR adds support for the `SSRBlock` and `SSRStream` components


packages/qwik/src/server/v2-ssr-container.ts
```diff
@@ -421,6 +421,10 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
     this.write(rawHtml);
   }
 
+  commentNode(text: string) {
+    this.write('<!--' + text + '-->');
```
mhevery, line 425: I think these comment nodes will confuse the current implementation of `materializeFromVNodeData,` so we will need to update that.

mhevery (changes requested):

mhevery (approved):

PR #6185, opened 2024-04-30
fix(starters): Fix vite types referencing in base starter

octet-stream:
Closes #6059

# Overview

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

It's been about a month since [this](https://github.com/QwikDev/qwik/pull/6059) PR was opened and to this point I doubt that it is going somewhere, so I decided to take over and fix this issue, using the same solution I have proposed [here](https://github.com/QwikDev/qwik/pull/6059#discussion_r1543971400), and I was able to verify it is working.

# Use cases and why

My [previous](https://github.com/QwikDev/qwik/pull/5942) attempt to fix types for Bun turned out to be incomplete, so this PR fixes the issue. Here's a brief list of changes:

- [x] Move `vite/client` types referencing to a separate file called qwik.env.d.ts. This file can be used for more global typings in a future, if needed.
- [x] Remove `types` field in `compilerOptions` for base app starter with a reference to that file in `includes` field instead.

As I said before, the `types` field narrows the list of the sources where TypeScript will look for typings for global scope, so I don't think it would be a good fit when there's many other typings that can be needed from `node_modules/@types`. In our case, we need to support both `@types/node` and `@types/bun` both of which has global types. See documentation for more info: https://www.typescriptlang.org/tsconfig#types 

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/qwik-binding-map.ts
+++ b/packages/qwik/src/optimizer/src/qwik-binding-map.ts
@@ -30,5 +30,15 @@ export const QWIK_BINDING_MAP = {
         "platformArchABI": "qwik.win32-x64-msvc.node"
       }
     ]
+  },
+  "linux": {
+    "x64": [
+      {
+        "platform": "linux",
+        "arch": "x64",
+        "abi": "gnu",
+        "platformArchABI": "qwik.linux-x64-gnu.node"
+      }
+    ]
   }
 };
--- a/starters/apps/base/tsconfig.json
+++ b/starters/apps/base/tsconfig.json
@@ -16,11 +16,10 @@
     "isolatedModules": true,
     "outDir": "tmp",
     "noEmit": true,
-    "types": ["node", "vite/client"],
     "paths": {
       "~/*": ["./src/*"]
     }
   },
   "files": ["./.eslintrc.cjs"],
-  "include": ["src", "./*.d.ts", "./*.config.ts"]
+  "include": ["qwik.env.d.ts", "src", "./*.d.ts", "./*.config.ts"]
 }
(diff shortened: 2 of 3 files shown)


packages/qwik/src/optimizer/src/qwik-binding-map.ts
```diff
@@ -30,5 +30,15 @@ export const QWIK_BINDING_MAP = {
...
+  },
+  "linux": {
+    "x64": [
+      {
+        "platform": "linux",
+        "arch": "x64",
+        "abi": "gnu",
+        "platformArchABI": "qwik.linux-x64-gnu.node"
+      }
+    ]
```
octet-stream, line 42: I think these lines appeared after I ran `pnpm build.local` command. It's not related to the PR itself, so I'm not sure if I should keep it there.

wmertens: yes those are good :)

starters/apps/base/tsconfig.json
```diff
@@ -16,11 +16,10 @@
...
     "outDir": "tmp",
     "noEmit": true,
-    "types": ["node", "vite/client"],
     "paths": {
       "~/*": ["./src/*"]
     }
   },
   "files": ["./.eslintrc.cjs"],
-  "include": ["src", "./*.d.ts", "./*.config.ts"]
+  "include": ["qwik.env.d.ts", "src", "./*.d.ts", "./*.config.ts"]
```
wmertens, line 24: isn't this already covered by `*.d.ts`?

wmertens (dismissed):
I like the idea of not pinning `types`, but can you add a comment to that effect in your new qwik.env file?

starters/apps/base/tsconfig.json
octet-stream: Yes, thanks for pointing this out. I already removed it.

PatrickJS (approved):

PR #6208, opened 2024-05-02
fix(v2): component rerender with the same key and different props

Varixo:
This PR fixes component rerender with the same key and different props

### DIFF
--- a/packages/qwik/src/core/v2/client/vnode-diff.ts
+++ b/packages/qwik/src/core/v2/client/vnode-diff.ts
@@ -1046,8 +1046,8 @@ function propsDiffer(src: Record<string, any>, dst: Record<string, any>): boolea
   if (!src || !dst) {
     return true;
   }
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -1046,8 +1046,8 @@ function propsDiffer(src: Record<string, any>, dst: Record<string, any>): boolea
   if (!src || !dst) {
     return true;
   }
-  let srcKeys = removeChildrenKey(Object.keys(src));
```
Varixo, line 1049: in props we don't have children anymore, so I assume this is not needed now. Let me know if I am wrong

wmertens: Right, the optimizer makes sure combined with _jsxC which checks at runtime. 

However, the children prop can still be in the varProps, it should just be ignored

mhevery (approved):

PR #6240, opened 2024-05-06
refactor(aws): handler alias

PatrickJS:
common aws config
- [x] default lambda export "handler"
- [ ] docs about deploying lambda


PatrickJS:
I'll make another pr for the docs

Issue #4908, opened 2023-08-04
[🐞] onSubmit$ attribute of the `Form component` doesn't work when using method `get`


gustavocadev:
### Which component is affected?

Qwik City

### Describe the bug

The onSubmit attribute of the `Form component` doesn't work when using method `get`
![image](https://github.com/BuilderIO/qwik/assets/84424661/b06e63b9-06bf-41f5-ae8c-275deb00c662)

### Reproduction

https://github.com/gustavocadev/qwik-bug-reproduction

### Steps to reproduce

`pnpm install` followed by `pnpm dev`

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
    Memory: 4.86 GB / 12.92 GB
  Binaries:
    Node: 18.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.6.7 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.6.5 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Spartan (44.22621.2070.0), Chromium (115.0.1901.188)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @builder.io/qwik: ^1.2.6 => 1.2.6
    @builder.io/qwik-city: ^1.2.6 => 1.2.6
    undici: 5.22.1 => 5.22.1
    vite: 4.4.0 => 4.4.0
```

PatrickJS:
this does work but only onSubmitCompleted$ event gets fired. I made a PR to fire the event before submit

PR #6241, opened 2024-05-06
feat(Form): multi onSubmit$ handlers

PatrickJS:
allow for an array of events for onSubmit$

also fixes https://github.com/QwikDev/qwik/issues/4908


packages/qwik-city/runtime/src/form-component.tsx
```diff
@@ -53,13 +53,39 @@ export const Form = <O, I>(
   key: string | null
 ) => {
   if (action) {
+    const isArrayApi = Array.isArray(onSubmit$);
+    // if you pass an array you can choose where you want action.submit in it
+    if (isArrayApi) {
+      const actionInArray = Array.isArray(onSubmit$) && onSubmit$.includes(action.submit as any);
```
hamatoyogi, line 59: I find this a bit confusing... there's already `isArrayApi` check, not sure it's needed again for `actionInArray`, or am i missing something?

PatrickJS: oh yeah

hamatoyogi (commented):
Might be worth adding a test as well to make sure it works and also one (currently failing, but could be skipped) for v2

PatrickJS:
I made it additional api and in v2 we can make it slightly breaking api

PatrickJS:
I added a test

PatrickJS:
ok test added and all green

wmertens (approved):
LGTM

Issue #3716, opened 2023-04-07
[🐞]I would expect a more prominent error.


mhevery:
### Which component is affected?

Qwik Runtime

### Describe the bug

https://stackblitz.com/edit/qwik-starter-tg16pg?file=src%2Froutes%2Findex.tsx

Click on counter
Notice that the error is only in the console.

Would expect a more prominent Error in the UI in dev mode???

### Reproduction

https://stackblitz.com/edit/qwik-starter-tg16pg?file=src%2Froutes%2Findex.tsx

### System Info

```shell
na
```

PR #6254, opened 2024-05-06
feat(qwikloader): emit error

PatrickJS:
fixes https://github.com/QwikDev/qwik/issues/3716


wmertens:
Is qerror with the data specific enough to know that it was a symbol error? 

I think we should also have a default handler in qwik city that hard reloads the page when the error is 404

PatrickJS:
yeah, I'm thinking we can use this and add it to the import for a hard reload (to fix the other problem).


this was meant for detecting an error

PatrickJS:
@wmertens ok I added an importError check we can later add a listener to check for it and reload for prod

wmertens (approved):
LGTM

PR #6265, opened 2024-05-07
revert: starter vite config server bundle optimizations

thejackshelton:
# Overview

The commented SSR configuration in vite.config.ts intended to enhance server code bundling for production builds by including all development dependencies and excluding specified production dependencies to optimize the server bundling.

This has led to consumers adding packages in the wrong location, and as a result these packages do not work in production.

This PR comments out that code, and labels it as an advanced config for when you understand which dependencies should be consumed either as a dependency, or dev dependency.

In the future, I believe we should uncomment this code by default, but only when better error handling is present, as it was difficult to debug. For example, a CLI message that tells us whether the wrong dependency was bundled or not.

For more context see:
https://discord.com/channels/990511757091033108/1040763063533588562/1237190861985550376

from the Qwikifiers discord

Original change:
https://github.com/QwikDev/qwik/pull/5961

cc @wmertens 

# Checklist:

### DIFF
--- a/starters/apps/base/vite.config.ts
+++ b/starters/apps/base/vite.config.ts
@@ -28,20 +28,24 @@ export default defineConfig(({ command, mode }): UserConfig => {
       // For example ['better-sqlite3'] if you use that in server functions.
       exclude: [],
     },
-    // This tells Vite how to bundle the server code.
-    ssr:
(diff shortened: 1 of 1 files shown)


wmertens (approved):
LGTM

PR #6281, opened 2024-05-08
fix(v2 optimizer): improved signal forwarding

wmertens:
I think this approach makes more sense, use fnSignal if there's anything in there that could be a signal.


wmertens:
Note that this also adds `_wrapProps()` to e.g. inline components, which will only work if creating a signal is safe in there.

Issue #6288, opened 2024-05-09
[✨] Throw server$ errors with a specific status code


DustinJSilk:
### Is your feature request related to a problem?

Currently all `server$` errors return an HTTP 500 error which can't be configured per request.

### Describe the solution you'd like

I'd like to be able to throw a custom error which includes a status code.

Option A:

```tsx
import { ServerError } from `@builder.io/qwik-city`

const myFunc = server$(() => {
  throw new ServerError(404, { /** Some response data */ })
})
```

Option B:

```tsx
const myFunc = server$(() => {
  throw this.error(404, { /** Some response data */ })
})
```

### Describe alternatives you've considered

Currently the choices are:
- Return 200 with a union type for both a successful and unsuccessful response
- Throw an error that is always a 500

wmertens:
I like it and prefer option A

PatrickJS:
ok I have it here. but you have to make sure you know how to detect the error from the response object json
https://github.com/QwikDev/qwik/pull/6290

```js
const myFunc = server$(() => {
  if (throwing) {
    throw new ServerError(404, { error: true, data: {} })
  }
  return { error: false, data: {}}
})


const data = await myFunc()
if (data.error) {
  console.log(data.error)
}
```
this pattern should be way better
```js
const myFunc = server$(() => {
  if (throwing) {
    const data = {};
    throw new ServerError(404, [data])
  }
  const data = {};
  return [null, data]
})


const [err, data] = await myFunc()
if (err) {
  console.log(err, data)
}
```

DustinJSilk:
Amazing, thank you @PatrickJS !

PatrickJS:
you have to import from `import { ServerError } from "@builder.io/qwik-city/middleware/request-handler";` in the next release

Issue #5656, opened 2023-12-31
[✨] Allow using GET with server$


manvalls:
### Is your feature request related to a problem?

POST requests are not easily cacheable

### Describe the solution you'd like

A way to tell server$ to use GET instead of POST

### Describe alternatives you've considered

Considered just using an endpoint, but I love the streaming functionality

wmertens:
Typically I'd cache on the client side in this case.

Doing a GET implies putting the variables in the URL and listening for GET in the handler as well. That latter could be a little bit annoying but probably ok.

I suppose we could do it, what do you think @mhevery ?

PatrickJS:
server$ needs a config for other features like changing origin, adding headers, query params etc

wmertens:
@PatrickJS I propose a default config via the qwik city provider and then maybe the first argument. Right now that can optionally be an abort signal and we could also make it be an options object

PR #6290, opened 2024-05-09
feat(server$): config argument, optional GET, ServerError

PatrickJS:
# Overview

fixes https://github.com/QwikDev/qwik/issues/5656
fixes https://github.com/QwikDev/qwik/issues/6288


PatrickJS:
in another pr I can add 3 levels of config. I also want to do this for better AbortSignal

DustinJSilk:
This is awesome! 🤩

Would it make sense to have ServerError take a generic argument so any serializable data can be returned?

```tsx
export class ServerError<T = Record<any, any>> extends Error {
  constructor(
    public status: number,
    public data: T
  ) {
    super();
  }
}
```

starters/apps/qwikcity-test/src/routes/(common)/server-func/server-configs/index.tsx
wmertens, line 1: You also need a test for ServerError and the data passed

wmertens (changes requested):
Looking forward to this!

wmertens:
Misko is ok with this API extension

starters/apps/qwikcity-test/src/routes/(common)/server-func/server-configs/index.tsx
PatrickJS: done

PatrickJS:
```ts
type ResponseTuple = [null | string, string];

const serverFunctionA = server$(async function a(): Promise<ResponseTuple> {
  throw new ServerError<[string]>(401, ["my error"]);
});

const serverFunctionB = server$(async function b(): Promise<ResponseTuple> {
  return [null, this.method || ""];
});

export const MultipleServerFunctionsInvokedInTask = component$(() => {
  const methodA = useSignal("");
  const methodB = useSignal("");

  useVisibleTask$(async () => {
    const [error /*, data */] = await serverFunctionA();
    if (error) {
      methodA.value = error;
    }
    await delay(1);
    //     err, method
    const [, method] = await serverFunctionB();
    methodB.value = method;
  });

  return (
    <div id="server-error">
      {methodA.value}
      {methodB.value}
    </div>
  );
});
```

DustinJSilk:
I'm curious why the error isn't thrown on the client so that it can be caught, and instead we return the error like the method was successful?

I'm wondering if this was the right decision for the given reasons:
- Does this go against usual JS conventions of catching errors rather than returning errors?
- Throwing an error and then not catching it later feels wrong - throwing should mean we catch
- We now have to handle both transient errors and ServerErrors - if the request fails we need to catch it, and if a ServerError is thrown we need to handle the value type.
- Changing code to throw ServerError instead of a native error means we have to update our code elsewhere to handle the new error response type, making it more of a breaking change

There is also actually a ErrorResponse class from a routeLoader$, which it might have made sense to reuse so that any API middleware, whether used in a server$ or routeLoader$, can throw the same class and result in the same status code/response handling.

I realise this might be late, but with the upcoming v2 changes, would it make sense to revisit this, or is there merit in the way it is currently handled?

PR #6298, opened 2024-05-11
test(optimizer/core): import attributes

PatrickJS:
add import attribute to rust

### DIFF
--- a/packages/qwik/src/optimizer/core/src/test.rs
+++ b/packages/qwik/src/optimizer/core/src/test.rs
@@ -1941,6 +1941,24 @@ export const Greeter = component$(() => {
 	});
 }
 
+[test]
+fn example_import_attributes() {
+	test_input!(TestInput {
+		code: r#"
+import { component$, $ } from '@builder.io/qwik';
+import json from "./foo.json" with { type: "json" };
+
+export const Greeter = component$(() => {
+    return json;
+});
+"#
+		.to_string(),
+		transpile_ts: true,
+		transpile_jsx: true,
+		..TestInput::default()
+	});
+}
+
 #[cfg(target_os = "windows")]
 #[test]
 fn issue_188() {


wmertens:
You need to run pnpm test.rust.update

PatrickJS:
ok just did but it looks like it fails when it should work (since import attributes changes to with keyword)

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_import_attributes.snap
```diff
@@ -0,0 +1,37 @@
...
+
+export const Greeter = component$(() => {
+    return json;
+});
+
+== DIAGNOSTICS ==
+
+[
+  {
+    "category": "error",
```
PatrickJS, line 20: this shouldn't error

wmertens: right, I'm trying to upgrade swc which supports `with` in the update-rust branch.

PatrickJS: awesome

PatrickJS:
waiting on this pr i think https://github.com/QwikDev/qwik/pull/6005

wmertens:
This is now handled via #6005. Not sure if we need a test for it, the entire functionality is provided by SWC

wmertens:
@PatrickJS do you want to rebase and `pnpm test.rust.update`? Should just work and then we can be sure we don't regress

gioboa:
is this PR still valid?

gioboa (commented):
is this PR still valid?

gioboa (commented):
@PatrickJS thanks for opening this PR.
I'm closing this for now because it's stale, feel free to create a new PR with your improvements. :muscle:

PR #6301, opened 2024-05-12
fix(qwik-city): enforce / on base paths

wmertens:
@mhevery not sure if this should go in only v2 because it might be slightly breaking.

### DIFF
--- a/packages/qwik/src/server/utils.ts
+++ b/packages/qwik/src/server/utils.ts
@@ -27,7 +27,7 @@ export function getBuildBase(opts: RenderToStringOptions) {
     }
     return base;
   }
-  return '/build/';
+  return `${import.meta.env.BASE_URL}build/`;
 }
 
 /** @public */
--- a/packages/qwik-city/buildtime/context.ts
+++ b/packages/qwik-city/buildtime/context.ts
@@ -45,6 +45,15 @@ function normalizeOptions(
   viteBasePath: string,
   userOpts: PluginOptions | undefined
 ) {
+  if (!(viteBasePath.startsWith('/') && viteBasePath.endsWith('/'))) {
+    // TODO v2: make this an error
+    console.error(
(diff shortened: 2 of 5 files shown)


packages/qwik/src/server/utils.ts
```diff
@@ -27,7 +27,7 @@ export function getBuildBase(opts: RenderToStringOptions) {
     }
     return base;
   }
-  return '/build/';
+  return `${import.meta.env.BASE_URL}build/`;
```
PatrickJS, line 30:
```suggestion
  return `${import.meta.env.BASE_URL || '/'}build/`;
```

wmertens: Isn't that always defined?

PatrickJS: nope

wmertens: So this [is determined by](https://vitejs.dev/guide/env-and-mode#:~:text=This%20is%20determined%20by%20the%20base%20config%20option.) [the base config option](https://vitejs.dev/config/shared-options.html#base) which defaults to `/`, and we check that it ends on a `/`, so I think it's always defined.

PatrickJS: can we still do warning for now and throw errors later

PatrickJS:
I think we might need q:origin because I want the base path to be the same but different origin. and if I want to have a certain version of base then I'll change the base path to that version in ssr entry

PatrickJS (approved):

gioboa (approved):

wmertens:
@PatrickJS could you elaborate on origin? I don't understand the use case

PatrickJS:
so the base path might be the same but you may change the origin to an assets domain. in one of my apps we're using base to change the origin to an assets cdn while the app is hosted on another origin domain

wmertens:
@PatrickJS I think you can achieve it without q:origin. Set vite base to the URL prefix of the app and set the renderToStream base to the full url of the origin+prefix+build/

Issue #6247, opened 2024-05-06
[🐞] SSG ignoring the index page when trailingSlash is false


franck-co:
### Which component is affected?

Qwik City (routing)

### Describe the bug

Given the following routes directory
```
routes
└─ (group)
   ├─ index.tsx
   └─ articles
      └─ index.tsx
```
SSG is generating `articles.html`  successfully but the top level `index.html` is missing when `trailingSlash` option is set to `false`.

I suspect the line below :
https://github.com/QwikDev/qwik/blob/25c0c98dda929392819263b4db33b9c88ecf9e90/packages/qwik-city/static/main-thread.ts#L143

When `trailingSlash` is enabled, `pathname` is `'/'` and it's working
When `trailingSlash` is disabled `pathname` is `''` and the condition evaluates to `false`, thus never adding the page to SSG queue.

### Reproduction

https://stackblitz.com/edit/qwik-starter-zmev6t

### Steps to reproduce

1. Run `npm install` then `npm run build`
2. Observe the `dist` folder not having index.html
3. Set `trailingSlash` to `true` in `vite.config.js`
4. Run `npm run build` again
5. `index.html` is now present in the `dist` folder 

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (2) x64 Intel Core Processor (Haswell, no TSX)
    Memory: 22.34 GB / 28.68 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    pnpm: 9.0.6 - ~/.local/share/pnpm/pnpm
  Browsers:
    Chromium: 124.0.6367.118
  npmPackages:
    @builder.io/qwik: 1.5.3 => 1.5.3 
    @builder.io/qwik-city: 1.5.3 => 1.5.3 
    @builder.io/qwik-labs: github:QwikDev/qwik-labs-build => 0.0.1 
    @builder.io/sdk-qwik: 0.14.21 => 0.14.21 
    typescript: 5.4.5 => 5.4.5 
    undici: ^6.15.0 => 6.15.0 
    vite: 5.2.11 => 5.2.11
```

[...]

PR #6304, opened 2024-05-13
fix: SSG ignoring the index page when trailingSlash is false

JerryWu1234:
# Overview

fix:#6247

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/utils/fs.ts
+++ b/packages/qwik-city/utils/fs.ts
@@ -33,7 +33,7 @@ export function getPathnameFromDirPath(opts: NormalizedPluginOptions, dirPath: s
     // remove grouped layout segments
     .filter((segment) => !isGroupedLayoutName(segment))
     .join('/');
-
+  if (pathname === '') return '/';
   return pathname;
 }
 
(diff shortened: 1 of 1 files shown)


gioboa (approved):
Thanks for your help.
Did you test this change locally?

JerryWu1234:
> Thanks for your help.
> 
> Did you test this change locally?

Of course I will send a picture that includes my testing

packages/qwik-city/utils/fs.ts
```diff
@@ -33,7 +33,9 @@ export function getPathnameFromDirPath(opts: NormalizedPluginOptions, dirPath: s
     // remove grouped layout segments
     .filter((segment) => !isGroupedLayoutName(segment))
     .join('/');
-
+  if (pathname === '') {
```
gioboa, line 36: we are always applying this change and not only when `opts.trailingSlash` is false but I think this is ok

JerryWu1234: I know. But basepathname defaults to a “/” in the project

PR #6305, opened 2024-05-13
refactor(qwikloader): add types and no handler qerror

PatrickJS:
# Overview
fix qwikloader types

### DIFF
--- a/packages/qwik/src/qwikloader.ts
+++ b/packages/qwik/src/qwikloader.ts
@@ -114,7 +114,12 @@ export const qwikLoader = (
             emitEvent('qerror', { importError: true, error, symbol, uri });
           }
         }
-        if (!handler) return;
+        if (!handler) {
+          const error = new Error('Handler not found for ' + url.href);
+          emitEvent('qerror', { error, href: url.href });
+          // break out of the loop if handler is not found
+          break;
+        }
         const previousCtx = doc[Q_CONTEXT];
         if (element[isConnected]) {
           const eventData = { symbol, element, reqTime };
--- a/packages/qwik/src/qwikloader.ts
+++ b/packages/qwik/src/qwikloader.ts
@@ -11,9 +11,17 @@ import type { QContext } from './core/state/context';
  * @param doc - Document to use for setting up global listeners, and to determine all the browser
  *   supported events.
  */
-export const qwikLoader = (doc: Document, hasInitialized?: number) => {
+export const qwikLoader = (
+  doc: Document & { __q_context__?: [Element, Event, URL] | 0 },
+  hasInitialized?: number
+) => {
   const Q_CONTEXT = '__q_context__';
-  const win = window as any;
+  type qWindow = Window & {
+    qwikevents: {
+      push: (...e: string[]) => void;
+    };
+  };
+  const win = window as unknown as qWindow;
   const events = new Set();
 
   // Some shortenings for minification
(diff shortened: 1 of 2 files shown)


packages/qwik/src/qwikloader.ts
```diff
@@ -101,21 +114,27 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
             emitEvent('qerror', { importError: true, error, symbol, uri });
           }
         }
-        const previousCtx = (doc as any)[Q_CONTEXT];
+        if (!handler) {
```
PatrickJS, line 117: this wasn't handled before. this will only happen during skew version issue or connection problems

packages/qwik/src/qwikloader.ts
```diff
@@ -36,13 +44,13 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
     );
   };
 
-  const resolveContainer = (containerEl: Element) => {
-    if ((containerEl as QContainerElement)[Q_JSON] === undefined) {
+  const resolveContainer = (containerEl: Element & QContainerElement) => {
```
wmertens, line 47: Can't we make QContainerElement be also Element?

packages/qwik/src/qwikloader.ts
```diff
@@ -57,12 +65,17 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
...
+    onPrefix: string,
+    ev: Event,
+    eventName = ev.type
+  ) => {
     const attrName = 'on' + onPrefix + ':' + eventName;
     if (element.hasAttribute('preventdefault:' + eventName)) {
       ev.preventDefault();
     }
-    const ctx = (element as any)['_qc_'] as QContext | undefined;
+    const ctx = element['_qc_'] as QContext | undefined;
```
wmertens, line 78: I think the cast is no longer needed?

packages/qwik/src/qwikloader.ts
wmertens: sync qrls can never be wrong because they're part of the container and import error already emits an error, so you're just making the loader bigger.

You could move line 101/114 into here and break

packages/qwik/src/qwikloader.ts
```diff
@@ -101,21 +114,27 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
...
         if (element[isConnected]) {
           const eventData = { symbol, element, reqTime };
           try {
-            (doc as any)[Q_CONTEXT] = [element, ev, url];
+            doc[Q_CONTEXT] = [element, ev, url];
             isSync || emitEvent<QwikSymbolEvent>('qsymbol', eventData);
             const results = handler(ev, element);
             // only await if there is a promise returned
-            if (isPromise(results)) {
+            if (isPromise(results as Promise<unknown>)) {
```
wmertens, line 131: how come it didn't need a cast before?

wmertens (dismissed):

packages/qwik/src/qwikloader.ts
PatrickJS: yup we should

packages/qwik/src/qwikloader.ts
PatrickJS: if (!handler) fixes this so we don't need to cast it anymore

packages/qwik/src/qwikloader.ts
```diff
@@ -101,11 +114,15 @@ export const qwikLoader = (doc: Document, hasInitialized?: number) => {
             emitEvent('qerror', { importError: true, error, symbol, uri });
           }
         }
-        const previousCtx = (doc as any)[Q_CONTEXT];
+        if (!handler) {
+          // break out of the loop if handler is not found
```
wmertens, line 118: so if the sync qrl is missing somehow (however unlikely), it will silently break. Wouldn't it be better to move the qerror event into this block?
Or else, just put the break after the qerror if we expect to always have sync

PatrickJS: oh thats what you meant. yeah I can check if error is defined otherwise define it

PatrickJS: well now it's slightly more code but it's more correct for us to detect the issue

PatrickJS:
merging since we aren't cutting a release yet (waiting on your updates). but I can update if we need to change any more

PR #6310, opened 2024-05-14
chore(qwik-city/middleware): add TextEncoderStream polyfill and tests

PatrickJS:
add _TextEncoderStream_polyfill and tests


PatrickJS:
@wmertens when you get time can you review

octet-stream (commented):
Saw this PR and decided to add a few bits after looking closer myself to `TextEncoderStream`. This implementation is shorter, and I believe it's fine to keep it as it is, but I have a few thoughts from what I understand:

1. The `TextEncoderStream` class is not an extension of the `TransformStream` (you can see that from the documentation on MDN [here](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoderStream)), but it has same generic interface with `readable` and `writable` properties (see the spec [here](https://encoding.spec.whatwg.org/#interface-textencoderstream)). I'm not sure if we are trying to achieve the most compatibility, but you can refer to [Deno's](https://github.com/denoland/deno/blob/f8956eb763159f325d5cb0d4c4281d2e640e2e71/ext/web/08_text_encoding.js#L394-L480) and [Node.js'](https://github.com/nodejs/node/blob/3f3226c8e363a5f06c1e6a37abd59b6b8c1923f1/lib/internal/webstreams/encoding.js#L38-L119) implementations if that's the case.
2. There are properties that I don't see on this class either. I left more comments on the specifics :)

PatrickJS:
yeah the props are there only to match the previous polyfill. so everything is just-in-case

octet-stream:
Oh, sure. But anyway here's TypeScript flavored version of TextEncoderPolyfill based on Node.js' implementation, just in case:

```ts
/**
 * TextEncoderStream polyfill based on Node.js' implementation https://github.com/nodejs/node/blob/3f3226c8e363a5f06c1e6a37abd59b6b8c1923f1/lib/internal/webstreams/encoding.js#L38-L119 (MIT License)
 */
export class TextEncoderStream {
  #pendingHighSurrogate: string | null = null

  #handle = new TextEncoder()

  #transform = new TransformStream<string, Uint8Array>({
    transform: (chunk, controller) => {
      // https://encoding.spec.whatwg.org/#encode-and-enqueue-a-chunk
      chunk = String(chunk)

      let finalChunk = ""
      for (const item of chunk) {
        const codeUnit = item.charCodeAt(0)
        if (this.#pendingHighSurrogate !== null) {
          const highSurrogate = this.#pendingHighSurrogate

          this.#pendingHighSurrogate = null
          if (codeUnit >= 0xDC00 && codeUnit <= 0xDFFF) {
            finalChunk += highSurrogate + item
            continue
          }

          finalChunk += "\uFFFD"
        }

        if (codeUnit >= 0xD800 && codeUnit <= 0xDBFF) {
          this.#pendingHighSurrogate = item
          continue
        }

        if (codeUnit >= 0xDC00 && codeUnit <= 0xDFFF) {
          finalChunk += "\uFFFD"
          continue
        }

        finalChunk += item
      }

      if (finalChunk) {
        controller.enqueue(this.#handle.encode(finalChunk))
      }
    },

    flush: controller => {
      // https://encoding.spec.whatwg.org/#encode-and-flush
      if (this.#pendingHighSurrogate !== null) {
        controller.enqueue(new Uint8Array([0xEF, 0xBF, 0xBD]))
      }
    }
  })

  get encoding() {
    return this.#handle.encoding
  }

  get readable() {
    return this.#transform.readable
  }

  get writable() {
    return this.#transform.writable
  }

  get [Symbol.toStringTag]() {
    return "TextEncoderStream"
  }
}
```

PatrickJS:
looks good lets try it

PatrickJS:
@octet-stream thanks for finding this better polyfill

thejackshelton (approved):
Tests pass locally and make sense on my end.

My understanding, is that this polyfill is removing the previous one, and is a more robust version that should support the 5% or so where this is not supported.

Is that the case? Anything I should look in-depth into? 

Otherwise LGTM! 🙌

PR #6212, opened 2024-05-03
docs: disable ChatGPT and Supabase

mhevery:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests / types / typos

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

- [ ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] Added new tests to cover the fix / functionality

PR #6314, opened 2024-05-15
refactor: remove Qwik GPT AI from the docs search

thejackshelton:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

We had a meeting regarding docs, and our consensus was that the **Qwik GPT** feature is not helpful in providing a good search experience to those looking to figure out *how or why* they should build with Qwik.

This PR tries to remove any lingering Qwik GPT code. While I'm sure some of us might get the feeling that we should comment it out and save it for later, I believe the reality is that would remain code in the codebase. We can always look in the file history and get it back if we really need it.

TODO: please review the `packages/docs/src/routes/query/[id]/index.tsx` file.

- What is this file for?
- What do we know is safe to remove?
- What do you think we might not need?

Happy to continue this PR once those questions are answered 👍 . Would greatly appreciate the time taken to review this.

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


PatrickJS:
is this the same?
https://github.com/QwikDev/qwik/pull/6212 because we can reopen the old one

thejackshelton:
> is this the same? #6212 because we can reopen the old one

Yep, it's a dupicate. Closing this one.

PR #6317, opened 2024-05-15
fix(v2): slot tests fixes

Varixo:
This PR fixes slot e2e tests


mhevery (approved):

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -621,9 +623,66 @@ export const vnode_diff = (
...
+    );
+
     return needsQDispatchEventPatch;
   }
 
+  function createElementWithNamespace(tag: string): Element {
+    const domParentVNode = vnode_getDomParentVNode(vParent);
+
+    const parentIsDefaultNamespace = domParentVNode
+      ? vnode_getElementName(domParentVNode) && vnode_isDefaultNamespace(domParentVNode)
```
Varixo, line 641: `vnode_getElementName(domParentVNode)` this will read a "real" element name if we don't have it, but I don't have idea how to do it another way

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -621,9 +623,66 @@ export const vnode_diff = (
...
+      : true;
+    const parentIsForeignObject = !parentIsDefaultNamespace
+      ? isForeignObjectElement(vnode_getElementName(domParentVNode!))
```
Varixo, line 644: here is the same situation

packages/qwik/src/core/v2/client/vnode.ts
```diff
@@ -394,6 +396,43 @@ export const vnode_ensureElementInflated = (vnode: VNode) => {
   }
 };
 
+export const vnode_getDOMChildVNodes = (
```
Varixo, line 399: this is copy pasted code `vnode_getDOMChildNodes `, but returns VNodes instead of elements

packages/qwik/src/core/v2/client/vnode.ts
```diff
@@ -1004,11 +1072,180 @@ export const vnode_insertBefore = (
...
 
+function vnode_getDomChildrenToInsert(
+  journal: VNodeJournal,
+  document: Document,
+  domParentVNode: ElementVNode,
+  newChild: VNode
+) {
+  const parentNodeIsDefaultNamespace = vnode_isDefaultNamespace(domParentVNode);
+  const parentIsForeignObject = !parentNodeIsDefaultNamespace
+    ? isForeignObjectElement(vnode_getElementName(domParentVNode))
```
Varixo, line 1083: here is the same problem with element name

packages/qwik/src/core/v2/client/vnode.ts
mhevery: Can we somehow merge the code so that it is not duplicate? I don't want to be fixing the same code twice.

Maybe have an extra argument which you can use if the VNode should be unwraped or not.

packages/qwik/src/core/v2/client/vnode.ts
```diff
@@ -686,6 +725,28 @@ export const vnode_getVNodeForChildNode = (
   return child as ElementVNode;
 };
 
+// this is trying to find the vnode for the element using Breadth-First Search
+export const vnode_getVNodeForElement = (
```
mhevery, line 729: Why do you need this?

packages/qwik/src/core/v2/client/vnode.ts
```diff
@@ -931,6 +994,7 @@ export const mapArray_get = <T>(
 
 export const vnode_insertBefore = (
   journal: VNodeJournal,
+  document: Document,
```
mhevery, line 997: You should not have to pass `document` around. Please revert this. If you need to get a hold of element you can always say `node.ownerDocument` https://developer.mozilla.org/en-US/docs/Web/API/Node/ownerDocument Because that is a rare case I don't mind reading the DOM for that.

packages/qwik/src/core/v2/client/vnode.ts
mhevery: This seems like a lot of code. Why is all of this needed? Also, can we move all of the code that morphs from one namespace into a different namespace into a separate file?

mhevery (changes requested):

packages/qwik/src/core/v2/client/vnode.ts
Varixo: I removed it. It was for finding the q:template element

mhevery (approved):

PR #6318, opened 2024-05-15
fix(image-size-runtime.html): exclude SVG images from Perf: properly …

williamsdyyz:
…size image warnings

SVG is a vector format. Attributes like width, height & viewbox are not relevant to the size in bytes of the image, so the Perf: properly size image warning is incorrect.

# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/image-size-runtime.html
+++ b/packages/qwik/src/optimizer/src/plugins/image-size-runtime.html
@@ -321,7 +321,7 @@
         }
         const realArea = info.width && info.height;
         const threshholdArea = realArea * 0.5;
-        const tooBig = browserArea < threshholdArea;
+        const tooBig = browserArea < threshholdArea && info.type !== 'svg';
         skip = false;
         if (layoutInvalidation || tooBig) {
           if (!overlay) {


thejackshelton:
Hey @williamsdyyz! Is there an example SVG you can give so that I can confirm whether the warning went away after this PR?

I tried reproducing, but was unsuccessful.

williamsdyyz:
This is one of the files I used

flagen-US

PatrickJS (approved):

thejackshelton:
@williamsdyyz sorry to bug you, but how do I reproduce the `perf: size too big` warning?

I have tried using it as an svg and add the svg as an image with a large, also trying to add large `width` and `height` attributes. Even adding massive images from unsplash, no luck there. 

And this is with 1.5.4, not with your current PR version. I can't seem to get the warning.

Maybe you have a reproduction I could view?

williamsdyyz:
@thejackshelton no problem.  I created a new Qwik project with `pnpm creater qwik@latest` and... couldn't reproduce it.

Did some digging. I had added Tailwind to my original project. Did more digging. Turns out it's one of Tailwind's default styles.

Try this code. The img with height:auto has the problem, the other doesn't.

```
import { component$ } from "@builder.io/qwik";
import flag from "./flagen-US.svg";

  return (
    <>
      <img src={flag} width={24} height={24} style={{ height: "auto" }} />
      <br />
      <br />
      <img src={flag} width={24} height={24} />
    </>
  );
});

```

thejackshelton:
Appreciate the patience here! Tested and everything looks good 👍 . 

Merging.

Issue #5473, opened 2023-11-22
[🐞] Library mode + preserveModules -> error in consumer project: Cannot resolve symbol s_c7PiYJ1s0GU  — QWIK ERROR Error: Code(31)


maiieul:
### Which component is affected?

Qwik Optimizer (rust)

### Describe the bug

Related to https://github.com/BuilderIO/qwik/issues/5097, but creating another issue because the two will probably be solved independently.

In an attempt to solve the big bundle and tree-shaking issue in qwik-ui, which is also an issue for any other qwik library using the library starter, @shairez and @wmertens found that setting `build.rollupOptions.output.preserveModules` to `true` can work around the issue. But the problem is that this introduces a namespace issue in preview and production with "error code 31" and the error message "Cannot resolve symbol s_c7PiYJ1s0GU".

So using 

```tsx
    rollupOptions: {
      output: {
        preserveModules: true,
        preserveModulesRoot: 'packages/kit-headless/src',
      },
    },
```

if the library has a component$ named `AccordionRoot`, and then I define `export const AccordionRoot = component$(...)` in the local app, the `AccordionRoot` component from the library will be overridden by the `AccordionRoot` in the local app, leading to "Cannot resolve symbol s_c7PiYJ1s0GU... Error code (31)" types of errors when symbols cannot reach the qwik-ui's AccordionRoot code as it isn't there.

We believe the code isn't there because it is the local project's AccordionRoot that is in the build output instead.

### Reproduction

https://github.com/maiieul/qwik-libraryMode-and-preserveModules-namespace-issue-repro

### To see the error message:

1) Clone the repo
2) `npm i` (npm will make the debugging step easier)
3) `npm run preview`

Go into src/components/accordion-root/index.tsx and change the const to `AccordionRoot2` and the error will go away.

### System Info

```shell
System:
    OS: Linux 6.5 Fedora Linux 38 (Workstation Edition)
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i5-1340P
    Memory: 17.00 GB / 31.04 GB
    Container: Yes
    Shell: 3.6.1 - /usr/bin/fish
  Binaries:
    Node: 20.9.0 - /usr/local/bin/node
    npm: 10.1.0 - /usr/local/bin/npm
    pnpm: 8.10.5 - ~/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 119.0.6045.159
  npmPackages:
[...]

PR #6326, opened 2024-05-16
fix(optimizer): transform code incrementally

wmertens:
This builds on top of #6539 (which was first part of this).

Qwik should not split files before other build plugins had a chance to change the code.

The problem is that Qwik outputs "entry files" that contain multiple QRL segments, possibly from several different files, and that Rollup doesn't allow changing a file once it was loaded.

So this PR will need to enforce ordering. Not done yet.


shairez:
This fixes this issue right?
#5473

scripts/submodule-core.ts
```diff
@@ -163,9 +163,10 @@ async function submoduleCoreProd(config: BuildConfig) {
 
   console.log('🐭 core.min.mjs:', await fileSize(join(config.distQwikPkgDir, 'core.min.mjs')));
 
-  // always set the cjs version (probably imported server-side) to dev mode
   let esmCode = await readFile(join(config.distQwikPkgDir, 'core.mjs'), 'utf-8');
   let cjsCode = await readFile(join(config.distQwikPkgDir, 'core.cjs'), 'utf-8');
+  // fixup the Vite base url
+  cjsCode = cjsCode.replaceAll('undefined.BASE_URL', 'globalThis.BASE_URL||"/"');
```
PatrickJS, line 169: 😅

packages/qwik/src/qwikloader.ts
```diff
@@ -102,10 +102,12 @@ export const qwikLoader = (
       for (const qrl of attrValue.split('\n')) {
         const url = new URL(qrl, base);
         const href = url.href;
-        const symbol = url.hash[replace](/^#?([^?[|]*).*$/, '$1') || 'default';
+        const match = /^#?(([^#]+)#)?([^?[|]*).*$/.exec(url.hash);
```
PatrickJS, line 105: can we add regexp online editor link. for example https://regexr.com/ lets you also add regexp tests for example https://regexr.com/816es (don't use mine we should make a qwik account with all them saved

PatrickJS:
this pr  https://github.com/QwikDev/qwik/pull/6017 will effect this one

wmertens:
@PatrickJS actually I think it supercedes it - I think the problem in dev mode is chunk sources from unknown parents and this solves that. 

Although a quick fix would be to add the prefix in the existing vite-server.ts

wmertens:
closing in favor of #6670

PR #6327, opened 2024-05-16
fix(qwik-city): Removed artificial recursive limiter from recursive path matching

gnemanja:
# What is it?

- [*] Feature / enhancement
- [ ] Bug

# Description

For recursive matching of catchall routes, there was an artificial recursion watchdog limiter (set to 5). So, a route like
/[...catchall]/a/b/c/d/e wouldn't work if catchall didn't have a value (for example: /a/b/c/d/e would fail).
To fix this, we just check if there are any chances for the route to still match instead of stopping at the X amount of time.

# Checklist:

- [*] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [*] I have performed a self-review of my own code
- [*] Added new tests to cover the fix / functionality

### DIFF
--- a/packages/qwik-city/runtime/src/route-matcher.ts
+++ b/packages/qwik-city/runtime/src/route-matcher.ts
@@ -144,8 +144,7 @@ function recursiveScan(
   }
   let pathIdx = pathLength;
   const sep = suffix + '/';
-  let depthWatchdog = 5;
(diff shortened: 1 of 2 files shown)


gioboa (approved):
Thanks @gnemanja 
It looks good to me

PatrickJS (approved):

Issue #6300, opened 2024-05-12
[🐞] Route is broken with deep nesting


mhuretski:
### Which component is affected?

Qwik City (routing)

### Describe the bug

Hi,

When using [...lang] catchall route at the root of the routing, deeply nested routes are broken if lang is not provided.

```
src/routes/[...lang]/
    │
    ├── (app)/
    │   ├── index.tsx
    │   └── layout.tsx
    │
    ├── some/
    │   └── deep/
    │       └── nested/
    │           ├── [id]/
    │           │   └── index.tsx (broken without "lang")
    │           └── index.tsx
    │
    └── working/
        └── nesting/
            └── [id]/
                └── index.tsx
```

/some/deep/nested/some_id - broken
/some_lang/some/deep/nested/some_id - OK
/some/deep/nested - OK
/some_lang/some/deep/nested - OK
/working/nesting/some_id - OK
/some_lang/working/nesting/some_id - OK

Catchall route doesn't seem to be a great fit as an optional route, perhaps adding something like [?lang] could simplify routing logic with deep nesting to avoid this issue?

### Reproduction

https://stackblitz.com/edit/qwik-starter-q65fv9?file=src%2Froutes%2F%5B...lang%5D%2F(app)%2Findex.tsx

### System Info

```shell
"@builder.io/qwik": "^1.5.4",
"@builder.io/qwik-city": "^1.5.4",
```

JerryWu1234:
may I pick this up ?

gioboa:
> may I pick this up ?

Go for it :+1: Thanks

JerryWu1234:
maybe it can directly close this Issue?

PatrickJS:
oh yeah this was fixed in 1.5.5

PR #6330, opened 2024-05-17
fix: a bug

JerryWu1234:
# Overview
#6300 

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik-city/runtime/src/route-matcher.ts
+++ b/packages/qwik-city/runtime/src/route-matcher.ts
@@ -144,7 +144,7 @@ function recursiveScan(
   }
   let pathIdx = pathLength;
   const sep = suffix + '/';
-  let depthWatchdog = 5;
+  let depthWatchdog = 8;
   while (pathIdx >= pathStart && depthWatchdog--) {
     const match = matchRoutePart(route, routeStart, routeLength, path, pathIdx, pathLength);
     if (match) {


packages/qwik-city/runtime/src/route-matcher.ts
```diff
@@ -144,7 +144,7 @@ function recursiveScan(
   }
   let pathIdx = pathLength;
   const sep = suffix + '/';
-  let depthWatchdog = 5;
+  let depthWatchdog = 8;
```
wmertens, line 147: I think you can set this to 50 and also add a console.error before line 160 when it gets to 0

JerryWu1234: Do you have any idea about what the content of the error throws?

PatrickJS: we merged #6327 which remove the limit and adds a test. do we want a limit?

JerryWu1234: So do I close this pr?

PatrickJS: I wanted to get his pr in since he was first-time contributor 😁 I think we should have a limit of 50 or some way to set a limit. I don't think any app will even reach 50

JerryWu1234: @PatrickJS  still needs this limit？So do i continue this PR ?

PatrickJS: @JerryWu1234 I'll close the PR I know you're working on a few other things 🙏 thanks so much for helping us

PR #6339, opened 2024-05-20
feat(vercel): no cache <any>service-worker.js files

PatrickJS:
avoid caching service-worker


gioboa (approved):
Nice catch @PatrickJS

PR #6344, opened 2024-05-20
fix(bun): Support application builds with Bun

### LINKED ISSUE #5917: [🐞] Can't build project via `bun run --bun`

Qwik Rollup / Vite plugin

### Describe the bug

I've been trying to build a project using Bun's [feature](https://bun.sh/docs/cli/run#bun) that allows to replace Node.js binary for a command. However, while Qwik has adapter for Bun's http server, it does not seem to have a full support for bun, because when I'm trying to build my project like this I get following error:

```
$ bun run --bun tsc --incremental --noEmit --pretty
$ bun run --bun vite build
vite v5.1.4 building for production...
✓ 0 modules transformed.
x Build failed in 4ms
error during build:
Error: [vite-plugin-qwik] Not implemented
    at <anonymous> (/Users/octetstream/projects/qwik-bun-build-issue/node_modules/@builder.io/qwik/optimizer.mjs:1435:35)
    at processTicksAndRejections (:12:39)
error: "vite" exited with code 1
error: script "build.client" exited with code 1
error: "qwik" exited with code 1
error: script "build" exited with code 1
```

The error is pretty clear - you don't seem to support bun runtime in qwik optimizer. I've been trying to add bun to `getPlatformInputFiles`, just to see it that would be enough (because Bun aims to be compatible with Node.js), and it seems like it is not. When I do so, a new error occurs:

```
$ bun run --bun tsc --incremental --noEmit --pretty
$ bun run --bun vite build
vite v5.1.4 building for production...
✓ 0 modules transformed.
x Build failed in 6ms
error during build:
undefined
error: "vite" exited with code 1
error: script "build.client" exited with code 1
error: "qwik" exited with code 1
error: script "build" exited with code 1
```

### Reproduction

https://github.com/octet-stream/qwik-bun-build-issue

### Steps to reproduce

1. Clone this repository via git clone `git@github.com:octet-stream/qwik-bun-build-issue.git`;
2. Install dependencis via `bun i`;
3. Run `bun run build` and the build process with fail with the error.

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 98.50 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.2 - ~/Library/Caches/fnm_multishells/2008_1708258753953/bin/node
    npm: 10.2.4 - ~/Library/Caches/fnm_multishells/2008_1708258753953/bin/npm
    pnpm: 8.15.4 - /opt/homebrew/bin/pnpm
    bun: 1.0.29 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 122.0.6261.69
    Safari: 17.3.1
  npmPackages:
    @builder.io/qwik: ^1.4.5 => 1.4.5 
    @builder.io/qwik-city: ^1.4.5 => 1.4.5 
    undici: * => 6.6.2 
    vite: ^5.0.12 => 5.1.4
```

### Additional Information

The current solution is to just keep using Node.js to build Qwik City apps. This also means Node.js should be presented during the build stage when deploying, which is not a huge problem when using Docker multilayer builds, I think. But even with Docker that would mean an unnecessary Node.js setup, just to build a project.

octet-stream:
# Overview

Fixes #5917 (partially)

While Qwik has adapter for Bun, there's no support for running Qwik's CLI via Bun itself. Because of this developers must depend on Node.js during deployments just to build their applications even if they don't use it on their servers.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This PR fixes support for Bun in Qwik by just removing `bun` from the results of `getEnv` function in `packages/qwik/src/optimizer/src/platform.ts`. I am not quite satisfied with the solution, but it works. Still, I will be much appreciated if somebody helps me find a better solution, because I am not much familiar with the codebase.

In addition, this PR limits the number of workers for SSG, because otherwise the build will just hang without any progress. I wasn't able to properly isolate this bug and versify whether it is an issue with Qwik or Bun itself, so I just added the limit. The fact that the same code works in Node.js suggest that it can be on Bun's side, so I will report this issue to them and we'll see.

Lastly, this PR adds another place where we patch TextEncoderStream for bun, because bun does's have their implementation.

Also, since Qwik already uses polyfills in other places, this is something that needs to be better organized. I suggest we add a function that will (conditionally) apply all needed polyfill. You can look it Remix as the example: https://remix.run/docs/en/main/other-api/node#polyfills - they have a function `installGlobals` that does the same thing.

# Use cases and why

Expected behavior was: `bun run --bun build` is able to run just as `pnpm build` without any issue

Actual behavior was: `bun run --bun build` fails with the error saying that the platform is not supported.

Currently to work around this issue, I had to setup Node.js in one of my Docker layers just to build my application during deployments.

# Checklist:


packages/qwik/src/optimizer/src/platform.ts
```diff
@@ -265,10 +265,6 @@ const getEnv = (): SystemEnvironment => {
     return 'deno';
   }
 
-  if (typeof Bun !== 'undefined') {
-    return 'bun';
```
PatrickJS, line 269: is there a way to make the pr work without doing this?

octet-stream: There should be a better way, the problem is likely on my side, because I have no full understanding of the code. I tried to add test the result of this function in where it is used, but this way it fails.

More specifically, here (this is where getSystem is called and then the platform is tested in if statements):

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/platform.ts#L38

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/platform.ts#L65

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/platform.ts#L167

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/platform.ts#L228

Then getSystem function is called here:

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/optimizer.ts#L17

And the result from `getEnv` function is used here:

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/plugins/plugin.ts#L309

and here

https://github.com/QwikDev/qwik/blob/a6a8525cd676c6e9c8e92ddf00466635d6c78c47/packages/qwik/src/optimizer/src/plugins/plugin.ts#L331C19-L331C37

It didn't work when I added bun to these tests and tried this locally, so I went with this solution. I definitely missed something.

PatrickJS: this is the only change preventing me from merging. can you just add if node or bun. rather than just removing bun

octet-stream: This is the result with current solution:

telegram-cloud-document-2-5362005861461937410

I will try again anyway, because what I did isn't good in any way, but it works.

octet-stream: Oh, ok. Maybe I didn't tested properly. It works. But shouldn't I also change plugin.ts same way?

PatrickJS (approved):

PatrickJS:
@octet-stream thanks!

PatrickJS:
@octet-stream this is in 1.5.5

PR #6345, opened 2024-05-20
fix(server$): return content type changes

PatrickJS:
when server$ lambda blows up the content type changes.

this is needed for debugging

### DIFF
--- a/packages/qwik-city/runtime/src/server-functions.ts
+++ b/packages/qwik-city/runtime/src/server-functions.ts
@@ -386,6 +386,12 @@ export const serverQrl = <T extends ServerFunction>(
             throw obj;
           }
           return obj;
+        } else if (contentType === 'text/plain') {
+          const str = await res.text();
+          if (res.status === 500) {
+            throw str;
+          }
+          return str;
         }
       }
     }) as ServerQRL<T>;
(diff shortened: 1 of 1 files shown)


packages/qwik-city/runtime/src/server-functions.ts
```diff
@@ -380,6 +380,18 @@ export const serverQrl = <T extends ServerFunction>(
             throw obj;
           }
           return obj;
+        } else if (contentType === 'application/json') {
```
PatrickJS, line 383: theses cases should be hit but in cases that I mentioned it's better that they're handled

thejackshelton (approved):
Looks good to me! More descriptive errors -> happiness :)

gioboa:
You are doing a great job @PatrickJS 🚀

packages/qwik-city/runtime/src/server-functions.ts
```diff
@@ -380,6 +380,18 @@ export const serverQrl = <T extends ServerFunction>(
...
+          const obj = await res.json();
+          if (res.status === 500) {
```
wmertens, line 385: Should be >=400, no?

PatrickJS: it should never hit this case but for example of the Lambda blows up then aws will change the return headers/value

PR #6350, opened 2024-05-21
fix(v2): delete HTMLFragment

Varixo:
This PR removes the HTMLFragment component. It is not needed.

### DIFF
--- a/packages/docs/src/routes/api/qwik/api.json
+++ b/packages/docs/src/routes/api/qwik/api.json
@@ -978,20 +978,6 @@
       "mdFile": "qwik.htmlelementattrs.md"
     },
     {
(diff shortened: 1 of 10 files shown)


gioboa (approved):
👏great

thejackshelton (approved):
Ah looks like CI is failing but I can't seem to see why.

Linting error in prettier, otherwise:

LGTM! 🙌

Varixo:
> Ah looks like CI is failing but I can't seem to see why.
> 
> Linting error in prettier, otherwise:
> 
> LGTM! 🙌

Yeah, this is weird because I didn't change anything inside the vnode-diff file 🤔

PR #6365, opened 2024-05-23
fix: utf-8 lowercase

PatrickJS:
fix utf-8 lowercase

> The default character encoding used in HTML5 is UTF-8. This means if you include <!DOCTYPE html> at the top of your HTML file (which declares that it's an HTML5 file), it'll automatically use UTF-8 unless specified otherwise.

> Most browsers use UTF-8 by default if no character encoding is specified. But because that's not guaranteed, it's better to just include a character encoding specification using the <meta> tag in your HTML file.

### DIFF
--- a/packages/docs/public/repl/~repl-server-host.html
+++ b/packages/docs/public/repl/~repl-server-host.html
@@ -1,7 +1,7 @@
 <!doctype html>
 <html>
   <head>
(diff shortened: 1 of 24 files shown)


thejackshelton (approved):
LGTM! 🙌

PR #6366, opened 2024-05-23
fix(prefetch): configure scope and give better errors

PatrickJS:
fix configuring scope value.


thejackshelton (approved):
LGTM 🙌 . Thank you for adding tests

PR #6378, opened 2024-05-24
fix(v2): fix double empty text node + initial qwik city fixes

Varixo:
This PR fixes double empty text node rendering problem and adds initial qwik city fixes

### DIFF
--- a/packages/qwik/src/core/index.ts
+++ b/packages/qwik/src/core/index.ts
@@ -81,7 +81,7 @@ export type { QwikDOMAttributes, QwikJSX, QwikJSX as JSX } from './render/jsx/ty
 export type { QwikIntrinsicElements } from './render/jsx/types/jsx-qwik-elements';
 export type { QwikHTMLElements, QwikSVGElements } from './render/jsx/types/jsx-generated';
 export { render2 as render } from './v2/client/dom-render';
-export { getDomContainer } from './v2/client/dom-container';
+export { getDomContainer, getQContainerElement } from './v2/client/dom-container';
 export type { RenderSSROptions, StreamWriter } from './render/ssr/render-ssr';
 export type { RenderOptions, RenderResult } from './render/dom/render.public';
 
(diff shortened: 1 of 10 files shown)


packages/qwik/src/core/v2/client/dom-container.ts
```diff
@@ -87,18 +87,12 @@ export function getDomContainerFromHTMLElement(qContainerElement: Element): ICli
...
     ? (vnode_getDomParent(element) as Element)
     : element;
-  while (
-    qContainerElement &&
-    (!qContainerElement.hasAttribute(QContainerAttr) ||
-      isQContainerElementWithValue(qContainerElement))
-  ) {
-    qContainerElement = qContainerElement.closest(QContainerSelector);
-  }
-  return qContainerElement;
```
Varixo, line 101: not sure why this while loop was here

mhevery: no idea

packages/qwik/src/core/index.ts
```diff
@@ -81,7 +81,7 @@ export type { QwikDOMAttributes, QwikJSX, QwikJSX as JSX } from './render/jsx/ty
 export type { QwikIntrinsicElements } from './render/jsx/types/jsx-qwik-elements';
 export type { QwikHTMLElements, QwikSVGElements } from './render/jsx/types/jsx-generated';
 export { render2 as render } from './v2/client/dom-render';
-export { getDomContainer } from './v2/client/dom-container';
+export { getDomContainer, getQContainerElement } from './v2/client/dom-container';
```
mhevery, line 84: You are making `getQContainerElement` public. I am not sure we should do that. We will need to document it and support it. I think it is better for it to stay private.

mhevery (changes requested):

packages/qwik/src/core/index.ts
Varixo: I have changed `getQContainerElement` from public to internal. It is okay?

mhevery (approved):

PR #6389, opened 2024-05-25
ci: added ability to publish test packages

shairez:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Added [pkg.pr.new](https://github.com/stackblitz-labs/pkg.pr.new) to be able to test PRs and latest main commits without the need to release a new version each time.

Great for testing new fixes before they go live

# Checklist:


.github/workflows/ci.yml
```diff
@@ -31,6 +31,14 @@ env:
   CARGO_INCREMENTAL: 0
 
 jobs:
+  check:
+    name: check
+    runs-on: ubuntu
+    steps:
+      - name: check
+        run: |
+          echo "node input: ${{ inputs.node_version }}"
```
wmertens, line 40: I assume you'll add something in this job?

.github/workflows/ci.yml
```diff
@@ -11,6 +11,11 @@ on:
       - 'build/**'
   workflow_dispatch:
     inputs:
+      node_version:
+        description: 'Node version to use for the build'
+        required: false
+        type: string
+        default: '20.x'
```
wmertens, line 18: 22 is faster

.github/workflows/ci.yml
```diff
@@ -481,6 +494,9 @@ jobs:
             packages/qwik-labs/vite/
             packages/qwik-labs/package.json
           if-no-files-found: error
+      - run: pnpx pkg-pr-new publish --compact ./packages/qwik/dist ./packages/qwik-city/dist dist-dev-eslint-plugin-qwik dist-dev-create-qwik
```
wmertens, line 497: What makes these dist-dev directories?

.github/workflows/ci.yml
shairez: This was just to check, but it's removed now

.github/workflows/ci.yml
shairez: the upload-artifacts task from the job before

package.json
```diff
@@ -108,6 +108,7 @@
     "monaco-editor": "^0.45.0",
     "mri": "1.2.0",
     "path-browserify": "1.0.1",
+    "pkg-pr-new": "^0.0.5",
```
gioboa, line 111: We are using `pnpx`, do we need this dependency?

.github/workflows/ci.yml
```diff
@@ -481,6 +486,9 @@ jobs:
             packages/qwik-labs/vite/
             packages/qwik-labs/package.json
           if-no-files-found: error
+      - run: pnpx pkg-pr-new publish --compact ./packages/qwik/dist ./packages/qwik-city/lib dist-dev-eslint-plugin-qwik dist-dev-create-qwik
```
gioboa, line 489: `pnpx` is [deprecated](https://github.com/pnpm/pnpm/pull/3652) 
Can we use the new notation pls?

gioboa (changes requested):
Nice job. I really like this idea.

.github/workflows/ci.yml
shairez: yes

package.json
shairez: I think it's because we want to keep the consistency of the version.

shairez:
Currently `pkg.pr.new` has a size limit we exceed.
until we'll sort it out, this PR stays in draft mode

package.json
gioboa: From my understanding the X command will use the latest version though

shairez: it should use the local one and only install the latest if a local one is not found

https://docs.npmjs.com/cli/v10/commands/npx

image

gioboa: Thanks for investigating into that. So keeping the dependency makes sense to me 😉

.github/workflows/ci.yml
gioboa: we can use `pnpm dlx pkg-pr-new@0.0.5` 
we can remove the dep if we will use the `@` notation

wmertens:
@shairez we can fix the size limit by not shipping the native plugins, only the wasm. For that, it should skip the native builds and you should overwrite `packages/qwik/src/optimizer/src/qwik-binding-map.ts` so it exports `{}`.

.github/workflows/ci.yml
shairez: done

shairez: OK I figured it out

.github/workflows/ci.yml
```diff
@@ -150,7 +155,12 @@ jobs:
           path: packages/eslint-plugin-qwik/dist/
           if-no-files-found: error
 
-  ############ BUILD WASM ############
+      - name: Publish packages for testing
+        run: pnpm dlx pkg-pr-new@0.0.8 publish --compact --pnpm ./packages/qwik/dist ./packages/qwik-city/lib ./packages/eslint-plugin-qwik/dist ./packages/create-qwik/dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
AmirSa12, line 161: This is not needed anymore, FYI :)

shairez: the github token?

shairez:
Thanks so much @Aslemammad for the awesome job you did in creating this and fixing it to make it work for us as well
Very much appreciated! 🙏🙌

.github/workflows/ci.yml
Aslemammad: yep!

Aslemammad:
🙌🙌🙌

Issue #6386, opened 2024-05-25
[🐞] Using components as a jsx output in server$ is not renderable


Twiggeh:
### Which component is affected?

Qwik Runtime

### Describe the bug

E.g. using a component in the server return, makes that jsx no longer renderable, with the error:
Rendering promises in JSX signals is not supported Error: Rendering promises in JSX signals is not supported

```jsx
const _getRooms = () => {
	if (!isServer) throw new Error("No access from client");
	if (!globalThis.rooms)
		globalThis.rooms = ["hello", "goodbye", "something else", "noooo"];
	return globalThis.rooms as string[];
};

const ExpComp = component$(({ rooms }: { rooms: string[] }) => {
	if (!rooms.length) return <div>No Rooms yet!</div>;

	return (
		<div>
			{rooms.map((g) => (
				<div key={g}>{g}</div>
			))}
		</div>
	);
});

const getMyExpensive = server$(() => {
	const rooms = _getRooms();
	return <ExpComp rooms={rooms}></ExpComp>;
});
```

This works in routeLoaders though, also works if I inline the component.

### Reproduction

https://github.com/Twiggeh/qwik-jsx-output

### Steps to reproduce

Its the default starter template with bun. You can also just copy the src/routes/index.tsx, since it contains all of the code to reproduce.

### System Info

```shell
replicable on stackblitz too

  System:
    OS: Linux 6.6 NixOS 24.05 (Uakari) 24.05 (Uakari)
    CPU: (16) x64 AMD Ryzen 9 5900HX with Radeon Graphics
    Memory: 25.61 GB / 30.78 GB
    Container: Yes
    Shell: 5.2.26 - /run/current-system/sw/bin/bash
  Binaries:
    Node: 20.11.1 - /run/current-system/sw/bin/node
    npm: 10.2.4 - /run/current-system/sw/bin/npm
    bun: 1.1.6 - /nix/store/bg0s001n46kdb5k7r1bmwabnbg5786l6-bun-1.1.6/bin/bun
  Browsers:
    Brave Browser: 122.1.63.162
```
```


### Additional Information

@PatrickJS asked me to put this as a bug from Discord: 
https://discord.com/channels/842438759945601056/1242809922463207494

PatrickJS:
This is fixed in v2

gioboa:
I did a double check with the latest v2 code and it's working fine.
v2 will be release soon so I marked this issue with `STATUS-3: solved by V2`, thanks.

PR #6390, opened 2024-05-25
feat(signals): promises JSX in signals

PatrickJS:
promises in signals

maybe partial fix for https://github.com/QwikDev/qwik/issues/6386


PatrickJS:
doesn't seem to work

Issue #6103, opened 2024-04-08
[🐞] Redirects Not Working 


JBischoff28:
### Which component is affected?

Qwik City (routing)

### Describe the bug

I am trying to redirect from within a routeAction$. Instead of redirecting the user, nothing happens at all. Sometimes the redirect will work, other times it will not. There has not been an identifiable pattern either. The issue has occured with both absolute and relative url paths. I have tried changing the redirect's status code, but that also had no effect.

This issue has been persistent in every project that I've used Qwik, so it is not project specific. I have also reached out on discord, and was told it was not an issue that I am making.

The following is a message that was sent to me by a Qwik team member:
ok looks to me that the makeQDataPath function isn't happy with your redirect URL. And looking at it, it only allows redirecting to absolute paths in the same domain... https://github.com/BuilderIO/qwik/blob/396e98ebab748109818f19523dc185d12bdb6aa1/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts#L546-L556

### Reproduction

npm create qwik@latest

### Steps to reproduce

1. Run "npm create qwik@latest" to start a new quick project.
2. Create a basic form.
3. Create an empty route action and attach the route action to the Form's action prop.
4. Try to redirect to another page using requestEvent.redirect
5. Sometimes the redirect will work, other times it will not. I've been using relative paths when I am redirecting to another location within the app, and absolute paths when redirecting outside the app.

### System Info

```shell
Nothing out of the ordinary. Use your standard development configs and you should get the same result.
```

taichi221228:
@JBischoff28 See [here](https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts#L503). The status code may be the reason for the redirection.

PatrickJS:
oh we can add a console warning or error in dev-only if the status codes aren't the redirecting kind

PatrickJS:
I can add more errors etc. but the issue is this line then
`const url = new URL(href, 'http://localhost');`

taichi221228:
Thanks for the answer!
The logs may indeed be unfriendly, as nothing is output when an unsuitable URL is specified as well as a status code.

> And looking at it, it only allows redirecting to absolute paths in the same domain...

I are currently trying to resolve this issue.
For reference, I checked the [redirect function in Next.js](https://nextjs.org/docs/app/api-reference/functions/redirect) and it does indeed support external links. 

[...]

PR #6410, opened 2024-05-27
fix(qwik-city): fix redirect with search params

octet-stream:
# Overview

I am not sure if my issue is related to #6103, so I decided open a PR with the fix for my issue. This PR changes how qwik-city-component handles URLSearchParams changes within same route segment.

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This PR replaced `isSamePathname` with `isSamePath` so that the URL will be updated after redirect.

# Use cases and why

Current behavior: When action redirects within same route segment, and only changes URLSearchParams, Qwik ignores the change of the URL, but the data on the page is updated.

After this change: The Qwik updates the URL, along with data refresh, just the way it works in browsers.

Reproduction demo can be found here: https://stackblitz.com/edit/qwik-redirects-er3vm8-36vtnr?file=src%2Froutes%2Flayout.tsx&view=editor

# Checklist:


JerryWu1234 (approved):

packages/qwik-city/runtime/src/qwik-city-component.tsx
```diff
@@ -263,7 +263,7 @@ export const QwikCityProvider = component$<QwikCityProps>((props) => {
         }
         const newHref = pageData.href;
         const newURL = new URL(newHref, trackUrl);
-        if (!isSamePathname(newURL, trackUrl)) {
+        if (!isSamePath(newURL, trackUrl)) {
           trackUrl = newURL;
```
JerryWu1234, line 267: because of this line,

```js
if (!isSamePathname(newURL, trackUrl)) {
          trackUrl = newURL;
          loadRoutePromise = loadRoute(qwikCity.routes, qwikCity.menus, qwikCity.cacheModules, trackUrl.pathname);
        }

        if(trackUrl.href !== newURL.href) {
          trackUrl = newURL;
        }
```

it would be better if it only reassigns this value `trackUrl`.
because it does need to recall loadRoute function
how do you think so

wmertens: @JerryWu1234 not sure what you mean here. This PR just seems to change the check for equality so that it also checks the search string to decide when to re-run route loaders. I don't see the `trackUrl.href` check?

wmertens (commented):
I'm not sure this isn't a breaking change, since now it will re-run route loaders where it wasn't doing that before.

If we apply this, then there's no way to opt out of this behavior. If you are updating the search string often, it will cause a lot of network requests, no?

Perhaps there is a way to opt out of route loading when the search string changes?

@mhevery thoughts?

packages/qwik-city/runtime/src/qwik-city-component.tsx
JerryWu1234: https://github.com/QwikDev/qwik/blob/a219b51f701bf29e8782dbbf5c0ada0928f34f39/packages/qwik-city/runtime/src/client-navigate.ts#L19

it is the reason why route didn't redirect.
So if newURL and  trackUrl aren't equal, we can only make `trackUrl ` get new URL.

so that [samePath](https://github.com/QwikDev/qwik/blob/a219b51f701bf29e8782dbbf5c0ada0928f34f39/packages/qwik-city/runtime/src/client-navigate.ts#L14C5-L14C49) would become true ,and route will redirect.
if so that, it doesn't need to call `loadRoute ` 

@wmertens

octet-stream:
> If you are updating the search string often, it will cause a lot of network requests, no?

In my case this is expected behavior, because loader depends on the `URLSearchParams` to serve a data and there's no way I can think of to know from the client if the server depends this data. This is also how this works in browsers: If you click a link, or return a redirect response with `Location` header, your browser will send a request to the url from the `Location` header, then update the URL in the address bar and render that page, just as if you clicked a link. You can open my repro, disable JavaScript and try update the filter. The current behavior in Qwik is that it uses default behavior for `fetch`, which is to follow redirects (fetch sends a request, and if 3xx status code occurs it checks for `Location` header, and if it set to a URL, it will automatically send a subsequent request, then return the result of that request as `Response`, I think this is where redirect behavior is described in spec https://fetch.spec.whatwg.org/#http-redirect-fetch). The only missing part is that Qwik will not update the URL after such response is received. At least this is how it works in my case (which is URLSearchParams change).

I can also confirm that there're might be use-cases when developers want to update URLSearchParams without triggering navigation, given that such discussions exist: https://github.com/remix-run/react-router/discussions/9851

packages/qwik-city/runtime/src/qwik-city-component.tsx
gioboa: I think the correct behavior is to re run the loader, we are doing the same with `actions` so for consistency I will prefer call `loadRoute`

gioboa (approved):
I tested your code locally and it's working fine.
This is a great fix btw. With the old code `routeLoader$` is running without changing the URL and this is not correct at all.
With your fix @octet-stream the problem is solved.

PR #6417, opened 2024-05-28
fix(v2): fix navigation scroll restore

Varixo:
This PR fixes the remaining e2e navigation tests

### DIFF
--- a/packages/qwik-city/runtime/src/qwik-city-component.tsx
+++ b/packages/qwik-city/runtime/src/qwik-city-component.tsx
@@ -56,6 +56,7 @@ import {
   getScrollHistory,
   saveScrollHistory,
   restoreScroll,
+  callRestoreScrollOnDocument,
 } from './scroll-restoration';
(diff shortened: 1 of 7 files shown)


packages/qwik-city/runtime/src/qwik-city-component.tsx
```diff
@@ -514,6 +515,9 @@ export const QwikCityProvider = component$<QwikCityProps>((props) => {
             const scrollState = currentScrollState(scroller);
             saveScrollHistory(scrollState);
             win._qCityScrollEnabled = true;
+            if (isBrowser) {
+              callRestoreScrollOnDocument();
+            }
```
Varixo, line 520: In V1 this logic is inside the qwik core in the notify-render. I don't like this, so I moved this here (tests are passing, so I guess it is ok)

Varixo: The only thing is transition api, there is no test for it

PR #6419, opened 2024-05-28
chore: qwik-auth v0.2.1

wmertens:
not sure if this works, my dev env is broken right now


PatrickJS:
not sure why cf is failing but looks good

wmertens:
@shairez can you try publishing from this?

wmertens:
ah it fails because the types are already created by the build apparently. So the dts plugin isn't needed.

```
9:15:16 PM: lib/index.qwik.cjs  11.19 kB │ gzip: 3.32 kB
9:15:16 PM: ✓ built in 12.51s
9:15:16 PM: ❌ Error: EEXIST: file already exists, mkdir '/opt/build/repo/packages/qwik-auth/lib/types'
9:15:16 PM:  Error: Error: EEXIST: file already exists, mkdir '/opt/build/repo/packages/qwik-auth/lib/types'
9:15:16 PM:     at panic (/opt/build/repo/scripts/util.ts:302:33)
9:15:16 PM:     at build (/opt/build/repo/scripts/build.ts:198:5)
```

PR #6427, opened 2024-05-31
feat(vite-server config): added the option to disable image dev tools for dev server

s-voloshynenko:
# Overview

Added the option to disable image dev tools for dev server.

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description
Added the option to disable image dev tools for dev server by extending `vite.config`:
```ts
...,
qwikVite({
  ...,
  devTools: {
    imageDevTools: false,
  },
}),
...,
```

By default it's enabled, as we always should force to build a good user experience.

TBH, I would rather to upgrade image dev tools with more functionality where we can define with jsx-comments which content is safe so that it can be skipped during validation check. On the other hand, devTools should be optional as clickToSource plugin so that developer can decide which tools are useful.

# Use cases and why

Qwik dev server has a really useful built-in tool to identify CLS issues, and it might be the case that we can't always determine the real width and height of an image. E.g. if we want to display a list of different dynamic logos in a single row that is limited to a specific height. To save aspect-ratio of the logo, width of the logo can vary. With the current configuration, you will see a number of CLS overlays but in fact it can be an expected behavior, especially if that content is not SSR-first.

I also saw questions in Discord chat how to disable it, so it might be useful.

# Checklist:


packages/docs/src/routes/api/qwik/index.md
```diff
@@ -3675,7 +3675,7 @@ Options for the prefetch service worker.
 </tbody></table>
 **Returns:**
 
-[JSXNode](#jsxnode)&lt;'script'&gt;
```
s-voloshynenko, line 3678: not sure why it was chaned, shall I revert it?

PatrickJS:
thanks, I'll need to use this one soon

PatrickJS (approved):

gioboa (approved):
Thanks @s-voloshynenko for this great PR

PR #6432, opened 2024-06-01
test: fix locale e2e test

shairez:
# Overview
# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Fixes the broken locale e2e test
# Use cases and why

### DIFF
--- a/starters/e2e/qwikcity/locale.e2e.ts
+++ b/starters/e2e/qwikcity/locale.e2e.ts
@@ -1,11 +1,12 @@
 import { expect, test } from "@playwright/test";
+import { QContainerSelector } from "../../../packages/qwik/src/core/util/markers";
 
 test.describe("Qwik City locale API", () => {
   test("pass locale to Qwik", async ({ page }) => {
     await page.goto("/qwikcity-test/locale");
     const locale = page.locator(".locale");
     await expect(locale).toHaveText("test-locale");
-    const qContainer = page.locator("[q\\:container=paused]");
+    const qContainer = page.locator(QContainerSelector);
     await expect(qContainer).toHaveAttribute("q:locale", "test-locale");
   });
 });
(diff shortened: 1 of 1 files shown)


starters/e2e/qwikcity/locale.e2e.ts
```diff
@@ -1,12 +1,11 @@
...
 
 test.describe("Qwik City locale API", () => {
-  // TODO(v2): fix this
-  test.skip("pass locale to Qwik", async ({ page }) => {
+  test("pass locale to Qwik", async ({ page }) => {
     await page.goto("/qwikcity-test/locale");
     const locale = page.locator(".locale");
-    const qContainer = page.locator("[q\\:container]");
     await expect(locale).toHaveText("test-locale");
+    const qContainer = page.locator("[q\\:container=paused]");
```
Varixo, line 8: yeah this is a good way to solve this, but I think we should use the same selector as the rest of the system
Look here https://github.com/QwikDev/qwik/blob/3feb2e6da2e4e9725dd8331e2730ba13e88e63d5/packages/qwik/src/core/util/markers.ts#L42

Varixo:
Could we skip `Cloudflare Pages` for v2 branch? Is it a good idea?

starters/e2e/qwikcity/locale.e2e.ts
shairez: good idea!

shairez:
@mhevery do you think we need the cloudflare pages on the v2 branch?

for some reason it's failing on not finding the CLI

starters/e2e/qwikcity/locale.e2e.ts
shairez: done

Varixo:
@shairez can we merge this?

shairez:
Yeah I'll do it

PR #6443, opened 2024-06-02
feat(v2): new signals implementation

mhevery:
# Overview

This is an exploration (may not land) the Signal implementation for something more clear / more maintainable.

Todo: 

- [x] Figure out why the unit test complains when removing `.only`
- [x] element subscriber type 
- [x] change all subscriber invocations to be the format that Signal2 uses
- [x] store support (deep)
- [x] (de)serialization 
- [x] replace original signal
- [x] remove original signal and rename

---

Note about cross-container support: because:

- containers can resume in any order
- qwikloader only handles events, not serialization
- versions may differ between containers

, the only way to communicate between containers is with custom events like `"q-signal:my- global-name"` that dispatch signal values. This would be another effect handled by a separate GlobalSignal.


JerryWu1234:
image

Will this PR also fix these test problems?

wmertens:
@JerryWu1234 right that's the idea, we were having trouble getting signals to work so we decided to improve the implementation

packages/qwik/src/core/v2/signal/v2-signal.ts
```diff
@@ -388,7 +430,8 @@ export class ComputedSignal2<T> extends Signal2<T> {
     throw new TypeError('ComputedSignal is read-only');
   }
 }
-
+// TO DISCUSS (with Misko): Shouldn't it be called a "WrappedSignal" ?
```
wmertens, line 433: Yes I'd prefer that too

packages/qwik/src/core/v2/shared/scheduler.ts
```diff
@@ -141,6 +141,11 @@ export interface Chore {
   $returnValue$: any;
 }
 
+export interface NodePropPayload {
+  value: string;
+  scopedStyleIdPrefix: string | null;
```
wmertens, line 146: Surround with $$, no?

wmertens (approved):

Issue #6423, opened 2024-05-29
[🐞] args of the `component$` are not mutable.


genki:
### Which component is affected?

Qwik Runtime

### Describe the bug

The props of the `component$` are given as the args of the QRL, so they should be mutable, but they are declared as `const` to be immutable through the `useLexicalScope`. This makes mismatch.
So if I try to change the value of the args, it causes `TypeError: Assignment to constant variable.` even if there's no error in compile time.

### Reproduction

https://stackblitz.com/edit/qwik-starter-c2m93t?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

Please see the link above.

### System Info

```shell
System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M2
    Memory: 93.02 MB / 24.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.12.1 - /opt/homebrew/opt/node@20/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.5.0 - /opt/homebrew/opt/node@20/bin/npm
    pnpm: 8.7.5 - ~/Library/pnpm/pnpm
    bun: 1.1.3 - ~/.bun/bin/bun
  Browsers:
    Chrome: 125.0.6422.113
    Safari: 17.4.1
  npmPackages:
    @builder.io/qwik: file:../clone/qwik/packages/qwik/dist => 1.5.3-dev20240507001309 
    @builder.io/qwik-city: file:../clone/qwik/packages/qwik-city/lib => 1.5.3-dev20240507001309 
    typescript: ^5.3.3 => 5.3.3 
    undici: ^5.28.3 => 5.28.4 
    vite: ^5.1.5 => 5.2.11
```

maiieul:
I'm not sure, this might be intended behavior to enable lazy-loading/lazy-execution. 

May I ask what's your use case? You could use a signal instead:

```tsx
import {
  Signal,
  component$,
  useSignal,
  useVisibleTask$,
} from '@builder.io/qwik';

type FooProps = {
  testSig: Signal<number>;
};

const Foo = component$<FooProps>(({ testSig }) => {
  // eslint-disable-next-line qwik/no-use-visible-task
  useVisibleTask$(() => {
    testSig.value = ++testSig.value;
  });
  return (
    <>
      <div>{testSig}</div>
    </>
  );
});

export default component$(() => {
[...]

PR #6460, opened 2024-06-05
docs(props): Immutability & Signals explanation and example.

codyroberts:
# Overview

The props section now clarifies that props are immutable and cannot be changed, and recommends readers to use signals in situations where the props need to be updated.

The example now shows two separate ways to use signals for props.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. Explains prop immutability
- 2. Recommends the use of signals
- 3. Gives examples of signal use  

# Checklist:


packages/docs/src/routes/docs/(qwik)/components/overview/index.mdx
```diff
@@ -113,40 +113,46 @@ export default component$(() => {
 
 ## Props
 
-Props are used to pass data from the parent into the component. Props are accessible via the `props` argument to the component$ function.
+Props are used to pass data from the parent into the component. They are accessible via the `props` argument to the `component$` function. Data passed as props is immutable, and cannot be changed. In cases where data sent as props needs to be updated, it is recommended to use signals.
```
wmertens, line 116: This is not actually the case. The data is mutable, but unless you use signals or stores, there won't be a reactive change. 
The props object itself is private to the component, it can also be mutated without problems, which might be handy when using props spreading.

packages/docs/src/routes/docs/(qwik)/components/overview/index.mdx
```diff
@@ -113,40 +113,46 @@ export default component$(() => {
...
 
-In this example a component `Item` declares optional `name`, `quantity`, `description`, and `price`.
+In the example below a component `Item` declares optional `name`, `quantity`, `description`, and `price` props. The `quantity` and `price` props are signals; `price` is a signal before being sent to `Item` and can be changed from the parent component. `quantity` is sent as a number value, and a new signal is created from the initial value inside of the `Item` component, where it can be changed.
```
wmertens, line 118: No, this is actually not a good idea. You should deference signals when you put them into props, and the optimizer will forward them.

Using signals as props is possible but it complicates the mental model and actual code without benefits.

wmertens (changes requested):
Thank you for your contribution, improving the docs is always welcome, but props should generally not be signals. 

An exception is the `bind:value` prop.

packages/docs/src/routes/docs/(qwik)/components/overview/index.mdx
shairez: @codyroberts maybe changing "immutable" to "non reactive" would be more accurate in that case?

@wmertens can you give a short example of the props spreading benefit please?

codyroberts: I tested it locally and I am still getting `Error: Code(17): props are immutable` when doing something simple like:

```js
export const Item = component$<ItemProps>((props) => {
  props.price = 10;
  return (
    <ul>
      <li>name: {props.name}</li>
      <li>quantity: {props.quantity}</li>
      <li>description: {props.description}</li>
      <li>price: {props.price}</li>
    </ul>
  );
});
```

I was making this change to the documentation in response to this thread:
https://github.com/QwikDev/qwik/issues/6423

packages/docs/src/routes/docs/(qwik)/components/overview/index.mdx
shairez: @codyroberts so in Qwik, there's an optimization happening behind the scenes which allows you to use `someProp={mySignal.value}` 

This won't actually create a subscription, but will pass the signal down for you and let you use that value wherever you need to use it and the subscription will be created for you.

@wmertens btw, it'll be "read only", meaning you can't really change that signal when you pass it's value as a prop.

If you want the ability to change it you need to pass it unwrapped (as the example shows) and then you can modify it.

codyroberts: Oh I see now, thank you! 

If necessary I could probably make the documentation clearer about the fact that if you don't want to use a signal and don't need to modify the original value then you can create a local variable and set it to the props value, then change that.

packages/docs/src/routes/docs/(qwik)/components/overview/index.mdx
wmertens: @codyroberts ah indeed you're right, I didn't notice that before. So indeed the props object is shallowly immutable, but not deeply. This does not error:

```tsx

export const Foo = component$((props: any) => {
  // mutate a given prop
  props.bar.hello = 3

  return <button {...props}>A button</button>
})

  return <Foo bar={{ hi: true }} />;
});
```

So deleting props isn't possible, you need to override via the props order.

packages/docs/src/routes/docs/(qwik)/components/overview/index.mdx
PatrickJS: "I could probably make the documentation clearer" yes please

shairez: > If necessary I could probably make the documentation clearer about the fact that if you don't want to use a signal and don't need to modify the original value then you can create a local variable and set it to the props value, then change that.

Thanks!
yeah most of the time if you just need the prop value you'd probably just de-structure it like:
```
component$(({something}: MyProps) => {
...
} )

// OR

const { something } = props;

```

That's what mostly we're seeing

codyroberts:
@shairez @wmertens  @PatrickJS 

I tried to clarify the things we talked about, and added a couple examples and corresponding demos. Please let me know what you think!

PatrickJS:
I'm good with everything so long as everyone else is

PatrickJS (approved):

gioboa (approved):
Thanks for your help 👏

shairez:
LGTM
Thanks @codyroberts ! 🙏

PR #6481, opened 2024-06-10
docs: improve readability

sajebehari:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwik)/faq/index.mdx
+++ b/packages/docs/src/routes/docs/(qwik)/faq/index.mdx
@@ -191,7 +191,7 @@ export const StructuralChange = component$(() => {
 
 ```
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/(qwik)/faq/index.mdx
```diff
@@ -191,7 +191,7 @@ export const StructuralChange = component$(() => {
...
+The important thing to understand is that vDOM doesn't cause performance issues in Qwik. In React, however, invalidating a root component results in vDOM for the whole tree getting created. In Qwik, the decision is made on a per-component basis.  And only for components that have structural change AND are  changing their structure. If a component is structural (vDOM) but no change in structure is detected, then Qwik skips the component. You can think of it as auto-memoization of all components, this means that vDOM is only employed when the view is changing structurally. This is rare because in most cases the view only changes its values.
```
gioboa, line 194:
```suggestion
The important thing to understand is that vDOM doesn't cause performance issues in Qwik. In React, however, invalidating a root component results in vDOM for the whole tree getting created. In Qwik, the decision is made on a per-component basis.  And only for components that have structural change AND are changing their structure. If a component is structural (vDOM) but no change in structure is detected, then Qwik skips the component. You can think of it as auto-memoization of all components, this means that vDOM is only employed when the view is changing structurally. This is rare because in most cases the view only changes its values. 
```
removed space

gioboa (approved):

PR #6543, opened 2024-06-14
revert: expose MDX transformer and context

gioboa:
Reverts QwikDev/qwik#6535


wmertens (approved):
Approved but why do you need to revert? We don't want to guarantee this API?

thejackshelton:
> Approved but why do you need to revert? We don't want to guarantee this API?

It exposes an internal API. Further context in core chat

PatrickJS (approved):

Issue #6546, opened 2024-06-14
[✨] Enable resuming within Shadow DOM


1000hz:
### Is your feature request related to a problem?

At Cloudflare, we are experimenting with embedded microfrontends and need to isolate them from each other and the parent application by rendering them within a `ShadowRoot`.

In this playground example, you can see that UI rendered within a shadow root fails to be resumed and is not properly interactive. Here, `<App>` is rendered within a `<template shadowrootmode="open">`, which places it inside of a shadow root:
https://qwik.dev/playground/#f=5Va9DoIwGNx5ioalmFQGExxMwM33IEKECELEjfDufj%2F9gyg6mrjR0l7b6%2FW%2B80Szd8cZ%2BWK45ipPC8qV4jUxGRhYxklXzVpGyNi7kLIKVqDFnFhwjdbMBIKXj0KnBg4fWYZ0y6I7DzG4cFtu0bhgoAyADAu8%2BgJ4VZoNy1kyIlnk96tEmFeP1d8liB5bNPSNCjhh2Zu0AWOkfdh%2BISKCisGKoMikDGxaKZwU9ySOQjb1pXpIcdBdGwsx6S8Hemrq3mmGo55pcxHUmnIqIvKYGB4B9%2FAVd8bffHp8Xih60k8oV2L0jjdxHKXF%2F812dknyy7YD6aJvMEwNVQ5lDbfddkWZhpBtb6Gv3rlDIa6Z%2B8m1ng

In our case, we're actually rendering the entire Qwik app within a shadow root. This further complicates things because the Qwik runtime is injected within the shadow root where any references to `document.currentScript` are `null`.

I also suspect there are problems with the event retargeting that occurs when an event crosses shadow root boundaries. The Qwik runtime will need to use `event.composedPath()` to see the original event target, as `event.target` will point to the shadow root host.

### Describe the solution you'd like

Ideally, we'd be able to instruct the Qwik runtime which root node it should attach its global event listeners on.

### Describe alternatives you've considered

-

PR #6547, opened 2024-06-15
feat(core): Add support to shadow DOM for QwikLoader.

mhevery:
feat(core): Add support to shadow DOM for QwikLoader.

Shadow DOM can be used to isolate an application. The isolation also
includes:
- Events don't bubble across shadow boundaries.
- QuerySelector doesn't cross shadow boundaries.

QwikLoader relies on both event bubbling and querySelector to process
events such as `on:click` and dispatch such as `on-window:resize`.

To enable QwikLoader to work with shadow DOM, it is necessary to tell
qwikloader that when it is registering an event listener, as well as
performing querySelector, it should do so not just on document but also
inside the shadow DOM. To do that it is the responsibility of the
developer to tell qwikloader about the shadow DOM roots.

CONSTRAINTS:
The qwikloader applications `q:container` must be fully inside the
shadow root. (i.e. the application itself must not cross the shadow DOM
boundaries.)

Developer responsibility: (choose one)

- Have qwik loader find the roots by annotate the elements which have 
  shadow DOM with `q:shadowRoot` attribute.
- Manually insert a `<script>` tag into HTML which will add the shadow DOM
  root into the qwikloader's `qwikevents` array.

Example:
Possible implementation
```html
<div id="my-child" q:shadowRoot>
  <template shadowrootmode="open">
     <div q:container="..." q:base="..."...>qwik-container here</div>
  </template>
</div>
```

Fixes: #6546

### DIFF
--- a/packages/qwik/src/qwikloader.ts
+++ b/packages/qwik/src/qwikloader.ts
@@ -18,11 +18,14 @@ export const qwikLoader = (
   const Q_CONTEXT = '__q_context__';
   type qWindow = Window & {
     qwikevents: {
-      push: (...e: string[]) => void;
+      events: Set<string>;
+      roots: Set<Node>;
+      push: (...e: (string | (EventTarget & ParentNode))[]) => void;
     };
   };
   const win = window as unknown as qWindow;
-  const events = new Set();
+  const events = new Set<string>();
+  const roots = new Set<EventTarget & ParentNode>([doc]);
 
   // Some shortenings for minification
   const replace = 'replace';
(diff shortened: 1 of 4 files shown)


mhevery:
So turns out that you can't use `documen.currentScript` inside the shadow root 🤦‍♂️ . So we will need to do this differently. 🤔 

I think a better way to do it is for the QwikLoader to query `q:shadow` and for each one recursively descend into the shadow root and if it can find `q:container` attribute, than it should automatically add the shadow-fragment into the tracking set of roots. I will do that change.

packages/qwik/src/qwikloader.ts
```diff
@@ -32,8 +35,20 @@ export const qwikLoader = (
   const isConnected = 'isConnected';
   const qvisible = 'qvisible';
   const Q_JSON = '_qwikjson_';
+  const qContainerAttr = '[q\\:container]';
+  const nativeQuerySelectorAll = (root: ParentNode, selector: string) =>
+    Array.from(root.querySelectorAll(selector));
   const querySelectorAll = (query: string) => {
-    return doc.querySelectorAll(query);
+    const elements: Element[] = [];
+    roots.forEach((root) => elements.push(...nativeQuerySelectorAll(root, query)));
```
PatrickJS-kunaico, line 43: since they're all roots and assumed to be independant the order of roots doesn't matter right?

mhevery: Correct. What are you getting at?

PR #6553, opened 2024-06-15
fix(v2): fix streaming tests

Varixo:
this PR fixes the streaming tests + some refactoring

### DIFF
--- a/packages/qwik/src/core/v2/client/vnode-diff.ts
+++ b/packages/qwik/src/core/v2/client/vnode-diff.ts
@@ -562,9 +562,9 @@ export const vnode_diff = (
     }
   }
 
-  /** @param tag Returns true if `qDispatchEvent` needs patching */
-  function createNewElement(jsx: JSXNode, tag: string): boolean {
-    const element = createElementWithNamespace(tag);
+  /** @param elementName Returns true if `qDispatchEvent` needs patching */
+  function createNewElement(jsx: JSXNode, elementName: string): boolean {
+    const element = createElementWithNamespace(elementName);
 
     const { constProps } = jsx;
     let needsQDispatchEventPatch = false;
@@ -609,7 +609,7 @@ export const vnode_diff = (
           continue;
         }
 
-        if (tag === 'textarea' && key === 'value') {
+        if (elementName === 'textarea' && key === 'value') {
           if (typeof value !== 'string') {
             if (isDev) {
               throw new Error('The value of the textarea must be a string');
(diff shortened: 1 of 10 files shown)


starters/e2e/e2e.streaming.e2e.ts
```diff
@@ -53,6 +52,7 @@ test.describe("streaming", () => {
     await expect(count).toHaveText("Rerender: 1");
 
     await rerender.click();
+    await page.waitForTimeout(5_000);
```
shairez, line 55: @Varixo does this mean that we are adding a 5 seconds delay? 

Can we instead wait for an expected element / text to show up or something?

Varixo: we need to wait for the resources, I will look at it if we can do this better

packages/qwik/src/server/v2-ssr-render2.ts
```diff
@@ -208,8 +208,8 @@ function handleStreaming(opts: RenderToStreamOptions, timing: RenderToStreamResu
           } else {
             enqueue(chunk);
           }
-          const chunkSize = networkFlushes === 0 ? initialChunkSize : minimumChunkSize;
-          if (forceFlush || bufferSize >= chunkSize) {
+          const maxBufferSize = networkFlushes === 0 ? initialChunkSize : minimumChunkSize;
+          if (count === 0 && (forceFlush || bufferSize >= maxBufferSize)) {
```
shairez, line 212: can we rename `count` to `openedSSRBlocks` ?

packages/qwik/src/server/v2-ssr-container.ts
```diff
@@ -1019,14 +1025,12 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
     const frame: ContainerElementFrame = {
       tagNesting: tagNesting,
       parent: this.currentElementFrame,
-      elementName: tag,
+      elementName: elementName,
```
igalklebanov, line 1028:
```suggestion
      elementName,
```

packages/qwik/src/server/v2-ssr-container.ts
```diff
@@ -970,29 +976,29 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
   }
 
   /**
-   * @param tag
+   * @param elementName
    * @param depthFirstElementIdx
    * @param isElement
    */
```
igalklebanov, line 982: let's align with current function's arguments..

why do we even need this js docs block? seems to not add much.

Varixo: yeah I will remove it

starters/e2e/e2e.streaming.e2e.ts
Varixo: I removed this, after lowering the waiting time this is not needed

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -562,9 +562,9 @@ export const vnode_diff = (
     }
   }
 
-  /** @param tag Returns true if `qDispatchEvent` needs patching */
-  function createNewElement(jsx: JSXNode, tag: string): boolean {
-    const element = createElementWithNamespace(tag);
+  /** @param elementName Returns true if `qDispatchEvent` needs patching */
```
igalklebanov, line 565: This is misleading.
The return value is only affected by `jsx` argument's `constProps`.

```suggestion
  /** 
   * Returns whether `qDispatchEvent` needs patching. This is true when one of the `jsx` argument's const props has the name of an event.
   * @returns {boolean}
   */
```

Varixo: This is because when we create an element we handle only the constProps. The varProps are handled later

igalklebanov: Yeah, but the jsdoc is for `function createNewElement(jsx: JSXNode, elementName: string): boolean {`. In which `needsQDispatchEventPatch` is only set to `true` once.

Varixo: yes, I will change the description

shairez:
@igalklebanov @Varixo please write a comment when this PR is ready to be merged, thanks!

Varixo:
> @igalklebanov @Varixo please write a comment when this PR is ready to be merged, thanks!

For me it is ready

igalklebanov:
> @igalklebanov @Varixo please write a comment when this PR is ready to be merged, thanks!

Yalla let's go! 🚀

PR #6556, opened 2024-06-16
refactor(optimizer): nicer module finding + fix dev

wmertens:
- refactored manifest generation
- dev server collects QRL parents during ssr build
- optimizer also stores dev info for noop QRLs
- dev server also uses dev info as an extra fallback
- dev source URLs are kept similar looking to the source tree
- QRLs are annotated with parent so they can be built on demand


gioboa (dismissed):
Look great to me 👍

packages/qwik/src/core/qrl/qrl.ts
```diff
@@ -225,26 +238,16 @@ export const serializeQRLs = (
...
-  const symbol =
-    symbolStartIdx == symbolEndIdx ? 'default' : qrl.substring(symbolStartIdx, symbolEndIdx);
-
-  const captureStartIdx = captureIdx;
-  const captureEndIdx = endIdx;
-  const capture =
-    captureStartIdx === captureEndIdx
-      ? EMPTY_ARRAY
-      : qrl.substring(captureStartIdx + 1, captureEndIdx - 1).split(' ');
+  const parse = /^(?<c>[^#[]*)#?((?<a>[^#]+)#)?(?<s>[^[]*)(\[(?<p>[^\]]*)\])?$/.exec(qrl);
```
Varixo, line 243: compiling regexp takes time so maybe this regexp could be outside the function?

packages/qwik/src/optimizer/src/manifest.ts
```diff
@@ -264,55 +279,87 @@ export function generateManifestFromBundles(
...
+
+    // Rollup doesn't provide the moduleIds in the outputBundle but Vite does
+    outputBundle.moduleIds ||= Object.keys(outputBundle.modules);
+
+    // Our special @qwik-entries chunk
+    // We need to parse this to get the minified locations of the js packets
+    if (outputBundle.moduleIds.some((id) => id.endsWith(QWIK_ENTRIES_ID))) {
+      const { code } = outputBundle;
+      const match =
+        /^(?<left>.*){\s*['"]begin-imports[^:]*:[^,}]*,?(?<entries>[^}]*)['"]end-imports['"]\s*:[^}]*}(?<right>.*)/m.exec(
```
Varixo, line 296: this regexp is created inside the loop, could we extract this to a variable and move outside the loop?

packages/qwik/src/optimizer/src/manifest.ts
```diff
@@ -264,55 +279,87 @@ export function generateManifestFromBundles(
...
+      // s_17zcY0gsYE4:t(()=>import("./q-Co_1fcGT.js").then(_=>_.cO),[]),
+
+      // locally embedded qrls look like
+      // o=_=>{},E=Object.freeze(Object.defineProperty({__proto__:null,s_H7LftCVcX8A:o},Symbol.toStringTag,{value:"Module"}));
+      // and then referenced as
+      // s_H7LftCVcX8A:t(()=>Promise.resolve().then(()=>E),void 0)
+      const parts = match.groups!.entries.split(',');
+      for (const part of parts) {
+        const info =
+          /\s*['"]?(?<symbol>[a-zA-Z0-9_]+)['"]?:(.*import\(['"]\.\/(?<path>[^'"]+))?(.*then\(.*=>.*\.(?<attr>\w+))?/.exec(
```
Varixo, line 319: here the same question

packages/qwik/src/core/qrl/qrl.ts
wmertens: no longer relevant, but it turns out that inline regex is faster than regex in const! https://www.measurethat.net/Benchmarks/Show/31074/0/regex-defined-inside-vs-outside-loop

mhevery (approved):

PR #6558, opened 2024-06-16
docs(qwik-nutshell): remove onInput$ reference to React onChange

maiieul:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

I deleted redundant bullet points, because without the context of the paragraph before it (which devs might be reading the bullet points after a search), it's not clear that the comparison is done against React anymore.

### Concerning the onChange:

The line deleted
> For `<input>`, the `onChange` event is called `onInput$` in Qwik.

assumes people are coming from React, where `onChange` is like html's `oninput` . Whereas Qwik's `onChange$` and `onInput$` actually reflect html's `onchange` and `oninput`, respectively 🕺.

Ref: 
https://www.w3schools.com/tags/att_onchange.asp
https://www.w3schools.com/tags/att_oninput.asp

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/guides/qwik-nutshell/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/guides/qwik-nutshell/index.mdx
@@ -25,14 +25,7 @@ created_at: '2023-03-30T19:49:50Z'
 
 ## Components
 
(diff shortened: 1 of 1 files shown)


gioboa (approved):
Your reasoning makes sense to me 👍

PR #6559, opened 2024-06-16
refactor: remove references to `currentScript`

mhevery:
`currentScript` is not available in shadow dom. For this reason we can't use it. This change removes all references to `currentScript`

The biggest change is that the `qFuncs` are no longer written to the container element but rather to the document. Because there is only one document the `qFuncs` property now includes the `container` `q:manifest-hash` to distinguish between different containers.


gioboa (dismissed):
It looks great to me :clap:

gioboa (approved):

PR #6563, opened 2024-06-17
fix(v2): fix toggle named projection

Varixo:
This PR fixes toggling named projection with an interaction


packages/qwik/src/core/v2/client/dom-container.ts
```diff
@@ -64,7 +64,7 @@ import {
   vnode_setProp,
   type VNodeJournal,
 } from './vnode';
-import { vnode_diff } from './vnode-diff';
+import { isSlotProp, vnode_diff } from './vnode-diff';
```
igalklebanov, line 67: are we sure we want `isSlotProp` to sit @ `vnode-diff`? 🤔

packages/qwik/src/core/v2/client/dom-container.ts
```diff
@@ -289,7 +289,7 @@ export class DomContainer extends _SharedContainer implements IClientContainer,
       vNode[VNodeProps.flags] |= VNodeFlags.Resolved;
       for (let i = vnode_getPropStartIndex(vNode); i < vNode.length; i = i + 2) {
         const prop = vNode[i] as string;
-        if (!prop.startsWith('q:')) {
+        if (isSlotProp(prop)) {
```
igalklebanov, line 292: 👨‍🍳 💋

packages/qwik/src/core/v2/client/dom-container.ts
Varixo: yeah, I will think about it later

Varixo: I moved this into separate file

PR #6564, opened 2024-06-17
fix(repl): code display in repl output

wmertens:
- click to file works again
- don't duplicate code DOM elements
- in mobile, put file paths at the top

And most notably:
- store now deeply updates so that code changes update the output


gioboa (approved):
Tested with the preview link and it's working fine
:clap: @wmertens Thanks

PR #6576, opened 2024-06-19
fix: moved express to dependencies section

thecoder93:
# Overview

Moved express to dependencies section

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

Express should be under dependencies

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/starters/adapters/express/package.json
+++ b/starters/adapters/express/package.json
@@ -4,12 +4,14 @@
     "build.server": "vite build -c adapters/express/vite.config.ts",
     "serve": "node server/entry.express"
   },
+  "dependencies": {
+    "express": "4.19.2"
+  },
   "devDependencies": {
     "@types/compression": "^1.7.2",
     "@types/express": "^4.17.19",
     "compression": "^1.7.4",
-    "dotenv": "^16.3.2",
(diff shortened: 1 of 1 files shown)


gioboa (approved):
That's great @thecoder93 Thanks

wmertens:
@thecoder93 @gioboa IMHO express should be in devdependencies, but only if vite is configured to bundle it into the server bundle.

We had it configured like that, which is why it was not in dependencies.

However that caused confusion when people mixed deps and devDeps. I'm not sure what the configuration is now.

@PatrickJS ?

nelsonprsousa:
> @thecoder93 @gioboa IMHO express should be in devdependencies, but only if vite is configured to bundle it into the server bundle.
> 
> We had it configured like that, which is why it was not in dependencies.
> 
> However that caused confusion when people mixed deps and devDeps. I'm not sure what the configuration is now.
> 
> @PatrickJS ?

Where can we learn more about dependencies vs devDependencies while working with qwik/vite? For me it looks strange to not have express as a dependency if it is the main entry for production?!

PR #6579, opened 2024-06-19
fix(qrl): correct qrl loading in vite dev mode

wmertens:
Qwik city preloaded spa-init.ts but that didn't include the parent info so it gave a harmless error.
Also, qrl paths with + turned into spaces, which broke loading

### DIFF
--- a/packages/qwik-city/runtime/src/spa-init.ts
+++ b/packages/qwik-city/runtime/src/spa-init.ts
@@ -3,7 +3,7 @@ import type { ScrollHistoryState } from './scroll-restoration';
 import type { ScrollState } from './types';
 
 import { isDev } from '@builder.io/qwik/build';
-import { $ } from '@builder.io/qwik';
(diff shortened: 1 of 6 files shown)


gioboa (approved):

gioboa:
Can you provide a short description to keep track of these changes in the long-term?
Looks good to me 👏

packages/qwik-city/runtime/src/spa-init.ts
```diff
@@ -14,7 +14,7 @@ import { $ } from '@builder.io/qwik';
 // - Robust, fully relies only on history. (scrollRestoration = 'manual')
 
 // ! DO NOT IMPORT OR USE ANY EXTERNAL REFERENCES IN THIS SCRIPT.
-export default $((container: HTMLElement) => {
```
PatrickJS, line 17: haha good use-case for event$

PatrickJS (approved):

PR #6582, opened 2024-06-20
fix: SW prefetch

gioboa:
Here is the test I've done with this code

image

### DIFF
--- a/packages/qwik/src/core/components/prefetch.ts
+++ b/packages/qwik/src/core/components/prefetch.ts
@@ -1,6 +1,6 @@
 import { isDev } from '@builder.io/qwik/build';
-import { _jsxC } from '../internal';
 import type { JSXNode } from '@builder.io/qwik/jsx-runtime';
+import { _jsxC } from '../internal';
 import { useServerData } from '../use/use-env-data';
 
 /**
@@ -45,7 +45,7 @@ export const PrefetchServiceWorker = (opts: {
   } else {
     // base: '/'
     // path: 'qwik-prefetch-service-worker.js
-    resolvedOpts.path = resolvedOpts.base + resolvedOpts.path;
+    resolvedOpts.path = resolvedOpts.base.replace('build/', '') + resolvedOpts.path;
   }
   // dev only errors
   if (isDev) {
@@ -147,14 +147,10 @@ export const PrefetchGraph = (
     path: 'qwik-prefetch-service-worker.js',
     ...opts,
   };
-  const args = [
-    'graph-url',
-    resolvedOpts.base,
-    resolvedOpts.base + `q-bundle-graph-${resolvedOpts.manifestHash}.json`,
-  ]
+  const args = ['graph-url', resolvedOpts.base, `q-bundle-graph-${resolvedOpts.manifestHash}.json`]
     .map((x) => JSON.stringify(x))
     .join(',');
-  const code = `(window.qwikPrefetchSW||(window.qwikPrefetchSW=[])).push(${args})`;
+  const code = `(window.qwikPrefetchSW||(window.qwikPrefetchSW=[])).push([${args}])`;
   const props = {
     dangerouslySetInnerHTML: code,
     nonce: opts.nonce,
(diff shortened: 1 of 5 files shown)


packages/qwik/src/core/render/jsx/types/jsx-qwik-attributes.ts
```diff
@@ -241,7 +241,6 @@ export interface QwikIntrinsicAttributes {
 
   /** Corresponding slot name used to project the element into. */
   'q:slot'?: string;
-  'q:shadowRoot'?: boolean;
```
PatrickJS, line 244: don't we want to keep this?

packages/qwik/src/core/components/prefetch.ts
```diff
@@ -43,7 +43,7 @@ export const PrefetchServiceWorker = (opts: {
     // allow different path and base
     resolvedOpts.path = opts.path;
   } else {
-    // base: '/'
+    // base: '/build/'
```
wmertens, line 46: @PatrickJS why?

packages/qwik/src/core/components/prefetch.ts
```diff
@@ -140,21 +144,18 @@ export const PrefetchGraph = (
...
-  const args = [
-    'graph-url',
-    resolvedOpts.base,
-    resolvedOpts.base + `q-bundle-graph-${resolvedOpts.manifestHash}.json`,
-  ]
+  const args = ['graph-url', resolvedOpts.base, `q-bundle-graph-${resolvedOpts.manifestHash}.json`]
     .map((x) => JSON.stringify(x))
     .join(',');
-  const code = `(window.qwikPrefetchSW||(window.qwikPrefetchSW=[])).push(${args})`;
+  const code = `(window.qwikPrefetchSW||(window.qwikPrefetchSW=[])).push([${args}])`;
```
wmertens, line 159: you can just stringify all of `args`

packages/qwik/src/core/components/prefetch.unit.tsx
```diff
@@ -24,15 +24,17 @@ describe('PrefetchServiceWorker', () => {
       expect(output.html).to.contain('<script nonce="1234" q:key="prefetch-service-worker">');
     });
     it('should render script with a scope', async () => {
+      // if the qwik app was isolated to /en/ folder
+      // scope will only run in /en/ pathname in the url
       const output = await renderToString(
-        <PrefetchServiceWorker base="/build/en/" scope="/en/" />,
+        <PrefetchServiceWorker base="/en/build/" scope="/en/" />,
```
wmertens, line 30: this contradicts the "service worker not in build" comment

packages/qwik/src/core/components/prefetch.ts
PatrickJS: oh yeah haha this is abusing how array is serialized

packages/qwik/src/server/prefetch-utils.ts
```diff
@@ -23,8 +23,8 @@ export function prefetchUrlsEventScript(base: string, prefetchResources: Prefetc
   const data: QPrefetchData = {
     bundles: flattenPrefetchResources(prefetchResources).map((u) => u.split('/').pop()!),
   };
-  const args = ['prefetch', base, data.bundles!].map((x) => JSON.stringify(x)).join(',');
-  return `window.qwikPrefetchSW||(window.qwikPrefetchSW=[]).push(${args});`;
+  const args = ['prefetch', base, ...data.bundles!].map((x) => JSON.stringify(x)).join(',');
+  return `(window.qwikPrefetchSW||(window.qwikPrefetchSW=[])).push([${args}]);`;
```
wmertens, line 27:
```js
const args = JSON.stringify(['prefetch', base, ...data.bundles!]);
return `window.qwikPrefetchSW||(window.qwikPrefetchSW=[]).push(${args});`;
```

packages/qwik/src/core/components/prefetch.unit.tsx
PatrickJS: well the problem is that base in this case is q:base which is /build/ not / so the / is /en/

wmertens: :no_mouth: I'm confused now

wmertens (dismissed):
LGTM

PatrickJS (approved):

PR #6587, opened 2024-06-21
fix(optimizer): make component entries more unique

wmertens:
...by adding the original filename.

This way multiple files in the same directory with the same component name won't cause problems.


thejackshelton-kunaico (approved):
Looks like wout magic. Approved

thejackshelton (approved):
LGTM! Fixes a needed issue in the optimizer

gioboa (approved):

PR #6596, opened 2024-06-24
fix(v2): fix vnode separators

Varixo:
This PR fixes vnode separators and `missing refElement` error in the docs

### DIFF
--- a/packages/qwik/src/core/v2/shared/vnode-data-types.ts
+++ b/packages/qwik/src/core/v2/shared/vnode-data-types.ts
@@ -36,10 +36,10 @@ export const VNodeDataSeparator = {
   ADVANCE_1024: /* ****** */ 43, // `+` is vNodeData separator skipping 512.
   ADVANCE_2048_CH: /* *  */ ',', // ',' is vNodeData separator skipping 1024.
   ADVANCE_2048: /* ****** */ 44, // ',' is vNodeData separator skipping 1024.
-  ADVANCE_4096_CH: /* *  */ `.`, // `.` is vNodeData separator skipping 2048.
-  ADVANCE_4096: /* ****** */ 46, // `.` is vNodeData separator skipping 2048.
-  ADVANCE_8192_CH: /* *  */ `/`, // `/` is vNodeData separator skipping 4096.
-  ADVANCE_8192: /* ****** */ 47, // `/` is vNodeData separator skipping 4096.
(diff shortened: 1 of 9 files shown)


packages/qwik/src/server/v2-ssr-container.ts
```diff
@@ -956,7 +956,7 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
     let skipCount = elementIdx - lastSerializedIdx;
     // console.log('emitVNodeSeparators', lastSerializedIdx, elementIdx, skipCount);
     while (skipCount != 0) {
-      if (skipCount >= 4096) {
```
Varixo, line 959: @mhevery I thing this is a good change, right? If skipCount is equal to 4096 then we want to execute code in the else block, right?

packages/qwik/src/core/v2/shared/vnode-data-types.ts
```diff
@@ -34,12 +34,12 @@ export const VNodeDataSeparator = {
   ADVANCE_512: /* ******* */ 42, // `*` is vNodeData separator skipping 256.
   ADVANCE_1024_CH: /* ** */ `+`, // `+` is vNodeData separator skipping 512.
   ADVANCE_1024: /* ****** */ 43, // `+` is vNodeData separator skipping 512.
-  ADVANCE_2048_CH: /* *  */ '`', // '`'` is vNodeData separator skipping 1024.
-  ADVANCE_2048: /* ****** */ 44, // '`'` is vNodeData separator skipping 1024.
+  ADVANCE_2048_CH: /* *  */ ',', // ',' is vNodeData separator skipping 1024.
+  ADVANCE_2048: /* ****** */ 44, // ',' is vNodeData separator skipping 1024.
   ADVANCE_4096_CH: /* *  */ `.`, // `.` is vNodeData separator skipping 2048.
   ADVANCE_4096: /* ****** */ 46, // `.` is vNodeData separator skipping 2048.
```
Varixo, line 40: I don't know why we are missing the `45` charcode?

mhevery: That is weird! I don't think we should be skipping anything. I think that is a bug and should be fixed.

packages/qwik/src/server/v2-ssr-container.ts
mhevery: yes, that is correct. Thank you for noticing.

mhevery (approved):

PR #6604, opened 2024-06-25
fix: disable PrefetchGraph in dev mode

gioboa:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/components/prefetch.ts
+++ b/packages/qwik/src/core/components/prefetch.ts
@@ -143,10 +143,10 @@ const PREFETCH_CODE = /*#__PURE__*/ ((
 export const PrefetchGraph = (
   opts: { base?: string; manifestHash?: string; manifestURL?: string; nonce?: string } = {}
 ) => {
-  if (isDev) {
+  const isTest = import.meta.env.TEST;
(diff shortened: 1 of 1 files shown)


packages/qwik/src/core/components/prefetch.ts
```diff
@@ -143,6 +143,13 @@ const PREFETCH_CODE = /*#__PURE__*/ ((
 export const PrefetchGraph = (
   opts: { base?: string; manifestHash?: string; manifestURL?: string; nonce?: string } = {}
 ) => {
+  const isTest = import.meta.env.TEST;
```
thejackshelton, line 146: same here, what is isTest?

thejackshelton (approved):

packages/qwik/src/core/components/prefetch.ts
gioboa: We have few tests for this component and we need to simulate the production mode even if we are testing

shairez (approved):

PR #6606, opened 2024-06-25
fix: disable PrefetchServiceWorker in dev mode

gioboa:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/components/prefetch.ts
+++ b/packages/qwik/src/core/components/prefetch.ts
@@ -30,6 +30,14 @@ export const PrefetchServiceWorker = (opts: {
   fetchBundleGraph?: boolean;
   nonce?: string;
 }): JSXNode<'script'> => {
+  const isTest = import.meta.env.TEST;
+  if (isDev && !isTest) {
+    const props = {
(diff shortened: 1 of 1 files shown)


packages/qwik/src/core/components/prefetch.ts
```diff
@@ -30,6 +30,14 @@ export const PrefetchServiceWorker = (opts: {
   fetchBundleGraph?: boolean;
   nonce?: string;
 }): JSXNode<'script'> => {
+  const isTest = import.meta.env.TEST;
```
thejackshelton, line 33: What's the point of this?

thejackshelton (approved):
LGTM other than the `isTest`

packages/qwik/src/core/components/prefetch.ts
gioboa: We have few test for this component and we need to simulate the production mode even if we are testing

shairez (approved):

PR #6613, opened 2024-06-28
feat(qwik-core): `useStylesWithScope` that introduces `:scope` pseudo class to scope styles to the component manually.

genki:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

The `useStyles$` can style entire page but it couldn't scope to components.
The `useStylesScoped$` can style scoped to the component but it couldn't style its child components.
So I added one more thing, the `useStyleWithScope$` that introduces the `:scope` pseudo class to scope styles manually.

By using this function, you can use `:scope` pseudo class to be replaced with `.${scopeId}`.

```css
:scope {
  background: red
}
```

The above css is to be like this:

```css
.⭐️5yhoz-0 {
  background: red
}
```

# Use cases and why

While the `useStylesScoped$` put the `scopeId` to every rules in the CSS, this function can select rules to put it. 
So you can scope styles to the component and its children such as `<Link>` too.

# Checklist:


genki:
Ah something went wrong in my git work. I will retry.

PR #6635, opened 2024-07-02
fix: show a warning instead of an error for rerunning tasks on a different host

dmitry-stepanenko:
This is a fix for a toggle e2e test

Instead of throwing an error we will show a warning now


Varixo:
I think we should remove "skip" from the toggle test too

Varixo (approved):

gioboa (approved):
Thanks @dmitry-stepanenko 👏🎉

PR #5883, opened 2024-02-23
fix(eslint): Allow JSXNode to serialize

genki:
# Overview

Allow the `JSXNode` to serialize.

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests / types / typos

# Description

Even though there is the `JSXNodeSerializer`, the eslint has been warning when you use the `JSXNode` across the `$` sign.
This PR solves this issue.

# Use cases and why

Fixes #5864 

# Checklist:

- [x] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] Added new tests to cover the fix / functionality

PR #6637, opened 2024-07-03
fix(eslint): Fixed lint problem about the `JSXNode`

### LINKED ISSUE #5940: [🐞] <Resource> passed as prop does not render in correct position

Qwik Rollup / Vite plugin

### Describe the bug

I was trying to pass some JSX as a prop when I encountered this bug (I couldn't use `<Slot />` because I needed to inspect the passed JSX).

If I understand correctly, any value passed as a prop should be serializable, and JSX (more specifically the `type JSXOutput = JSXNode | string | number | boolean | null | undefined | JSXOutput[];`) *is* serializable:
* [`_verifySerializable`](https://github.com/BuilderIO/qwik/blob/1995753ff587d9938dcc08515ca64693a805d14f/packages/qwik/src/core/state/common.ts#L38) explicitly lists nullish values (`null` and `undefined`), (contiguous) arrays, booleans, strings and numbers
* the [`serializers`](https://github.com/BuilderIO/qwik/blob/1995753ff587d9938dcc08515ca64693a805d14f/packages/qwik/src/core/container/serializers.ts#L533) array includes `JSXNodeSerializer` that takes care of remaining `JSXNode`s

In the following reproduction, the app component takes two copies of JSX `<Inner />` that has a `<Resource>` call inside, and passes them to `Outer`, one as a slotted child and the other as the `content` prop. `Outer` tries to render each JSX in a `<div>`.

**Expected behavior**: in the HTML tab, each JSX should be rendered in its respective `<div>`.
**Actual behavior**: the `content` JSX is rendered *outside of* its `<div>`.

This bug only happens in server-side rendering; both JSX renders correctly if I insert a `useVisibleTask$` to force client-side rendering.

Please feel free to close this issue if my understanding is wrong or I made a mistake somewhere else.

### Reproduction

https://qwik.dev/examples/introduction/hello-world/#f=7VY7DsIwDJ3ts0SNEJ1Q0h0WBhYOwFwWKlWKcneeP6TNzITE1vpb2%2B%2B53oFmHI9tUEw6mcvtfl1ewG5g2gAROqRCJUAP6KtJ%2BAuUMSUk1EWijMwlnWdZLHGqUFJ7g2FUy4lRCFuXzTV3Uf1KQUUe8rSvqupA0H%2FSMydJJYhux44Li4YY3L26Dk3cEttX9Yl91JIWFtji0HcMN4sDotCHV62FRFgvC%2B4Q%2BA36WCF7zmKAn8Ejl1WcV0jjn30%2Fy743

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz
    Memory: 1.16 GB / 7.86 GB
  npmPackages:
    @builder.io/qwik: ^1.4.5 => 1.4.5
    @builder.io/qwik-city: ^1.4.5 => 1.4.5
    undici: * => 6.6.2
    vite: ^5.0.12 => 5.1.4
```

### Additional Information

_No response_

genki:
# Overview

I have fixed the lint issue about the serialization of th JSXNode (reported #5940)
Here is the reproduction of the issue. This code crashes.
https://stackblitz.com/edit/qwik-starter-ptzf8r?file=src%2Froutes%2Findex.tsx

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

It was needed to check the symbol name if it is the `JSXNode` before checking of the callable things
in order to avoid to be checked the internal structures of JSXNode and being kicked out.

# Use cases and why

Relating #5940

# Checklist:


packages/eslint-plugin-qwik/src/validLexicalScope.ts
```diff
@@ -486,7 +489,6 @@ const ALLOWED_CLASSES = {
   Set: true,
   Map: true,
   Uint8Array: true,
-  JSXNodeImpl: true,
```
wmertens, line 489: why do you remove this?

wmertens:
can you run `pnpm api.update` so it doesn't have the api changes? Those get added by the build somewhere and we don't know why.

packages/eslint-plugin-qwik/src/validLexicalScope.ts
genki: I think the change now covers such the case.
I had added the `JSXNodeImpl` to the `ALLOWED_CLASSES` in order to fix the issue previously here:
https://github.com/QwikDev/qwik/pull/5883
But now I found the right answer to do this, so I have deleted it.

genki:
@wmertens 
I have misunderstood about the order of execution of tasks `api.update` and `build.full` that was `api.update` then `build.full`. But the right way was inverse.
Now I did it.

PR #6643, opened 2024-07-04
fix(rollup): force single output

### LINKED ISSUE #6620: [🐞] image optimizer with imagetools in cloudflare addressed wrong srcset

Qwik Runtime

### Describe the bug

In cloudflare, I see in the build logs that build directory webp images are being built correct in "build/q-[hash].webp", but in browser, developer tools says srcset ist set to "/assets/[filename]-[hash].webp".

dev and preview have no problem with it.

example here (Demo app):
https://qwik-app-9et.pages.dev/

image

image

### Reproduction

https://qwik-app-9et.pages.dev/

### Steps to reproduce

cloudflare build

### System Info

```shell
System:
    OS: Windows 11 10.0.22621
    CPU: (8) x64 Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
    Memory: 20.31 GB / 31.90 GB
  Binaries:
    Node: 20.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.5.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (126.0.2592.81)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @builder.io/qwik: ^1.5.7 => 1.5.7
    @builder.io/qwik-city: ^1.5.7 => 1.5.7
    typescript: 5.3.3 => 5.3.3
    undici: * => 6.19.2
    vite: ^5.2.10 => 5.3.1
```

### Additional Information

_No response_

gioboa:
Close #6620

# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


wmertens:
We should either fix the image plugin or completely revert the multi output PR and error when trying to set up multiple outputs.

gioboa:
@KingSora Hi, multiple outputs are working strangely.
I forced the code to use the first output and it solves the issue.
Can you verify why with an array it's not working?

---

With single output (correct)
`srcSet: '/build/q-uOW82FhY.webp 200w, /build/q-4fKdvPuU.webp 400w, .... `

with array (wrong)
`srcSet: '/assets/thunder-uOW82FhY.webp 200w, /assets/thunder-4fKdvPuU.webp 400w, ...`

---
Thanks

KingSora:
@gioboa I'm currently on vacation and wont be able to checkt that for at least another week... I also cant think of a reason why this behavior occurs

wmertens:
Maybe it's the vite image tools plugin that doesn't handle outputs well and falls back to some defaults. 

However I can't find it in the source https://github.com/JonasKruckenberg/imagetools/blob/main/packages/vite/src/index.ts#L11

wmertens:
See https://github.com/QwikDev/qwik/pull/6452

JonasKruckenberg:
I really don't have nearly enough context to answer this question, so I don't want to straight up dismiss this as something imagetools isn't at fault for. But as far as I remember, especially in a regular production build we don't do anything weird. 
All the plugin does for multiple images is calling [`emitFile`](https://github.com/JonasKruckenberg/imagetools/blob/2da03b253eba04210f55f3abd08934105646bc0d/packages/vite/src/index.ts#L171) for each output image and then generating a JS array from the resulting `__VITE_ASSET__` urls, nothing fancy really.

KingSora:
@gioboa @wmertens I've now investigated and possibly found the issue here:

The reason this is happening is how the [`viteAdapter`](https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/adapters/shared/vite/index.ts#L14) works. In particular its usages for e.g. [vercel](https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/adapters/vercel-edge/vite/index.ts#L15).

Its returning a partial vite config in the `config` hook:
```ts
build: {
  ssr: true,
  outDir,
  rollupOptions: {
    output: {
      format: 'es',
      hoistTransitiveImports: false,
    },
  },
},
```

This partial config gets merged by vite into the final config object. The final merged `build.rollupOptions.output` field then looks like this:
```ts
[
  {
    assetFileNames: 'build/q-[hash].[ext]',
    chunkFileNames: 'q-[hash].js',
    dir: '/project/.vercel/output/functions/_qwik-city.func'
  },
  { format: 'es', hoistTransitiveImports: false }
]
```

As you can see there are now two entries, and this is exactly what causes the behavior you are observing. When the `output` field is not an array the final merged `build.rollupOptions.output` field is:
```ts
{
  assetFileNames: 'build/q-[hash].[ext]',
  chunkFileNames: 'q-[hash].js',
  dir: '/project/.vercel/output/functions/_qwik-city.func',
  format: 'es', 
  hoistTransitiveImports: false
}
```

So what happens is that vite sees that the output field is an array and simplay merges the defined output object by appending it to the array. Before the output field was an object and was merged by combining the two objects.

----

This situation is a bit unfortunate and I can think of two possible solutions for this behavior:

1. The `qwikVite` plugin doesn't always return an array as output.. e.g. if the `output` field is an object the plugin won't transform it into an array with a single entry (as it is happening now)
2. The `adapters` somehow account for multiple outputs

gioboa:
Thanks for your investigation 😉 @KingSora 
We will discuss inside the core team how to tackle it

gioboa:
@JonasKruckenberg Thanks for your fast response.
@KingSora probably found the issue

gioboa:
I close this one in favour of the new one that is fixing the problem entirely.

KingSora:
@gioboa #6723 should fix the behavior while also keeping the functionality for an array of outputs.. please take a look :)

PR #6647, opened 2024-07-05
refactor: tiny adjustment for document testing

arlyxiao:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/testing/document.ts
+++ b/packages/qwik/src/testing/document.ts
@@ -20,53 +20,54 @@ export function createDocument(opts?: MockDocumentOptions) {
  * @public
  */
 export function createWindow(opts: MockDocumentOptions = {}): MockWindow {
-  const win = createDocument(opts).defaultView!;
-  return win;
+  return createDocument(opts).defaultView!;
 }
 
 export function ensureGlobals(doc: any, opts?: MockDocumentOptions) {
-  if (!doc[QWIK_DOC]) {
-    if (!doc || doc.nodeType !== 9) {
-      throw new Error(`Invalid document`);
-    }
-
-    doc[QWIK_DOC] = true;
-
-    const loc = normalizeUrl(opts?.url);
-
-    Object.defineProperty(doc, 'baseURI', {
-      get: () => loc.href,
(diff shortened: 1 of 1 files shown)


packages/qwik/src/testing/document.ts
```diff
@@ -20,53 +20,54 @@ export function createDocument(opts?: MockDocumentOptions) {
...
-  return win;
+  return createDocument(opts).defaultView!;
 }
 
 export function ensureGlobals(doc: any, opts?: MockDocumentOptions) {
-  if (!doc[QWIK_DOC]) {
-    if (!doc || doc.nodeType !== 9) {
-      throw new Error(`Invalid document`);
-    }
+  if (!doc || doc.nodeType !== 9) {
```
gioboa, line 27: We are changing the behaviour here. If doc[QWIK_DOC] we need to return doc.defaultView

gioboa (changes requested):
Thanks @arlyxiao for your contribution 👏
I added my feedback

packages/qwik/src/testing/document.ts
```diff
@@ -20,53 +20,54 @@ export function createDocument(opts?: MockDocumentOptions) {
...
+    throw new Error(`Invalid document`);
+  }
+
+  if (doc && doc[QWIK_DOC]) {
+    return doc.defaultView;
```
arlyxiao, line 32: @gioboa  Here returns doc.defaultView

gioboa: I see but we are throwing an error few line before even if doc[QWIK_DOC] is defined. In the old behaviour we were throwing only if not doc[QWIK_DOC]

arlyxiao: > I see but we are throwing an error few line before even if doc[QWIK_DOC] is defined. In the old behaviour we were throwing only if not doc[QWIK_DOC]

yep. You're right. Updated.

gioboa (approved):
Thanks @arlyxiao

PR #6649, opened 2024-07-07
fix(v2): attribute setting and CSR event registration

Varixo:
This PR fixes attribute toggling and CSR event registration


packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -592,6 +598,9 @@ export const vnode_diff = (
             HANDLER_PREFIX + ':' + scope + ':' + eventName,
             value
           );
+          if (eventName) {
+            registerQwikLoaderEvent(eventName);
```
wmertens, line 602: So we don't need to add the scope to the event name, right?

wmertens (approved):
LGTM 🙌

packages/qwik/src/core/v2/client/vnode-diff.ts
Varixo: Yes

PR #6680, opened 2024-07-13
chore: fix "STATUS-1" labels removal with workflow

dmitry-stepanenko:


gioboa (approved):
Great @dmitry-stepanenko

PR #6687, opened 2024-07-15
fix(qwik-city): Prevent unwanted navigation caused by internal search params

### LINKED ISSUE #6660: [🐞] Breaking change introduced in 1.7.0 - `qaction` in url is appended using `routeAction` and `globalAction`

Qwik Runtime

### Describe the bug

1.6.0 didn't have this issue

### Reproduction

> https://stackblitz.com/edit/qwik-starter-cidktr

Thanks @blecorre-mnp for the repro

### Steps to reproduce

> You can see that once you click on the submit buttons, you'll have the ```&qaction=...``` appended to the url.

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 96.05 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.15.0 - ~/.nvm/versions/node/v20.15.0/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v20.15.0/bin/npm
    pnpm: 9.4.0 - /opt/homebrew/bin/pnpm
    bun: 1.1.18 - /opt/homebrew/bin/bun
  Browsers:
    Safari: 17.5
  npmPackages:
    @builder.io/qwik: ^1.7.0 => 1.7.0
    @builder.io/qwik-city: ^1.7.0 => 1.7.0
    typescript: 5.5.3 => 5.5.3
    undici: ^6.19.2 => 6.19.2
    vite: ^5.3.3 => 5.3.3
```

### Additional Information

_No response_

octet-stream:
# Overview

Fixes #6660 issue appeared after #6410 fix was released. The later works as expected, but it led to unwanted redirects caused by internal URLSearchParams in response URL.

# Changes

- [x] Remove `qaction` from URLSearch param in `QwikCityProvider` component 

# What is it?

- [ ] Feature / enhancement
- [x] Bug

# Description

This PR removes `qaction` from URLSearchParams in the URL returned from action response. I honestly don't think this is a perfect solution, but it'll do as quick fix for now. Please, let me know if I can make this better anyhow.

I should also mention that this is a SPA navigation fix. When JS is disabled, or browser navigation is triggered, the `qaction` parameter will still show up in the address bar after the action is fired, because this is how browsers work with forms. I think this is also something we should consider to change, so this behavior world be the same for SPA and MPA navigation. It can be solved by redirecting browser to a URL without internal params after action is fired.

# Use cases and why

Expected behavior: Sending forms should not cause to navigation by internal search params, such as `qaction`

Actual behavior (since 1.7.0): When form is sent via JS the URL in address bar updates with `qaction` added to search params every time the form is sent (this parameter is added multiple times).

# Checklist:


shairez:
Thanks for the PR @octet-stream ! great job!

So the part the is adding the `QACTION_KEY` is here - 
https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/runtime/src/utils.ts#L38

@wmertens @gioboa I'm not sure why we need to add it to the client side url though... 

Because if JS is disabled, the form will get submitted by the `actionPath` which gets set here - 
https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/runtime/src/server-functions.ts#L62

and then being used here for "no JS" cases - 
https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/runtime/src/form-component.tsx#L86

Maybe I'm missing something but if not, maybe we can just remove it because it isn't needed?

----

Regardless, we need to add another issue to remember to solve the `?qaction=hash` chaining on multiple submission in "no JS" mode to the same page.

octet-stream:
> Regardless, we need to add another issue to remember to solve the ?qaction=hash chaining on multiple submission in "no JS" mode to the same page.

I don't see `qaction` being chained when JS is disabled if that's what you mean. The issue only occurs with JS.

But yeah, if we can just not add it there, then this should be a better solution. Thank you for additional information!

octet-stream:
Though I see this clientUrl being used in different places within the loadClientData:

1. https://github.com/QwikDev/qwik/blob/563ddc512a9dec38194f10914ad5988f523bd71d/packages/qwik-city/runtime/src/use-endpoint.ts#L35
2. https://github.com/QwikDev/qwik/blob/563ddc512a9dec38194f10914ad5988f523bd71d/packages/qwik-city/runtime/src/use-endpoint.ts#L51
3. https://github.com/QwikDev/qwik/blob/563ddc512a9dec38194f10914ad5988f523bd71d/packages/qwik-city/runtime/src/use-endpoint.ts#L73
4. https://github.com/QwikDev/qwik/blob/563ddc512a9dec38194f10914ad5988f523bd71d/packages/qwik-city/runtime/src/use-endpoint.ts#L79

I'm not sure what these supposed to do `exactly` (I mean when exactly this code runs), but I see that it sets a url for fetch and sets/gets/deletes value from `CLIENT_DATA_CACHE`.

wmertens:
closing in favor of #6694 which fixes the problem at the source. I did cherry pick the tests though, thanks!

PR #6694, opened 2024-07-16
fix(qwik-city): strip internal parameters for compare

### LINKED ISSUE #6660: [🐞] Breaking change introduced in 1.7.0 - `qaction` in url is appended using `routeAction` and `globalAction`

### LINKED ISSUE #6677: [🐞] in dev mode, ignore dist/

Qwik Runtime

### Describe the bug

When running in dev mode, and there's a stale dist/ folder with SSG files in there, those files get served.

In dev, only public/ should be served.

### Reproduction

not yet

### Steps to reproduce

drive by issue creation, no repro, could be just an issue with this proprietary repo

### System Info

```shell
macos with latest qwik
```

### Additional Information

_No response_

wmertens:
Followup for #6410 and #6687
Fixes #6660
Fixes #6677

### DIFF
--- a/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts
+++ b/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts
@@ -144,15 +144,15 @@ const _resolveRequestHandlers = (
     }
 
     if (collectActions) {
-      const loaders = Object.values(routeModule).filter((e) =>
-        checkBrand(e, 'server_loader')
-      ) as LoaderInternal[];
-      routeLoaders.push(...loaders);
-
-      const actions = Object.values(routeModule).filter((e) =>
-        checkBrand(e, 'server_action')
-      ) as ActionInternal[];
-      routeActions.push(...actions);
+      for (const module of Object.values(routeModule)) {
+        if (typeof module === 'function') {
+          if (module.__brand === 'server_loader') {
+            routeLoaders.push(module as LoaderInternal);
+          } else if (module.__brand === 'server_action') {
+            routeActions.push(module as ActionInternal);
+          }
+        }
+      }
     }
   }
 };
@@ -407,7 +407,9 @@ export function getPathname(url: URL, trailingSlash: boolean | undefined) {
       url.pathname = url.pathname.slice(0, -1);
     }
   }
-  return url.toString().substring(url.origin.length);
+  // strip internal search params
+  const search = url.search.slice(1).replaceAll(/&?q(action|data|func)=[^&]+/g, '');
+  return `${url.pathname}${search ? `?${search}` : ''}${url.hash}`;
 }
 
 export const encoder = /*#__PURE__*/ new TextEncoder();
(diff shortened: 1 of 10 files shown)


shairez:
seems like a duplicate of #6687 no?

btw did you see this?
https://github.com/QwikDev/qwik/pull/6687#issuecomment-2229606481

gioboa (approved):
That's great 👏 thanks @wmertens

packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -407,7 +407,9 @@ export function getPathname(url: URL, trailingSlash: boolean | undefined) {
       url.pathname = url.pathname.slice(0, -1);
     }
   }
-  return url.toString().substring(url.origin.length);
+  // strip internal search params
+  const search = url.search.slice(1).replaceAll(/&?q(action|data|func)=[^&]+/g, '');
```
octet-stream, line 411: I would still suggest using URLSearchParams and constants instead. But they are in runtime package.

Something like this:

```ts
[QACTION_KEY, QFN_KEY, QDATA_KEY].forEach(key => url.searchParams.delete(key))
```

The browsers and other runtimes have API that makes it easy to work with search params, no need to wrestle with strings.

wmertens: I first had this, but even when it's fully supported it will change the search string a bit. That's why the test doesn't have `=` for the second parameter.

When you remove searchParams, it will convert the rest to search parameters and `?myquery` becomes `?myquery=`. Since we're a framework I want to avoid messing with the URL whenever we can.

shirotech:
Very nice, hope this can be released soon. Have been waiting for a specific fix for awhile in anticipation. Great work team!

PR #6704, opened 2024-07-18
fix(changesets): version monorepo root

wmertens:
the root version is used in several places and is also useful for humans

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,4 +1,5 @@
 packages:
+  - '.'
   - 'packages/*'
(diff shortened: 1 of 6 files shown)


pnpm-workspace.yaml
```diff
@@ -1,4 +1,5 @@
 packages:
+  - '.'
```
shairez, line 2: Did you verify what you see when you run `pnpm change` after this modification?

wmertens: no but we can see what it will do in the changeset PR

shairez (approved):

PR #6706, opened 2024-07-18
fix(dev): docs.dev playground + small things

wmertens:
- allow HMR for `entry.ts` routes in dev
- speed up dev docs build by only building `repl-sw.js`, and only when needed
- speed up search results by doing SPA navigation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -731,15 +731,18 @@ jobs:
       - lint-package
       - build-docs
       - build-insights
-    if: |
-      always() &&
-      (needs.lint-package.result == 'success' || needs.lint-package.result == 'skipped') &&
-      (needs.test-e2e.result == 'success' || needs.test-e2e.result == 'skipped') &&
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -503,8 +503,32 @@ jobs:
...
+        uses: actions/cache/restore@v4
+        with:
+          key: ${{ needs.changes.outputs.hash-docs }}
+          path: |
+            packages/docs/dist
+            packages/docs/server
+
+      # not the official version, so be careful when updating
+      - name: Deploy to Cloudflare Pages
         uses: AdrianGonz97/refined-cf-pages-action@c7b767eef599f127d8ce2388d4ed2270d3b750af
```
Varixo, line 532: out of curiosity - why we are not using the official one?

Varixo (approved):

.github/workflows/ci.yml
wmertens: the official one doesn't post the build result in a message and hasn't been updated in two years

PR #6711, opened 2024-07-19
fix(v2): use on events fixes

Varixo:
This PR fixes few edge cases for the `useOn` hook

### DIFF
--- a/packages/qwik/src/core/use/use-on.ts
+++ b/packages/qwik/src/core/use/use-on.ts
@@ -1,7 +1,4 @@
-import { assertQrl } from '../qrl/qrl-class';
 import type { QRL } from '../qrl/qrl.public';
-import { getContext, HOST_FLAG_NEED_ATTACH_LISTENER } from '../state/context';
-import { type Listener, normalizeOnProp } from '../state/listeners';
 import { useInvokeContext } from './use-core';
 import { type KnownEventNames } from '../render/jsx/types/jsx-qwik-events';
 import type {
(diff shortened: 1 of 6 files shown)


packages/qwik/src/core/v2/shared/component-execution.ts
```diff
@@ -153,6 +179,10 @@ function findFirstStringJSX(jsx: JSXOutput): JSXNode<string> | null {
       queue.push(jsx.children);
     } else if (Array.isArray(jsx)) {
       queue.push(...jsx);
+    } else if (isPromise(jsx)) {
+      return maybeThen<JSXOutput, JSXNode<string> | null>(jsx, (jsx) => findFirstStringJSX(jsx));
+    } else if (isSignal(jsx)) {
+      return findFirstStringJSX(untrack(() => jsx.value as JSXOutput));
```
Varixo, line 185: This is the important change. We could get the jsx from a promise or a signal value, so we need to check if the returned value is a JSXNode.

packages/qwik/src/core/v2/shared/component-execution.ts
```diff
@@ -104,45 +118,65 @@ export const executeComponent2 = (
...
+            if (jsxElement) {
+              addUseOnEvent(jsxElement, 'document:onQinit$', useOnEvents[key]);
+            }
+          } else if (isDev) {
+            logWarn(
+              'You are trying to add an event using `useOn` hook, ' +
+                'but a node to which you can add an event is not found. ' +
+                'Please make sure that the component has a valid element node. '
+            );
+          }
```
Varixo, line 154: I have added a warning message for dev mode if someone is trying to use `useOn` hook on the invisible component. We can't handle that, because we don't have any element node to add an event to.

packages/qwik/src/core/use/use-on.ts
```diff
@@ -121,43 +108,64 @@ const createEventName2 = (
...
+    if (!events) {
+      onMap![eventName] = events = [];
+    }
+    events.push(eventQrl);
+  };
+
+  return {
+    isAdded: addedFlags[seqIdx],
+    addEvent,
+  };
```
Varixo, line 168: Does anyone has any better idea?

packages/qwik/src/core/use/use-on.ts
```diff
@@ -121,43 +108,64 @@ const createEventName2 = (
...
 };
```
Varixo, line 169: Does anyone has any better idea?

mhevery (approved):

PR #6722, opened 2024-07-25
fix(v2): qwik react

Varixo:
This PR should fix most problems with qwik react in v2


packages/qwik/src/core/component/component.public.ts
```diff
@@ -186,25 +181,7 @@ export const componentQrl = <PROPS extends Record<any, any>>(
...
-        props,
-        [OnRenderProp]: componentQrl,
-        [QSlot]: props[QSlot],
-      },
-      null,
-      props.children,
-      flags,
-      finalKey
-    ) as any;
-  }
```
Varixo, line 207: for some reason for the qwik-react here is returning error. This is not needed in v2, so I removed it

packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -232,6 +232,7 @@ export const vnode_diff = (
           }
         } else if (jsxValue === SkipRender) {
           // do nothing, we are skipping this node
+          journal = [];
```
Varixo, line 235: is there any better way to ignore journal? @mhevery

wmertens (approved):
Great! BTW qwik-react will need a version bump for v2 compatibility, might as well call it 1.0.0 now

LGTM

mhevery (approved):

PR #6737, opened 2024-07-29
fix: Remove Cannot resolve symbol from the integration tests in v2

JerryWu1234:
# Overview

https://github.com/orgs/QwikDev/projects/1/views/1?pane=issue&itemId=68407338

# What is it?

- [x] Feature / enhancement
- [x] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:


Varixo (changes requested):
This is not the correct fix. You shoudn't remove the error message from the code. Instead you need to find why this message is appearing and do something inside the tests

JerryWu1234:
image

>This is not the correct fix. You shoudn't remove the error message from the code. Instead you need to find why this message is appearing and do something inside the tests

Sorry, I got it wrong.

I finish it now. I tweak the unit tests

packages/qwik/src/core/v2/tests/render-api.spec.tsx
```diff
@@ -46,15 +46,16 @@ const defaultManifest: QwikManifest = {
...
+    s_click: 's_click.js',
+  },
   version: '1',
 };
 
 const defaultCounterManifest: QwikManifest = {
   ...defaultManifest,
-  mapping: {
-    click: 'click.js',
-  },
```
wmertens, line 57: If you're moving everything into the default manifest, the countermanifest is no longer useful, so you should remove it.

I think it's fine to do it, the manifest may contain entries that aren't used.

wmertens:
Please fix the formatting. I recommend you enable format-on-save in your editor, with Prettier support enabled.

wmertens (approved):
LGTM

PR #6739, opened 2024-07-29
ci: made changesets work with dist tag

shairez:
This PR reverts the default "version setting" behavior to add the `dev-timestamp` suffix to the packages' version number.

But, if the CI identifies that the latest "push" (merge) removed the md files from the `.changesets` folder, and there are no other .md files there, it'll set the `disttag` to `latest`.

This will make sure Qwik doesn't add the `dev-timestamp` suffix but instead leave it intact


wmertens (approved):
LGTM

PR #6745, opened 2024-07-30
chore: added pretty-quick and a pre-commit hook

shairez:
# Overview

The end of all CI linting errors

Every local pnpm install will also install a pre-commit hook that will run `pretty-quick --staged` which will fix prettier errors only for staged files.

oh happy days!  🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,9 @@
       ]
     }
   },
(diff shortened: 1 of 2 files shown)


gioboa (approved):
@shairez 👏 thanks

PR #4998, opened 2023-08-24
feat: add Valibot support with `valibot$` to qwik-city

brandonpittman:
# Overview

Adds Valibot support with a `valibot$` function similarly to how Zod is supported by `zod$`.

Fixes #4924 

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Docs / tests / types / typos

# Description

Requires: 

- `valibot`

# Use cases and why

So you can use Valibot to validate your actions

# Checklist:

- [x] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [ ] Added new tests to cover the fix / functionality

PR #6720, opened 2024-07-24
fix(server-functions): omitted any from fieldErrors

tzdesign:
Typescript crashes because any could be infinite

fix #6719

# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Docs / tests / types / typos

# Description

See

- #6719 

# Checklist:

- [x] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] Added new tests to cover the fix / functionality

Issue #5784, opened 2024-01-27
[🐞] Differentiate input and output types in actions for zod validator


octet-stream:
### Which component is affected?

Qwik City (routing)

### Describe the bug

Sometimes input and output types in `zod` schema may differentiate, but actions ignore this on type-level. From the looks of [actions code](https://github.com/BuilderIO/qwik/blob/20af284eba25b7470844bb6e44bf0176c07ac18b/packages/qwik-city/runtime/src/types.ts#L349), I assume that you use `zod.infer` to get the output type for given schema, then pass it both to the action input and output. I suggest you replace `zod.infer` with `zod.input` and `zod.output` (and do the same for [upcoming `valibot` support](https://github.com/BuilderIO/qwik/pull/4998), because it also has different utilities to infer input and output types: `Input` and `Output`), and also treat actions `Input` and `Output` as tho different type parameters: `action.submit` should take `Input` type, and action implementation should get `Output` type.

### Reproduction

https://stackblitz.com/edit/qwik-starter-pyfl6d?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

1. Open the reproduction link, then open `src/routes/index.tsx` file and look at the types output for the first argument of `projectCreate.submit` method (just place cursor over the method). It shows `isPublic` property as required.
2. Open `src/zod/ProjectCreateInput.ts` file and look at the type output for `IProjectCreateInput` and `OProjectCreateInput` types. Notice how they are different: `isPublic` is **optional** for *input* type, while being **required** for the *output*

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 45.17 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.3.0 - ~/Library/Caches/fnm_multishells/9803_1704038600457/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - ~/Library/Caches/fnm_multishells/9803_1704038600457/bin/npm
    pnpm: 8.14.3 - /opt/homebrew/bin/pnpm
    bun: 1.0.23 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 120.0.6099.234
    Safari: 17.2.1
  npmPackages:
    @builder.io/qwik: 1.4.2 => 1.4.2 
    @builder.io/qwik-auth: 0.1.3 => 0.1.3 
[...]

PR #6752, opened 2024-08-01
feat: add `valibot$` validator and fix types of `zod$` implementation

fabian-hiller:
# Overview

Adds Valibot support with a `valibot$` validator similarly to how Zod is supported with `zod$`.

Related #4998, Credits @brandonpittman

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

The current implementation of `zod$` used type casting. I have fixed the types and solved the problems that led to the current type-casting solution. I also added support for my library Valibot with a `valibot$` validator that works similarly to the `zod$` validator.

# Checklist:


fabian-hiller:
This is a draft PR because it is missing docs and probably a few other things. I prefer to get a review and confirmation that this will be merged before finalizing this PR.

wmertens:
Can you run `pnpm api.update` so we can see what the changes are?

wmertens:
LGTM - we could add it as an alpha preview feature in v1.8 IMHO, saying that we're looking for feedback

fabian-hiller:
Thank you for your review!

> Can you run `pnpm api.update` so we can see what the changes are?

Will do this tomorrow.

> we could add it as an alpha preview feature in v1.8 IMHO, saying that we're looking for feedback

Not sure if this is necessary but we can do it.

fabian-hiller:
How quickly can this be merged? Should I continue with the documentation?

packages/qwik-city/src/runtime/src/server-functions.ts
```diff
@@ -224,53 +229,73 @@ export const validatorQrl = ((
...
+      const hasArrayType = issue.path.some((path) => path.type === 'array');
+      if (hasArrayType) {
+        const keySuffix = issue.expected === 'Array' ? '[]' : '';
+        const key =
+          issue.path
+            .map((item) => (item.type === 'array' ? '*' : item.key))
+            .join('.')
+            .replace(/\.\*/g, '[]') + keySuffix;
+        acc[key] = acc[key] || [];
+        if (Array.isArray(acc[key])) {
```
wmertens, line 244: am I reading this correctly that it won't tell you which array item has the issue? Is that not possible to do?

wmertens (commented):
I think we can merge this as a new feature provided we're sure that the error output will not change later.

Then we'd have to have some tests that verify the error shape.

wmertens:
I think it straightforward enough to merge asap, it could be part of 1.8.0. Documentation would be great, please add some tests though, see https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/core/render/jsx/types/jsx-types.unit.tsx for examples on how to test types, as well as real tests for the error shape.
Could you also `pnpm change add` to add a changeset?

packages/qwik-city/package.json
```diff
@@ -9,6 +9,7 @@
     "source-map": "^0.7.4",
     "svgo": "^3.3",
     "undici": "*",
+    "valibot": "0.37.0",
```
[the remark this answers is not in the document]
fabian-hiller: I used a fixed version number because the same approach was taken with Zod. As long as there are no breaking changes, it doesn't matter. So we can change it to `"^0.33.0"` or `">=0.33.0 <1"` and update to v1 when it is available. Please let me know how to proceed.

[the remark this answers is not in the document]
wmertens: With zod it actually needs the fixed version because the minor update breaks the types :( With valibot I hope that won't happen :)

packages/qwik-city/src/runtime/src/server-functions.ts
fabian-hiller: That's true and I think it's bad, but I just followed the Zod implementation because changing it would lead to breaking changes due to the `ValidatorErrorKeyDotNotation` type.

fabian-hiller:
Will do. Is it true that the Zod implementation is not tested, or am I just not finding the tests?

packages/qwik-city/src/runtime/src/server-functions.ts
wmertens: Hmm, could we not have a different error signature for valibot? Make it a generic type?

wmertens:
> Will do. Is it true that the Zod implementation is not tested, or am I just not finding the tests?

indeed no tests currently, but added in #6720

packages/qwik-city/src/runtime/src/server-functions.ts
fabian-hiller: We could do this, but it would lead to a major rewrite of the types in `types.ts`. Also, switching from `zod$` to `valibot$` gets confusing as the error implementation is inconsistent. I have heard that there are plans to rewrite Qwik City. This is something we could tackle as part of that.

fabian-hiller: I recommend choosing a unified format without edge cases, or making it custom and just building on top of the schema library's error formatting. Both Zod and Valibot contain a `flatten` function/method.

fabian-hiller:
@wmertens does my last `pnpm change add` commit looks good to you or is something missing?

packages/qwik-city/package.json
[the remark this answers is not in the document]
fabian-hiller: I don't expect breaking changes. Should I change it to `>=0.33.0 <2`?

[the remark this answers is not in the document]
wmertens: Yes please

packages/qwik-city/src/runtime/src/server-functions.ts
wmertens: I prefer a per-validator format. That way if people want to roll their own they can, and valibot can already provide nicer errors.

wmertens:
> @wmertens does my last `pnpm change add` commit looks good to you or is something missing?

Could you write it as a report of things changed instead? So not "I changed X to Y" but instead "X is now Y".

packages/qwik-city/src/runtime/src/server-functions.ts
fabian-hiller: I agree, but how should I proceed with this PR? I don't think we should break the current `zod$` implementation. Also, it might be complicated to change this only for Valibot because we probably need to change/break a few types.

wmertens: But if each validator has its own return type, we're not breaking zod$ are we?

fabian-hiller: The current `zod$` implementation used generic names for types like `GetValidatorType` and `TypedDataValidator`. We could of course add `GetValibotValidatorType` and `TypedValibotDataValidator` and add new overload signatures only for Valibot, but that would make the code even more complicated.

Since `zod$` does not use a Zod-specific return type, I just "plugged" the `valibot$` implementation into the `zod$` one and improved a few things. It might be easier to stick with my current solution for now and rewrite it at some point.

wmertens: But if we do it now, it's a new feature, and if we do it later, it's a breaking feature, no?

fabian-hiller: Yes, that's true, but we'll probably break it for `zod$` at some point anyway.

wmertens: If we have this generic per-validator API we wouldn't have to, we'd just make zod2$.

fabian-hiller: That's true, but I'm not sure we should increase the number for each breaking change. In the end, I'm fine with either solution. So feel free to make a final decision, and I will try my best to implement it.

fabian-hiller: Another breaking change we should probably consider in the long run is to remove the export of Zod's `z`. I am not sure if such a library should be "baked" into Qwik City. It's probably better to just use the version of Zod that the user has deliberately installed. But that's another topic.

tzdesign:
@wmertens @fabian-hiller is there something I can help with?

I need this isAny check in the types to upgrade qwik. We still hang on 1.5.6 with this neat JSX security issue.

fabian-hiller:
Feel free to take a look at the currently unresolved error shape implementation and share your thoughts on how to proceed: https://github.com/QwikDev/qwik/pull/6752#pullrequestreview-2215059841

packages/qwik-city/src/runtime/src/server-functions.ts
maiieul: If we want to strictly follow semver we can create zod2$ and in V2 deprecate zod2$ and use zod2$'s code for zod$. That way we don't have to do a breaking change for valibot$ and it can start to share code with zod2$ (+ the benefits for users of zod2$) right now.

But if it's just easier for now to go with zod$ and valibot$ that's good too. Valibot users are quite bleeding-edgy so they shouldn't complain if there's a breaking change for valibot$ in the future.

I think you can take the decision @fabian-hiller since you have more context. Both are fine imo. Easier and cleaner code might be more important than another optimization atm.

> Another breaking change we should probably consider in the long run is to remove the export of Zod's z
Would be good to have those zod breaking changes in the same release. Could you create an issue so that we don't forget?

brandonpittman: > But if it's just easier for now to go with zod$ and valibot$ that's good too. Valibot users are quite bleeding-edgy so they shouldn't complain if there's a breaking change for valibot$ in the future.

Let’s not adopt a  breaking change is okay mentality from the get-go. The recent Valibot API change was a not-so-small problem in our apps at work recently.

octet-stream:
Looks like this PR fixes type inference for zod validator, and if so, then it will solve this issue: https://github.com/QwikDev/qwik/issues/5784

packages/qwik-city/src/runtime/src/server-functions.ts
```diff
@@ -224,53 +229,73 @@ export const validatorQrl = ((
...
 
+const flattenValibotIssues = (issues: v.GenericIssue[]) => {
+  return issues.reduce<Record<string, string | string[]>>((acc, issue) => {
+    if (issue.path) {
+      const hasArrayType = issue.path.some((path) => path.type === 'array');
+      if (hasArrayType) {
+        const keySuffix = issue.expected === 'Array' ? '[]' : '';
+        const key =
+          issue.path
+            .map((item) => (item.type === 'array' ? '*' : item.key))
```
tzdesign, line 240: @fabian-hiller I don't know much about valibot, what is that key? Is it a wrong valibot version, that it fails?

Sorry to get back to you guys so late. We had 10 hour shifts fixing the qwik-shop.

tzdesign (changes requested):
@fabian-hiller can you fix the key thing, I would than check it out in our large code base again.

packages/qwik-city/src/runtime/src/server-functions.ts
fabian-hiller: Which version of Valibot do you use? We will hit v1 RC with a stable API soon.

fabian-hiller:
I will try to go over everything in the next few days and share my thoughts on how to proceed.

packages/qwik-city/src/runtime/src/server-functions.ts
tzdesign: @fabian-hiller none. This might be the problem?

fabian-hiller: I am not an expert on the Qwik repo and cannot help with that. I don't think this "problem" is related to this PR.

fabian-hiller: Okay. I think I know what you mean now. I will take a look.

fabian-hiller:
@wmertens I have thought about this PR and think it is best if we mark the new `valibot$` adapter as "alpha" and merge it as is. The main reason is that this PR improves many types and fixes some problems with the current `zod$` implementation without introducing breaking changes. If we were to start rewriting everything to support the default error format of the validator being used, this PR would probably never get merged as I have less time at the moment. After merging, someone else could change the error format and figure out how to rewrite the types.

fabian-hiller:
@wmertens I am not sure what to do with the tests. This is because the type `ValidatorErrorType` does not match the implementation of `flattenZodIssues` in edge cases if I understand the code correctly. Before investing a day of work in fixing the types and writing meaningful tests, I would rather not write tests, merge this PR as is and create a new PR for Qwik City v2 that fixes and changes the implementation of `zod$` and `valibot$` by using the native error format of the validation library. Otherwise, we will be forced to write the tests twice.

fabian-hiller:
My current recommendation is to review and merge this PR as is. I plan to create another PR for Qwik City v2 that will change the error formatting to the native output, further improve the types, and add unit and type tests.

wmertens (approved):
:+1: :cowboy_hat_face:  :+1:

brandonpittman:
Happy to report this is working as I had hoped it would last year. Thank you @fabian-hiller for seeing it over the finish line! 🥳

brandonpittman:
This is working really well. When's it going to be out of _experimental_?

fabian-hiller:
I am not sure it will ever be. Maybe a `schema$` function build on top of the Standard Schema interface I am working on with Colin, David and others is the way to go for Qwik.

packages/qwik-city/package.json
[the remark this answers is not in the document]
shairez: @fabian-hiller @wmertens apparently there were some changes done for 1.0.0-alpha that caused one of my projects to fail using `globalAction$` because it was upgraded from an older qwik-city version but kept using the older version because it was in the range.

I think we need to get to the standard-schema solution asap to make zod / valibot optional

Issue #6756, opened 2024-08-02
[✨] Add ability to set build.client and build.types command


azat-io:
### Is your feature request related to a problem?

Hello!

I have a project that contains some library code, and I also have documentation for the project stored in the same repository.

Currently, my scripts look like this:

```json
{
  "scripts": {
    "docs:dev": "vite --open --mode ssr --config vite.config.docs.ts",
    "docs:build": "qwik build",
    "build.client": "vite build --config vite.config.docs.ts",
    "build.types": "tsc --incremental --noEmit",
    "docs:debug": "node --inspect-brk ./node_modules/vite/bin/vite.js --mode ssr --force",
    "docs:preview": "qwik build preview && vite preview --open --config vite.config.docs.ts",
    "test:format": "prettier --check \"**/*.{js,ts,jsx,tsx,css,json,md,yml}\"",
    "test:css": "stylelint **/*.css",
    "test:js": "eslint \"**/*.{js,ts,tsx}\"",
    "test:types": "tsc --noEmit --pretty",
    "test:unit": "vitest --run --coverage",
    "test": "pnpm run /^test:/"
  }
}
```

Thus, anything that refers to documentation in scripts starts with the prefix "docs:".

I would like to be able to define script names for the client build and types myself.

### Describe the solution you'd like

```json
{
    "docs:build": "qwik build --client-script docs:client --types-script docs:types",
    "docs.client": "vite build --config vite.config.docs.ts",
    "docs.types": "tsc --incremental --noEmit",
}
```

### Describe alternatives you've considered

N/A.

wmertens:
Yes sounds good. PR welcome, the code is under qwik/src/cli/

thasmo:
It would be nice if build scripts can be run independently so it can be even more flexible; i.e. other tools (e.g. turborepo, etc.) can be used to control the build cycle.

`qwik build-client`, `qwik build-types` or similar.

azat-io:
I made a temporary solution that allows you to set the command name:
#6758

PR #6758, opened 2024-08-03
feat: add ability to set cli commands

azat-io:
# Overview

#6756

# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Add the ability to specify a command name for the Qwik application build.

# Use cases and why

This can be useful if the repository contains code for more than just the site.

For example, library code and site code with documentation. It would be convenient to separate commands by name.

# Checklist:

### DIFF
--- a/.changeset/beige-pigs-joke.md
+++ b/.changeset/beige-pigs-joke.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': minor
+---
+
+Add ability to set CLI commands.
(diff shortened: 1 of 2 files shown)


wmertens:
I'm not convinced - you still have to call the qwik script with the paramaters and have an npm script defined.

Can you give a real scenario and then we can brainstorm what the ideal CLI API for that would be?

wmertens:
Hi @azat-io, I propose you create an RFC discussion so we can flesh out what it should look like, see https://github.com/QwikDev/qwik-evolution?tab=readme-ov-file#qwik-rfc-process

In the meantime I'll close this PR but once we agree on how to proceed we can continue on this

PR #6761, opened 2024-08-04
fix(vite): ignore missing qrl segments

wmertens:
It's ok for a qrl export to be missing when it got tree shaken


packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -746,7 +746,14 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
         .map((mod) => mod.hook)
         .filter((h) => !!h) as HookAnalysis[];
 
-      const manifest = generateManifestFromBundles(path, hooks, injections, rollupBundle, opts);
+      const manifest = generateManifestFromBundles(
+        path,
+        hooks,
```
shairez, line 751: I would rename this to `segments` as well

wmertens:
@shairez done, went all the way finally...

shairez (approved):

PR #6763, opened 2024-08-04
feat(v2): new server$, actions and loaders serialization

Varixo:
Change the old serialization to the new v2 serialization for the `server$`, actions and loaders


wmertens (approved):
LGTM

PR #6768, opened 2024-08-05
test(docs): see if it builds in pr

wmertens:
do not merge


wmertens:
:tada: it built the docs

PR #6773, opened 2024-08-07
feat: add preserve modules to lib starter

thejackshelton:
# Overview

Adds preserve modules to the library starters. It also makes the entryname apparent in the build, otherwise you would have:
`index.qwik2.mjs` and so forth.

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/.changeset/wicked-foxes-behave.md
+++ b/.changeset/wicked-foxes-behave.md
@@ -2,4 +2,4 @@
 'create-qwik': patch
 ---
 
-feat: improves library bundling / tree-shaking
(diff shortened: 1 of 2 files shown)


gioboa (dismissed):
@thejackshelton yessss :+1:

wmertens:
Needs a changeset with why

shairez (changes requested):
Like Wout said, this needs a changeset

thejackshelton:
Good catch will add one here soon

shairez (approved):

PR #6778, opened 2024-08-08
fix: a bug 4956

JerryWu1234:
# Overview

# What is it?

- [ ] Feature / enhancement
- [ ] Bug

# Description

# Use cases and why

- 1. One use case
- 2. Another use case

# Checklist:

### DIFF
--- a/packages/qwik/src/core/container/serializers.ts
+++ b/packages/qwik/src/core/container/serializers.ts
@@ -39,6 +39,7 @@ import { assertString, assertTrue } from '../error/assert';
 import { Fragment, JSXNodeImpl, isJSXNode } from '../render/jsx/jsx-runtime';
 import type { JSXNode } from '@builder.io/qwik/jsx-runtime';
 import { Slot } from '../render/jsx/slot.public';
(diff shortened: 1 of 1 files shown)


packages/qwik/src/core/container/serializers.ts
```diff
@@ -193,7 +194,12 @@ const URLSerializer = /*#__PURE__*/ serializer<URL>({
   $prefix$: '\u0005',
   $test$: (v) => v instanceof URL,
   $serialize$: (obj) => obj.href,
-  $prepare$: (data) => new URL(data),
+  $prepare$: (data) => {
+    if (isBrowser && !isDev) {
+      return new URL(data + document.location.search);
```
wmertens, line 199: I don't think this is the solution, now you're adding the location to all URLs.

Furthermore, containers only get serialized on the server, so this code doesn't do anything.

Can you add a test first and then try to fix it?

wmertens (changes requested):

JerryWu1234:
have already had tests

wmertens (approved):
LGTM

PR #6797, opened 2024-08-15
fix(qwik-city): Avoided unexpected caching for dynamic q-data

### LINKED ISSUE #6789: [🐞] SPA reload randomly fails to update the values loaded by the `routeLoader$`

Qwik City (routing)

### Describe the bug

I am making app that uses `routeLoader$` for loading dynamic data.
I found the SPA reload sometimes fails to update the data like this.

https://github.com/user-attachments/assets/a88f5efb-625d-4b67-8c1f-c87978e3cc29

Note that the 3rd reload (around 0:07 at the video) trigger the call of the route loader and logged the count at server side but doesn't change the values in client side.
The next click updates the values but doesn't trigger the call.

I expected the each reload can update the value.

### Reproduction

https://github.com/genki/qtest

### Steps to reproduce

Please clone the repo above and
```sh
pnpm i
pnpm run dev
```
then open `http://localhost:5173/`

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 117.77 MB / 24.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.12.1 - /opt/homebrew/opt/node@20/bin/node
    npm: 10.5.0 - /opt/homebrew/opt/node@20/bin/npm
    pnpm: 9.6.0 - ~/Library/pnpm/pnpm
    bun: 1.1.21 - ~/.bun/bin/bun
  Browsers:
    Chrome: 127.0.6533.101
    Safari: 17.5
  npmPackages:
    @builder.io/qwik: ^1.8.0 => 1.8.0 
    @builder.io/qwik-city: ^1.8.0 => 1.8.0 
    typescript: 5.4.5 => 5.4.5 
    undici: * => 6.19.5 
    vite: ^5.2.10 => 5.3.5
```

### Additional Information

Interestingly, this behaviour doesn't occur at the stackbliz.
https://stackblitz.com/edit/github-ahtxgw?file=src%2Froutes%2Findex.tsx

genki:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Infra

# Description

Inside the `loadClientData`, the `fetch()` had been called without the cache control options even if it is needed to avoid caching.
So I added it if the `opts.clearCache` is set.

Fixes #6789

# Checklist:


packages/qwik/src/optimizer/src/qwik-binding-map.ts
```diff
@@ -30,15 +30,5 @@ export const QWIK_BINDING_MAP = {
         "platformArchABI": "qwik.win32-x64-msvc.node"
       }
     ]
-  },
```
wmertens, line 33: Don't remove this

packages/docs/src/routes/api/qwik/index.md
```diff
@@ -3679,7 +3679,7 @@ Options for the prefetch service worker.
 </tbody></table>
 **Returns:**
 
-[JSXNode](#jsxnode)&lt;'script'&gt;
+JSXNode&lt;'script'&gt;
```
wmertens, line 3682: Can you run `pnpm api.update` so it doesn't do this? Something in the build causes this, I wish it didn't :(

packages/docs/src/repl/worker/app-bundle-client.ts
```diff
@@ -19,7 +19,7 @@ export const appBundleClient = async (
     srcInputs: getInputs(options),
     // Older versions don't support `segment`
     entryStrategy:
-      options.entryStrategy.type === 'segment' ? { type: 'hook' } : options.entryStrategy,
+      options.entryStrategy?.type === 'segment' ? { type: 'hook' } : options.entryStrategy,
```
wmertens, line 22: Oh wow, it's not always defined?

packages/qwik-city/src/runtime/src/use-endpoint.ts
```diff
@@ -28,7 +28,7 @@ export const loadClientData = async (
   let resolveFn: () => void | undefined;
 
   if (!qData) {
-    const fetchOptions = getFetchOptions(opts?.action);
+    const fetchOptions = getFetchOptions(opts?.action, opts?.clearCache) ?? {};
```
wmertens, line 31: Why the {}? No need to create objects here, no? And if it is really needed, better to add it in getFetchOptions

wmertens (changes requested):
The fix looks great but we need a few changes

packages/qwik/src/optimizer/src/qwik-binding-map.ts
genki: I see. This might have been automatically modified by `pnpm run build.full`.
I will manually revert this, but it could solve?
I am a bit confusion about the current build process has changed from before.

packages/docs/src/repl/worker/app-bundle-client.ts
genki: I had to fix this in order to suppress error from the eslint because it is defined as `entryStrategy?: EntryStrategy;`

packages/qwik-city/src/runtime/src/use-endpoint.ts
genki: Yeah, this is not needed. I remove it.
This is a remnant of a tentative implementation that was modifying opts inplace.

packages/docs/src/repl/worker/app-bundle-client.ts
wmertens: Weird that TS didn't encounter this problem when it got merged

wmertens (approved):
LGTM 🙌 Thanks!

PR #6810, opened 2024-08-20
feat: add migrate-v2 command

dmitry-stepanenko:


shairez:
@dmitry-stepanenko the new package for city is `@qwik.dev/router` 😊

packages/qwik/src/cli/migrate-v2/run-migration.ts
```diff
@@ -0,0 +1,35 @@
...
+    initialValue: true,
+  });
+
+  if (isCancel(proceed) || !proceed) {
+    bye();
+  }
+
+  try {
+    replacePackage('@builder.io/qwik-city', '@qwik.dev/city', versions['@qwik.dev/city']);
+    replacePackage('@builder.io/qwik', '@qwik.dev/qwik', versions['@qwik.dev/qwik']);
```
ianlet, line 28: should we add `@builder.io/qwik-auth` -> `@auth/core` as well? considering it was removed in v2

PatrickJS: yes please add

ianlet: @shairez anything related to `qwik-labs` or other packages to add here?

ianlet:
`QwikCityProvider` was renamed to `QwikRouterProvider`. Ideally, this renaming should be part of this migrate command as well.

packages/qwik/src/cli/migrate-v2/run-migration.ts
shairez: so... I think `qwik-auth` should be migrated to `@auth/qwik` but there are a bunch of extra migration steps for that, so I don't know if we want to get into that via this script.

Regarding `qwik-labs`, all of its features have been folded into either `core` or `router`.
But the only used one I know of is `insights`.

I've documented its changes here - 
https://github.com/QwikDev/qwik/pull/7018

again, it's not widely used, so I don't know if it worth the effort to make it automatic like the other more used ones.

shairez:
Thanks @ianlet !

FYI qwik-react was renamed to `@qwik.dev/react`

dmitry-stepanenko:
@shairez @PatrickJS @ianlet hey folks, WDYT if we change the target of this PR to `main`? I think that users should be able to install latest v1 and then run `npm run qwik migrate-v2`. It makes more sense than installing v2 manually

I also added logic to update the following names
```
QwikCityProvider => QwikRouterProvider
qwikCity => qwikRouter
QwikCityVitePluginOptions => QwikRouterVitePluginOptions
QwikCityPlugin => QwikRouterPlugin
createQwikCity => createQwikRouter
QwikCityNodeRequestOptions => QwikRouterNodeRequestOptions
```

Logic for the `@builder.io/qwik-auth` wasn't added, let's discuss it separately

dmitry-stepanenko:
closing in favour of https://github.com/QwikDev/qwik/pull/6810, which targets v1 app

Issue #6777, opened 2024-08-07
[🐞] TypeError: Cannot convert undefined or null to object


harshmangalam:
### Which component is affected?

Qwik Runtime

### Describe the bug

I was trying to init Qwik in codesandbox but `pnpm run dev` exit with error
```
error when starting dev server:
TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at Object.validateSource (file:///project/workspace/node_modules/.pnpm/@builder.io+qwik@1.5.4_undici@6.16.0/node_modules/@builder.io/qwik/optimizer.mjs:1944:41)
    at async Context.buildStart (file:///project/workspace/node_modules/.pnpm/@builder.io+qwik@1.5.4_undici@6.16.0/node_modules/@builder.io/qwik/optimizer.mjs:3513:7)
    at async Promise.all (index 2)
    at async hookParallel (file:///project/workspace/node_modules/.pnpm/vite@5.2.11/node_modules/vite/dist/node/chunks/dep-cNe07EU9.js:50688:9)
    at async Object.buildStart (file:///project/workspace/node_modules/.pnpm/vite@5.2.11/node_modules/vite/dist/node/chunks/dep-cNe07EU9.js:51040:13)
    at async file:///project/workspace/node_modules/.pnpm/vite@5.2.11/node_modules/vite/dist/node/chunks/dep-cNe07EU9.js:65087:13
    at async httpServer.listen (file:///project/workspace/node_modules/.pnpm/vite@5.2.11/node_modules/vite/dist/node/chunks/dep-cNe07EU9.js:65105:17)
 ELIFECYCLE  Command failed with exit code 1.
```

### Reproduction

[...]

PR #6812, opened 2024-08-21
fix(optimizer): don't crash when no input defined

JerryWu1234:
#6777 
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Infra

# Description

# Checklist:


packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -575,7 +575,7 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
       let { code } = transformedModule[0];
       const { map, segment } = transformedModule[0];
 
-      if (devServer) {
+      if (devServer && !opts.csr) {
```
wmertens, line 578: Actually this needs to be fixed in a different way. The input can be a string too.

We need a function that gets the string or first object member or undefined. 

Then here and in other places where the first input is taken, use that function.

I think there's always an input even when csr but I might be wrong.

JerryWu1234: I agree with you. But there isn't any information about SCR in document.
so that I hope it would be better if there is more information about SCR I can make reference

JerryWu1234: > Actually this needs to be fixed in a different way. The input can be a string too.
> 
> We need a function that gets the string or first object member or undefined.
> 
> Then here and in other places where the first input is taken, use that function.
> 
> I think there's always an input even when csr but I might be wrong.

I agree with you. But there isn't any information about SCR in document.
so that I hope it would be better if there is more information about SCR I can make reference

JerryWu1234: > Actually this needs to be fixed in a different way. The input can be a string too.
> 
> We need a function that gets the string or first object member or undefined.
> 
> Then here and in other places where the first input is taken, use that function.
> 
> I think there's always an input even when csr but I might be wrong.

image

Could you give me more information about how develops it?

wmertens: Hmm ok I was mistaken. So just grab the first input as before but check if there is one `if (devServer && firstInput)`

JerryWu1234: I got it

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -574,8 +574,9 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
       debug(`load()`, 'Found', id);
       let { code } = transformedModule[0];
       const { map, segment } = transformedModule[0];
+      const firstInput = Object.values(opts.input)?.[0];
```
wmertens, line 577: Object.values will always be an array so no ?. needed, no?

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -574,8 +574,9 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
...
 
-      if (devServer) {
+      if (devServer && firstInput) {
         const firstInput = Object.values(opts.input)[0];
```
wmertens, line 580: You need to remove this one since you're getting it earlier

JerryWu1234:
this error because it can't be `null` 

```js
Object.values(null)
```

VM790:1 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.values (<anonymous>)
    at <anonymous>:1:8

JerryWu1234:
image
I checked it after modified

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -428,7 +428,7 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
         };
       }
 
-      const firstInput = Object.values(opts.input)[0];
+      const firstInput = Object.values(opts.input || [])[0];
```
wmertens, line 431: `const firstInput = opts.input && Object.values(opts.input)[0];`

This way you don't create objects for no reason

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -574,9 +574,9 @@ export function createPlugin(optimizerOptions: OptimizerOptions = {}) {
       debug(`load()`, 'Found', id);
       let { code } = transformedModule[0];
       const { map, segment } = transformedModule[0];
+      const firstInput = Object.values(opts.input || [])[0];
```
wmertens, line 577: Same. Are these the only two places checking opts.input? I thought there were more

wmertens (changes requested):

packages/qwik/src/optimizer/src/plugins/vite-dev-server.ts
```diff
@@ -142,7 +142,7 @@ export async function configureDevServer(
           return;
         }
 
-        const firstInput = Object.values(opts.input)[0];
+        const firstInput = Object.values(opts.input || [])[0];
```
wmertens, line 145: Ah, I didn't notice this one on my phone 😅

sapphi-red:
@wmertens May I ask you to take another look? The qwik template in create-vite is broken for a while.

JerryWu1234:
> @wmertens May I ask you to take another look? The qwik template in create-vite is broken for a while.


This Pr will cause another problem?

JerryWu1234:
> @wmertens May I ask you to take another look? The qwik template in create-vite is broken for a while.

Has This caused another problem?

sapphi-red:
@JerryWu1234 No. The qwik template has been broken for a while because of https://github.com/QwikDev/qwik/issues/6777 and I found that this PR.

wmertens (approved):
LGTM

PR #6824, opened 2024-08-26
Add fetchpriority option for link + Fix prefetch url in dev mode

GrandSchtroumpf:
# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [ ] Infra

# Description
prefetch strategy `linkInsert` always has `fetchpriority` set to "auto" which might result into busying the browser at load time. With `linkFetchPriority` we can set the bundle prefetch to "low" to give an indication to the browser.

# Checklist:

- [x] I have made corresponding changes to the Qwik docs
- [ ] Added new tests to cover the fix / functionality -> no existing test infra for this par for the code

### DIFF
--- a/.changeset/shaggy-apes-kneel.md
+++ b/.changeset/shaggy-apes-kneel.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+sync$ QRLs will now be serialized into the HTML in a shorter form
--- a/.changeset/silver-countries-kiss.md
+++ b/.changeset/silver-countries-kiss.md
@@ -1,5 +1,5 @@
 ---
-'@builder.io/qwik-city': minor
+'@builder.io/qwik-city': patch
 ---
 
 Prevent unexpected caching for q-data.json
--- a/packages/qwik-city/src/runtime/src/router-outlet-component.tsx
+++ b/packages/qwik-city/src/runtime/src/router-outlet-component.tsx
@@ -41,13 +41,16 @@ export const RouterOutlet = component$(() => {
             document:onQCInit$={spaInit}
             document:onQInit$={sync$(() => {
               // Minify window and history
-              ((window: ClientSPAWindow, history: History & { state?: ScrollHistoryState }) => {
-                if (!window._qcs && history.scrollRestoration === 'manual') {
-                  window._qcs = true;
+              // Write this as minified as possible, the optimizer does not really minify this code.
+              ((w: ClientSPAWindow, h: History & { state?: ScrollHistoryState }) => {
+                if (!w._qcs && h.scrollRestoration === 'manual') {
+                  // true
+                  w._qcs = !0;
 
-                  const scrollState = history.state?._qCityScroll;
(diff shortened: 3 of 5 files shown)


thejackshelton:
LGTM! Except for the DOMContentLoaded listener. Which if I recall we were using just to test?

Verified that this does fix the paths in dev with modulepreload, dev mode is super snappy 😄

GrandSchtroumpf:
@thejackshelton I removed the addEventListener and verify that `modulepreload` href works great for dev & production now.
Using `html-append` or `js-append` with `modulepreload` makes first interaction in devmode much faster.

thejackshelton (approved):
LGTM

packages/docs/src/routes/api/qwik/index.md
```diff
@@ -3679,7 +3679,7 @@ Options for the prefetch service worker.
 </tbody></table>
 **Returns:**
 
-[JSXNode](#jsxnode)&lt;'script'&gt;
+JSXNode&lt;'script'&gt;
```
wmertens, line 3682: Please run `pnpm api.update` again, these get messed up by the build somehow :(

wmertens (changes requested):
LGTM besides the api update

packages/docs/src/routes/api/qwik/index.md
GrandSchtroumpf: fixed with #[26db654](https://github.com/QwikDev/qwik/pull/6824/commits/26db6545b56a337cedb4c97b2b496066515579c3)

thejackshelton (dismissed):

shairez (approved):
thanks @GrandSchtroumpf ! 🙏

PatrickJS-kunaico (approved):

shairez:
For some reason, the tests are failing after I merged `main`

PatrickJS (approved):

packages/qwik/src/server/prefetch-strategy.ts
```diff
@@ -78,7 +79,7 @@ function addBundle(
   buildBase: string,
   bundleFileName: string
 ) {
-  const url = buildBase + bundleFileName;
+  const url = qDev ? bundleFileName : buildBase + bundleFileName;
```
wmertens, line 82: I think this also needs the qrl parent, see #6936 and https://github.com/QwikDev/qwik/blob/6ba3052817979c1b24a25f3a113bbcd0849fa71c/packages/qwik/src/optimizer/src/plugins/vite-dev-server.ts#L76

wmertens (changes requested):
can you run `pnpm api.update`? And if you could make the prefetch work that would be nice, but actually I wonder if we should even do that, since we also don't run the service worker in dev.

packages/qwik/src/server/prefetch-strategy.ts
GrandSchtroumpf: The browser is able to load the file without the _qrl_parent query params. Why do we need this query params when loading the module ?

PR #6825, opened 2024-08-26
feat(qwik-city): usePreventNavigate

wmertens:
this enables preventing navigation when the application state is dirty.

The API needs to be ratified still, so marking as draft even though everything works.

### DIFF
--- a/.changeset/wise-olives-compete.md
+++ b/.changeset/wise-olives-compete.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik-city': minor
+---
+
+`usePreventNavigate` lets you prevent navigation while your app's state is unsaved. It works asynchronously for SPA navigation and falls back to the browser's default dialogs for other navigations.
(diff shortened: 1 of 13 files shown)


packages/docs/src/routes/docs/(qwikcity)/routing/index.mdx
```diff
@@ -238,6 +239,97 @@ export default component$(() => {
 
 > The `Link` component uses the `useNavigate()` hook [internally](https://github.com/QwikDev/qwik/blob/e452582f4728cbcb7bf85d03293e757302286683/packages/qwik-city/runtime/src/link-component.tsx#L33).
 
+### Preventing navigation
```
shairez, line 242: I think we should add this under "experimental features" section in the docs with a big disclaimer.

shairez (approved):
LGTM

Thanks for the hard work on this @wmertens ! 🙏

PR #6839, opened 2024-08-30
docs(bun): remove deprecated `sharp` instructions

antoinepairet:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Infra

# Description

`sharp` is a native package that requires binaries to run. Before version 0.33.0, `sharp` relied on a postinstall script to download binaries. This made it incompatible with early versions of `bun` which did not execute `postinstall` scripts. In the meantime, `sharp` relies on `optionalDependencies` to download the binaries. As of Bun v1.0.16, the top 500 npm packages with lifecycle scripts are allowed by default. There is no more need to add `sharp` in the `trustedDependencies` of `package.json`.

References:
- https://github.com/oven-sh/bun/issues/3783
- https://github.com/oven-sh/bun/issues/4959
- https://github.com/lovell/sharp/issues/3750#issuecomment-1751172728
- https://bun.sh/docs/install/lifecycle
- https://github.com/oven-sh/bun/blob/main/src/install/default-trusted-dependencies.txt

# Checklist:

- [x] I have made corresponding changes to the Qwik docs


gioboa (changes requested):
Thanks @antoinepairet 👏 it looks great to me 
I think we should update the second file too 

Screenshot_20240830-122250

wmertens:
@antoinepairet could you apply the trustedDependencies change to the second file?

gioboa (approved):

Issue #6844, opened 2024-08-30
[🐞] MDX layouts currently ignored by transformer.


danielvaijk:
### Which component is affected?

Qwik City (routing)

### Describe the bug

A regression was introduced in version `1.5.0` of `qwik-city` where the default export for an MDX layout is currently unhandled, resulting in it being ignored. I noticed that [this PR](https://github.com/QwikDev/qwik/pull/5882/files) seems to have introduced the changes.

### Reproduction

https://github.com/danielvaijk/vandijk.sh

### Steps to reproduce

1. Create an MDX file with a default export for a layout (as defined [here](https://mdxjs.com/docs/using-mdx/#layout)).
2. You'll notice the resulting HTML does not include the layout.

### System Info

```shell
System:
    OS: macOS 14.6.1
    CPU: (8) arm64 Apple M1
    Memory: 321.98 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.0 - ~/.nvm/versions/node/v18.18.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.0/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
    bun: 1.0.6 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 128.0.6613.85
    Safari: 17.6
  npmPackages:
    @builder.io/qwik: file:.yalc/@builder.io/qwik => 1.8.0
    @builder.io/qwik-city: file:.yalc/@builder.io/qwik-city => 1.8.0
    typescript: ^5.2.2 => 5.5.2
    undici: 5.28.4 => 5.28.4
    vite: 5.3.5 => 5.3.5
```

### Additional Information

In the system info, you'll notice that I used Yalc to install a local version of Qwik in my project. I've managed to fix this locally and a Pull Request will follow shortly for this ticket.

gioboa:
Fixed by #6845

PR #6845, opened 2024-08-30
fix: correctly handle MDX layouts

danielvaijk:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Infra

# Description

This PR has fixes for two things:

1. I noticed that the link scripts in the `package.json` were broken since the relative path was incorrect.
2. A fix for https://github.com/QwikDev/qwik/issues/6844.

# Checklist:

### DIFF
--- a/packages/qwik-city/src/buildtime/markdown/mdx.ts
+++ b/packages/qwik-city/src/buildtime/markdown/mdx.ts
@@ -4,7 +4,6 @@ import { rehypeSyntaxHighlight } from './syntax-highlight';
 import type { BuildContext } from '../types';
 import { parseFrontmatter } from './frontmatter';
 import { getExtension } from '../../utils/fs';
-import { createHash } from 'node:crypto';
 import type { CompileOptions } from '@mdx-js/mdx';
 
 export async function createMdxTransformer(ctx: BuildContext): Promise<MdxTransform> {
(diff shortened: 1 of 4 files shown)


packages/qwik-city/src/buildtime/markdown/mdx.ts
```diff
@@ -79,10 +79,14 @@ export async function createMdxTransformer(ctx: BuildContext): Promise<MdxTransf
...
       const newDefault = `
-const WrappedMdxContent = () => {
-  return _jsxC(RenderOnce, {children: _jsxC(_createMdxContent, {}, 3, null)}, 3, ${JSON.stringify(key)});
+export default function MDXContent() {
+  const key = ${JSON.stringify(key)};
+  const content = _jsxC(_createMdxContent, {}, 3, null);
+
+  return typeof MDXLayout === "undefined" ?
+    _jsxC(RenderOnce, { children: content }, 3, key) :
+    _jsxC(RenderOnce, { children: _jsxC(MDXLayout, { children: content }, 3, key)});
```
wmertens, line 88:
```js
  const children = typeof MDXLayout === "undefined" ? content : _jsxC(MDXLayout, { children: content }, 3)
  return _jsxC(RenderOnce, children, 3, key);
```

wmertens (changes requested):
Nice but the return value is wrong I think, it needs to have the key on the top level component, no? And the static subtree flag should apply to all

packages/qwik-city/src/buildtime/markdown/mdx.ts
danielvaijk: I'm not 100% sure tbh. I do know the previous revision worked and so does the one with your suggestion, so I'll take the latter due to my lack of understanding of the rendering nuances of `_jsxC`. Thanks!

wmertens:
@danielvaijk the E2E tests are failing, perhaps you can run them locally to find out the problem?

wmertens (dismissed):
I fixed the E2E tests by no longer using internal API, which also helps with v2.

LGTM!

LGTM

wmertens (dismissed):

wmertens (dismissed):

wmertens (approved):

PR #6853, opened 2024-09-02
fix(prefetch-sw): only prefetch direct imports

wmertens:
+ use prefetch sw with docs

It was adding dynamic imports, which is not correct. Those should be handled by Link prefetching and Insights.

Furthermore, the graph is reduced by not including dependencies that other dependencies already import.


packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -1119,30 +1119,48 @@ function absolutePathAwareJoin(path: Path, ...segments: string[]): string {
...
-  for (const bundleName in graph) {
+  if (!graph) {
+    return [];
+  }
+  const names = Object.keys(graph).sort();
+  const map = new Map<string, { index: number; deps: Set<string> }>();
+  const clearTransitiveDeps = (deps: Set<string>, seen: Set<string>, bundleName: string) => {
+    const bundle = graph[bundleName];
+    for (const dep of bundle.imports || []) {
+      if (deps.has(dep)) {
```
shairez, line 1130: the name `deps` here is confusing, I would rename it to `parentDeps` and `parentBundle` so it'll be easier to follow.

(or you could change the `dep` in the loop to `childDep` )

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -1119,30 +1119,48 @@ function absolutePathAwareJoin(path: Path, ...segments: string[]): string {
...
   const bundleGraph: QwikBundleGraph = [];
   const graph = manifest.bundles;
-  const map = new Map<string, { index: number; deps: string[] }>();
-  for (const bundleName in graph) {
+  if (!graph) {
+    return [];
+  }
+  const names = Object.keys(graph).sort();
+  const map = new Map<string, { index: number; deps: Set<string> }>();
+  const clearTransitiveDeps = (deps: Set<string>, seen: Set<string>, bundleName: string) => {
```
shairez, line 1127: changing this into a `function` instead of a `const` and moving it further down (after it gets called) will be better for readability I believe

mhevery (changes requested):
I don't think removing dynamic import is the right fix here. It is true that we are currently over fetching do to dynamic imports, but I think better fix is to encode which imports are dynamic into the graph (for example negative numbers are dynamic imports)

SW can than make decisions on this.

Also it would be great if we could differentiate dynamic import form `useTask` (which will very likely be needed) from a route navigation, which is much less likely to be needed.

This is why I think the fix here is to encode more information so that SW can make more intelligent decisions.

In the past we ran into issues where SW would fetch static imports but not dynamic and than the site was slow because the initial click handler would get processed but subsequent `useTask` etc... would be slow.

wmertens:
We can differentiate imports, handlers, segments, and dynamic imports. We can differentiate by negative numbers indicating the type of what follows, -1 handlers, -2 segments, -3 dynamic.

Alternatively, Qwik loader can just tell us which handlers are actually on the page and those then get fetched opportunistically, no differentiation needed, only direct imports. What do you think?

Ah and we can sort by module size, should we fetch big or small modules first?

Not for this PR, but the prefetch SW is also missing:
- limit concurrent downloads
- clear speculative fetches on navigation
- a way for qwik city to declare a path-to-imports mapping for link prefetching. I'm thinking qwik-city extends the manifest with paths and their imports. 
- support for insights

packages/qwik/src/optimizer/src/plugins/vite.ts
wmertens: Actually I prefer it the other way round, I don't like how `function` looks and I like to know what a function does before it gets called...

shairez: I understand. the reason I like the convention of "top down" is because it's easier this way to understand the flow of the whole thing before reading the specific implementation details.

This could start getting really hard to follow especially when dealing with longer functions you need to pass all the way just to figure out what's the main function is actually doing. 

We can move this discussion to discord and maybe create a coding style guide for the repo based on our decisions

wmertens:
@mhevery I now detect tasks (by the fact that they also export _hW) and fetch them with lower priority. Other dynamic imports are left up to the qprefetch events.

WDYT?

shairez (dismissed):

mhevery (approved):
My understanding is that you and Shai had a chat and decided to add this as a temporary work around, and than fix it later. In that case this is good to go.

PR #6854, opened 2024-09-02
fix: qwikVite transforming id's with ?raw

thejackshelton:
Bug in either vite or how we handle the transform of `?raw`. Prevents `?raw` imports from being transformed w/ qwikVite

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Infra

# Description

# Checklist:


maiieul:
We might want to change/remove the logic that transforms those `?raw` in the first place @wmertens

maiieul:
Also this bug might be happening with other import formats such as `?jsx`, etc.

wmertens (approved):

Issue #6004, opened 2024-03-13
[🐞] Qwik server crashes when entering multiple slashes at the end of a home route


shukklin:
### Which component is affected?

Qwik Runtime

### Describe the bug

If you enter several slashes at the end of the home page route and than press the Enter key, Qwik server will crash with an error:
```
{
    TypeError [ERR_INVALID_URL]: Invalid URL
    at __node_internal_ (https://vitejsvitedcpt7f-1uhl.w-credentialless-staticblitz.com/blitz.810981ba.js:36:5406)
    at Object.<anonymous> (https://vitejsvitedcpt7f-1uhl.w-credentialless-staticblitz.com/blitz.810981ba.js:36:4168)
    at new URL (https://vitejsvitedcpt7f-1uhl.w-credentialless-staticblitz.com/blitz.810981ba.js:48:8018)
    at eval (file:///home/projects/vitejs-vite-dcpt7f/node_modules/@builder.io/qwik/optimizer.mjs:2625:15) {
  input: '//',
  code: 'ERR_INVALID_URL'
}
```

<img width="800" alt="image" src="https://github.com/BuilderIO/qwik/assets/54322808/0696f0c2-b9e9-4dcb-8ead-23c48c8992da">

### Reproduction

https://stackblitz.com/edit/vitejs-vite-dcpt7f?file=vite.config.js

### Steps to reproduce

1. Run npm install && npm run dev.
2. Add several slashes at the end of the home page route (for example, //).
3. Press the enter key.
4. Check server terminal output and you'll see an error "ERR_INVALID_URL".

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 2.11 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.5.0 - ~/.nvm/versions/node/v21.5.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v21.5.0/bin/npm
    pnpm: 8.15.4 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 122.0.6261.112
    Safari: 17.3.1
  npmPackages:
    @builder.io/qwik: ^1.5.1 => 1.5.1 
    @builder.io/qwik-city: ^1.5.1 => 1.5.1 
    @builder.io/qwik-react: ^0.5.4 => 0.5.4 
    undici: ^6.7.0 => 6.7.0 
[...]

PR #6855, opened 2024-09-03
fix: a bug about crashes down when there are two slashes.

JerryWu1234:
# What is it?
#6004 
- [x] Feature / enhancement
- [x] Bug
- [x] Infra

# Description

# Checklist:


JerryWu1234:
image

.changeset/forty-humans-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': major
+---
+
+crash down when there are two slashes at once
```
wmertens, line 5: I don't understand what this means :( Can you phrase it differently? Should we have a test for this issue?

packages/qwik/src/optimizer/src/plugins/image-size-server.ts
```diff
@@ -232,11 +235,19 @@ export const getImageSizeServer = (
...
         text = text.slice(0, offset) + imgTag + text.slice(end);
         fs.writeFileSync(filePath, text);
-      } catch (e) {
-        console.error('Error auto fixing image', e, url);
+      } else {
+        next();
+      }
+    } catch (e) {
+      if (e instanceof Error) {
+        server.ssrFixStacktrace(e);
```
wmertens, line 243: why is this needed? There's no ssr code running in here is there?

packages/qwik/src/optimizer/src/plugins/image-size-server.ts
```diff
@@ -232,11 +235,19 @@ export const getImageSizeServer = (
...
+        await formatError(sys, e);
+      }
+      next(e);
+    } finally {
+      if (typeof (res as QwikViteDevResponse)._qwikRenderResolve === 'function') {
```
wmertens, line 248: I'm surprised this is needed, and also, shouldn't this only happen when it does something?

In any case I don't like that it's here, because this module is completely unrelated to SSR, so having this here is quite unmaintainable.

packages/qwik/src/optimizer/src/plugins/image-size-server.ts
```diff
@@ -107,33 +109,34 @@ export async function getInfoForSrc(src: string) {
...
   sys: OptimizerSystem,
   rootDir: string,
-  srcDir: string
+  srcDir: string,
+  server: ViteDevServer
 ): Connect.NextHandleFunction => {
   return async (req, res, next) => {
-    const fs: typeof import('fs') = await sys.dynamicImport('node:fs');
-    const path: typeof import('path') = await sys.dynamicImport('node:path');
+    try {
```
wmertens, line 116: So is the fix that errors are caught now? Then the changelog could be "be more robust about errors in the qwik city image info service"

wmertens (changes requested):

JerryWu1234:
image
here as result

JerryWu1234:
image
However I tested in my local

.changeset/forty-humans-jump.md
wmertens, line 5: Ok I think you can change this to something like "gracefully handle image dimensions service errors"

packages/qwik/src/optimizer/src/plugins/image-size-server.ts
```diff
@@ -107,33 +108,34 @@ export async function getInfoForSrc(src: string) {
 export const getImageSizeServer = (
   sys: OptimizerSystem,
   rootDir: string,
-  srcDir: string
+  srcDir: string,
+  server: ViteDevServer
```
wmertens, line 112: You're no longer using server, right? So this can go again

wmertens (changes requested):
Getting there 😊

JerryWu1234:
@wmertens everything is fine

wmertens (approved):
LGTM

PR #6868, opened 2024-09-06
chore(optimizer): remove _hW export

wmertens:
it is not used directly, and the prefetch mechanisms take care of the import waterfall.


wmertens:
LOL no idea how, but this makes the E2E tests fail. At least now we know that those imports are necessary :relaxed:

PR #6872, opened 2024-09-08
chore: move optimizer to qwik-optimizer WIP

wmertens:
trying to make the optimizer into a separately versioned package

rebased from an earlier attempt so there are some silly reverts that need fixing


wmertens:
will do this later sometime

Issue #6856, opened 2024-09-03
`tsm` is no longer supported


wmertens:
### Is your feature request related to a problem?

[we can't update vfile](https://github.com/vfile/vfile/issues/90) because `tsm` doesn't support its import map.

### Describe the solution you'd like

We're only using tsm for builds, so we can easily replace it.

### Describe alternatives you've considered

use tsnode, or maybe via some small build we don't need anything?

wmertens:
@JerryWu1234 yes that would be great, thank you ☺️

shairez:
fixed by #6877 
Thanks @JerryWu1234 !

PR #6877, opened 2024-09-12
 INFRA: migration from tsm to tsx

JerryWu1234:
#6856

# What is it?

- [x] Feature / enhancement
- [x] Bug
- [ ] Infra

# Description

# Checklist:


JerryWu1234:
> just out of curiosity, what's the problem with TSM? was thinking that it looked promising for some of our other projects

”because tsm doesn't support its import map.“  this can answer your question ?

wmertens:
@intellix https://github.com/vfile/vfile/issues/90#issuecomment-2242667474

tsconfig.json
```diff
@@ -40,6 +40,16 @@
...
   },
+  "ts-node": {
+    "experimentalSpecifierResolution": "node",
+    "transpileOnly": true,
+    "esm": true,
+    "swc": true,
+    "compilerOptions": {
+      "jsxFactory": "qwikJsx.h",
+      "jsxFragmentFactory": "qwikJsx.Fragment"
+    }
```
wmertens, line 51: Can you try adding https://www.npmjs.com/package/ts-node#require here so that the runBefore is always executed?

package.json
```diff
@@ -144,6 +146,7 @@
     "prettier-plugin-jsdoc": "1.3.0",
     "pretty-quick": "4.0.0",
     "prompts": "2.4.2",
+    "regenerator-runtime": "^0.14.1",
```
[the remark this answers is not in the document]
JerryWu1234: image

here
image

JerryWu1234:
@wmertens I tested `pnpm run build.core` and `pnpm run build.full`  looks greet. and quick than ever

JerryWu1234:
image

@wmertens I don't how to fix it, and other it looks greet

package.json
```diff
@@ -187,35 +190,35 @@
   },
   "private": true,
   "scripts": {
-    "api.update": "tsm scripts/index.ts --tsc --api --dev",
-    "build": "tsm scripts/index.ts",
+    "api.update": "node --loader ts-node/esm --require ./scripts/runBefore.ts scripts/index.ts --tsc --api --dev",
```
octet-stream, line 193: The `--loader` flag is [discouraged](https://nodejs.org/api/cli.html#--experimental-loadermodule) in favor of `--import`, are you sure this is the right one? Also ts-node does not work with the --import flag. See: https://github.com/TypeStrong/ts-node/issues/1909

I suggest to consider these other options instead:

1. [tsimp](https://www.npmjs.com/package/tsimp) - uses tsc + caching;
2. [tsx](https://www.npmjs.com/package/tsx) - uses esbuild;
3. Anything Node.js itself has to offer: [`--experimental-strip-types`](https://nodejs.org/api/cli.html#--experimental-strip-types) or [`--experimental-transform-types`](https://nodejs.org/api/cli.html#--experimental-transform-types) flag, but they are available since Node v22 end they are experimental;

.changeset/loud-deers-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-qwik': major
```
wmertens, line 2:
```suggestion
'create-qwik': patch
```

package.json
wmertens: @JerryWu1234 Did you see this? I think tsimp is the best fit in our case, because it's only for build tooling, is light-weight and it will auto-start a daemon that caches.

Because we need to preload some globals, we still need the same approach as tsm.cjs, but it would instead be e.g. `scripts/tsimp.cjs`, which would `import "tsimp"` and then splice() `process.argv[1]` (I think) and then `await import(process.argv[1])` which would run the script as if node ran it directly.

wmertens (changes requested):
I think it would be best to start from scratch with https://www.npmjs.com/package/tsimp as described in the comment. It doens't need swc added etc, and can be a drop-in replacement/speedup.

package.json
```diff
@@ -155,10 +159,10 @@
     "terser": "5.31.3",
     "tmp": "0.2.3",
     "tree-kill": "1.2.2",
-    "tsm": "2.3.0",
+    "ts-node": "10.9.2",
```
wmertens, line 162: this can go

wmertens (dismissed):

package.json
ianlet: Joining in on the fun :steam_locomotive: 

`tsimp` doesn't work with [Typescript v5.6.2](https://github.com/tapjs/tsimp/issues/29) and the maintainer stated that [the project is staled and isn't stable](https://github.com/tapjs/tsimp/issues/29#issuecomment-2397517830)

So back to square one!

JerryWu1234:
@wmertens Please help me check this PR, as it is easy to get conflicts.

.changeset/loud-deers-run.md
```diff
@@ -0,0 +1,5 @@
+---
+'create-qwik': patch
+---
+
+migration from tsm to ts node
```
ianlet, line 5:
```suggestion
migration from tsm to tsx
```

ianlet: Also the PR title needs to be adjusted: `build: switching tsm to tsx`. Or something along those lines.

shairez: well currently we support `FEAT:` for features and `FIX:` for bugs fixes and `DOCS:` for documentation changes. 

Maybe we can add `INFRA:` if you think it's important

but for this PR I think we can merge as it is... WDYT?

ianlet: yes no problem for me :)

JerryWu1234:
@shairez  done

shairez (approved):

JerryWu1234:
> thanks @JerryWu1234 ! good job!

thanks, I think this PR is my biggest PR currently

PR #6880, opened 2024-09-12
feat(qwik): feature flags

wmertens:
Allow putting features behind feature flags

WIP


.changeset/curvy-frogs-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+Introducing the `unstableFeatures[]` option to the Vite plugin. This allows you to opt in to features that are not guaranteed to have a stable API.
```
gioboa, line 5: IMHO `experimentalFeatures` sounds better to me.

thejackshelton:
I like what Astro does in this case.

```tsx
    experimental: {
        i18nDomains: true
    },
```

An experimental object in the vite config that is typed

wmertens:
@thejackshelton I can do that too, but I prefer the shorter `['valibot', 'preventNavigate']` which doesn't need `true` added. It's also typed BTW.

thejackshelton:
> @thejackshelton I can do that too, but I prefer the shorter `['valibot', 'preventNavigate']` which doesn't need `true` added. It's also typed BTW.

In that case something like:

```
experimental: ['valibot', 'preventNavigate']
```

works for me too

.changeset/curvy-frogs-return.md
Varixo: yeah, for me `experimentalFeatures` is better too

shairez:
Great job @wmertens 

I also think `experimentalFeatures` is better than `unstableFeatures` (if we're changing it, I suggest to change it across the entire PR)

shairez (approved):

PatrickJS:
awesome

PatrickJS (dismissed):

gioboa (approved):

PR #6886, opened 2024-09-18
add documentation on now to track useStore

czuma:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ x ] Docs / tests / types / typos
- [ ] Infra

# Description

this add documentation on a case you want to track useStore and act when some of it's internal values change.  This is added because the anonymous function expresed don't work and you need to provide the track function the object directly. more info here
https://discord.com/channels/842438759945601056/1161720833035743272

# Checklist:

- [ x ] I have made corresponding changes to the Qwik docs

### DIFF
--- a/packages/docs/src/routes/docs/(qwik)/components/tasks/index.mdx
+++ b/packages/docs/src/routes/docs/(qwik)/components/tasks/index.mdx
@@ -19,7 +19,8 @@ contributors:
   - adamdbradley
   - aendel
   - jemsco
-updated_at: '2023-10-18T07:33:22Z'
+  - czuma
+updated_at: '2024-09-18T02:08:00Z'
 created_at: '2023-03-31T02:40:50Z'
 ---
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/(qwik)/components/tasks/index.mdx
```diff
@@ -232,6 +233,7 @@ const delay = (time: number) => new Promise((res) => setTimeout(res, time));
 
 > Sometimes it is required to only run code either in the server or in the client. This can be achieved by using the `isServer` and `isBrowser` booleans exported from `@builder.io/qwik/build` as shown above.
 
+> If you want to track `useStore()` objects and execute something when any of their internal values change, use `track(myExampleStore)` directly without the anonymous function. This prevents `track()` from losing the reference to the original reactive object.
```
wmertens, line 236: This is inaccurate. What this non-function form does is tell `track` to treat it as a signal/store and subscribe to it. The function form would have to be `track(() => Object.keys(store))`.

In any case, it still doesn't subscribe to deep changes.

PR #6888, opened 2024-09-18
fix(optimizer): ignore unknown deps in graph

wmertens:
There have been reports of a problem building the graph

EDIT: confirmed working


thejackshelton-kunaico (approved):
LGTM. Let's change the `weird but ok` comment to something like:

```
// return when bundle is not found in the graph
```

thejackshelton (approved):
LGTM. Let's change the `weird but ok` comment to something like:

```
// return when bundle is not found in the graph
```

shairez (approved):

PR #6901, opened 2024-09-22
fix(v2): only wrapped signal as a subscriber

Varixo:
Only WrappedSignal should be a subscriber. Signal and ComputedSingal shoudn't


wmertens (commented):
@Varixo we don't need extra tests for this? And is a ComputedSignal a WrappedSignal?

Varixo:
> @Varixo we don't need extra tests for this? And is a ComputedSignal a WrappedSignal?

there is already a test for cleaning effectdependencies/subscribers. I wonder how we can test it more

And yes in JSX ComputedSignal is a WrappedSignal.

Maybe I am wrong and all signal should be a subscribers?

wmertens:
nono, not all signals need to be subscribers.

wmertens (approved):
LGTM

PR #6905, opened 2024-09-23
fix(v2): component rerender with inline component inside

Varixo:
Fix component rerender with inline component inside

### DIFF
--- a/packages/qwik/src/core/v2/client/vnode-diff.ts
+++ b/packages/qwik/src/core/v2/client/vnode-diff.ts
@@ -5,7 +5,14 @@ import { assertDefined, assertFalse, assertTrue } from '../../error/assert';
 import type { QRLInternal } from '../../qrl/qrl-class';
 import type { QRL } from '../../qrl/qrl.public';
 import { dangerouslySetInnerHTML, serializeAttribute } from '../../render/execute-component';
-import { Fragment, JSXNodeImpl, isJSXNode, type Props } from '../../render/jsx/jsx-runtime';
+import {
+  Fragment,
+  JSXNodeImpl,
+  directGetPropsProxyProp,
+  getMergedProps,
+  isJSXNode,
+  type Props,
+} from '../../render/jsx/jsx-runtime';
 import { Slot } from '../../render/jsx/slot.public';
 import type { JSXNode, JSXOutput } from '../../render/jsx/types/jsx-node';
 import type { JSXChildren } from '../../render/jsx/types/jsx-qwik-attributes';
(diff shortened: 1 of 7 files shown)


packages/qwik/src/core/v2/client/vnode-diff.ts
```diff
@@ -1061,7 +1072,7 @@ export const vnode_diff = (
         host,
         (component$Host || container.rootVNode) as HostElement,
         component as OnRenderFn<unknown>,
-        jsxValue.props
+        getMergedProps(currentJSX.props)
```
wmertens, line 1075: This breaks referencing props inside inline components:

https://qwikdev-build-v2-fix-inline.qwik-8nx.pages.dev/examples/reactivity/counter/#f=Q0o0JsYW2BKNDrzKxZt8IP4MzgCWhkB%2FVoPL6VorCG0FqQpq4b5HrpnVobVGMVCrug6kwtdESmug5oGdM7jYh5pqow8WA0YHMWkWm12gKALZAW2AgD1nC%2FenBsj7IGcrGCiA0jVGdIGKQHjAgr0MVm9bDTZJD%2BJKRNAXIEUKpPGEaDNUQxyKpFFbu9YOLAlvacGjEGqQjT7MAaA4HIkZz8KSLhkPnOpSi6if6wA

should output "render show 0" but outputs "render show [WrappedSignal INVALID [object Object]]"

Let's add a test case for props inside inline :)

packages/qwik/src/core/v2/ssr/ssr-render-jsx.ts
```diff
@@ -233,7 +232,7 @@ function processJSXNode(
             ssr.closeFragment();
           }
         } else if (type === SSRComment) {
-          ssr.commentNode((jsx.props.data as string) || '');
+          ssr.commentNode(directGetPropsProxyProp(jsx.varProps, jsx.constProps, 'data') || '');
```
wmertens, line 235: I think it would be better to have a function that gets the jsxnode and dereferences const/var itself.

packages/qwik/src/core/v2/ssr/ssr-render-jsx.ts
```diff
@@ -491,14 +490,17 @@ function getSlotName(host: ISsrNode, jsx: JSXNode, ssr: SSRContainer): string {
...
   }
-  return (jsx.props.name as string) || QDefaultSlot;
+  return directGetPropsProxyProp(jsx.varProps, jsx.constProps, 'name') || QDefaultSlot;
 }
 
 function appendQwikInspectorAttribute(jsx: JSXNode) {
   if (isDev && qInspector && jsx.dev && jsx.type !== 'head') {
     const sanitizedFileName = jsx.dev.fileName?.replace(/\\/g, '/');
     const qwikInspectorAttr = 'data-qwik-inspector';
-    if (sanitizedFileName && !(qwikInspectorAttr in jsx.props)) {
```
wmertens, line 501: I would expect it's faster to do
```ts
if (sanitizedFileName && !(qwikInspectorAttr in jsx.varProps) && !(qwikInspectorAttr in jsx.constProps)) {
```

but actually we never put it in varProps so this could be
```ts
    if (
      sanitizedFileName &&
      !jsx.constProps || !(qwikInspectorAttr in jsx.constProps))
    ) {
      (jsx.constProps ||= {})[qwikInspectorAttr] =
        `${sanitizedFileName}:${jsx.dev.lineNumber}:${jsx.dev.columnNumber}`;
    }
```

wmertens (changes requested):

PR #6909, opened 2024-09-24
Managing multiple redirects in Qwik

thejackshelton:
# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Infra

# Description

# Checklist:

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/guides/redirects/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/guides/redirects/index.mdx
@@ -52,3 +52,74 @@ throw redirect(302, '/login');
 If you do not provide a status code, Qwik City will default to a `302` Found status.
 
 Read more about redirect status codes [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages).
+
+## Managing multiple redirects
+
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/(qwikcity)/guides/redirects/index.mdx
```diff
@@ -52,3 +52,74 @@ throw redirect(302, '/login');
...
+        })),
+    (error) => {
+      // Logs issues with redirects without failing the deployment.
+      // Remove this function to stop deployment on redirect issues.
+      console.error("Error setting up redirects", error);
+      return [];
+    },
+  );
+
+  const redirectRules = await redirectData;
```
ianlet, line 102:
```suggestion
  async function fetchRules(): Promise<
    { source: string; destination: string; permanent: boolean }[]
  > {
    // Fetch data from a CMS or API, and add more rules as needed.
    // Filter and map your data to make it easier to handle it, as simulated here:
    return [
      { source: "/old-path", destination: "/new-path", permanent: true },
      {
        source: "/another-old-path",
        destination: "/another-new-path",
        permanent: false,
      },
    ];
  }

  const redirectRules = await fetchRules();
```
Does it really add value to the example to filter/map rules here after we fetch the rules? Especially considering that we simulate the fetched rules with only valid data.

As the goal of this example seems to be only to demonstrate how to handle multiple redirects, I would skip that part and consider `fetchRules` only returns valid data.

Though we could add a comment inside `fetchRules` to invite the readers to get & filter/rules their own properly.

ianlet (approved):

gioboa (approved):
👏nice use case
Thanks @thejackshelton

PatrickJS (approved):

PR #6925, opened 2024-09-29
fix(v2): warning instead of error for tag nesting

Varixo:
Print a warning instead of throwing an error for tag nesting


mhevery (changes requested):
Unfortunately this has to be an error. If we let this through the browser will rearrange the tags which will change the offsets and break deserialization. Incorrect nesting means broken dearrialization in many cases. 

So we should not merge this.

PR #6932, opened 2024-10-03
fix: do not serialize methods of objects passed to styles

jakovljevic-mladen:
# What is it?

- [ ] Feature / enhancement
- [x] Bug fix
- [ ] Infra

# Description

It is completely valid to write a code like this:

```tsx
export const cmp = component$(() => {
  const style = {
    hasOwnProperty: () => false,
    foo: () => {
      return 'bar';
    },
    color: 'red',
  };

  return <a style={style} href="/some-link">Link</a>;
});
```

This gets serialized to DOM like this:

image

To prove this wrong, I added some tests for `stringifyStyle` function that converts whatever it gets from user to a usable string value that later gets stored to DOM. Without this fix, tests `"should ignore object methods"` and `"should stringify object with custom hasOwnProperty method"` would fail.

Side note: since I wanted to add a lot of other tests for `stringifyStyle` (and `setValueForStyle` as well), I noticed that error with code [`QError_stringifyClassOrStyle`](https://github.com/QwikDev/qwik/blob/baed4526582fb3c64f6531eae1dbff2c3d194bcc/packages/qwik/src/core/error/error.ts#L59) does not mention style property, so I updated error message in `packages/qwik/src/core/error/error.ts`. Does that make any problem for public facing APIs? If not, then I'd like it to stay because it mentions `style` as well.

# Checklist:

### DIFF
--- a/.changeset/tender-files-press.md
+++ b/.changeset/tender-files-press.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+Do not allow object methods to be serialized with style prop
--- a/packages/qwik/src/core/error/error.ts
+++ b/packages/qwik/src/core/error/error.ts
@@ -5,7 +5,7 @@ export const codeToText = (code: number, ...parts: any[]): string => {
   if (qDev) {
     // Keep one error, one line to make it easier to search for the error message.
     const MAP = [
-      'Error while serializing class attribute', // 0
+      'Error while serializing class or style attributes', // 0
       'Can not serialize a HTML Node that is not an Element', // 1
       'Runtime but no instance found on element.', // 2
       'Only primitive and object literals can be serialized', // 3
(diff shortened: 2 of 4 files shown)


wmertens:
First of all, thank you for your PR!

But, we're about ready to release v2 alpha and we'd need to re-implement this for v2. The code is on the build/v2 branch and serialization all happens in `shared-serialization.ts`.

Also I am not sure that approach will work there, it will probably strip all functions from objects instead of just for style objects. Normally, qwik should refuse to SSR when it encounters a function but right now we're just ignoring them because we're serializing too much state.

How common is this problem? Is there a workaround?

jakovljevic-mladen:
@wmertens,

> How common is this problem? Is there a workaround?

Not sure how common the problem is, but people can certainly encounter it. I don't think this issue is problematic at all, but this serialization doesn't have to happen, that's all.

Also, I tested this against changes on V2 branch, and this issue is present there as well:

image

And I went to the v2 branch and I noticed the same util functions exist there as well, so since they're not removed, they're probably still being used in v2 as well. Anyway, since this issue is present in v1, this patch can still be considered for v1.

wmertens (approved):

PR #6933, opened 2024-10-05
feat: minify scroll restoration shim

thejackshelton:
Scroll restoration shim code currently isn't minified in production. This PR creates a minified version, and uses the previous version in dev mode.

image

# What is it?

- [ ] Feature / enhancement
- [ ] Bug
- [ ] Infra

# Description

# Checklist:

### DIFF
--- a/packages/qwik-city/src/runtime/src/spa-shim.ts
+++ b/packages/qwik-city/src/runtime/src/spa-shim.ts
@@ -6,7 +6,7 @@ import { getPlatform } from '@builder.io/qwik';
 
 import init from './spa-init';
 
-const SPA_SHIM_MINIFIED = `(t,e,l)=>{((r,c,o)=>{if(!r._qcs&&"manual"===c.scrollRestoration){r._qcs=!0;let s=c.state?._qCityScroll;s&&r.scrollTo(s.x,s.y);let n=o.currentScript;if(n){let i=n.closest("[q\\\\:container]"),a=new URL(e,new URL(t,o.baseURI));import(a.href).then(t=>t[l](i))}}})(window,history,document)};`;
+const SPA_SHIM_MINIFIED = `(t,e,l)=>{((r,c,o,s,i)=>{if(!r._qcs&&"manual"===c.scrollRestoration){r._qcs=!0;s=c.state?._qCityScroll;s&&r.scrollTo(s.x,s.y);s=o.currentScript;if(s){i=s.closest("[q\\\\:container]"),s=new URL(e,new URL(t,o.baseURI));import(s.href).then(t=>t[l](i))}}})(window,history,document)};`;
 
 export default (base: string) => {
   if (isServer) {
--- a/packages/qwik-city/src/runtime/src/spa-shim.ts
+++ b/packages/qwik-city/src/runtime/src/spa-shim.ts
@@ -40,14 +40,7 @@ export const shim = async (base: string, path: string, symbol: string) => {
       const container = script!.closest('[q\\:container]')!;
       const url = new URL(path, new URL(base, document.baseURI));
 
-      if (isDev) {
-        // Bypass dev import hijack. (not going to work here)
-        // eslint-disable-next-line no-new-func
-        const imp = new Function('url', 'return import(url)');
-        (await imp(url.href))[symbol](container);
-      } else {
-        (await import(url.href))[symbol](container);
-      }
+      (await import(url.href))[symbol](container);
     }
   }
 };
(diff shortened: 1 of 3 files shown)


packages/qwik-city/src/runtime/src/spa-shim.ts
```diff
@@ -6,11 +6,13 @@ import { getPlatform } from '@builder.io/qwik';
 
 import init from './spa-init';
 
+const SPA_SHIM_MINFIED = `const shim=async(t,o,n)=>{if(!window._qcs&&"manual"===history.scrollRestoration){window._qcs=!0;const s=history.state?._qCityScroll;s&&window.scrollTo(s.x,s.y);const i=document.currentScript;if(i){const s=i?.closest("[q\\:container]"),c=new URL(o,new URL(t,document.baseURI));if(isDev){const t=new Function("url","return import(url)");(await t(c.href))[n](s)}else(await import(c.href))[n](s)}}};`;
```
wmertens, line 9: You should minify this more, you can pass window, document and history as props in an IIF. Also, the isDev should be evaluated here, not on the client

thejackshelton:
```tsx
(function (w, h, d) {
  const shim = async (base: string, path: string, symbol: string) => {
    if (!(w as ClientSPAWindow)._qcs && h.scrollRestoration === 'manual') {
      // TODO Option to remove this shim especially for MFEs, like loader, for now we only run once.
      (w as ClientSPAWindow)._qcs = true;
  
      const scrollState = (h.state as ScrollHistoryState)?._qCityScroll;
      if (scrollState) {
        w.scrollTo(scrollState.x, scrollState.y);
      }
  
      const script = d.currentScript as HTMLScriptElement;
      if (script) {
        // Inside shadow DOM, we can't get a hold of a container. So we can't
        // load the SPA shim.
        const container = script!.closest('[q\\:container]')!;
        const url = new URL(path, new URL(base, d.baseURI));
  
        if (isDev) {
          // Bypass dev import hijack. (not going to work here)
          // eslint-disable-next-line no-new-func
          const imp = new Function('url', 'return import(url)');
          (await imp(url.href))[symbol](container);
        } else {
          (await import(url.href))[symbol](container);
        }
      }
    }
  };
})(window, history, document);
```

Does this look correct before I minify it? 
image

wmertens: I think so 😅

wmertens (approved):
Ideally we'd have an e2e test for dev, but since we don't have that yet we'll just have to test manually

packages/qwik-city/src/runtime/src/spa-shim.ts
thejackshelton: oh wait I didn't stringify it

```tsx
(function (w, h, d) {
  const shim = async (base: string, path: string, symbol: string) => {
    if (!(w as ClientSPAWindow)._qcs && h.scrollRestoration === 'manual') {
      // TODO Option to remove this shim especially for MFEs, like loader, for now we only run once.
      (w as ClientSPAWindow)._qcs = true;

      const scrollState = (h.state as ScrollHistoryState)?._qCityScroll;
      if (scrollState) {
        w.scrollTo(scrollState.x, scrollState.y);
      }

      const script = d.currentScript as HTMLScriptElement;
      if (script) {
        // Inside shadow DOM, we can't get a hold of a container. So we can't
        // load the SPA shim.
        const container = script!.closest('[q\\:container]')!;
        const url = new URL(path, new URL(base, d.baseURI));

        if (isDev) {
          // Bypass dev import hijack. (not going to work here)
          // eslint-disable-next-line no-new-func
          const imp = new Function('url', 'return import(url)');
          (await imp(url.href))[symbol](container);
        } else {
          (await import(url.href))[symbol](container);
        }
      }
    }
  };
  shim.toString();
})(window, history, document);
```

packages/qwik-city/src/runtime/src/spa-shim.ts
```diff
@@ -6,11 +6,13 @@ import { getPlatform } from '@builder.io/qwik';
 
 import init from './spa-init';
 
+const SPA_SHIM_MINIFIED = `(t,e,l)=>{((r,c,o)=>{if(!r._qcs&&"manual"===c.scrollRestoration){r._qcs=!0;let s=c.state?._qCityScroll;s&&r.scrollTo(s.x,s.y);let n=o.currentScript;if(n){let i=n.closest("[q\\\\:container]"),a=new URL(e,new URL(t,o.baseURI));import(a.href).then(t=>t[l](i))}}})(window,history,document)};`;
```
wmertens, line 9: We don't need let, and it's leaking `a`:

```suggestion
const SPA_SHIM_MINIFIED = `(t,e,l)=>{((r,c,o,s,i)=>{if(!r._qcs&&"manual"===c.scrollRestoration){r._qcs=!0;s=c.state?._qCityScroll;s&&r.scrollTo(s.x,s.y);s=o.currentScript;if(s){i=s.closest("[q\\\\:container]"),s=new URL(e,new URL(t,o.baseURI));import(s.href).then(t=>t[l](i))}}})(window,history,document)};`;
```

wmertens:
Actually, let's merge #6937 instead. It minifies + allows disabling entirely.

PR #6937, opened 2024-10-06
feat(qwik-city): `noSPA` (experimental)

wmertens:
This disables the SPA init, slightly reducing code size and startup speed in MPA-only applications.

This is an experimental feature.


thejackshelton (approved):
This PR is inspired by #6933. Confirmed that this code is not present in the source when noSPA is enabled, and the SPA shim code is minified.

Tried this in a production app and didn't run into any problems. LGTM for experimental use 👍

gioboa (approved):

Issue #6921, opened 2024-09-28
[🐞] Unknown messages throw an error on PrefetchServiceWorker


DustinJSilk:
### Which component is affected?

Qwik Runtime

### Describe the bug

Unknown messages to the PrefetchServiceWorker should thrown an error with `console.error('UNKNOWN MESSAGE:', msg);`

Because of this [line here](https://github.com/QwikDev/qwik/blob/a86f56e7117e9ac566c756b21dd76acaed3ea306/packages/qwik/src/prefetch-service-worker/process-message.ts#L65), if the message is not an array, slice doesn't exist and the function throws an error which is difficult to debug:
```ts
  state.$log$('received message:', type, msg[1], msg.slice(2));
```

The type should be set as unknown and then validated first as a Qwik message before continuing.

### Reproduction

N/A

### System Info

```shell
na
```

maiieul:
Closing since qwik uses `<link rel="modulepreload">` by default since 1.14.

PR #6942, opened 2024-10-07
fix: validate prefetch-sw messages

DustinJSilk:
# What is it?

- [ ] Feature / enhancement
- [x] Bug
- [ ] Infra

# Description

fixes: https://github.com/QwikDev/qwik/issues/6921

If an unknown message is sent to the SW which isn't an array, Qwik throws an error attempting to slice the array. If it is an array type, Qwik will log an error. This PR will check if messages are Qwik messages before queuing them and ignore unknown messages.

# Checklist:

### DIFF
--- a/packages/qwik/src/optimizer/src/qwik-binding-map.ts
+++ b/packages/qwik/src/optimizer/src/qwik-binding-map.ts
@@ -30,15 +30,5 @@ export const QWIK_BINDING_MAP = {
         "platformArchABI": "qwik.win32-x64-msvc.node"
       }
     ]
-  },
-  "linux": {
-    "x64": [
-      {
-        "platform": "linux",
-        "arch": "x64",
-        "abi": "gnu",
-        "platformArchABI": "qwik.linux-x64-gnu.node"
-      }
-    ]
   }
 };
--- a/packages/qwik/src/prefetch-service-worker/setup.ts
+++ b/packages/qwik/src/prefetch-service-worker/setup.ts
@@ -1,5 +1,5 @@
 import { directFetch } from './direct-fetch';
-import { drainMsgQueue } from './process-message';
+import { drainMsgQueue, type SWMessages } from './process-message';
 import { createState, type SWState } from './state';
 
 export const setupServiceWorker = (swScope: ServiceWorkerGlobalScope) => {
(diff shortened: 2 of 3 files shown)


packages/qwik/src/optimizer/src/qwik-binding-map.ts
```diff
@@ -30,15 +30,5 @@ export const QWIK_BINDING_MAP = {
...
-  },
-  "linux": {
-    "x64": [
-      {
-        "platform": "linux",
-        "arch": "x64",
-        "abi": "gnu",
-        "platformArchABI": "qwik.linux-x64-gnu.node"
-      }
-    ]
```
DustinJSilk, line 42: This was removed automatically when building, should this change be reverted?

ianlet: Yes, please :smile:

.changeset/lucky-snails-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': minor
```
wmertens, line 2:
```suggestion
'@builder.io/qwik': patch
```

.changeset/lucky-snails-think.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Validate prefetch service worker messages and ignore unknown messages
```
wmertens, line 5:
```suggestion
Prefetch service worker is now robust again unknown messages
```

packages/qwik/src/prefetch-service-worker/setup.ts
```diff
@@ -46,3 +48,7 @@ export const setupServiceWorker = (swScope: ServiceWorkerGlobalScope) => {
     event.waitUntil(swScope.clients.claim());
   });
 };
+
+function isQwikMessages(msg: any): msg is SWMessages {
+  return msg[0] && Object.values(MsgType).includes(msg[0][0]);
```
wmertens, line 53: every message that comes in you generate an array of all the values in the enum. How about renaming the enum values so they are lowercase and you can do `!!MsgType[msg[0][0]]`?

wmertens (changes requested):

packages/qwik/src/prefetch-service-worker/setup.ts
DustinJSilk: Ah yes you’re right. Hmm, the tricky part is that some of the enum values can’t be valid enum keys, like `graph-url`. We could generate the array once outside of the function. I’m not sure what a better solution could be, what do you think?

gioboa:
Thanks for your help @DustinJSilk :pray: 
Closing since qwik uses <link rel="modulepreload"> by default since 1.14.

Issue #6950, opened 2024-10-08
[📖] broken demo for Middlewares#error


ianlet:
### Suggestion

The `error()` demo in the [Middleware](https://qwik.dev/docs/middleware/#error) page is missing/broken.

![Screenshot 2024-10-08 at 17-24-22 Middleware Guides 📚 Qwik Documentation](https://github.com/user-attachments/assets/cf5d1cb5-e41d-496c-8937-8fe6d7e89263)

PR #6951, opened 2024-10-09
fix: a bug that error in the middleware occurs 404 

JerryWu1234:
#6950 
# What is it?

- [x] Feature / enhancement
- [x] Bug
- [x] Infra

# Description

# Checklist:


.changeset/gorgeous-garlics-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': major
```
wmertens, line 2: This should normally be `patch`, but in this case it's a fix for the docs so no changeset is needed, can you remove this file?

packages/docs/src/routes/demo/qwikcity/middleware/layout.tsx
```diff
@@ -8,8 +8,7 @@ export const onRequest: RequestHandler = async ({ next }) => {
       error &&
       typeof error === 'object' &&
       'message' in error &&
-      typeof error.message === 'string' &&
-      error.message == 'ERROR: Demonstration of an error response.'
+      typeof error.message === 'string'
```
wmertens, line 11: I don't understand, now all errors will be ignored? What problem is this fixing?

wmertens (changes requested):

packages/docs/src/routes/demo/qwikcity/middleware/layout.tsx
JerryWu1234: image
this error because is error.message was ignored.

JerryWu1234: > but now not only this example error is ignored, but all errors. Which other errors are happening and why should they be ignored?

image

this is only thing I can trace back to.

I guess this just block error message in build

JerryWu1234:
@shairez @wmertens 

I have added a test for this PR.

shairez (approved):
Good job Jerry!

PR #6964, opened 2024-10-10
fix(store): don't mutate store target while wrapping in the store proxy

Varixo:
- freeze data shared between server side renders
- don't mutate store target while wrapping in the store proxy

### DIFF
--- a/packages/qwik/src/core/shared/shared-serialization.ts
+++ b/packages/qwik/src/core/shared/shared-serialization.ts
@@ -254,6 +254,8 @@ const inflate = (container: DeserializeContainer, target: any, typeId: TypeIds, 
         (effects as any)[STORE_ARRAY_PROP] = storeEffect;
       }
       handler.$effects$ = effects as any;
+
+      container.$storeProxyMap$.set(value, target);
       break;
     }
     case TypeIds.Signal: {
@@ -614,7 +616,6 @@ export interface SerializationContext {
   $resources$: Set<ResourceReturnInternal<unknown>>;
   $renderSymbols$: Set<string>;
   $storeProxyMap$: ObjToProxyMap;
-  $serializedStoreProxyMap$: WeakSet<object>;
 
   $getProp$: (obj: any, prop: string) => any;
   $setProp$: (obj: any, prop: string, value: any) => void;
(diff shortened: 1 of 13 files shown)


packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -801,7 +800,13 @@ export const createSerializationContext = (
       } else if (isJSXNode(obj)) {
         discoveredValues.push(obj.type, obj.props, obj.constProps, obj.children);
       } else if (Array.isArray(obj)) {
-        discoveredValues.push(...obj);
+        for (const o of obj) {
+          if (storeProxyMap.has(o)) {
+            discoveredValues.push(storeProxyMap.get(o));
+          } else {
+            discoveredValues.push(o);
```
wmertens, line 807:
```
            discoveredValues.push(storeProxyMap.get(o) || o);
```

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -801,7 +800,13 @@ export const createSerializationContext = (
       } else if (isJSXNode(obj)) {
         discoveredValues.push(obj.type, obj.props, obj.constProps, obj.children);
       } else if (Array.isArray(obj)) {
-        discoveredValues.push(...obj);
+        for (const o of obj) {
```
wmertens, line 803: this is wrong, it should just push like before, we need to look at the objects themselves one by one

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -760,8 +758,16 @@ export const createSerializationContext = (
       } else if (obj instanceof Error) {
         discoveredValues.push(...Object.values(obj));
       } else if (isStore(obj)) {
-        discoveredValues.push(getStoreTarget(obj));
-        discoveredValues.push(getStoreHandler(obj)!.$effects$);
+        const target = getStoreTarget(obj)!;
+        const effects = getStoreHandler(obj)!.$effects$;
+        discoveredValues.push(target, effects);
```
wmertens, line 763: Small optimization: the effects will always be a unique array so skip it

```suggestion
        discoveredValues.push(target, ...effects);
```

wmertens (approved):
❤️

PR #6973, opened 2024-10-13
fix(slots): get the right component frame while processing slots

### LINKED ISSUE #6972: [🐞] v2: nested projections

Qwik Runtime

### Describe the bug

     In the case of the logical tree with nested projections in the same component, we need to
     persist the depth of the projection

     Nested projection VNodes can occur only in case of nested slots. For example:

     ```tsx
     <Comp1> <-- has slot
       <Comp2> <-- has slot
          <Slot />
       </Comp2>
     </Comp1>
     ```
     
     Reproduction:
 ```tsx
import { Slot } from '@builder.io/qwik';

  return (
    <CompOne>
      <div>Hey</div>
    </CompOne>
  );
});

export const CompOne = component$(() => {
  return (
    <CompTwo>
      <CompThree>
        <Slot />
      </CompThree>
    </CompTwo>
  );
});

export const CompTwo = component$(() => {
  return (
    <div>
      <Slot />
    </div>
  );
});

export const CompThree = component$(() => {
  return (
    <div>
      <Slot />
    </div>
  );
});
     ```
     

### Reproduction

N/A

### Steps to reproduce

Have a nested projection in the same logical component

### System Info

```shell
N/A
```

### Additional Information

_No response_

Varixo:
Fixes #6972

### DIFF
--- a/packages/qwik/src/core/api.md
+++ b/packages/qwik/src/core/api.md
@@ -543,6 +543,32 @@ export const _isJSXNode: <T>(n: unknown) => n is JSXNode<T>;
 // @public (undocumented)
 export const isSignal: (value: any) => value is Signal<unknown>;
 
+// Warning: (ae-internal-missing-underscore) The name "ISsrComponentFrame" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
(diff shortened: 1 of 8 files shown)


wmertens (approved):
🙌

PR #6983, opened 2024-10-16
fix(signals): should early resolve computed qrl

Varixo:
System throws an error with qrl promise when the qrl is not resolved. To prevent that we should early resolve computed qrl while computed deserialization

### DIFF
--- a/packages/qwik/src/core/shared/shared-serialization.ts
+++ b/packages/qwik/src/core/shared/shared-serialization.ts
@@ -42,6 +42,7 @@ import {
   type SyncQRLInternal,
 } from './qrl/qrl-class';
 import type { QRL } from './qrl/qrl.public';
+import { ChoreType } from './scheduler';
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -282,6 +283,7 @@ const inflate = (container: DeserializeContainer, target: any, typeId: TypeIds,
       computed.$untrackedValue$ = d[1];
       computed.$invalid$ = d[2];
       computed.$effects$ = d.slice(3);
+      (container as DomContainer).$scheduler$?.(ChoreType.QRL_RESOLVE, null, computed.$computeQrl$);
```
wmertens, line 286: also add `d[0].resolve()` here so it already tries to download it in this tick

+ add why it does this: to prevent anything from firing while computed qrls load so that they won't throw when not expected

wmertens (approved):
👍

PR #6988, opened 2024-10-17
fix(qwik-city): allow qwik to process .mdx files

wmertens:
this allows embedded qrl functionality.

This may or may not be a good idea.

### DIFF
--- a/.changeset/three-ears-cry.md
+++ b/.changeset/three-ears-cry.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik-city': patch
+---
+
+FIX: .md and .mdx files are now also being processed by the Qwik optimizer after converting from Markdown. This allows embedding qwik components.
--- a/packages/qwik-city/src/buildtime/vite/plugin.ts
+++ b/packages/qwik-city/src/buildtime/vite/plugin.ts
@@ -1,6 +1,6 @@
 import swRegister from '@qwik-city-sw-register-build';
 import { createMdxTransformer, type MdxTransform } from '../markdown/mdx';
-import { basename, join, resolve, extname } from 'node:path';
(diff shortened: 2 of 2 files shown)


shairez:
@wmertens can you elaborate on it a bit?
We can already embed Qwik components in mdx files no?

thejackshelton-kunaico:
> @wmertens can you elaborate on it a bit? We can already embed Qwik components in mdx files no?

This might be related to the session wout and I paired on, where specifically in Qwik UI MDX breaks in v2 because of the specific recma config, which has a qrl in the mdx.

e.g what allows us to do the custom styling for each mdx tag (blockquote, paragraph, etc.)

thejackshelton:
> @wmertens can you elaborate on it a bit? We can already embed Qwik components in mdx files no?

@shairez 

This might be related to the session wout and I paired on, where specifically in Qwik UI MDX breaks in v2 because of the specific recma config, which has a qrl in the mdx.

e.g what allows us to do the custom styling for each mdx tag (blockquote, paragraph, etc.)

gioboa (commented):
is it still valid?

wmertens:
I'm not sure if we should do it. It's costing extra processing time and most of the time it's not needed

wmertens:
I'm not sure if we should do it. It's costing extra processing time and most of the time it's not needed

gioboa:
I see, an extra cost it's not great.
if we aren't solving a bug, let's close it.

gioboa (commented):
I see, an extra cost it's not great.
if we aren't solving a bug, let's close it.

wmertens:
Another reason is that we can optimize static content a lot if the optimizer processes it. Let's do it.

maiieul:
Closing as stale.

PR #6989, opened 2024-10-17
Bind-in-jsx

wmertens:
this makes bind:value and bind:checked work even when passed as spread props.

WIP because the REPL needs to work with it too


wmertens:
closing in favor of #7321

PR #6990, opened 2024-10-19
docs: clarify wording

brakmic:
# What is it?

- Docs

# Description

The original text wrongly implied that Qwik's tools increase JavaScript, which goes against what Qwik is all about. I reworded it to clarify that the issue applies to general tools, not Qwik's.


gioboa (approved):
Thanks @brakmic

PR #6991, opened 2024-10-19
fix: solve issue with Cache-Control header deletion

nelsonprsousa:
# What is it?

- Bug (kinda)

# Description
Currently, it is impossible to define a Cache-Control header using the `request.cacheControl` on any method that takes a request event other than success (HTTP 200). Looks like qwik-city, internally, deletes whatever is defined.

I am not entirely sure about the `error()` and `fail()` handlers, which are also forcing the removal of the header. I'd say that engineers could potentially decide to cache 404s, for example, and the framework should not enforce this kind of rules. However, I am not sure about this premise for 4xx or 5xx.

Redirects (HTTP 301, HTTP 302, HTTP 307, HTTP 308, at least) should be totally configurable by software engineers.

# Checklist

### DIFF
--- a/packages/qwik-city/src/middleware/request-handler/request-event.ts
+++ b/packages/qwik-city/src/middleware/request-handler/request-event.ts
@@ -207,6 +207,10 @@ export function createRequestEvent(
         }
         headers.set('Location', fixedURL);
       }
+      // Fallback to 'no-store' when end user is not managing Cache-Control header
+      if (statusCode > 301 && !headers.get('Cache-Control')) {
+        headers.set('Cache-Control', 'no-store');
+      }
       exit();
       return new RedirectMessage();
     },
--- a/packages/qwik-city/src/middleware/request-handler/request-event.ts
+++ b/packages/qwik-city/src/middleware/request-handler/request-event.ts
@@ -194,7 +194,6 @@ export function createRequestEvent(
 
     error: (statusCode: number, message: string) => {
       status = statusCode;
-      headers.delete('Cache-Control');
       return new ErrorResponse(statusCode, message);
     },
 
@@ -208,10 +207,6 @@ export function createRequestEvent(
         }
         headers.set('Location', fixedURL);
       }
-      headers.delete('Cache-Control');
-      if (statusCode > 301) {
-        headers.set('Cache-Control', 'no-store');
-      }
       exit();
       return new RedirectMessage();
     },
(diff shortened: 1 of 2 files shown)


nelsonprsousa:
After a more in-depth analysis, I found that we currently have three methods that are hard deleting user-defined `Cache-Control` headers:

- `redirect`
- `error`
- `fail`

In my opinion, at least `redirect` and `error` should not enforce this rule. For HTTP 301, 302, 307, 308, and possibly other 3xx status codes, it should be the user's responsibility to define the cache control. Additionally, when we do something like `throw error(HttpStatusCode.NOT_FOUND, 'Not Found');` (e.g., when the `id` in a route like `url.com/:id` does not exist), that response should be cacheable. This means we should remove the hardcoded rule from the `error` handler as well.

As for the `fail` handler, I don’t have a strong opinion. It may be used more in the 5xx range, such as for internal server errors, where invalidating the cache could be the right approach.

Looking forward to your feedback @wmertens @thejackshelton @maiieul @mhevery

gioboa (changes requested):
Nice, thanks @nelsonprsousa 
Can you add a changeset please?([follow this tutorial](https://go.screenpal.com/watch/cZivIcVPJQV))

nelsonprsousa:
> Nice, thanks @nelsonprsousa Can you add a changeset please?([follow this tutorial](https://go.screenpal.com/watch/cZivIcVPJQV))

Sure, I can.

What do you think about the other handlers that are actually doing the same thing? Particularly `error`? 

`fail` I am tempted to leave it as is (even though I still think it is quite a bold move from a framework to perform this kind of operations).

nelsonprsousa:
@gioboa I removed the hard code removal of the Cache-Control header from qwik-city as agreed 👍 

Also added a changeset.

LMK if everything's ok 👍

packages/qwik-city/src/middleware/request-handler/request-event.ts
```diff
@@ -208,10 +207,6 @@ export function createRequestEvent(
         }
         headers.set('Location', fixedURL);
       }
-      headers.delete('Cache-Control');
-      if (statusCode > 301) {
```
gioboa, line 212: Can we add `headers.set('Cache-Control', 'no-store');` fallback if the header is not present and statusCode > 301 ?

gioboa (changes requested):
It looks great to me. Can you add the fallback please?

packages/qwik-city/src/middleware/request-handler/request-event.ts
gioboa: Thanks 👍

gioboa (approved):
Great @nelsonprsousa 👍

PR #6992, opened 2024-10-20
fix(computed): retry qrl execution on throwing promise

Varixo:
When we are executing the `computedIfNeeded` function we can throw a promise (a qrl) if the qrl is not resolved, then we should retry execution of it.

This is a workaround until a complex fix will be implemented.


wmertens (approved):
LGTM !

PR #6994, opened 2024-10-22
chore(qwik-auth) update dependencies due to security alert

tzdesign:
the cookie package had a validation issue see https://github.com/jshttp/cookie/pull/167


ianlet:
Thanks for this update @tzdesign :raised_hands: 

`qwik-auth` is deprecated and should be replaced by https://www.npmjs.com/package/@auth/qwik. So it would be better if you update your project to use this package instead. That way, you can benefit from security fixes and other patches.

tzdesign:
@ianlet why don't we remove it from the monorepo?

ianlet:
@tzdesign it was removed in v2. So as soon as v2 is released it won't be an issue anymore.

shairez:
thanks @tzdesign 

Yeah, as @ianlet said, this is package has been removed from the monorepo in V2, so closing this PR for now

If you think there's a good reason to merge this, let me know and I will

Issue #4918, opened 2023-08-09
[🐞] useComputed$ needs to do an equals check


FlatMapIO:
### Which component is affected?

Qwik Runtime

### Describe the bug

Currently, computed does not do an equals check, so even if the result is always true, it will still trigger the task to rerun. This is clearly not the expected behavior. I hope useComputed can add an equals option and set it to Object.is, otherwise it seems pointless for it to exist.

### Reproduction

https://stackblitz.com/edit/qwik-starter-2t6x22?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

```tsx
export default component$(() => {
  const countRef = useSignal(0);
  const isGreetOneRef = useComputed$(() => {
    return countRef.value > 1;
  });
  useTask$(({ track }) => {
    track(isGreetOneRef);
    console.log(" run task ");
  });
  return (
    <div>
      <button onClick$={() => (countRef.value = countRef.value + 1)}>
        incr
      </button>
    </div>
  );
});

```

### System Info

```shell
NONE
```

shairez:
Thanks for the report @FlatMapIO !

Reviewed it, seems like it's worth investigating further

gioboa:
Same behavior with v1.5.7.
@Varixo is it fixed in v2? can you check it please? Thanks

Varixo:
> Same behavior with v1.5.7. @Varixo is it fixed in v2? can you check it please? Thanks

We should check this after v2 signals implementation

Varixo:
Thank you for the clear reproduction! This should be fixed in v2 release, so I am going to close that issue

PR #6996, opened 2024-10-23
fix: should not trigger effects if computed value is not changed

Varixo:
Fixes: #4918

### DIFF
--- a/packages/qwik/src/core/signal/signal.ts
+++ b/packages/qwik/src/core/signal/signal.ts
@@ -345,12 +345,12 @@ export const triggerEffects = (
             container.$scheduler$(ChoreType.QRL_RESOLVE, null, effect.$computeQrl$);
           }
         }
-        if (effect instanceof ComputedSignal || effect instanceof WrappedSignal) {
-          try {
-            retryOnPromise(() => effect.$invalidate$());
-          } catch (e) {
-            logError(e);
-          }
+        try {
+          retryOnPromise(() =>
+            (effect as ComputedSignal<unknown> | WrappedSignal<unknown>).$invalidate$()
+          );
+        } catch (e) {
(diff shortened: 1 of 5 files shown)


packages/qwik/src/core/signal/signal.ts
```diff
@@ -345,15 +345,12 @@ export const triggerEffects = (
...
-        (effect as ComputedSignal<unknown> | WrappedSignal<unknown>).$invalid$ = true;
-        const previousSignal = signal;
-        try {
-          signal = effect;
-          effect.$effects$?.forEach(scheduleEffect);
-        } catch (e: unknown) {
-          logError(e);
-        } finally {
-          signal = previousSignal;
+        if (effect instanceof ComputedSignal || effect instanceof WrappedSignal) {
```
wmertens, line 348: Isn't this always true?

packages/qwik/src/core/signal/signal.ts
```diff
@@ -345,15 +345,12 @@ export const triggerEffects = (
...
+          try {
+            retryOnPromise(() => effect.$invalidate$());
```
wmertens, line 350: Should force() not also use this API then?

wmertens (commented):
Looks great, just two questions

packages/qwik/src/core/signal/signal.ts
Varixo: right, we can just cast it

packages/qwik/src/core/signal/signal.ts
Varixo: `force` don't use computeIfNeeded so probably not needed

wmertens (approved):
LGTM 🎉

PR #6999, opened 2024-10-25
Update debouncer type accuracy

CharlesStover:
# What is it?

- Docs

# Description

- Replaces `any` type with inferred type. `debounce` now knows its type arguments and knows that it loses its return type.
- Replaces unnecessary `Number` conversion.
  - `setTimeout` incorrectly returns `NodeJS.Timeout`, because TypeScript assumes it is running on a Node server; but `window.setTimeout` correctly returns `number`, because TypeScript infers it is running in the browser.

### DIFF
--- a/packages/docs/src/routes/docs/cookbook/debouncer/index.mdx
+++ b/packages/docs/src/routes/docs/cookbook/debouncer/index.mdx
@@ -31,12 +31,17 @@ The Qwik framework provides unique capabilities for managing state and effects i
 
 <CodeFile src="/src/routes/demo/cookbook/mediaController/index.tsx">
 ```tsx
-export const useDebouncer = (fn: QRL<(args: any) => void>, delay: number) => {
+export const useDebouncer = <A extends readonly unknown[], R>(
+  fn: QRL<(...args: A) => R>,
+  delay: number,
+): QRL<(...args: A) => void> => {
   const timeoutId = useSignal<number>();
 
-  return $((args: any) => {
-    clearTimeout(timeoutId.value);
-    timeoutId.value = Number(setTimeout(() => fn(args), delay));
+  return $((...args: A): void => {
+    window.clearTimeout(timeoutId.value);
+    timeoutId.value = window.setTimeout((): void => {
+      void fn(...args);
+    }, delay);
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/cookbook/debouncer/index.mdx
```diff
@@ -31,12 +31,17 @@ The Qwik framework provides unique capabilities for managing state and effects i
 
 <CodeFile src="/src/routes/demo/cookbook/mediaController/index.tsx">
 ```tsx
-export const useDebouncer = (fn: QRL<(args: any) => void>, delay: number) => {
+export const useDebouncer = <A extends readonly unknown[], R>(
```
wmertens, line 34: I've never used read-only on an args array, what does it do?

wmertens (changes requested):
Looks great but can you add some tests? See https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/core/render/jsx/types/jsx-types.unit.tsx for examples

ianlet:
> Looks great but can you add some tests? See https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/core/render/jsx/types/jsx-types.unit.tsx for examples

Does it really need tests? It's a docs example, not an actual feature offered by Qwik.

packages/docs/src/routes/docs/cookbook/debouncer/index.mdx
CharlesStover: Using it in a function definition is typically good, because it allows callers to pass readonly values:

```
function mutableArg(arr: unknown[]): void;
function readonlyArg(arr: readonly unknown[]): void;

const mutableArr: unknown[] = [];
const readonlyArr: readonly unknown[] = [];

mutableArg(mutableArr); // ✅
mutableArg(readonlyArr); // ❌

readonlyArg(mutableArr); // ✅
readonlyArg(readonlyArr); // ✅
```

In the case of `arguments` or `...args`, I don't think this impacts the caller at all. It just makes the arguments array read-only.

wmertens (approved):
LGTM

PR #7003, opened 2024-10-27
fix(dev): don't lowercase segment filenames

wmertens:
This caused lookups of parents to fail due to the case not matching the case on disk.

There's no reason to lowercase the segment names since they are virtual. It also makes the source panel in the devtools nicer to read.


shairez (approved):
LGTM, thanks!

PR #7004, opened 2024-10-27
fix(build): always have qwik binary available

wmertens:
To have the bin `qwik` working during installs, the file it points to from package.json must exist.

We already had a perpetual script in the qwik package but it got ignored at some point.


shairez (approved):
Thanks Wout!

PR #7007, opened 2024-10-28
chore: remove internals from core.d.t.s

wmertens:
- promote `sync$`, `QRLEventHandlerMulti` and the locale functions to public
- add /internal for importing internals, which is just core but with the extended d.ts
- move all `@alpha` tags to `@beta` because only that is supported for tag trimming in api-documenter
- only export trimmed d.ts for qwik-city

The end result is that when picking imports from core, you only see useful imports.


Varixo (approved):
LGTM

PR #7014, opened 2024-10-30
added vite as peer to qwik-react and qwik-labs

shairez:
`vite` is now a peer dependency of `qwik`, `qwik-city`, `qwik-react` and `qwik-labs`, so that there can be no duplicate imports. This should not have consequences, since all apps also directly depend on `vite`.

# What is it?

- Infra

# Description


wmertens (approved):
LGTM

PR #7024, opened 2024-10-31
fix(ecosystem): rename flowbite icon name

jcfauchet:
# What is it?

- Docs

# Description

Rename the flowbite icon correctly


gioboa (approved):
Nice @jcfauchet

PR #7026, opened 2024-10-31
reduce the fear level of the warning of useComputed async

shairez:
# What is it?

- Infra

# Description

When devs see warnings, they aren't used to see the word `Error` in it. 

So I removed it so devs won't think that they're app is broken or something

### DIFF
--- a/packages/qwik/src/core/use/use-task.ts
+++ b/packages/qwik/src/core/use/use-task.ts
@@ -1,17 +1,10 @@
-import { newInvokeContext, invoke, waitAndRun, untrack, useInvokeContext } from './use-core';
-import { logError, logErrorAndStop, logOnceWarn } from '../util/log';
-import { delay, safeCall, maybeThen, isPromise } from '../util/promises';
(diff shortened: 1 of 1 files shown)


packages/qwik/src/core/use/use-task.ts
```diff
@@ -763,7 +763,8 @@ export const runComputed = (
       const stack = new Error(
         'useComputed$: Async functions in computed tasks are deprecated and will stop working in v2. Use useTask$ or useResource$ instead.'
       ).stack;
-      logOnceWarn(stack);
+      const lessScaryStack = stack ? stack.replace(/^Error:\s*/, '') : '';
```
wmertens, line 766: if there's no stack it should just return I suppose?

shairez: yeah I guess :)

shairez: although I guess we should just log the warning without the stack... I'll add it

wmertens (approved):
LGTM!

PR #7037, opened 2024-11-03
fix(dev): let vite do all dev url handling

wmertens:
we now take the verbatim parent url to build the qrls, and let vite decode it.

This makes it work in windows and with the different relative/absolute urls that vite generates.


shairez (approved):
great job @wmertens ! 👏

PR #7040, opened 2024-11-03
v2: make optimizer async + remove fs

wmertens:
This removes the unused cli and `fs` functions from the optimizer, and it makes the optimizer async with the `tokio` rust package which manages a thread pool and automatically runs uses all available cores for transforming.

However, it doesn't result in that much speedup because the optimization is fast and most of the build time is spent on other things. Insights build is unchanged and docs build seems to be 10s faster.


Varixo (approved):
LGTM

PR #7044, opened 2024-11-05
chore: move internals out of JSXNode

wmertens:
remove `immutableProps` and `flags` from the public type.

This makes it so that v1 and v2 can be type compatible.

Technically it's a breaking API change but these fields were never meant for public consumption and they're still present.


shairez (approved):
thanks @wmertens !

Issue #7047, opened 2024-11-05
[🐞] Removing in a list of input remove the last input


GrandSchtroumpf:
### Which component is affected?

Qwik Runtime

### Describe the bug

When working with a list of input bound to a `useStore` removing one item from the list always remove the last 
```jsx
export default component$(() => {
  const form = useStore<{ list: string[] }>({
    list: [],
  });
  return (
    <form>
      <ul>
        {form.list.map((_, i) => (
          <li key={i}>
            <input value={form.list[i]} />
            <button type="button" onClick$={() => form.list.splice(i, 1)}>
              Remove
            </button>
          </li>
        ))}
      </ul>
      <button type="button" onClick$={() => form.list.push('')}>
        Add
      </button>
    </form>
  );
});
```
I believe the error comes from `key={i}`. We cannot use the value as a `key` as it would rerender on each keystrike and would unfocus.

### Reproduction

https://stackblitz.com/edit/github-dmhnex?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

1. Click on "Add" multiple time
2. Enter different values in each input
3. Click on "Remove" in one of the input (except the last one)
4. The last input has been remove instead of the one you clicked on

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 AMD Ryzen 9 7950X3D 16-Core Processor
    Memory: 18.06 GB / 31.18 GB
  Binaries:
    Node: 22.8.0 - ~\.nvm\versions\node\v22.8.0\bin\node.EXE
    npm: 10.9.0 - ~\.nvm\versions\node\v22.8.0\bin\npm.CMD
    pnpm: 9.12.2 - ~\.nvm\versions\node\v22.8.0\bin\pnpm.CMD
  Browsers:
    Edge: Chromium (129.0.2792.65)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    @builder.io/qwik: https://pkg.pr.new/@builder.io/qwik@f72fda5 => 1.10.0
    @builder.io/qwik-city: https://pkg.pr.new/@builder.io/qwik-city@f72fda5 => 1.10.0
    typescript: 5.4.5 => 5.4.5
    undici: * => 6.20.0
    vite: 5.3.5 => 5.3.5
```

JerryWu1234:
I'll  take a look

JerryWu1234:
@GrandSchtroumpf 
```js
[...]

PR #7049, opened 2024-11-07
bug: Ensure q:key is included when a user adds a new key

JerryWu1234:
image

Fix ensure q:key is included when a user adds a new key

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


wmertens:
This looks correct, but why is this needed? Is there a bug related to it?

Keys can be really long and this would potentially add a lot of bytes to the HTML output.

JerryWu1234:
> This looks correct, but why is this needed? Is there a bug related to it?
> 
> Keys can be really long and this would potentially add a lot of bytes to the HTML output.

https://github.com/QwikDev/qwik/issues/7047

image

it's weird,

When a list is rendered on the client side, adding more items to the list triggers a re-render. However, only the first <li> element has a `q:key` property in the initial render, while the rest lack it.

PR #7057, opened 2024-11-10
fix: calling component as a function

Varixo:
This PR fixes calling a component as a function by returning a jsxSplit with correct props inside component

### DIFF
--- a/packages/qwik/src/core/tests/inline-component.spec.tsx
+++ b/packages/qwik/src/core/tests/inline-component.spec.tsx
@@ -423,23 +423,39 @@ describe.each([
   });
 
   it('should render component$ inside inlined wrapper', async () => {
-    const ComplexWrapper = (props: PublicProps<{}>, key: string | null, flags: number) => {
-      const cmpFn = component$(() => {
-        return <Slot />;
+    interface ComplexWrapperProps {
+      foo: string;
+    }
+
+    const ComplexWrapper = (
+      props: PublicProps<ComplexWrapperProps>,
+      key: string | null,
+      flags: number
+    ) => {
+      const cmpFn = component$<ComplexWrapperProps>(({ foo }) => {
+        return (
+          <div>
+            {foo}: <Slot />
+          </div>
+        );
       });
       return cmpFn(props, key, flags);
     };
 
     const { vNode } = await render(
-      <ComplexWrapper>
-        <div id="1">Test</div>
+      <ComplexWrapper foo="bar">
+        <div>
+          bar: <div id="1">Test</div>
+        </div>
       </ComplexWrapper>,
       { debug }
     );
     expect(vNode).toMatchVDOM(
       <InlineComponent>
         <Component>
-          <div id="1">Test</div>
+          <div>
+            bar: <div id="1">Test</div>
+          </div>
         </Component>
       </InlineComponent>
     );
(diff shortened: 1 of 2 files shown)


packages/qwik/src/core/tests/inline-component.spec.tsx
```diff
@@ -419,4 +421,27 @@ describe.each([
       </Component>
     );
   });
+
+  it('should render component$ inside inlined wrapper', async () => {
+    const ComplexWrapper = (props: PublicProps<{}>, key: string | null, flags: number) => {
+      const cmpFn = component$(() => {
+        return <Slot />;
```
ianlet, line 428: If you render something else than the `<Slot />` here, I'm not sure it actually works as expected.

Here's an example:
```tsx
  it('should render component$ inside inlined wrapper', async () => {
    interface ComplexWrapperProps {
      foo: string;
    }

    const ComplexWrapper = (
      props: PublicProps<ComplexWrapperProps>,
      key: string | null,
      flags: number
    ) => {
      const cmpFn = component$<ComplexWrapperProps>(({ foo }) => {
        return (
          <div>
            {foo}: <Slot />
          </div>
        );
      });
      return cmpFn(props, key, flags);
    };

    const { vNode } = await render(<ComplexWrapper foo="bar">Test</ComplexWrapper>, { debug });
    expect(vNode).toMatchVDOM(
      <InlineComponent>
        <Component>
          <div>bar: Test</div>
        </Component>
      </InlineComponent>
    );
  });
```

Varixo: hey I changed to that test case, but everything is still working. The solution is a bit different than yours, so maybe this is the reason

shairez:
@ianlet thank you for reviewing
we'll merge this but if you find any other cases where this is not working, please let us know and link to the issue, thanks!

packages/qwik/src/core/tests/inline-component.spec.tsx
```diff
@@ -419,4 +421,43 @@ describe.each([
...
+        );
+      });
+      return cmpFn(props, key, flags);
+    };
+
+    const { vNode } = await render(
+      <ComplexWrapper foo="bar">
+        <div>
+          bar: <div id="1">Test</div>
+        </div>
```
ianlet, line 449: oh this shouldn't be here :open_mouth: that's why this test is passing

with this rendered content, the expected `vNode` should be:
```tsx
      <InlineComponent>
        <Component>
          <div>
            bar: 
            <div>
              bar: <div id="1">Test</div>
            </div>
          </div>
        </Component>
      </InlineComponent>
```
and not the one below.

Here, try with this test instead:
```tsx
  it('should render component$ inside inlined wrapper', async () => {
    interface ComplexWrapperProps {
      foo: string;
    }

    const ComplexWrapper = (
      props: PublicProps<ComplexWrapperProps>,
      key: string | null,
      flags: number
    ) => {
      const cmpFn = component$<ComplexWrapperProps>(({ foo }) => {
        return (
          <div>
            {foo}: <Slot />
          </div>
        );
      });
      return cmpFn(props, key, flags);
    };

    const { vNode } = await render(<ComplexWrapper foo="bar">Test</ComplexWrapper>, { debug });

    expect(vNode).toMatchVDOM(
      <InlineComponent>
        <Component>
          <div>bar: Test</div>
        </Component>
      </InlineComponent>
    );
  });
```
This should fail.

shairez: Thanks @ianlet 

@Varixo do you know how to fix this?

Varixo: not yet :D

PR #7059, opened 2024-11-11
add monorepo support to the CLI 

shairez:
# What is it?

- Feature / enhancement

# Description

added `projectDir` param to the cli, so the `add` command could generate files into a sub folder of a monorepo. 

plus, added `alwaysInRoot` prop to the integrations config on the CLI, to mark what should always stay in the root folder even when using the `--projectDir=` param.

 # Checklist


starters/features/drizzle/package.json
```diff
@@ -22,7 +22,11 @@
         " Check out the Drizzle docs for more info:",
         "   - https://orm.drizzle.team/docs/overview"
       ]
-    }
+    },
+    "alwaysInRoot": [
+      "drizzle",
+      "drizzle.config.ts"
+    ]
```
wmertens, line 29: really? We have drizzle in Insights and it's in the insights package dir

starters/features/prisma/package.json
```diff
@@ -20,7 +20,10 @@
         " Check out the Prisma docs for more info:",
         "   - https://www.prisma.io/docs/getting-started"
       ]
-    }
+    },
+    "alwaysInRoot": [
+      "prisma"
```
wmertens, line 25: same

wmertens (dismissed):
looking good but not entirely sure about some of the root files

LGTM

starters/features/drizzle/package.json
shairez: good catch, thanks
fixing

wmertens (approved):
LGTM ⚡

PR #7061, opened 2024-11-13
fix: component as function call

Varixo:
Return a function with the `SERIALIZABLE_STATE` as a `_jsxSplit` type to process function call as a component$


wmertens (approved):
LGTM

wmertens:
@Varixo this can be merged right? I seem to recall that there was some question about it?

PR #7066, opened 2024-11-16
fix: fix function serialization

Varixo:
Don't add `ELEMENT_SEQ_IDX`, `USE_ON_LOCAL_SEQ_IDX` and `ELEMENT_PROPS` for inline components


wmertens (approved):
🔥

PR #7071, opened 2024-11-19
fix store subscription for `prop in store` cases

revintec:
# What is it?

- Bug

# Description

the code should be self explanatory
add the missing store subscription for `prop in store` cases

# Checklist

the fix is tested in local dev env. I haven't contributed to qwik for some time, so please notify me if this PR needs improvement, I'll change accordingly

### DIFF
--- a/.changeset/proud-carrots-grab.md
+++ b/.changeset/proud-carrots-grab.md
@@ -1,5 +1,5 @@
 ---
-"@builder.io/qwik": patch
+'@builder.io/qwik': patch
 ---
 
 Fix: add subscription when doing `"prop" in store`
--- a/.changeset/proud-carrots-grab.md
+++ b/.changeset/proud-carrots-grab.md
@@ -1,0 +1,5 @@
+---
+"@builder.io/qwik": patch
+---
+
+Fix: add subscription when doing `"prop" in store`
--- a/packages/qwik/src/core/state/store.ts
+++ b/packages/qwik/src/core/state/store.ts
@@ -194,22 +194,22 @@ export class ReadWriteProxyHandler implements ProxyHandler<TargetType> {
 
   has(target: TargetType, prop: string | symbol): boolean {
     if (prop === QOjectTargetSymbol) {
-        return true;
+      return true;
     }
     const invokeCtx = tryGetInvokeContext();
     if (invokeCtx) {
-        const subscriber = invokeCtx.$subscriber$;
-        if (subscriber) {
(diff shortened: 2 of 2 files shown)


wmertens:
How does this work? To me it seems like moving the condition will cause it to subscribe less instead of more? 

And a changeset would be nice, you can add one with `pnpm change`

revintec:
> How does this work? To me it seems like moving the condition will cause it to subscribe less instead of more?
> 
> And a changeset would be nice, you can add one with `pnpm change`

it's not moving the conditions around, instead it adds `this.$manager$.$addSub$` call in `has()`
previously code like `if("key" in store)...` would not get rerun when the key get set/added into the store
with this fix, they correctly do now

you maybe mistaken because I didn't squash the commits, the commits can be squashed automatically if this PR is merged according to this https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

I've created a screenshot for your convenience
image

revintec:
> How does this work? To me it seems like moving the condition will cause it to subscribe less instead of more?
> 
> And a changeset would be nice, you can add one with `pnpm change`

and I'm having trouble running `pnpm change` for now, I'll try again later

wmertens:
@revintec oh you're right, I was looking at only the diff of a single commit.

Good find, thanks!

I see that the problem also happens in v2 https://qwikdev-build-v2.qwik-8nx.pages.dev/playground/#f=Q0o0xubG2BKNDrzKpbBRAq3twSbZwg3VqM4Aec8KXFDWanLhbr3YQBoeiPq2GmI22ES9JGCBYgsxxC4xJQVYJxTBmypIjR9QKQNxQkZiMbAeAltd7RXs76cHaaFkplVqKIHFlYBFI0StZi2ijYTNGKBdWAwBimI3ArmZhTvfGFoitQxJyTcokTOaa2iRawA

Do you feel up to also fixing it in the build/v2 branch and adding a test? The code is different, but it should be pretty easy to find where to apply this, and you can see other tests for it. You can use the playground code for the test.

I'll add a changeset and merge.

wmertens (dismissed):
LGTM

wmertens (approved):

wmertens:
actually, I ported it in https://github.com/QwikDev/qwik/pull/7077/commits/a5972bafcd53d5c8eed31587e27d284177763ae1, so no PR needed.

PR #7078, opened 2024-11-20
fix: server$ not throwing for errors above 500

DustinJSilk:
# What is it?

- Bug

# Description

Server functions only throw errors in the browser with 500 status code errors.
For errors that are greater than 500, such as a 504, the server functions will return the text which means you are unable to correctly catch and handle the errors.

# Checklist


DustinJSilk:
@shairez @wmertens a gentle nudge, any chance this could get reviewed?

PR #7096, opened 2024-11-25
feat: rely on tags to retrieve version of v2 app

dmitry-stepanenko:


packages/qwik/src/cli/migrate-v2/versions.ts
```diff
@@ -1,6 +1,7 @@
-export const versions = {
-  '@qwik.dev/core': '2.0.0-alpha.1',
-  '@qwik.dev/router': '2.0.0-alpha.1',
-  '@qwik.dev/react': '2.0.0-alpha.1',
-  'eslint-plugin-qwik': '2.0.0-alpha.1',
-};
+export const versionTagPriority = ['latest', 'v2', 'rc', 'beta', 'alpha'];
```
dmitry-stepanenko, line 1: qwik tag will be resolved dynamically based on this order, but only for the v2 version.

In other words if at some point "latest" will point to v3, it will try to get the v2 with any other tag based on this priority. If nothing is found, script will fall back to using "2.0.0"

shairez:
amazing job @dmitry-stepanenko ! thank you very much! 🙏

shairez (approved):
tested as well, works like a charm!

PR #7105, opened 2024-11-27
chore: better html streaming error

Varixo:


packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -461,8 +461,10 @@ function choreComparator(a: Chore, b: Chore, shouldThrowOnHostMismatch: boolean)
         // On server we can't schedule task for a different host!
         // Server is SSR, and therefore scheduling for anything but the current host
         // implies that things need to be re-run nad that is not supported because of streaming.
-        const errorMessage =
-          'SERVER: during HTML streaming, it is not possible to cause a re-run of tasks on a different host';
+        const errorMessage = `SERVER: during HTML streaming, it is not possible to cause a re-run of tasks on a different host.
+          You are trying to modify node which was already streamed to the client.
```
shairez, line 465: maybe instead of "modify node" we could say: "You are trying to change a state that has been already streamed to the client" 

WDYT?

PR #7106, opened 2024-11-27
fix: serialization of an array of element refs

Varixo:
Fix the case where the signal value is an array of element refs


packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -1114,10 +1114,18 @@ function serialize(serializationContext: SerializationContext): void {
...
         // TODO maybe we don't need to store all vnode data if it's only a ref
         serializationContext.$addRoot$(v);
         v = new DomVRef(v.id);
+      } else if (Array.isArray(v)) {
+        for (let i = 0; i < v.length; i++) {
+          if ($isSsrNode$(v[i])) {
+            serializationContext.$addRoot$(v[i]);
+            v[i] = new DomVRef(v[i].id);
+          }
+        }
```
Varixo, line 1128: I am not sure if we should do it this way or handle all ssr nodes inside signal value

wmertens (approved):
👏

PR #7109, opened 2024-11-28
fix(core): throw when importing twice

wmertens:
this should make it easier to detect build misconfigurations


gioboa (approved):
it looks great to me @wmertens :robot:

starters/dev-server.ts
```diff
@@ -208,16 +204,7 @@ export {
...
-           * normally qwik finds qwik-city via package.json but we don't want that
-           * because it causes it to try to lookup the special qwik city imports
-           * even when we're not actually importing qwik-city
-           */
-          disableVendorScan: true,
-          vendorRoots: enableCityServer ? [qwikCityMjs] : [],
-          entryStrategy: {
-            type: "segment",
-          },
+          entryStrategy: { type: "segment" },
```
gioboa: Please have a double check here. I used your old code @wmertens

wmertens: looks good

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -282,11 +276,11 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
 
       const updatedViteConfig: UserConfig = {
         ssr: {
-          noExternal: [QWIK_CORE_ID, QWIK_CORE_SERVER, QWIK_BUILD_ID, ...vendorIds],
+          noExternal: [QWIK_CORE_ID, QWIK_CORE_SERVER, QWIK_BUILD_ID],
```
gioboa: Please have a double check here. I used your old code @wmertens

gioboa (changes requested):
docs dev is not working

gioboa (approved):

shairez:
@wmertens summary of our meeting:

- Moving it to v2 instead of v1
- Enhancing the error message to provide a more thorough solution

Thanks!

PR #7114, opened 2024-11-30
chore: enable service worker for the docs

Varixo:


thejackshelton:
You need to remove the module preload links in entry.ssr if you're going to add this back.

maiieul (approved):

PR #7117, opened 2024-12-01
chore: improve theme-management cookbook

Shane-Donlon:
# What is it?
- Docs

# Description
Previous version of this set the theme to light if no theme was found in local storage. Updated version checks for preferences and sets accordingly.

That is: if nothing saved in local storage and prefers dark mode, local storage is set to dark etc....
# Checklist

-  [x] Tested locally and seems to work just fine on a new qwik app.

### DIFF
--- a/packages/docs/src/routes/docs/cookbook/theme-management/index.mdx
+++ b/packages/docs/src/routes/docs/cookbook/theme-management/index.mdx
@@ -56,23 +56,25 @@ module.exports = {
 Then we have to add this code in the head tag of our ```root.tsx``` file.
 
 ```js
-<script
+       <script
           dangerouslySetInnerHTML={`
         (function() {
           function setTheme(theme) {
             document.documentElement.className = theme;
             localStorage.setItem('theme', theme);
           }
-          var theme = localStorage.getItem('theme');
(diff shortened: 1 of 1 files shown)


gioboa (approved):
Thanks @Shane-Donlon 
Kudos for your first PR in the project :clap:

gioboa:
@Shane-Donlon your change is pretty minimal but it the description you are mentioning more things.
Am I missing something?

Shane-Donlon:
> @Shane-Donlon your change is pretty minimal but it the description you are mentioning more things. Am I missing something?



Thanks Giorgio. It is a really small change, I was trying to describe the current bug in the code and what the PR does.
Example before the PR .

1. Checks local storage.
2. if theme is true then adds theme
3. else adds light theme

Problem: This would automatically add light as the theme regardless of dark mode preferences.

The updated function does the below
1. Checks local storage.
2. if theme is true then adds theme
3. else checks for dark mode preferences
4. if prefers dark mode
5. Sets theme to dark
6. Else sets to light.

gioboa:
Thanks @Shane-Donlon :clap:

PR #7132, opened 2024-12-05
fix: ref inside context

Varixo:


packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -577,9 +577,11 @@ type SsrNode = {
 };
 
 type DomRef = {
-  id: string;
+  ssrNode: SsrNode;
 };
 
+let isDomRef: ((obj: unknown) => obj is DomRef) | null = null;
+
```
Varixo, line 584: I am not sure if this is a good idea

wmertens: I don't see why not, although you could initially put the false function in there

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -839,8 +841,8 @@ export const createSerializationContext = (
         discoveredValues.push(obj.$el$, obj.$qrl$, obj.$state$, obj.$effectDependencies$);
       } else if (isSsrNode(obj)) {
         discoveredValues.push(obj.vnodeData);
-      } else if (isDomRef(obj)) {
-        discoveredValues.push(obj.id);
+      } else if (isDomRef!(obj)) {
+        discoveredValues.push(obj.ssrNode.id);
```
wmertens, line 845: Btw shouldn't we rename this by using `$...$`

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -1538,6 +1541,8 @@ export const canSerialize = (value: any): boolean => {
       return true;
     } else if (value instanceof Uint8Array) {
       return true;
+    } else if (isDomRef?.(value)) {
```
wmertens, line 1544: ... Then you don't need to ?.

PR #7139, opened 2024-12-07
fix(qwik): migrate-v2 cmd enhancements

dmitry-stepanenko:


packages/qwik/src/cli/migrate-v2/replace-package.ts
```diff
@@ -9,8 +9,14 @@ function updateFileContent(path: string, content: string) {
...
 }
 
-export function replacePackage(oldPackageName: string, newPackageName: string): void {
-  replacePackageInDependencies(oldPackageName, newPackageName);
+export function replacePackage(
+  oldPackageName: string,
+  newPackageName: string,
+  skipDependencies = false
+): void {
+  if (skipDependencies) {
```
shairez, line 17: shouldn't it be "not" like `if (!skipDependencies)` ?

dmitry-stepanenko: nice catch, thanks!

shairez (approved):

PR #7148, opened 2024-12-11
fix: replacing projection content with null or undefined

### LINKED ISSUE #7147: [🐞] V2 - Error with a <Slot /> that is conditionally rendered

Qwik Runtime

### Describe the bug

If a component has a `<Slot />` and this is conditionally rendered, after it has been rendered, even if the condition changes to remove the contents of the `<Slot />`, Qwik does not remove the contents of the `<Slot />`, it still keeps it.

This error occurs in these cases:

`{condition ? CONTENT : undefined} `

`{condition && CONTENT}`

But the error does not occur in this case:

`{condition ? CONTENT : <></>}`

### Reproduction

https://qwikdev-build-v2.qwik-8nx.pages.dev/playground/#f=7Va7DoJAEOz9io0VKCFcjCRowC%2Bw8wdMlFiYkBg65N%2Fd3TnuzohECzupuNxjZ7hhdgLR5GY1JprEd9%2BEeCSqnogn4Htgs2W%2BkyJysYTlpOudK4ukZpjSaorXhhMR2kdaBZAWQHyAsF3pTtwtzjWb1qmK8kIOnYpNgs9%2F2k5OTdkA2Np3FOnKkWSlwQqwY9r4gr27Exzr7k%2BDlc8TXUgHD0hJc%2BIaJe2P7SWtr01zi%2FQVE7xpQSbLYlqSUWJ%2B85CHsBK08QScyqFCReuMGdpRQGFrt%2FVxr%2F9wOQf4eaXYXUh8iZXvfcIUQRL%2BxieeJPh3iV%2B4xAM

### Steps to reproduce

Spam the "Click" button, and you will see that even if the random value is less than 50 the content of the `<Slot />` never disappears.

### System Info

```shell
npmPackages:
    @qwik.dev/core: 2.0.0-alpha.2
    @qwik.dev/router: 2.0.0-alpha.2
    typescript: 5.7.2
    undici: 7.0.0 
    vite: 6.0.1
```

### Additional Information

_No response_

Varixo:
Before we ignored children if they were equal to null or undefined. We should only ignore them if we don't have old projection slots and the children are null

closes #7147

### DIFF
--- a/.changeset/twenty-goats-flow.md
+++ b/.changeset/twenty-goats-flow.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: replacing projection content with null or undefined
--- a/packages/qwik/src/core/client/vnode-diff.ts
+++ b/packages/qwik/src/core/client/vnode-diff.ts
@@ -402,51 +402,52 @@ export const vnode_diff = (
   /////////////////////////////////////////////////////////////////////////////
 
   function descendContentToProject(children: JSXChildren, host: VirtualVNode | null) {
-    if (!Array.isArray(children)) {
-      children = [children];
-    }
-    if (children.length) {
-      const createProjectionJSXNode = (slotName: string) => {
-        return new JSXNodeImpl(Projection, EMPTY_OBJ, null, [], 0, slotName);
-      };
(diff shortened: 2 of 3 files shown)


shairez (approved):
Thanks @Varixo ! 🙏

PR #7149, opened 2024-12-12
refactor: remove processJsx method

Varixo:
This method is not needed and complicates the code unnecessarily


shairez:
Thanks @Varixo ! 🙏

PR #7152, opened 2024-12-13
docs: improve NavLink cookbook

Shane-Donlon:
# What is it?

- Docs

# Description
## Bug in current code
activeClass is an optional prop (string)

If isActive is true but no activeClass provided "undefined" would be passed in as a class
image

## Description of updated code:
if isActive is true and activeClass is true (non-empty string) then add activeClass to link

image of code example without active class
dev tools response to above code
image of code example with active class
dev tools response to above code

# Checklist


gioboa (approved):
Nice catch @Shane-Donlon Thanks

PR #7155, opened 2024-12-13
fix: prevent multiple store deserialization

Varixo:
The getter after deserialization (for example propsproxy) was executed every time.


shairez (approved):
LGTM, thanks!

PR #7159, opened 2024-12-13
fix migrating from V1 with module resolution "node'

### LINKED ISSUE #7130: [🐞] V2 migration tool should also update "moduleResolution" to "Bundler" in tsconfig.json

Qwik Rollup / Vite plugin

### Describe the bug

I installed qwik application in version 1.2.10 [when the default](https://github.com/QwikDev/qwik/blob/v1.2.10/starters/apps/base/tsconfig.json#L12) `"moduleResolution"` in `tsconfig.json` value was set to `"node"`. After migrating to V2 alpha, this does not work producing an error on any line that has `import ... from '@qwik.dev/core';` where an error says:

```
TS2307: Cannot find module @qwik.dev/ core or its corresponding type declarations.
There are types at
.../node_modules/@qwik.dev/core/public.d.ts
, but this result could not be resolved under your current moduleResolution setting. Consider updating to node16, nodenext, or bundler
```

Updating `moduleResolution` value to `Bundler` fixes the problem, so I believe V2 migration tool should consider updating `moduleResolution` field to `Bundler` among other things it does.

### Reproduction

https://qwik.new/

### Steps to reproduce

Not really sure, but this should be correct:

- Install Qwik version 1.2.10.
- Update to Qwik 1.11.0 manually (make sure `moduleResolution` is set to `node` in `tsconfig.json`.
- Run `qwik migrate-v2`.
- Open any file that has something like this: `import ... from '@qwik.dev/core';`.
- Error is present below `'@qwik.dev/core'`.

### System Info

```shell
V2
```

### Additional Information

_No response_

shairez:
# What is it?

- Bug

# Description

Actually fixes #7130
Now migrating from V1 with `moduleResolution: "node"` in the `tscofnig.json` won't break typescript. 

# Checklist

### DIFF
--- a/.changeset/friendly-gorillas-walk.md
+++ b/.changeset/friendly-gorillas-walk.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+FIX: types error when migrating to V2 with `moduleResulution: "node"`
--- a/packages/qwik-router/package.json
+++ b/packages/qwik-router/package.json
@@ -158,7 +158,8 @@
       "types": "./lib/service-worker.d.ts",
       "import": "./lib/service-worker.mjs",
       "require": "./lib/service-worker.cjs"
-    }
+    },
+    "./package.json": "./package.json"
   },
   "files": [
     "adapters",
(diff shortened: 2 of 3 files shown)


wmertens (approved):
❤️

PR #7168, opened 2024-12-16
fix mdx not rendering

shairez:
# What is it?

- Bug

# Description

For some reason,  the Qwik UI docs didn't render with version 1.10.0 because they are based on mdx.

Apparently #6845 broke it.
But this seems to fix it.

Couldn't setup a breaking test for this one (it happens only in dev mode, and the CLI e2e passed with my attempt to break it). 

It also seems like this is solved in V2, so this will be a quick (qwik) fix until V2 is released and then we'll dig into it more.

# Checklist

### DIFF
--- a/packages/qwik-city/src/buildtime/markdown/mdx.ts
+++ b/packages/qwik-city/src/buildtime/markdown/mdx.ts
@@ -5,6 +5,7 @@ import type { BuildContext } from '../types';
 import { parseFrontmatter } from './frontmatter';
 import { getExtension } from '../../utils/fs';
 import type { CompileOptions } from '@mdx-js/mdx';
+import { createHash } from 'node:crypto';
 
 export async function createMdxTransformer(ctx: BuildContext): Promise<MdxTransform> {
   const { compile } = await import('@mdx-js/mdx');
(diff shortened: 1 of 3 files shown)


maiieul:
@shairez I'm not sure this is necessary. The Qwik UI docs use an mdx recma plugin hack that I implemented back then. So your fix here would be fixing a hack inside of Qwik UI in a way 😅.  I got the Qwik UI docs to work by simply removing the plugin. This seems to work now since Woot changed the underlying implementation. I was going to make a PR, but no time for it when I was in China 🙈

shairez:
thanks @maiieul 
well that explains the tests passing 😅

closing

shairez:
Re-opening after discussion with Maieul

packages/qwik-city/src/buildtime/markdown/mdx.ts
```diff
@@ -5,6 +5,7 @@ import type { BuildContext } from '../types';
 import { parseFrontmatter } from './frontmatter';
 import { getExtension } from '../../utils/fs';
 import type { CompileOptions } from '@mdx-js/mdx';
+import { createHash } from 'node:crypto';
```
thejackshelton, line 8: Is it ok if we rely on node here for the hash? 

```
Math.random().toString(26).split(".").pop()
```

Maybe something like this is preferred

packages/qwik-city/src/buildtime/markdown/mdx.ts
```diff
@@ -69,11 +70,21 @@ export async function createMdxTransformer(ctx: BuildContext): Promise<MdxTransf
...
+        .digest('base64')
+        .slice(0, 8)
+        .replace('+', '-')
+        .replace('/', '_');
+      const addImport = `import { jsx, _jsxC, RenderOnce } from '@builder.io/qwik';\n`;
+      const newDefault = ` 
 const WrappedMdxContent = () => {
-  const content = _createMdxContent({});
-  return typeof MDXLayout === 'function' ? jsx(MDXLayout, {children: content}) : content;
+  const content = _jsxC(RenderOnce, {children: _jsxC(_createMdxContent, {}, 3, null)}, 3, ${JSON.stringify(key)});
```
thejackshelton, line 83: These are only added to the v1 codebase right? Otherwise in v2 these would break with different transform functions being used like jsxsplit

packages/qwik-city/src/buildtime/markdown/mdx.unit.ts
```diff
@@ -60,10 +60,13 @@ describe('mdx', async () => {
           })]
         });
       }
-
+       
       const WrappedMdxContent = () => {
-        const content = _createMdxContent({});
-        return typeof MDXLayout === 'function' ? jsx(MDXLayout, {children: content}) : content;
+        const content = _jsxC(RenderOnce, {children: _jsxC(_createMdxContent, {}, 3, null)}, 3, "eB2HIyA1");
```
thejackshelton, line 65: great job on the unit test. Looks like if we do run into a problem with the transforms we'll know pretty quickly with this

thejackshelton (commented):
Other than the feedback above, LGTM

packages/qwik-city/src/buildtime/markdown/mdx.ts
shairez: because it's in build time, I think it's OK (I also checked and this should be supported by both deno and bun)

packages/qwik-city/src/buildtime/markdown/mdx.ts
shairez: yes

gioboa (approved):
Thanks @shairez 
Well done ✅

PR #7176, opened 2024-12-18
fix: repl diagnostics

Varixo:


gioboa (approved):
It looks good to me 
Thanks @Varixo

PR #7182, opened 2024-12-20
fix: input's value is string when passing number

JerryWu1234:
# What is it?

- Bug

# Description

[Repro](https://qwikdev-build-v2.qwik-8nx.pages.dev/playground/#f=Q0o0xiZw7yAnGh1E7UtZqwQSEonAYIAbqGEIdIUCVCYJRcYIKIM7nJKKEF6rTtQD5sXS1FoFbYXqJBjbFi4OFIaJYtNukwluLIBSoq0SpM5SUkgCZkkrsB7b6sRa4tUmwdQiwhhnxjC0RGr6kZIxUAJ%2FNFvQIlsAAA)

# Checklist


packages/qwik/src/server/ssr-container.ts
```diff
@@ -1181,17 +1181,20 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
...
+            const strValue = escapeHTML(serializedValue);
+            this.write(strValue);
+            this.write('');
+          } else if (serializedValue !== true) {
             this.write('="');
             const strValue = escapeHTML(String(serializedValue));
             this.write(strValue);
-
             this.write('"');
           }
```
Varixo, line 1199: we should not repeat the code. Maybe something like this
```ts
this.write('=');
const strValue = escapeHTML(typeof serializedValue === 'number' ? serializedValue : String(serializedValue));
this.write(strValue);
this.write('');
```

Varixo (commented):
Thanks for that PR. Looks great, but we should add unit tests for it

JerryWu1234:
> Thanks for that PR. Looks great, but we should add unit tests for it

The structure of Dom from the util test still is a string when even I pass a number.

JerryWu1234:
There are still other problems in another place.

packages/qwik/src/server/ssr-container.ts
Varixo: or even better, just return the correct value from the `serializeAttribute`

shairez:
Thanks @JerryWu1234 ! 🙏

Issue #7183, opened 2024-12-20
[🐞] server$ middleware cannot catch errors


DustinJSilk:
### Which component is affected?

Qwik City (routing)

### Describe the bug

Middleware in a plugin file is able to intercept and catch routeLoader$ errors, but server$ can only call next() but catching errors doesn't work - the error is always returned immediately to the browser.

The usecase for this is to catch specific network errors (in this case ConnectErrors from connectrpc) and convert them into ServerErrors to return the correct status codes, and also to log errors with structured logging.

For example:

```tsx
import { type RequestHandler } from "@builder.io/qwik-city";

export const onRequest: RequestHandler = async ({ url, next }) => {
  const isServer = url.search.startsWith("?qfunc=");

  if (isServer) {
    console.log("server$ middleware triggerd");
  } else {
    console.log("routeLoader$ middleware triggerd");
  }

  try {
    await next();
  } catch (err) {
    if (isServer) {
      // <------------------------ This never runs ------------------->
      console.log("caught server$ error");
    } else {
      console.log("caught routeLoader$ error");
    }
  }
};
```

### Reproduction

https://github.com/DustinJSilk/qwik-server-middleware-error/tree/main

### Steps to reproduce

Run the app and view errors in the terminal

### System Info

```shell
N/A
```

JerryWu1234:
@DustinJSilk fixed it by your PR

Issue #7165, opened 2024-12-15
[🐞] Throwing a ServerError only works with production builds and not during dev


DustinJSilk:
### Which component is affected?

Qwik Rollup / Vite plugin

### Describe the bug

Throwing a `ServerError` from a `server$` function only uses the correct status code in production builds, not during development.

The issue seems to be that vite copies the ServerError class, so when Qwik checks if it is a server error with `err instanceof ServerError`, the thrown instance is different to the class it is compared with. I could be wrong but this is what it looks like.

### Reproduction

N/A

### System Info

```shell
N/A
```

gioboa:
This issue seems solved by the PR mentioned above.
There is also a RFC to standardise server errors, so I close this. Thanks

PR #7185, opened 2024-12-20
fix: standardise error handling

DustinJSilk:
# What is it?

- Bug

# Description

This fixes a few issues with server$ error handling

- 4xx ServerErrors are returning a a value instead of throwing an error on the frontend
- server$ middleware could not catch errors - the server$ function would simply exit early. https://github.com/QwikDev/qwik/issues/7183
- You can now also throw a ServerError in a routeLoader$ to change the status code and return some serialised data. This is simply an added bonus but should pave the way for more standardised errors across loaders, actions, and server$ functions

# Checklist


DustinJSilk:
I get this error when running pnpm change: `Failed to find where HEAD diverged from origin/upcoming. Does origin/upcoming exist?`

DustinJSilk:
I would add a fix for this: https://github.com/QwikDev/qwik/issues/7165, but i have no idea how to fix it.

DustinJSilk:
@wmertens fyi - returning errors as values was intentional in this PR: https://github.com/QwikDev/qwik/pull/6290.
I dont think this was the right decision, as it becomes really difficult to work with. Now, errors behave like javascript errors and the user can catch them and use a type guard on the data instead.

DustinJSilk:
@shairez any chance we can get this PR reviewed?

wmertens:
for the origin/upcoming, try `git fetch origin upcoming`.

So if I understand correctly, the only somewhat breaking change is that 4xx errors now throw on the client side, right? That seems more like a bugfix than a breaking change.

DustinJSilk:
Thanks @wmertens !

Yes, the only breaking change is that 4xx errors now throw on the client. It should be classified as a bug fix.

I've added a changeset now, I had to temporarily change the changset config `"baseBranch": "origin/upcoming",` to `"baseBranch": "upcoming",` to get it to work

packages/qwik-city/src/middleware/request-handler/error-handler.ts
```diff
@@ -8,15 +8,6 @@ export class ServerError<T = Record<any, any>> extends Error {
   }
 }
 
-export class ErrorResponse extends Error {
```
DustinJSilk, line 11: Deprecate this, don't removfe

packages/qwik-city/src/middleware/request-handler/user-response.ts
```diff
@@ -65,12 +70,17 @@ async function runNext(requestEv: RequestEventInternal, resolve: (value: any) =>
     if (e instanceof RedirectMessage) {
       const stream = requestEv.getWritableStream();
       await stream.close();
-    } else if (e instanceof ErrorResponse) {
-      console.error(e);
```
DustinJSilk, line 69: Try and find the logging class if it exists

starters/apps/qwikcity-test/src/routes/(common)/server-func/server-error/index.tsx
```diff
@@ -3,34 +3,47 @@ import { server$ } from "@builder.io/qwik-city";
...
+const serverFunctionA = server$(async function a(): Promise<string> {
+  throw new ServerError<ErrorReason>(401, {
+    reason: "my error",
+    middleware: "uncaught",
+  });
 });
 
-const serverFunctionB = server$(async function b(): Promise<ResponseTuple> {
-  return [null, this.method || ""];
+const serverFunctionB = server$(async function b(): Promise<string> {
```
DustinJSilk, line 18: Add a test for routerLoader$ errors

packages/qwik-city/src/middleware/request-handler/user-response.ts
DustinJSilk: There is no logger in Qwik City

packages/qwik-city/src/middleware/request-handler/user-response.ts
```diff
@@ -65,12 +70,18 @@ async function runNext(requestEv: RequestEventInternal, resolve: (value: any) =>
...
+    } else if (e instanceof ServerError) {
       if (!requestEv.headersSent) {
-        const html = getErrorHtml(e.status, e);
         const status = e.status as StatusCodes;
-        requestEv.html(status, html);
+        const accept = requestEv.request.headers.get('Accept');
+        if (accept === '*/*') {
```
DustinJSilk, line 77: I wonder what the Accept header is for SSR vs CSR pages and if server$ functions should actually send an Accept of application/qwik-json?

shairez (dismissed):
great job Dustin!

wmertens (changes requested):
Looks great, just two small things

DustinJSilk:
Thanks @wmertens and @shairez!

DustinJSilk:
Gentle nudge @wmertens

wmertens (approved):
LGTM

DustinJSilk:
@shairez can you merge these changes please

shairez:
Thanks again @DustinJSilk ! great job!

DustinJSilk:
Thanks both!

packages/docs/src/routes/docs/(qwikcity)/error-handling/index.mdx
```diff
@@ -0,0 +1,112 @@
...
+    // This is required for dev environments due to an issue with vite: https://github.com/vitejs/vite/issues/3910
+    (isDev && err instanceof Error && err.constructor.name === "ServerError")
+  );
+}
+
+function isRedirectMessage(err: unknown): err is RedirectMessage {
+  return (
+    err instanceof RedirectMessage ||
+    // This is required for dev environments due to an issue with vite: https://github.com/vitejs/vite/issues/3910
+    (isDev && err instanceof Error && err.constructor.name === "RedirectMessage")
```
wmertens, line 109: @DustinJSilk I visited that issue but I don't understand how it is related?

DustinJSilk: Vite makes a copy of dependencies in dev mode (in node_modules/vite i think).

This means there are 2 definitions of ServerError during dev: one exists in the Qwik code, and another gets copied into the Vite cache.

So when a ServerError is thrown, it's a Qwik ServerError instance, but when you import ServerError in the middleware, you're comparing it against the Vite cached ServerError which means instanceof doesn't work.

PR #7190, opened 2024-12-24
fix: destructured props for inline components

Varixo:
Convert destructured props for an inline component to single prop

### DIFF
--- a/packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_lightweight_functional.snap
+++ b/packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_lightweight_functional.snap
@@ -9,7 +9,7 @@ snapshot_kind: text
 
 import { $, component$ } from '@qwik.dev/core';
 
-export const Foo = component$(({color}) => {
+export const Foo = component$((props) => {
 	return (
 		<div>
 			<Button {...props} />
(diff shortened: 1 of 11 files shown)


wmertens (approved):
:muscle:

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_lightweight_functional.snap
```diff
@@ -9,7 +9,7 @@ snapshot_kind: text
 
 import { $, component$ } from '@qwik.dev/core';
 
-export const Foo = component$(({color}) => {
+export const Foo = component$((props) => {
```
wmertens, line 12: This one was wrong actually, maybe we want to use some props? Not sure if important

Varixo: What do you mean? Later we use `...props` and never the `color`, so I think now it is correct?
or do you mean that this error is intentional?

wmertens: No it was destructuring color but using props

Varixo: yeah, and it is wrong, right? or should I revert it?

PR #7191, opened 2024-12-24
fix: convert destructured string prop to props variable

Varixo:


wmertens (approved):
:heart: :rocket: :fireworks:

PR #7193, opened 2024-12-25
fix: serialize var prop

Varixo:


wmertens (approved):
I'm not too familiar with this part of the code but it makes sense to me.

:+1:

PR #7196, opened 2024-12-26
fix: textarea with null value

Varixo:


packages/qwik/src/core/shared/utils/jsx-filename.ts
```diff
@@ -0,0 +1,12 @@
+import type { DevJSX } from '../jsx/types/jsx-node';
+
+export function getFileNameFromJsx(jsxDev?: DevJSX): string | null {
```
wmertens, line 3: Can you rename filename to fileLoc or something similar so it's clear the position is added?

wmertens (approved):
Very nice! Just a suggestion for the name of the variable other than that good to merge

PR #7198, opened 2024-12-26
fix: updating signal-based var props

Varixo:
We should track signal changes for var props as well


wmertens (approved):
Wow it wasn't doing that 😅

PR #7208, opened 2024-12-30
fix: moving existing virtual node during vnode diffing

Varixo:
Previously we were incorrectly creating a new virtual node instead of moving the existing one


wmertens (approved):
:+1: :+1: :+1:

PR #7217, opened 2024-12-31
fix: convert destructured array's props to signal

Varixo:


packages/qwik/src/optimizer/core/src/transform.rs
```diff
@@ -2340,6 +2336,7 @@ fn is_return_static(expr: &Option<Box<ast::Expr>>) -> bool {
 			callee: ast::Callee::Expr(box ast::Expr::Ident(ident)),
 			..
 		})) => ident.sym.ends_with('$') || ident.sym.ends_with("Qrl") || ident.sym.starts_with("use"),
+		Some(box (ast::Expr::Array(_) | ast::Expr::Object(_))) => true,
```
wmertens, line 2339: shouldn't this be called recursively? `[Math.random()]`?

Varixo: Okay I see. I've changed test & implementation for that

wmertens (approved):
LGTM

PR #7223, opened 2025-01-02
feat: custom serialization

wmertens:
This PR adds symbols to mark objects as (no/yes)serializable via symbols instead of via a Set, adds a symbol prop that will get called on serialization for custom serialization, and provides a new Signal type that lazily manages a non-serializable value.

Note that this is basically the same as `useComputed$`, except that it is invalidated during SSR so that when the value gets read on the client it will always run the compute function.

Also note that `useComputed$` now passes the previous value to the compute function, to keep the implementation compact. This seems like it might be useful, should we document this?

Everything works, looking for comments on implementation, naming etc.

## TODO

- [x] update eslint to accept serialized signals and the new serializer symbols
- [ ] check crash that was reported


Varixo:
Looks great, but what about deserialization those objects?

wmertens:
> Looks great, but what about deserialization those objects?

That's what the createSerialized$ does

thejackshelton:
Does this mean we might be able to serialize functions from other libraries, like vanilla js?

wmertens:
@Varixo @thejackshelton gaah I had forgotten to push the actual commit :sweat_smile:

wmertens:
> Does this mean we might be able to serialize functions from other libraries, like vanilla js?

Yes exactly. You need to add a symbol prop that serializes the value and then provide a deserializer that will get called lazily.

packages/qwik/src/core/use/use-serializer.ts
Varixo, line 1: Shouldn't we name it the same as the hook? I mean `use-serialized.ts`?

packages/qwik/src/core/tests/use-computed.spec.tsx
```diff
@@ -470,4 +472,50 @@ describe.each([
...
+    );
+    await trigger(container.element, 'button', 'click');
+    expect(vNode).toMatchVDOM(
+      <>
+        <button>
+          <Signal ssr-required>{'3'}</Signal>
+        </button>
+      </>
+    );
+  });
```
Varixo, line 510: maybe new test file `use-serialized.spec.ts`?

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -1130,6 +1146,20 @@ function serialize(serializationContext: SerializationContext): void {
...
+      let result = serializationResults.get(value);
+      // special case: we unwrap Promises
+      if (isPromise(result)) {
+        const promiseResult = promiseResults.get(result)!;
+        if (!promiseResult[0]) {
+          console.error(promiseResult[1]);
+          throw qError(QError.serializerSymbolRejectedPromise);
+        }
+        result = promiseResult[1];
+      }
```
Varixo, line 1159: do we need test for this case?

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -831,11 +846,31 @@ export const createSerializationContext = (
...
+                  if (data === undefined) {
+                    data = NEEDS_COMPUTATION;
+                  }
+                  serializationResults.set(obj, data);
+                  discoveredValues.push(data);
+                })
+            );
+          } else {
+            discoveredValues.push(v);
+          }
```
Varixo, line 873: I wonder if we could organise it somehow? It is very complicated right now.
What about doing just
```ts
if (obj instanceof WrappedSignal) {

} else if (obj instanceof ComputedSignal) {

} else if ...
```

packages/qwik/src/core/signal/signal.ts
```diff
@@ -566,13 +569,105 @@ export class WrappedSignal<T> extends Signal<T> implements Subscriber {
...
+       * This function will be called with serialized data to reconstruct an object.
+       *
+       * If it is created for the first time, it will get the `initial` data or `undefined`.
+       *
+       * If it uses signals or stores, it will be called when these change, and then the second
+       * argument will be the previously constructed object.
+       *
+       * This function must not return a promise.
+       */
+      deserialize: (data: S | undefined, previous: T | undefined) => T;
```
thejackshelton, line 601: So on first render, the previous param is -> initial data / inital (3rd argument param of useSerializer$)?

packages/qwik/src/core/signal/signal.ts
```diff
@@ -566,13 +569,105 @@ export class WrappedSignal<T> extends Signal<T> implements Subscriber {
...
+      /**
+       * This function will be called with the custom object to get the serialized data. You can
+       * return a promise if you need to do async work.
+       *
+       * The result may be anything that Qwik can serialize.
+       *
+       * If you do not provide it, the object will be serialized as `undefined`.
+       */
+      serialize?: (customObject: T) => S | Promise<S>;
```
thejackshelton, line 610: What's an example of when you may need the deserialize function but not the serialize function?

thejackshelton (approved):
Very thorough PR. Awesome work ⚡

thejackshelton-kunaico (changes requested):
.

packages/docs/src/routes/api/qwik/index.md
```diff
@@ -4386,7 +4447,7 @@ Creates a computed signal which is calculated from the given function. A compute
 The function must be synchronous and must not have any side effects.
 
 ```typescript
-useComputed$: <T>(qrl: import("./use-computed").ComputedFn<T>) => T extends Promise<any> ? never : import("..").ReadonlySignal<T>
+useComputed$: <T>(qrl: ComputedFn<T>) => T extends Promise<any> ? never : ReadonlySignal<T>
```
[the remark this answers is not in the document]
wmertens: it doesn't change, it just doesn't need to import it any more (not sure why but I'm happy)

packages/qwik/src/core/signal/signal.ts
wmertens: on first run, the data is initial|undef. After serialization, data is what was serialized.

If the function uses reactive scope, it will rerun and then `current` will be defined and `data` will be undefined.

packages/qwik/src/core/signal/signal.ts
wmertens: When you're just constructing an object from scope data and you don't have any state to serialize.

ianlet:
I tried the changes locally and I'm having the following issue:

image

Plus a linting issue when using the `SerializeSignal` in a `useTask$`.
```
ESLint: 
When referencing "mapbox" inside a different scope (useTask$), Qwik needs to serialize the value, however "mapbox. force" is a function, which is not serializable. Check out https:// qwik. dev/ docs/ advanced/ dollar/ for more details.
(qwik/ valid-lexical-scope)
```

Here'es the code:
```tsx
import {
  component$,
  useSerializer$,
  useSignal,
  useTask$,
} from "@qwik.dev/core";
import type { StyleSpecification } from "mapbox-gl";
import mapboxgl from "mapbox-gl";

interface QwikMapboxProps {
  token: string;
  mapStyle: string | StyleSpecification;
}

interface SerializationState {
  token: string;
  mapStyle?: string | StyleSpecification;
}

export const QwikMapbox = component$<QwikMapboxProps>(({ token, mapStyle }) => {
  const ref = useSignal<HTMLDivElement>();

  const mapbox = useSerializer$({
    initial: {
      token,
      mapStyle,
    },
    serialize: (map: mapboxgl.Map): SerializationState => {
      return {
        token,
        mapStyle: map.getStyle() || undefined,
      };
    },
    deserialize: (
      data: SerializationState | undefined,
      current: mapboxgl.Map | undefined,
    ): mapboxgl.Map => {
      if (current) {
        return current;
      }

      return new mapboxgl.Map({
        container: "map",
        style: data!.mapStyle,
        accessToken: data!.token,
      });
    },
  });

  useTask$(() => {
    mapbox.value.on("load", () => console.log("MAP LOADED"));
  });

  return <div ref={ref} style={{ width: "800px", height: "600px" }}></div>;
});
```

Link to MRE: https://github.com/ianlet/qwik-custom-serder-scope-issues

ianlet:
Also, after playing with it a little bit, I really think the `deserialize` function should _not_ handle both the actual `deserialization` and the `update` when tracked signals change. Its conceptually 2 different things and will trigger 2 different implementation anyway for developers who will use it. And the fact that we had to document 2 examples on how to use it (without tracking and with tracking) reinforces this idea.

So, IMO it should be 2 different callbacks as it is 2 different use cases, triggered after 2 different events.

wmertens:
I addressed all the comments I think.

wmertens:
@ianlet BTW even when only using mapbox on the client, this is still useful. Just make the deserialize `isBrowser ? makeMapbox() : undefined` and it will just work on the client when you need it.

ianlet:
> @ianlet BTW even when only using mapbox on the client, this is still useful. Just make the deserialize `isBrowser ? makeMapbox() : undefined` and it will just work on the client when you need it.

I just updated my test repo to use the new syntax with `update` and I changed it to apply your `isServer` suggestion but I do so, the `deserialize` function is never called on the client so the value never gets initialized.

Also, having to handle the `undefined` everywhere in that case makes it a bit heavy, but it's a detail.

```tsx
    const mapbox = useSerializer$<
      mapboxgl.Map | undefined,
      SerializationState | undefined
    >(() => ({
      initial: {
        token,
        mapStyle,
        center,
      },
      serialize: (map: mapboxgl.Map | undefined) => {
        if (!map) return;

        return {
          token,
          mapStyle: map.getStyle() || undefined,
          center: map.getCenter().toArray(),
        };
      },
      deserialize: (state: SerializationState | undefined) => {
        if (isServer) return;

        const s = state || { token, mapStyle, center };

        return new mapboxgl.Map({
          container: ref.value!,
          style: s.mapStyle,
          center: s.center,
        });
      },
      update: (map: mapboxgl.Map | undefined) => {
        if (!map) return;

        map.setStyle(mapStyle);
        map.setCenter(center);
      },
    }));
```

Note: ESLint is still not happy about tracking the `SerializerValue`:
```tsx
    useTask$(({ track }) => {
      const map = track(mapbox);

      map?.on("load", () => console.log("MAP LOADED"));
    });
```
```
5:28:38 PM [vite] warning: When referencing "mapbox" inside a different scope (useTask$), Qwik needs to serialize the value, however "mapbox.force" is a function, which is not serializable.
Check out https://qwik.dev/docs/advanced/dollar/ for more details.
  Plugin: vite-plugin-qwik
  File: qwik-custom-serder/src/components/qwik-mapbox.tsx:65:25
  63 |
  64 |      useTask$(({ track }) => {
  65 |        const map = track(mapbox);
     |                           ^
  66 |
  67 |        map?.on("load", () => console.log("MAP LOADED"));
```

packages/eslint-plugin-qwik/tests/valid-lexical-scope/valid-no-serialize-symbol.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export interface Value {
+  value: number;
+  obj1: NoSerSym;
+  obj2: SerSym;
+}
+
+export const HelloWorld = component$(() => {
+  const state: Value = { value: 12, obj1: new NoSerSym(), obj2: new SerSym() };
```
mhevery, line 26: I am a bit confused as to what happens here:
- What will the `obj1` be after resuming? Presumably `undefined` 
- What will the `obj2` be after resuming? Not sure how the deserializer will restore the value.

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -297,13 +300,17 @@ const inflate = (
...
+    case TypeIds.SerializerSignal:
     case TypeIds.ComputedSignal: {
       const computed = target as ComputedSignal<unknown>;
-      const d = data as [QRLInternal<() => {}>, any, unknown?];
+      const d = data as [QRLInternal<() => {}>, EffectSubscriptions[] | null, unknown?];
       computed.$computeQrl$ = d[0];
-      computed.$effects$ = d[1];
-      if (d.length === 3) {
+      computed.$effects$ = new Set(d[1]);
+      if (d.length >= 3) {
```
mhevery, line 310: `>=3` seems like magic here. Can we pull this out to new function which has a descriptive name so it is clear what condition are we checking here.

packages/qwik/src/core/shared/shared-serialization.ts
```diff
@@ -957,6 +991,7 @@ const discoverValuesForVNodeData = (vnodeData: VNodeData, discoveredValues: unkn
 };
 
 const promiseResults = new WeakMap<Promise<any>, [boolean, unknown]>();
+const serializationResults = new WeakMap<object, unknown>();
```
mhevery, line 994: Can we add comment here as to what `serializationResults` stores?

packages/qwik/src/core/api.md
```diff
@@ -111,14 +111,31 @@ export interface CorrectedToggleEvent extends Event {
 // @public
 export const createComputed$: <T>(qrl: () => T) => T extends Promise<any> ? never : ComputedSignal<T>;
 
+// Warning: (ae-forgotten-export) The symbol "ComputedSignal_2" needs to be exported by the entry point index.d.ts
 // Warning: (ae-internal-missing-underscore) The name "createComputedQrl" should be prefixed with an underscore because the declaration is marked as @internal
 //
 // @internal (undocumented)
-export const createComputedQrl: <T>(qrl: QRL<() => T>) => T extends Promise<any> ? never : ComputedSignal<T>;
+export const createComputedQrl: <T>(qrl: QRL<() => T>) => ComputedSignal_2<T>;
```
mhevery, line 118: Why do we have `ComputedSignal_2` here? Is that OK?

packages/qwik/src/optimizer/src/plugins/rollup.ts
```diff
@@ -247,11 +247,11 @@ export function normalizeRollupOutputOptionsObject(
 }
 
 export function createRollupError(id: string, diagnostic: Diagnostic) {
-  const loc = diagnostic.highlights[0] ?? {};
+  const loc = diagnostic.highlights?.[0];
```
mhevery, line 250: Why are there changes to the optimeter here? This seems unrelated?

mhevery (changes requested):
Very excited about this change. Left few clarifing questions.

packages/qwik/src/optimizer/src/qwik-binding-map.ts
Varixo, line 1: This file should be reverted

packages/qwik/src/core/api.md
wmertens: Hmm this is the class and the signal having the same name, I'll fix

packages/qwik/src/optimizer/src/plugins/rollup.ts
wmertens: it broke in a test due to highlights not always being defined. I presumed it was due to the test environment being different, and the highlights are just passed on from swc so we don't control them.

packages/qwik/src/core/signal/signal-api.ts
```diff
@@ -1,12 +1,33 @@
 import type { QRLInternal } from '../shared/qrl/qrl-class';
 import type { QRL } from '../shared/qrl/qrl.public';
-import { ComputedSignal, Signal, throwIfQRLNotResolved } from './signal';
+import {
+  ComputedSignalImpl,
+  SerializerSignalImpl,
+  Signal as SignalImpl,
```
Varixo, line 6: should we change name of the Signal class to SignalImpl isntead just renaming import? (to be consistent)

Varixo (approved):
LGTM

wmertens:
@ianlet Thank you for your playgrounds, the problem is that you didn't use `foo.value` in the deserializer, and so there's no subscription being created.

I added an eslint rule (yey vibe coding) that warns about this.

PR #7232, opened 2025-01-07
fix: set correct vCurrent after moving the found node

Varixo:
When we rerender a component there is such as situation:
```tsx
const counter = useSignal(0);
...
<CmpA key={counter.value} />
<button onClick$={() => counter.value++}></button>
```
Step 1:
Add a new component before the old one
```html
<CmpA key="1"></CmpA>
<CmpA key="0"> // <-- vCurrent and jsx is a CmpA
  <div></div>
</CmpA>
<button></button>
```
Step 2:
Go to next sibling, find the button and move it up:
```html
<CmpA key="1"></CmpA>
<button></button> <-- jsx is a button
<CmpA key="0"> <-- but the vCurrent is still the old CmpA
  <div></div>
</CmpA>
```

After moving the button we should reassign the vCurrent. It should point now to the button instead


packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -724,6 +724,14 @@ export const vnode_diff = (
       } else {
         // Existing keyed node
         vnode_insertBefore(journal, vParent as ElementVNode, vNewNode, vCurrent);
+        // We are here, so jsx is different from thevCurrent, so now we want to point to the moved node.
```
shairez, line 727: typo

PR #7234, opened 2025-01-08
fix: find correct context after rendering empty array

Varixo:
Fix a problem with slot prop for ssr. Empty arrays should be considered as empty content.

### DIFF
--- a/packages/qwik/src/server/ssr-node.ts
+++ b/packages/qwik/src/server/ssr-node.ts
@@ -141,7 +141,7 @@ export class SsrComponentFrame implements ISsrComponentFrame {
     if (isJSXNode(children)) {
       const slotName = this.getSlotName(children);
       mapArray_set(this.slots, slotName, children, 0);
-    } else if (Array.isArray(children)) {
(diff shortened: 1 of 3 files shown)


packages/qwik/src/server/ssr-node.ts
```diff
@@ -141,7 +141,7 @@ export class SsrComponentFrame implements ISsrComponentFrame {
     if (isJSXNode(children)) {
       const slotName = this.getSlotName(children);
       mapArray_set(this.slots, slotName, children, 0);
-    } else if (Array.isArray(children)) {
```
thejackshelton, line 144: I'm a bit confused. The test mentions we should add slot information for an empty array, but in ssr node we're adding an additional guard to only execute when it's not empty.

wmertens (approved):
Nice!

packages/qwik/src/server/ssr-node.ts
Varixo: yeah, the code which is responsible to add this is below in the else block :D

PR #7236, opened 2025-01-09
feat: log a warning instead of throwing an error for server host mismatch error

Varixo:
Log a warning instead of throwing an error for server host mismatch error

### DIFF
--- a/packages/qwik/src/core/shared/error/error.ts
+++ b/packages/qwik/src/core/shared/error/error.ts
@@ -50,7 +50,6 @@ export const codeToText = (code: number, ...parts: any[]): string => {
       "Element must have 'q:container' attribute.", // 42
       'Unknown vnode type {{0}}.', // 43
       'Materialize error: missing element: {{0}} {{1}} {{2}}', // 44
-      'SsrError: {{0}}', // 45
       'Cannot coerce a Signal, use `.value` instead', // 46
       'useComputedSignal$ QRL {{0}} {{1}} returned a Promise', // 47
       'ComputedSignal is read-only', // 48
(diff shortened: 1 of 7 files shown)


packages/qwik/src/core/shared/error/error.ts
```diff
@@ -50,7 +50,6 @@ export const codeToText = (code: number, ...parts: any[]): string => {
       "Element must have 'q:container' attribute.", // 42
       'Unknown vnode type {{0}}.', // 43
       'Materialize error: missing element: {{0}} {{1}} {{2}}', // 44
-      'SsrError: {{0}}', // 45
```
wmertens, line 53: You can't just delete a line here, each error must be on a single line so that error links work.

If you renumber all errors it's ok but once it's stable you shouldn't renumber any more. What you can do is use old numbers for new errors I suppose.

packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -511,11 +492,8 @@ function choreComparator(
           You are attempting to change a state that has already been streamed to the client.
           This can lead to inconsistencies between Server-Side Rendering (SSR) and Client-Side Rendering (CSR).
           Problematic Node: ${aHost.toString()}`;
-        if (shouldThrowOnHostMismatch) {
-          throw qError(QError.serverHostMismatch, [errorMessage]);
-        }
         logWarn(errorMessage);
-        return null;
+        return 1;
```
wmertens, line 496: this makes sorting unstable. If you don't see a difference, return `0`.

wmertens (changes requested):

packages/qwik/src/core/shared/scheduler.ts
wmertens: actually shouldn't we still sort by host depth?

and the warning should happen when ssrnodes that already streamed are marked dirty, not when tasks rerun, no?

packages/qwik/src/core/shared/error/error.ts
Varixo: yeah I know, it is alpha thats why I removed the error

packages/qwik/src/core/shared/scheduler.ts
Varixo: for some reason returning 0 is not working, I will look at it

wmertens (approved):
LGTM !

PR #7248, opened 2025-01-12
Updated markers.ts to employ the '⚡️' prefix for component styles, enhancing consistency with the Qwik Framework.

sreeisalso:
# What is it?

- Feature / enhancement

# Description

created PR as per discord discussion https://discord.com/channels/842438759945601056/1326121240074780722/1327350404635168852


Shane-Donlon:
Hi Everyone,

While this Change is for an icon, and doesn't require a docs Change in itself.. 

I do think there's also an improvement opportunity for the Docs.

This change was sparked by the question

'What does the star icon represent when inspecting the DOM?' (See Discord link in description)

Inferring from the conversation it seems that this is not clear what the purpose of the emoji is.

While the change in icon should make it more intuitive that it is intentionally added by Qwik I do think we should take the opportunity in this PR to address the initial question too.

For clarity on those that don't know who I am.

I am not from core leadership, but rather offering a question here.

I'll defer the decision on docs change in this PR to Core Leadership, and if it is needed in a separate PR that's okay too. But that second PR will be a dependency on this PR.(As the icon is changing)

wmertens (changes requested):
Actually I think it should not interpolate the prefix in the tests, so that it's easy to see exactly what is being rendered (just like before).

Other than that, it's a theoretically breaking change because some people might be using CSS with those selectors, so let's change the changeset to `minor` instead of `patch`

wmertens:
as for the docs, yes, let's add a little explanation to the `useStyles$` hook to indicate that it's done to clearly show it's generated by Qwik and to prevent clashes.

shairez:
@Shane-Donlon yeah good idea! feel free to add a PR for the docs

shairez:
@wmertens if this has a potential to be a breaking change (although slim), I suggest - let's add it to V2 instead of V1 (it's not a critical bug fix, just a fun change).

WDYT @sreeisalso ?

sreeisalso:
I'm just curious and investigating the Qwik source code. I came across a PR request on Discord and made the necessary change. I'm open to whatever the core team decides.

sreeisalso:
> Actually I think it should not interpolate the prefix in the tests, so that it's easy to see exactly what is being rendered (just like before).

@wmertens but this will make difficult to change the marker, what will be the other option? can I define const in each test file instead of import?

wmertens:
@sreeisalso It's trivial to change the marker with a global search and replace, plus we won't change it any more most likely.

sreeisalso:
@wmertens is this good?

wmertens:
@sreeisalso we discussed this yesterday and we'll merge it in v2 because of the slightly breaking change. Do you want to do the rebase or shall I do it? I'd like to have this in 1 commit, rebased on `build/v2`.

sreeisalso:
@wmertens I'm not good at rebase, please do it yourself, let me know if any changes required.

sreeisalso:
@wmertens check pr #7304

shairez:
thanks @sreeisalso 
closing in favor of #7304

PR #7250, opened 2025-01-13
fix: the useOn hook not working with Slot

### LINKED ISSUE #7230: [🐞] V2 -  useOnDocument is not executed

Qwik Runtime

### Describe the bug

In the example I have a provider that contains a `useOnDocument('load')` and this provider is in the main `layout`, but the `useOnDocument` is never executed.

But if the `useOnDocument('load')` is placed in an `index` of a page, it is executed.

### Reproduction

https://stackblitz.com/edit/github-6f6maw-tnantexa?file=src%2Fproviders%2Fbreakpoint.tsx

### Steps to reproduce

Enter the playground and you will see that the console never executes the `load` message.

You can also check that when the `useOnDocument(‘load’)` is not executed, the pixel information is not displayed.

### System Info

```shell
npmPackages:
    @qwik.dev/core: 2.0.0-alpha.4
    @qwik.dev/router: 2.0.0-alpha.4
    typescript: 5.7.2
    undici: 7.0.0 
    vite: 6.0.7
```

### Additional Information

_No response_

JerryWu1234:
# What is it?

fixes #7230

fix: modify jsx output when it is Slot.
example : 
```js
</Slot>
```

```js
<>
  <script>
  </Slot>
<>
```

# Description

# Checklist


Varixo (changes requested):
Hey, thank you for your contribution, but this is not a good solution. We can't add special condition for every edge case in the vnode-diff.

Tip: the problem is somewhere in this function https://github.com/QwikDev/qwik/blob/ad4444c330abfb995866614187f66cc33d7bbeca/packages/qwik/src/core/shared/component-execution.ts#L133

JerryWu1234:
> Hey, thank you for your contribution, but this is not a good solution. We can't add special conditionl for every edge case in the vnode-diff.
> 
> 
> 
> Tip: the problem is somewhere in this function https://github.com/QwikDev/qwik/blob/ad4444c330abfb995866614187f66cc33d7bbeca/packages/qwik/src/core/shared/component-execution.ts#L133

There are a couple of problems. Q: template also has a problem with usedocument

Varixo:
> > Hey, thank you for your contribution, but this is not a good solution. We can't add special conditionl for every edge case in the vnode-diff.
> > Tip: the problem is somewhere in this function https://github.com/QwikDev/qwik/blob/ad4444c330abfb995866614187f66cc33d7bbeca/packages/qwik/src/core/shared/component-execution.ts#L133
> 
> There are a couple of problems. Q: template also has a problem with usedocument. However I will find a better way to do it. And will tell you when I finish it

I think we need to just modify jsx output to be:
```
<>
  <script ... />
  <Slot />
</>
```
instead of just `<Slot />`

Varixo:
if you need more help with that let me know

packages/qwik/src/core/tests/use-on.spec.tsx
```diff
@@ -647,4 +649,76 @@ describe.each([
     await trigger(document.body, 'div', 'click');
     await expect(document.querySelector('div')).toMatchDOM(<div>1</div>);
   });
+
+  it('issue 7230, when multiple useOnWindow are used in a component that is not rendered, it should add multiple script nodes', async () => {
```
Varixo, line 653: please rename it to:
`#7230 - when multiple useOn are used in a component that is not rendered, it should add multiple script nodes`

packages/qwik/src/core/tests/use-on.spec.tsx
```diff
@@ -647,4 +649,76 @@ describe.each([
...
+            </Component>
+            <script type="placeholder" hidden></script>
+            <script type="placeholder" hidden></script>
+            <script type="placeholder" hidden></script>
+          </Component>
+        </Component>
+      </Component>
+    );
+  });
+  it('issue 7230, when useOnDocument is used in a component that is not rendered, it should add a script node', async () => {
```
Varixo, line 693: please rename it to:
`#7230 - when useOnDocument is used in a component that is not rendered, it should add a script node`

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -101,7 +102,13 @@ export const executeComponent = (
       (jsx) => {
         const useOnEvents = container.getHostProp<UseOnMap>(renderHost, USE_ON_LOCAL);
         if (useOnEvents) {
-          return maybeThen(addUseOnEvents(jsx, useOnEvents), () => jsx);
+          let _jsx = jsx;
+          if (_jsx && (_jsx as JSXNodeInternal<FunctionComponent>).type === Slot) {
+            _jsx = _jsxSorted(Fragment, null, null, [jsx], 0, null);
+          }
```
Varixo, line 108: We should do it in a different cleaner way. We already do such as comparisons in the `addScriptNodeForInvisibleComponents`.
Probably we need to change the `addScriptNodeForInvisibleComponents` or the `addUseOnEvent`.

Varixo (changes requested):
Good job! I have added some comments

packages/qwik/src/core/shared/component-execution.ts
JerryWu1234: done

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -176,7 +182,7 @@ function addUseOnEvents(
         }
       }
     }
-    return jsxElement;
+    return _jsx || jsx;
```
Varixo, line 185: Just `_jsx` isn't enough?

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -133,8 +136,9 @@ export const executeComponent = (
 function addUseOnEvents(
   jsx: JSXOutput,
   useOnEvents: UseOnMap
-): ValueOrPromise<JSXNodeInternal<string> | null> {
+): ValueOrPromise<JSXNodeInternal<string> | null | JSXOutput> {
   const jsxElement = findFirstStringJSX(jsx);
+  let _jsx = jsx;
```
Varixo, line 141: I think we don't use `_` convention. Maybe we can name it `jsxResult`?

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -241,11 +252,12 @@ function addScriptNodeForInvisibleComponents(jsx: JSXOutput): JSXNodeInternal<st
     } else {
       jsx.children = [jsx.children, jsxElement];
     }
-    return jsxElement;
+    return [jsxElement, jsx];
   } else if (Array.isArray(jsx) && jsx.length) {
     // get first element
-    return addScriptNodeForInvisibleComponents(jsx[0]);
+    const [jsxElement, _jsx] = addScriptNodeForInvisibleComponents(jsx[0]);
+    return [jsxElement, jsx];
```
Varixo, line 259: just `return addScriptNodeForInvisibleComponents(jsx[0])` doesn't work?

packages/qwik/src/core/ssr/ssr-render-jsx.ts
```diff
@@ -300,7 +301,8 @@ function processJSXNode(
             options.styleScoped,
             options.parentComponentFrame
           );
-          const jsxOutput = applyQwikComponentBody(ssr, jsx, type);
+
+          const jsxOutput: any = applyQwikComponentBody(ssr, jsx, type);
```
Varixo, line 305: Can you remove `any` type here?

packages/qwik/src/core/tests/use-on.spec.tsx
```diff
@@ -21,7 +23,7 @@ Error.stackTraceLimit = 100;
 
 describe.each([
   { render: ssrRenderToDom }, //
-  { render: domRender }, //
+  { render: domRender }, ///
```
Varixo, line 26: Remove additional `/`

Varixo (changes requested):
Awesome work! Nice job! 😃 
I added some comments

packages/qwik/src/core/shared/component-execution.ts
JerryWu1234: yes.

package.json
```diff
@@ -235,7 +235,7 @@
     "tsc.trace": "tsc -p tsconfig.json --traceResolution > tsc.log",
     "tsc.watch": "tsc --noEmit --watch --preserveWatchOutput",
     "update.qwik.builds": "tsx --require ./scripts/runBefore.ts scripts/update-qwik-builds.ts packages/docs && tsx scripts/update-qwik-builds.ts packages/insights; pnpm install",
-    "vitest": "vitest"
+    "vitest": "vitest run use-visible-task.spec"
```
wmertens, line 238: Remove this

You can just do `npx vitest use-visible` BTW

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -101,7 +102,9 @@ export const executeComponent = (
       (jsx) => {
         const useOnEvents = container.getHostProp<UseOnMap>(renderHost, USE_ON_LOCAL);
         if (useOnEvents) {
-          return maybeThen(addUseOnEvents(jsx, useOnEvents), () => jsx);
+          return maybeThen(addUseOnEvents(jsx, useOnEvents), (_jsx) => {
+            return _jsx;
```
wmertens, line 106: Actually, the maybeThen is only here to catch errors, and if it errors it will now return the error object. 

@Varixo can this even error?

wmertens (commented):
Looks great I just wonder about that maybeThen

packages/qwik/src/core/shared/component-execution.ts
Varixo: we use this because we need to await for addUseOnEvents if it is a promise

Varixo: @JerryWu1234 Wout is right. Can we just return `addUseOnEvents(jsx, useOnEvents)` directly?

JerryWu1234: done

devcaeg:
Hello!

Sorry for being nosy, but the “return <Slot />” error not only affected “useOn” but also the “useTask$”, “useTaskVisible$”, was it taken into account in this PR?

PR #7259, opened 2025-01-15
fix: retry vnode diffing on promise throw

Varixo:
- throwing a promise is our way to await for QRL to resolve. We need to catch the promise while vnode diffing
- during effects cleanup we shouldn't read the props directly and sometimes create a new subscription


wmertens (commented):
LGTM

PR #7269, opened 2025-01-19
refactor(core): schedule QRLs instead of direct execution

wmertens:
This makes it possible to call functions from core in QRLs, and uses it to schedule event handlers properly.

This removes the haphazard _hW exports from segments and replaces them with a single core export `_task`.

This is cleaner and allows core to provide qrl functions.


packages/qwik/src/core/client/queue-qrl.ts
```diff
@@ -0,0 +1,27 @@
...
+  // This will already check container
+  const [runQrl] = useLexicalScope<[QRLInternal<(...args: unknown[]) => unknown>]>();
+  const context = getInvokeContext();
+  const el = context.$element$!;
+  const containerElement = _getQContainerElement(el) as HTMLElement;
+  const container = getDomContainer(containerElement);
+
+  const scheduler = container.$scheduler$;
+  if (!scheduler) {
+    throw new Error('No scheduler found');
```
Varixo, line 23: maybe we should throw the `qError` here?

packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -247,6 +262,10 @@ export const createScheduler = (
     if (isTask) {
       (hostOrTask as Task).$flags$ |= TaskFlags.DIRTY;
     }
+    if (type === ChoreType.RUN_QRL) {
+      // Bind the qrl to the current context
+      targetOrQrl = (targetOrQrl as QRLInternal<(...args: unknown[]) => unknown>).getFn();
+    }
```
Varixo, line 268: Can't we do that inside the `executeChore` function?

packages/qwik/src/optimizer/src/manifest.ts
```diff
@@ -270,7 +274,13 @@ export function generateManifestFromBundles(
     return canonPath(bundle.fileName);
   };
   // We need to find our QRL exports
-  const qrlNames = new Set([...segments.map((h) => h.name)]);
+  const qrlNames = new Set(segments.map((h) => h.name));
+  for (const symbol of extraSymbols) {
+    qrlNames.add(symbol);
+  }
+  const taskNames = new Set(
+    segments.filter((h) => /use[a-zA-Z]*Task(_\d+)?$/.test(h.displayName)).map((h) => h.name)
```
Varixo, line 282: maybe we can create regexp once and reuse it? not sure how about performance here

wmertens: it gets optimized, I tested it a while back

Issue #7260, opened 2025-01-15
[🐞] V2 Error using Tasks


damianpumar:
### Which component is affected?

Qwik Runtime

### Describe the bug

With this example, https://qwik.dev/docs/getting-started/#5-tasks-and-invoking-server-code with version 2 does not work
<img width="1439" alt="Screenshot 2025-01-15 at 23 53 40" src="https://github.com/user-attachments/assets/138c5b6f-406f-4d10-8377-ca042a4696a5" />

### Reproduction

-

### System Info

```shell
System:
    OS: macOS 15.0
    CPU: (10) arm64 Apple M2 Pro
    Memory: 142.47 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
    pnpm: 9.14.2 - ~/Library/pnpm/pnpm
    bun: 1.1.36 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 131.1.73.101
    Chrome: 131.0.6778.264
    Firefox Nightly: 135.0a1
    Safari: 18.0
  npmPackages:
    typescript: 5.4.5 => 5.4.5 
    vite: 5.4.10 => 5.4.10
```

### Additional Information

I'm working on fix it

JerryWu1234:
I can try it

JerryWu1234:
Let me discover

JerryWu1234:
Let me try it

Varixo:
@JerryWu1234 Damian said at the end he is working on it already

PR #7278, opened 2025-01-25
Fix vNode serialization issue on server$

damianpumar:
Fix issue #7260 skipping the vNode serialization from server$


Varixo (approved):
LGTM

PR #7283, opened 2025-01-27
fix(core dev): copy optimizer from CI build if dev

wmertens:
this makes sure that recent optimizer changes are used when not building the optimizer

### DIFF
--- a/scripts/binding-platform.ts
+++ b/scripts/binding-platform.ts
@@ -1,9 +1,8 @@
-import { type BuildConfig, ensureDir } from './util';
 import spawn from 'cross-spawn';
-import { join } from 'node:path';
-import semver from 'semver';
+import { copyFile, writeFile } from 'fs/promises';
 import { existsSync } from 'node:fs';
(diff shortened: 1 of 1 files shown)


scripts/binding-platform.ts
```diff
@@ -55,33 +54,78 @@ export async function copyPlatformBindingWasm(config: BuildConfig) {
...
+  if (version.startsWith('2')) {
+    // 6903 is the PR that builds v2
+    packageName = `@qwik.dev/core@${isDev ? '6903' : version}`;
+  } else {
+    packageName = `@builder.io/qwik@${isDev ? 'main' : version}`;
+  }
+
+  let cacheVersionDir: string;
+  if (isDev) {
+    // We fetche from pkg.pr.new which is a CDN for the CI builds
```
gioboa, line 74:
```suggestion
    // We fetch from pkg.pr.new which is a CDN for the CI builds
```

gioboa (dismissed):
Looks great @wmertens

PR #7284, opened 2025-01-27
fix: don't wrap and serialize functions that are attribute values

Varixo:
Do not convert attribute value function call to _fnSignal. This way it will create component-level subscription like in v1 and should work after resuming the application


wmertens (approved):
LGTM

PR #7289, opened 2025-01-28
Expose missing types & fix package.json internal expose uri

GrandSchtroumpf:
# What is it?

# Description
Expose `_serialize`, `_deserialize` & `_getContextElement` to make the `createWorker$` from RFC works with v2
Expose `OnVisibleTaskOptions` because it was missing

# Checklist

### DIFF
--- a/packages/qwik/public.d.ts
+++ b/packages/qwik/public.d.ts
@@ -1,7 +1,4 @@
 export {
-  _serialize,
-  _deserialize,
-  _getContextElement,
   $,
   ClassList,
   Component,
--- a/packages/qwik/public.d.ts
+++ b/packages/qwik/public.d.ts
@@ -1,4 +1,7 @@
 export {
+  _serialize,
+  _deserialize,
+  _getContextElement,
   $,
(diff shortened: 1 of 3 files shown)


packages/qwik/public.d.ts
```diff
@@ -1,4 +1,7 @@
 export {
+  _serialize,
+  _deserialize,
+  _getContextElement,
```
Varixo, line 4: This is wrong. We can't add this to the public API, because it's internal API. I think the functions are already exported as internal API here 
https://github.com/QwikDev/qwik/blob/acf5375d70471641fc8e7e4f3c80f6ea93194b1b/packages/qwik/src/core/internal.ts#L12
https://github.com/QwikDev/qwik/blob/acf5375d70471641fc8e7e4f3c80f6ea93194b1b/packages/qwik/src/core/internal.ts#L37

So you need to import it from `@qwik/core/internal`

Varixo (changes requested):
Thank you for your PR. I added some comments

PR #7297, opened 2025-01-29
feat: disable font preloading option in qwikVite

thejackshelton:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


thejackshelton:
cc @QwikDev/core-veterans the font preloading code will be removed entirely post v2 and replaced with fontless. 

I was considering doing an e2e test that verifies the link tags existence, but if this code is going to be removed entirely in v2, maybe a test in vitest with a mock of whether the code executes or not would be better?

I have verified that when `disableFontPreload: true` is set in qwikVite that it no longer preloads these fonts.

I should also mention that the code that preloads these fonts is already flaky as is, some self hosted fonts don't get preloaded altogether, so this functionality is currently brittle to begin with. I have a demo project that was able to reproduce the font preloading behavior.

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -1000,6 +1001,15 @@ interface QwikVitePluginCommonOptions {
...
   experimental?: (keyof typeof ExperimentalFeatures)[];
+
+  /**
+   * Disables automatic preloading of font assets (WOFF/WOFF2/TTF) found in the build output. When
+   * enabled, the plugin will not add `<link rel="preload">` tags for font files in the document
+   * head.
+   *
+   * Disabling may impact Cumulative Layout Shift (CLS) metrics.
+   */
+  disableFontPreload?: boolean;
```
wmertens, line 1012: Since CI passes and there's no API changes, I think you need to put this jsdoc elsewhere

thejackshelton: Where would I put it? That seemed to be the place where the other comments were and the functionality is working

wmertens (approved):
LGTM

PR #7306, opened 2025-02-03
fix(ci): make pnpm work again

wmertens:
this fixes the problems caused by https://github.com/nodejs/corepack/issues/612

We no longer rely on corepack to get the correct pnpm version, and we hope that `pnpm/action-setup@v4` will do the right thing.


gioboa (approved):
It looks good to me 
@wmertens thanks

PR #7310, opened 2025-02-04
fix: rendering attribute value from array of classes from spread props

Varixo:


packages/qwik/src/core/shared/shared-container.ts
```diff
@@ -48,6 +49,10 @@ export abstract class _SharedContainer implements Container {
     return trackSignalAndAssignHost(signal, subscriber, property, this, data);
   }
 
+  serializeAttributeValue(key: string, value: any, styleScopedId?: string | null) {
```
wmertens, line 52: Why not simply re-export?

Varixo: yes, I forgot about reverting that change

wmertens (approved):
LGTM

Issue #7216, opened 2024-12-31
[🐞] Cannot overwrite event handlers with spread props


wmertens:
### Which component is affected?

Qwik Optimizer (rust)

### Describe the bug

If you do `<div onChange$={handler} {...props} />` then the optimizer converts it into basically `<div {...props} onChange$={handler} />`.

### Reproduction

https://qwik.dev/playground/#f=Q040ZsbUTDTg5gFmykFUl9UamrZ2yHW0ekamumatQrWenh5Yc%2B1oKhsmqQwA

### System Info

```shell
v1 and v2
```

JerryWu1234:
What is your expected value?

JerryWu1234:
@wmertens Could you please tell me what you expected ?

JerryWu1234:
the version 2 was fixed.

PR #7316, opened 2025-02-06
fix: optimizer is now better at recognizing constProp

JerryWu1234:
# What is it?

fixes: #7216 

Optimize the Optimize, has been better at recognizing constProp

# Description

# Checklist


.changeset/lemon-tools-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: a bug #7216, that const prop was putted in the var
```
wmertens, line 5: `FIX: optimizer is now better at recognizing constProp`

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__ternary_prop.snap
```diff
@@ -65,15 +65,15 @@ export const Cmp_component_4ryKJTOKjWE = ()=>{
         toggleSig
     ]);
     return /*#__PURE__*/ _jsxSorted("button", null, {
+        onClick$: handleClick$,
```
wmertens, line 68: This is now wrong. This function is a qrl with const captures, so it should be const.

packages/qwik/src/optimizer/core/src/transform.rs
```diff
@@ -1389,7 +1386,16 @@ impl<'a> QwikTransform<'a> {
...
 									));
-									event_handlers.push(converted_prop);
+									if !is_const {
+										static_listeners = false;
+										var_props.push(converted_prop.clone().fold_with(self));
+									} else if spread_props_count > 0 {
+										maybe_const_props
+											.push(converted_prop.clone().fold_with(self));
+									} else {
+										const_props.push(converted_prop.clone().fold_with(self));
```
wmertens, line 1396: you can just always push to maybe_const_props, no need to check for the spread props again.

wmertens (changes requested):
Good start, but it looks like `_create_synthetic_qsegment` is returning is_const=false because it's a simple identifier, and it doesn't know if the identifier is const.

So we need to pass the const identifiers to that function.

It's important to get this right because putting a handler in a variable before passing it as a handler is very common.

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__ternary_prop.snap
wmertens: oops I misread this, it's still const just ordered differently

wmertens (approved):
LGTM

PR #7143, opened 2024-12-10
Upgrade eslint from 8 to 9

JerryWu1234:
# What is it?

- upgrade the ESlint version from 8 to 9, 

# Description

# Checklist

- [ ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [ ] I performed a self-review of my own code
- [ ] I added a changeset with `pnpm change`
- [ ] I made corresponding changes to the Qwik docs
- [ ] I added new tests to cover the fix / functionality

PR #7318, opened 2025-02-06
fix: upgraded to vite 6

sreeisalso:
# What is it?

- Feature / enhancement

# Description

upgraded to vite 6

# Checklist


package.json
```diff
@@ -174,7 +174,7 @@
     "typescript": "5.4.5",
     "undici": "*",
     "vfile": "6.0.2",
-    "vite": "5.3.5",
+    "vite": "catalog:",
```
sreeisalso, line 177: @Varixo @wmertens  is this allowed, or shoud I cahnge it to 6.1.0

wmertens: no you should use `pnpm deps` and upgrade via that. Then you should also make sure that the published deps allow vite 5 still.

package.json
```diff
@@ -28,7 +28,7 @@
           "dependencies": [
             "vite"
           ],
-          "pinVersion": "^5"
+          "pinVersion": "^6"
```
wmertens, line 31: this should stay ^5 unless we somehow become incompatible with vite 5

sreeisalso: ok

sreeisalso:
sorry for wasting the resource, I'm bumping into more errors, will analyse code and do the changes slowly.

sreeisalso:
will try again once #7143 is merged into main

PR #7323, opened 2025-02-07
fix: create svg nested children with correct namespace

Varixo:
In some cases vNodes didn't have the correct flags for the namespace


gioboa (approved):
It looks great 👍 
Thanks @Varixo

PR #7329, opened 2025-02-10
Fire the hashchange event even if the navigation changes only `location.hash`

genki:
# What is it?

- Bug fix

# Description

I am making an app with Qwik that is using SPA navigation between pages including URL fragment.
Currently, the SPA navigation doesn't fire the `hashchange` event if the change of URL is only the hash fragment because it is using `history.pushState`.
This is not the standard behaviour of the browser.

So I have changed this
* only if the change is consist of the fragment of URL.
* then navigate by using `location.hash = ...` due to fire the hashchange event.

# Checklist


JerryWu1234:
it would be great if you add a test for it

JerryWu1234:
@genki Could you add a test for it ?

genki:
@JerryWu1234 Sorry currently I can't take a time for a while 🥲

wmertens (changes requested):
LGTM but really needs a test

gioboa (changes requested):
may I ask you to add the missing test here please?

PR #7334, opened 2025-02-12
perf: don't bundle core in server or optimizer

wmertens:


packages/qwik/src/core/shared/utils/styles.ts
```diff
@@ -93,7 +92,7 @@ export function serializeAttribute(
   } else if (value === true && isPreventDefault(key)) {
     value = '';
   }
-  return value;
+  return value as string | null;
```
Varixo, line 95: why we need this cast?

Varixo (approved):

PR #7341, opened 2025-02-14
chore(docs): update rehype-pretty-code

sreeisalso:
# What is it?

- Feature / enhancement

# Description

  Updated Rehype-pretty-code to latest, no braking changes but reduced the dependency size

# Checklist


gioboa (approved):
Thanks @sreeisalso for your help

Shane-Donlon:
Hey Everyone,

I am not sure if a roll-back is the answer, but it may be.

It seems since this PR that the docs site has the yellow highlight in it.
Which visually this has poor contrast.

Example:
[https://qwik.dev/docs/components/events/](https://qwik.dev/docs/components/events/)


screenshot showing the issue as described above


I did a roll-back on this locally and it has removed this.

This is being caused by the inclusion of the HTML `<mark>` element with yellow being the default Agent Style for this element.

html source code with mark element being highlighted

While we can fix this in CSS (visually) 
Which would be this file 

[https://github.com/QwikDev/qwik/blob/main/packages/docs/src/components/code-sandbox/index.css](https://github.com/QwikDev/qwik/blob/main/packages/docs/src/components/code-sandbox/index.css)

Then we do need to consider this from an accessibility perspective and is the `<mark>` element appropriate here.

I'm not sure which approach is the best approach to take here, so thought I'd ask.

Let me know what you think!!
Thanks!

gioboa:
Thanks for the report @Shane-Donlon , nice catch
Can you open a PR to fix the accessibility issue please?

gioboa:
Yep, go for it @sreeisalso 👏

Shane-Donlon:
@sreeisalso 
Thank you so much! 
The roll-back was temporary to buy me some time while I was researching this, so you definitely saved me a weekend of headaches :rofl: 


Although if you do have some good learning resources I'd definitely appreciate some links so that I can avoid doing this (roll-back) in future!  :heart:

Issue #7227, opened 2025-01-05
[🐞] `useErrorBoundary` has no effect


ryuujo1573:
### Which component is affected?

Qwik City (routing)

### Describe the bug

<pre>
qwik vite-dev-server symbolMapper: parent not provided for error, falling back to mapper.
qwik vite-dev-server symbolMapper: unknown qrl requested without parent: error
</pre>

### Reproduction

https://stackblitz.com/edit/github-ur8va7yg?file=src%2Froutes%2Flayout.tsx

### System Info

```shell
System:
    OS: macOS 14.4.1
    CPU: (8) arm64 Apple M2
    Memory: 110.02 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - ~/Library/pnpm/node
    Yarn: 1.22.22 - ~/Library/pnpm/yarn
    npm: 10.9.0 - ~/Library/pnpm/npm
    pnpm: 9.4.0 - ~/Library/pnpm/pnpm
    bun: 1.1.8 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.205
    Chrome Canary: 128.0.6540.0
    Edge: 131.0.2903.112
    Firefox Nightly: 127.0a1
    Safari: 17.4.1
```

### Additional Information

There's no available official document as for now.

damianpumar:
Hello @ryuujo1573

Thank you for reporting this issue, but I can't reproduce it.

I created this error boundary component:

```
export const ErrorBoundary = component$(() => {
  const error = useErrorBoundary();

  if (error.error) {
    return <div>Error: {error.error}</div>;
  }

  return <Slot />;
});
```

Then, I threw an error from the `src/routes/index.tsx` component, and everything seems to work fine!

damianpumar:
Update:
@ryuujo1573 
After testing this ErrorBoundary, I'm getting this warning: 
`qwik vite-dev-server symbolMapper: unknown qrl requested without parent: error`

It works, but I see this warning, I will be reviewing it.! 

Thanks!

damianpumar:
Hello @ryuujo1573 
After test it several times, I'm still getting the same error, I will be working on fix it.

Thanks!

damianpumar:
> Hello @ryuujo1573 
> After test it several times, I'm still getting the same error, I will be working on fix it.
> 
> Thanks!

Fixed!

PR #7342, opened 2025-02-14
fix: error boundary

damianpumar:
This PR fix https://github.com/QwikDev/qwik/issues/7227


.changeset/afraid-months-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/qwik-city': patch
+'@builder.io/qwik': patch
+---
+
+Fix error boundary
```
Varixo, line 6: we should use `fix: ` prefix for changes
Maybe we could also inform users more about changes?

packages/qwik/src/core/use/use-error-boundary.ts
```diff
@@ -1,16 +1,11 @@
...
-    error: undefined,
-  });
-  useOn('error-boundary', qrl('/runtime', 'error', [store]));
-  useContextProvider(ERROR_CONTEXT, store);
+export const useErrorBoundary = () => {
+  const error = useStore<ErrorBoundaryStore>({ error: undefined });
+  useContextProvider(ERROR_CONTEXT, error);
 
-  return store;
+  return useContext(ERROR_CONTEXT);
```
Varixo, line 10: Can we just return `error` object here?

damianpumar: Let me try!
Thanks mate for your suggestion!

Varixo (approved):

damianpumar:
Any news about this fix @QwikDev/api-guards  ?

PR #7344, opened 2025-02-16
fix: Remove all default content type fastify parsers

nelsonprsousa:
Currently, Fastify/Qwik can not parse body on `onPost` requests due to the following exception:

`SyntaxError: Unexpected end of JSON input`

By removing the default content type parsers seems to let Fastify+Qwik handle the request body as expected.


nelsonprsousa:
@wmertens

wmertens (approved):
LGTM, thanks @nelsonprsousa !

PR #7346, opened 2025-02-16
fix(core): performance improvements

Varixo:
Speed up CSR rendering.
For 10k interactive components down from 600ms to 250ms.
- rewrite cleaning up subscriptions
- use for of instead of foreach
- change a little bit VNodes structure for props
- reuse existing subscription for signals and stores


wmertens (approved):
👏 LGTM let's gooooo

PR #7368, opened 2025-02-23
fix: signals back refs serialization

Varixo:


packages/qwik/src/core/signal/signal.ts
```diff
@@ -32,7 +32,7 @@ import type { OnRenderFn } from '../shared/component.public';
 import { _EFFECT_BACK_REF, NEEDS_COMPUTATION } from './flags';
 import { QError, qError } from '../shared/error/error';
 import { isDomContainer } from '../client/dom-container';
-import type { BackRef } from './signal-cleanup';
+import { BackRef } from './signal-cleanup';
```
wmertens, line 35: Can't it still be a type?

Varixo: right

wmertens (approved):
↩️

Issue #7373, opened 2025-02-24
[🐞] v2 Error “Uncaught (in promise) TypeError: Cannot read properties of undefined (reading ‘0’)”


devcaeg:
### Which component is affected?

Qwik Runtime

### Describe the bug

In the production (compiled and minified) version of the project, the following error appears:

```ts
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
```
This error is flagged in the compiled code at the following line:

```ts
, ce = e => (e[_.flags] & O.Element) === O.Element.
```
This issue only occurs in production, and the cause is still unclear.

While debugging the error in production (which is challenging due to code minification), the functions that appear to fail include:

- node_isElementVNode
- vnode_isMaterialized
- Other functions starting with vnode_is

### Reproduction

Missing

### System Info

```shell
System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M2
    Memory: 469.25 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    bun: 1.2.3 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 133.0.6943.127
    Edge: 133.0.3065.82
    Safari: 18.3
 npmPackages:
    @qwik.dev/core: https://pkg.pr.new/QwikDev/qwik/@qwik.dev/core@da11736
    @qwik.dev/router: https://pkg.pr.new/QwikDev/qwik/@qwik.dev/router@da11736
    typescript: 5.7.3 => 5.7.3 
    vite: 6.1.1 => 6.1.1
```

Varixo:
after #7376 is probably fixed, reopen if needed

PR #7376, opened 2025-02-24
fix: undefined or null as projection child

Varixo:


devcaeg:
If I am not mistaken, this PR also solves issue #7373

I must do more tests to confirm this.

wmertens (approved):

PR #7378, opened 2025-02-25
fix: RequestEventAction has missed a type `resolveValue`

### LINKED ISSUE #7377: [🐞] "resolveValue" Function Is Missing in TypeScript Types and Fails in "routeAction$" / "globalAction$"

Qwik City (routing)

### Describe the bug

The [resolveValue](https://qwik.dev/docs/route-loader/#access-the-routeloader-data-within-another-routeloader) function is documented as being available for `routeLoader$`, `routeAction$`, and `globalAction$`. However, there are two major issues:  

1. **Missing TypeScript Typings**  
   - While `resolveValue` exists in the `requestEvent` of `routeAction$` and `globalAction$`, it is missing from TypeScript types. This results in TypeScript failing to recognize it, leading to type errors.  

2. **Functionality Not Working in Actions**  
   - When attempting to use `resolveValue` in `routeAction$` or `globalAction$`, it throws an error instead of resolving the expected value.  

### **Expected Behavior**  
The code should work similarly to `routeLoader$`, allowing the resolution of a loader's value within an action.

### **Actual Behavior**  
The following error is thrown:  

```
[vite] Internal server error: You can not get the returned data of a loader that has not been executed for this request.
```

This suggests that `resolveValue` is either not supported or has unintended behavior when used in `routeAction$` or `globalAction$`.

### Reproduction

https://stackblitz.com/edit/github-4169k7qj?file=src%2Froutes%2Findex.tsx

### **Steps to Reproduce**  
Using `resolveValue` inside a `routeAction$` function:  

```ts
export const useAction = routeAction$(
  async (data, request) => {
    const user = await request.resolveValue(useUser);
    console.log(user);
  }
);
```

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (8) x64 12th Gen Intel(R) Core(TM) i3-12100F
    Memory: 3.85 GB / 15.82 GB
  Binaries:
    Node: 23.6.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.1.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.15.3 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (131.0.2903.112)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @builder.io/qwik: ^1.12.0 => 1.12.0
    @builder.io/qwik-city: ^1.12.0 => 1.12.0
    vite: 6.1.0 => 6.1.0
```

### Additional Information

Being able to access `routeLoader$` values inside `routeAction$` and `globalAction$` is highly beneficial, as it allows reusing computed loader data without needing to fetch or compute it again.

Would it be possible to officially support this functionality, or is this an unintended limitation? Additionally, the TypeScript types should be updated to reflect `resolveValue`'s presence in `requestEvent`.

JerryWu1234:
# What is it?

fixes #7377 

if routeAction$ wants to use `resolveValue` function，routeLoader$ has be loaded first, 
so I move the `routerLoader` logic before `method === 'POST'`, However, in some situations, it has to remain in its original position.

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/packages/docs/global.d.ts
+++ b/packages/docs/global.d.ts
@@ -3,3 +3,8 @@ declare module '*?raw-source' {
   const url: string;
   export default url;
 }
+
+declare module '*?compiled-string' {
+  const str: string;
+  export default str;
+}
--- a/packages/docs/src/repl/bundler/index.ts
+++ b/packages/docs/src/repl/bundler/index.ts
@@ -4,6 +4,13 @@ import type { ReplInputOptions, ReplResult } from '../types';
 import { getDeps } from './bundled';
 import type { InitMessage, BundleMessage, OutgoingMessage } from './bundler-worker';
 
+import ssrWorkerStringPre from './repl-ssr-worker?compiled-string';
+import listenerScript from './client-events-listener?compiled-string';
+
+export const ssrWorkerString = ssrWorkerStringPre
+  .replace(/globalThis\.DO_NOT_TOUCH_IMPORT/g, 'import')
+  .replace('globalThis.LISTENER_SCRIPT', JSON.stringify(listenerScript));
+
 const bundlers = new Map<string, Bundler>();
 
 class Bundler {
--- a/packages/docs/src/repl/bundler/repl-ssr-worker.ts
+++ b/packages/docs/src/repl/bundler/repl-ssr-worker.ts
@@ -1,8 +1,6 @@
 // SSR Worker - handles server-side rendering execution
 // MUST be served from /repl/ so that its imports are intercepted by the REPL service worker
 import type { QwikManifest } from '@builder.io/qwik/optimizer';
-// @ts-expect-error
-import listenerScript from './client-events-listener?compiled-string';
 
 // Worker message types
 interface MessageBase {
@@ -67,8 +65,8 @@ self.onmessage = async (e: MessageEvent<IncomingMessage>) => {
 async function executeSSR(message: InitSSRMessage): Promise<{ html: string; events: any[] }> {
   const { baseUrl, manifest, entry } = message;
 
-  // @ts-expect-error - we prevent Vite from touching this import and replace it later
-  const module = await DO_NOT_TOUCH_IMPORT(`/repl/${replId}-ssr/${entry}`);
+  // We prevent Vite from touching this import() and replace it after bundling
+  const module = await (globalThis as any).DO_NOT_TOUCH_IMPORT(`/repl/${replId}-ssr/${entry}`);
   const server = module.default;
 
   const render = typeof server === 'function' ? server : server?.render;
@@ -103,7 +101,10 @@ async function executeSSR(message: InitSSRMessage): Promise<{ html: string; even
   });
 
   // Inject the event listener script
-  ssrResult.html = ssrResult.html.replace('</body>', `<script>${listenerScript}</script></body>`);
+  ssrResult.html = ssrResult.html.replace(
+    '</body>',
+    `<script>${(globalThis as any).LISTENER_SCRIPT}</script></body>`
+  );
 
   // Restore console methods
   console.log = orig.log;
--- a/packages/docs/src/repl/ui/repl-instance.ts
+++ b/packages/docs/src/repl/ui/repl-instance.ts
@@ -1,7 +1,7 @@
 /** Maintains the state for a REPL instance */
 
 import { isServer, unwrapStore } from '@builder.io/qwik';
-import { getBundler } from '../bundler';
+import { getBundler, ssrWorkerString } from '../bundler';
 import { registerReplSW } from '../register-repl-sw';
 import type { RequestMessage, ResponseMessage } from '../repl-sw';
 import type { ReplAppInput, ReplResult, ReplStore } from '../types';
@@ -10,10 +10,6 @@ import type {
   InitSSRMessage,
   OutgoingMessage as SSROutgoingMessage,
 } from '../bundler/repl-ssr-worker';
-// @ts-expect-error - we don't have types for this yet
-import ssrWorkerStringPre from '../bundler/repl-ssr-worker?compiled-string';
-
-const ssrWorkerString = ssrWorkerStringPre.replace(/DO_NOT_TOUCH_IMPORT/g, 'import');
 
 let channel: BroadcastChannel;
 let registered = false;


wmertens:
the problem is that the resolveValue is not actually working even though it's documented. So this PR needs to implement it too.

JerryWu1234:
> the problem is that the resolveValue is not actually working even though it's documented. So this PR needs to implement it too.

Is there a code reference I could follow ?

wmertens:
@JerryWu1234 best to search the code for `resolveValue` and see how it is implemented, and then try to implement it in the same way in the other locations.

JerryWu1234:
> @JerryWu1234 best to search the code for `resolveValue` and see how it is implemented, and then try to implement it in the same way in the other locations.

thanks for your tips, it's useful

JerryWu1234:
@wmertens @Varixo please help to review it when you are available

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -257,9 +227,49 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
...
+            loaders[selectedActionId] = actionResolved;
+            isAction = true;
+          }
+        }
+      }
+    }
+    if (routeLoaders.length > 0 && isAction) {
+      await Promise.all(loader() || []);
     }
+    isAction = false;
```
gioboa, line 272: Do we need to reset it to false?

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -257,9 +227,49 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
...
+            }
+
+            loaders[selectedActionId] = actionResolved;
+            isAction = true;
+          }
+        }
+      }
+    }
+    if (routeLoaders.length > 0 && isAction) {
+      await Promise.all(loader() || []);
```
gioboa, line 270: we already run loaders at line 189 :thinking: is it correct?

packages/qwik-city/src/middleware/request-handler/request-event.ts
```diff
@@ -280,7 +279,7 @@ export function createRequestEvent(
       return writableStream;
     },
   };
-  return Object.freeze(requestEv);
+  return requestEv;
```
gioboa, line 282: Can you explain me the `Object.freeze` removal please?

gioboa (changes requested):
Thanks @JerryWu1234 for your help :tada: 
I added few comments

packages/qwik-city/src/middleware/request-handler/request-event.ts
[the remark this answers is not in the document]
wmertens: @JerryWu1234 please put this back

packages/qwik/src/optimizer/src/qwik-binding-map.ts
wmertens, line 1: please remove these changes from the PR (run `git checkout main -- packages/qwik/src/optimizer/src/qwik-binding-map.ts`)

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -184,43 +185,12 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
...
-            }
-            loaders[selectedActionId] = actionResolved;
-          }
-        }
-      }
-    }
 
     if (routeLoaders.length > 0) {
-      const resolvedLoadersPromises = routeLoaders.map((loader) => {
+      await Promise.all(loader() || []);
```
wmertens, line 190: I'd prefer `await Promise.all(routeLoaders.map((loader) => ... (previous code) )`

.changeset/eight-garlics-deliver.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/qwik-city': patch
+'@builder.io/qwik': patch
+---
+
+FIX: if routeAction$ wants to use resolveValue function，routeLoader$ has be loaded first,
```
wmertens, line 6: I don't understand how this PR implements what you write. How about this?
```
FIX: add type for `routeAction$`'s `requestEvent.resolveValue`
```

wmertens (changes requested):

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
JerryWu1234: if routeAction$ wants to use resolveValue function，routeLoader$ has be loaded first,
so I move the routerLoader logic before method === 'POST', However, in some situations, it has to remain in its original position.

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
JerryWu1234: We need to prevent it from running multiple times.

.changeset/eight-garlics-deliver.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/qwik-city': patch
+'@builder.io/qwik': patch
```
Varixo, line 3: I think this is qwik-city only change

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -184,6 +185,48 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
         );
       }
     }
+
+    if (routeLoaders.length > 0) {
+      await Promise.all(routeLoaders.map(loader));
+    }
+    function loader(loader: LoaderInternal) {
```
wmertens: please put this inside the map function, it's confusing to have it outside and after, especially with the name `loader` that you also shadow inside it

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -213,53 +257,17 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
...
-            }
-            return resolvedLoader;
-          });
-
-        return loaders[loaderId];
-      });
-
-      await Promise.all(resolvedLoadersPromises);
+    if (routeLoaders.length > 0 && isAction) {
+      await Promise.all(routeLoaders.map(loader));
```
wmertens: I don't understand why we run the routeloaders again? Seems wasteful?
(but now I understand why you moved it to a separate function)

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
wmertens: how does that happen?

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
JerryWu1234: > please put this inside the map function, it's confusing to have it outside and after, especially with the name `loader` that you also shadow inside it

loader be used by two place.
I can't put this inside map.

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
JerryWu1234: > I don't understand why we run the routeloaders again? Seems wasteful? (but now I understand why you moved it to a separate function)

in some edge cases.  routeLoader function has to run again.
Before I just moved that code before POST, but There is still some fail tests.

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
JerryWu1234: @wmertens I changed that name for clarify

wmertens:
@JerryWu1234 so can you explain the fix? You now run the routeLoaders before the action, but then you run them again after the action, why?

wmertens:
Ah because the action can change the loaders, right? 

In that case, I think it's better to make the first routeLoaders run on demand. We should not add the resolveValue synchronous API, but instead an async call like getLoaderValue(loader).

Right now, we're doubling the cost of loaders in actions in this PR

JerryWu1234:
> Ah because the action can change the loaders, right? 
> 
> 
> 
> In that case, I think it's better to make the first routeLoaders run on demand. We should not add the resolveValue synchronous API, but instead an async call like getLoaderValue(loader).
> 
> 
> 
> Right now, we're doubling the cost of loaders in actions in this PR

It's a long time PR, so I  don't remember very clearly. Let me re-read the code tomorrow.

JerryWu1234:
> Ah because the action can change the loaders, right?
> 
> In that case, I think it's better to make the first routeLoaders run on demand. We should not add the resolveValue synchronous API, but instead an async call like getLoaderValue(loader).
> 
> Right now, we're doubling the cost of loaders in actions in this PR

```js

export const useAddUser = routeAction$(
  async (data, requestEvent) => {
    const res = await requestEvent.resolveValue(useUser);
    res.firstName = data.firstName;
    res.lastName = data.lastName;
    return {
      success: true,
      userID: res,
    };
  },
  zod$({
    firstName: z.string(),
    lastName: z.string(),
  }),
);
```

you meant that user has to use `getLoaderValue ` instead of `resolveValue`?
@wmertens

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -268,6 +234,41 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
 
       await Promise.all(resolvedLoadersPromises);
     }
+
+    if (method === 'POST') {
```
gioboa, line 238: moved the POST section after the loaders resolution.

JerryWu1234:
@gioboa 
image

image

https://github.com/JerryWu1234/qwik/blob/377cd5dc92cfdc6daa3cf3a3f3c7d80b7c730b8c/packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts#L267



There are some `LoaderInternal`s still needs to be loaded after POST.

So Why I executed `routeLoaders` twice

wmertens:
hmm looks like the e2e tests actually aren't passing, I tried 3 times and normally it's only the macos one that is slow

JerryWu1234:
> hmm looks like the e2e tests actually aren't passing, I tried 3 times and normally it's only the macos one that is slow

this method isn't fully fix this problem.
I tried my local. the wrong is I wrote e2e that didn't pass .

gioboa:
Nice @JerryWu1234 
We don't want to run loaders twice, before and after because the call must be unique.
Immagine a db in the loader, should run only one time. Why is it failing the test? That's the question I guess

JerryWu1234:
> Why is it failing the test? That's the question I guess

```js
export const globalAction = globalAction$(
  async (data, requestEvent) => {
    const res = await requestEvent.resolveValue(useUserGlobal);
    if (res) {
      res.firstName = data.globalfirstName;
      res.lastName = data.globallastName;
    }
    return {
      success: true,
      userID: res,
    };
  },
  zod$({
    globalfirstName: z.string(),
    globallastName: z.string(),
  }),
);
```

because in globalAction function that involved `routerloader`.

gioboa:
I see, we already executed the loader though 🤔

PR #7384, opened 2025-02-27
fix: using routeLoader$ result as component prop

Varixo:
Fix incorrectly unwrapping WrappedSignal value for routeLoader$ data


wmertens (approved):
LGTM 🫎

PR #7386, opened 2025-02-28
fix: wrap WrappedSignal without unwrap flag

Varixo:


wmertens (approved):
LGTM

markovickosta:
It is working as expected, thank your for quick fix!

Issue #6585, opened 2024-06-20
[🐞] The tagged function depending on the signal value doesn't make reactive.


genki:
### Which component is affected?

Qwik Runtime

### Describe the bug

I am making a tag function that is reactively changing the content when the signal value changes.
It works when the function called as the normal function, but when it is called as the tag function then it doesn't work.
This may be relating to #5453

Here's the code snippet

```tsx
const useFoo = (count:Signal<number>) => {
  const tag = (s:string|TemplateStringsArray) => {
    const value = typeof s === "string" ? s : s[0];
    return `${value}-${count.value}`;
  }
  return tag;
}

export default component$(() => {
  const count = useSignal(0);
  const foo = useFoo(count);
  return (
    <>
      <p>{foo("test")}</p>
      <p>{foo`test`}</p>
      <div style={{ padding: "1rem" }}>Hello World</div>
      <button onClick$={() => {count.value++}}>Count up</button>
    </>
  );
});
```

### Reproduction

https://stackblitz.com/edit/qwik-starter-b8qgsc?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

Please visit the link above and click the `Count up` button to check if there's difference.

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 109.73 MB / 24.00 GB
    Shell: 3.6.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.12.1 - /opt/homebrew/opt/node@20/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.5.0 - /opt/homebrew/opt/node@20/bin/npm
    pnpm: 9.3.0 - ~/Library/pnpm/pnpm
    bun: 1.1.3 - ~/.bun/bin/bun
  Browsers:
    Chrome: 126.0.6478.114
    Safari: 17.5
  npmPackages:
    @builder.io/qwik: file:../clone/qwik/packages/qwik/dist => 1.5.6-dev20240611011722 
    @builder.io/qwik-city: file:../clone/qwik/packages/qwik-city/lib => 1.5.6-dev20240611011722 
    typescript: ^5.3.3 => 5.3.3 
    undici: ^5.28.4 => 5.28.4 
    vite: ^5.3.1 => 5.3.1
```

wmertens:
Indeed this is a bug. The difference is this:

```js
[...]

PR #7390, opened 2025-03-02
fix: don't wrap template literals with a function call inside in a si…

Varixo:


wmertens:
I think this breaks https://github.com/QwikDev/qwik/issues/6585 again

wmertens (approved):
LGTM

PR #7401, opened 2025-03-07
chore: refactor vSiblings

Varixo:
Remove vSiblings logic, they are no longer needed. Replace them with a Map for keyed nodes and with an array for non-keyed nodes to speed up diffing


wmertens (approved):
LGTM!

Issue #6115, opened 2024-04-11
[✨] Update your template to include eslint.config.js by default


hosseinitabar:
### Is your feature request related to a problem?

Yes. I used `npm create qwik` and it created a project based on `1.5.2` version. Yet when I updated it to use `eslint 9.0.0` and ran `npm run build` I got this error:

> node:748) ESLintIgnoreWarning: The ".eslintignore" file is no longer supported. Switch to using the "ignores" property in "eslint.config.js"

### Describe the solution you'd like

Please remove `.eslintignore` and `.eslintconfig.cjs` from your template and replace them with a single `eslint.config.js`.

### Describe alternatives you've considered

Right now we should manually do this in every project.

hosseinitabar:
@gioboa, I'm sorry I don't know how to do that.

JerryWu1234:
let me try it

JerryWu1234:
Will this bug from v1 be merged into v2?

gioboa:
Yep, it will be merged. Thanks for your help @JerryWu1234

JerryWu1234:
because the eslint 9.0 was a breaking change version, and I have used this library `@eslint/eslintrc` which the official have offered for migrate to v9.0,
But it's still unstable  

https://github.com/eslint/eslintrc/issues/158

wmertens:
I think we should use the ts one that helps with the types https://typescript-eslint.io/getting-started/

wmertens:
BTW @JerryWu1234 you need to rename your file to .cjs or it won't work in type:module projects. And maybe you should make it ESM with .mjs, but then __dirname needs to be replaced with the equivalent import.meta statement

JerryWu1234:
> I think we should use the ts one that helps with the types https://typescript-eslint.io/getting-started/

we just upgrade the Eslint's grade. and follow[ this doc](https://eslint.org/docs/latest/use/migrate-to-9.0.0)  don't you?,
So why we have to refer a new library?

>BTW @JerryWu1234 you need to rename your file to .cjs or it won't work in type:module projects. And maybe you should make it ESM with .mjs, but then __dirname needs to be replaced with the equivalent import.meta statement

I tried both ways, it's still wrong.

JerryWu1234:
https://github.com/typescript-eslint/typescript-eslint/issues/8211

still going on

PatrickJS:
yup it would be awesome to convert those two files into one

PR #7418, opened 2025-03-13
Migrate to ESLint 9 with flat config and fix monorepo linting

better-salmon:
# What is it?

- Bug

# Description

This PR fixes a significant issue where ESLint wasn't functioning properly (silently crashing) across most of the monorepo files. The changes include:

1. Migration from legacy ESLint config to ESLint 9 with flat configuration
2. Improved lint script execution by properly separating local ESLint configs from the root config (ensuring they no longer overlap — errors in IDE and from the script will be the same)
3. Removal of unnecessary eslint-disable comments throughout the codebase
4. Fixed all new errors introduced by the newer config that were previously missed while ESLint wasn't working
5. Use `eslint.config.js` by default in starters (Fix: #6115)

These improvements ensure proper linting across the entire codebase and enable better code quality enforcement going forward.

# Checklist


package.json
```diff
@@ -208,7 +197,7 @@
     "build.clean": "rm -rf packages/qwik/dist/ && rm -rf packages/qwik-city/lib/ && rm -rf packages/docs/dist/ && rm -rf packages/insights/dist/ && rm -rf packages/qwik-labs/lib/ && rm -rf packages/qwik-labs/vite/",
     "build.cli": "tsx --require ./scripts/runBefore.ts scripts/index.ts --cli --dev",
     "build.cli.prod": "tsx --require ./scripts/runBefore.ts scripts/index.ts --cli",
-    "build.core": "tsx --require ./scripts/runBefore.ts scripts/index.ts --tsc --build --qwikcity --api --platform-binding",
+    "build.core": "tsxs --require ./scripts/runBefore.ts scripts/index.ts --tsc --build --qwikcity --api --platform-binding",
```
wmertens, line 200: typo?

packages/docs/src/routes/api/qwik/api.json
```diff
@@ -1774,7 +1774,7 @@
...
...script\nPrefetchServiceWorker: (opts: {\n    base?: string;\n    scope?: string;\n    path?: string;\n    verbose?: boolean;\n    fetchBundleGraph?: boolean;\n    nonce?: string;\n}) => JSXNode<'script'>\n```\n\n\n<table><thead><tr><th>\n\nParameter\n\n\n</th><th>\n\nType\n\n\n</th><th>\n\nDescription\n\n\n</th></tr></thead>\n<tbody><tr><td>\n\nopts\n\n\n</td><td>\n\n{ base?: string; scope?: string; path?: string; verbose?: boolean; fetchBundleGraph?: boolean; nonce?: string; }\n\n\n</td><td>\n\nOptions for the prefetch service worker.\n\n- `base` - Base URL for the service worker `import.meta.env.BASE_URL` or `/`<!-- -->. Default is `import.meta.env.BASE_URL` - `scope` - Base URL for when the service-worker will activate. Default is `/` - `path` - Path to the service worker. Default is `qwik-prefetch-service-worker.js` unless you pass a path that starts with a `/` then the base is ignored. Default is `qwik-prefetch-service-worker.js` - `verbose` - Verbose logging for the service worker installation. Default is `false` - `nonce` - Optional nonce value for security purposes, defaults to `undefined`<!-- -->.\n\n\n</td></tr>\n</tbody></table>\n**Returns:**\n\nJSXNode&lt;'script'&gt;",
```
wmertens, line 1777: please run `pnpm api.update`, the build unfortunately changes this file

packages/insights/src/db/sql-user.ts
```diff
@@ -110,7 +110,7 @@ export const dbAddUserToApplication = async (email: string, publicApiKey: string
       .insert(userApplicationMap)
       .values({ userId: user.userId, applicationId: app?.applicationId })
       .execute();
-  } catch (e) {
+  } catch (_e) {
```
wmertens, line 113: better to just remove it, no?

wmertens (changes requested):
this is great! but some small nitpicks

packages/insights/src/db/sql-user.ts
better-salmon: I've noticed that the `no-unused-vars` rule is disabled in several places, so I've disabled it here as well for consistency. In my opinion, it's better to maintain uniform rules across all packages in the monorepo. However, enabling this rule everywhere right now would introduce many errors, so perhaps it's best to address that separately in another PR.

packages/docs/src/routes/api/qwik/api.json
better-salmon: Done

package.json
better-salmon: Fixed

better-salmon:
@wmertens, I also noticed that your monorepo setup isn't optimally configured. I could set up [turborepo](https://turbo.build/repo/docs) and shared ESLint and TypeScript configurations. However, since this would be a sizable PR, I wanted to confirm with you before starting any work 🙂

maiieul:
@better-salmon a decision was made in the past to not use Nx to avoid unnecessary overhead. Maybe it would work well, but maybe it could cause bugs that are hard to pinpoint. Same with Turborepo I believe. Also such tooling might turn off potential contributions from devs that don't know the tool in question.

TLDR: I think that if you rebase we can merge your PR :)

packages/insights/src/db/sql-user.ts
wmertens: I meant `} catch {` without the `(e)`

wmertens (approved):
LGTM

Issue #7086, opened 2024-11-22
[🐞] onResolved cannot return a Promise


maiieul:
### Which component is affected?

Qwik Runtime

### Describe the bug

Trying to await a function inside `onResolved` in a <Resource> I get

> Type '(appointments: MyCustomType) => Promise<JSXOutput>' is not assignable to type '(value: MyCustomType) => JSXOutput'.
  Type 'Promise<JSXOutput>' is not assignable to type 'JSXOutput'.ts(2322)
core.d.ts(2280, 5): The expected type comes from property 'onResolved' which is declared here on type 'IntrinsicAttributes & ResourceProps<MyCustomType>'

But even though typescript complains, the code works on my end. So I guess it's just an easy fix on the ts side.

### Reproduction

https://stackblitz.com/edit/github-b7kqvv?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

Open stackblitz repro, maybe edit the index.tsx file (add a space anywhere to get the ts highlighter).

### System Info

```shell
🙅‍♂️
```

Varixo:
This is probably fixed in v2 :D

maiieul:
Although the type fix is probably an easy fix in v1 too no?

wmertens:
Ok I see - but why use an async function here? The render functions are like inline components and those aren't supposed to be async, because it breaks reactivity tracking.

So not sure we should support this @maiieul @Varixo

maiieul:
Hmmm good point. I intuitively went for async inside onResolved to track some analytics data in our codebase. It does work for this use-case but if it breaks reactivity in certain cases then we need to document not to do that and instead create a component$ that can be passed to the `onResolved`.

maiieul:
@wmertens are you sure that the render functions behave like inline components and that it breask reactivity tracking? Shouldn't we have a test breaking in that case?

It seems to me that doing
```tsx
        onResolved={async (data) => {
          return <p>Data: {data}</p>;
        }}
```

is different from 

```tsx
        onResolved={async (data) => {
          return <InlineComponent data={data} />;
        }}

...

const InlineComponent = async ({data}: {data:string}) => {
  return <p>Data: {data}</p>
}
```

PR #7426, opened 2025-03-15
fix: fix a resource a type error when using async _resolved function

JerryWu1234:
# What is it?

fixes: #7086 

follow version 2 to fix this ts error 

# Description

# Checklist


JerryWu1234:
@maiieul please help me to review

packages/qwik/src/core/api.md
```diff
@@ -886,7 +886,7 @@ export interface ResourceProps<T> {
     // (undocumented)
     onRejected?: (reason: Error) => JSXOutput;
     // (undocumented)
-    onResolved: (value: T) => JSXOutput;
+    onResolved: (value: T) => JSXOutput | Promise<JSXOutput>;
```
maiieul, line 889: I believe we should also change the type for `onPending` and `onRejected`

packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
```diff
@@ -672,6 +672,29 @@ test('DelayResource', async () => {
   );
 });
 
+test('AsyncResource', async () => {
+  await testSSR(
+    <body>
+      <ul>
+        <AsyncResource text="thing" />
+        <AsyncResource text="thing" />
```
maiieul, line 680: Why do you have `<AsyncResource text="thing" />`twice here?

packages/qwik/src/core/use/use-resource.ts
```diff
@@ -252,10 +252,15 @@ export interface ResourceProps<T> {
...
+  // Resource path
+  return jsx(Fragment, {
+    children: getResourceValueAsPromise(props),
+  });
+};
+function getResourceValueAsPromise<T>(props: ResourceProps<T>): Promise<JSXOutput> | JSXOutput {
   const resource = props.value as ResourceReturnInternal<T> | Promise<T> | Signal<T>;
-  let promise: Promise<T> | undefined;
-  if (isResourceReturn(resource)) {
+  if (isResourceReturn(resource) && resource.value) {
```
maiieul, line 262: Do we need to add the `&& resource.value`?

JerryWu1234: > Do we need to add the `&& resource.value`?

Maybe resource is signal @maiieul

packages/qwik/src/core/use/use-resource.ts
```diff
@@ -191,8 +191,8 @@ export const useResource$ = <T>(
 /** @public */
 export interface ResourceProps<T> {
   readonly value: ResourceReturn<T> | Signal<Promise<T> | T> | Promise<T>;
-  onResolved: (value: T) => JSXOutput;
-  onPending?: () => JSXOutput;
+  onResolved: (value: T) => JSXOutput | Promise<JSXOutput>;
+  onPending?: () => JSXOutput | Promise<JSXOutput>;
```
wmertens, line 195: I don't understand why this is needed. The resource holds a Promise, but the Resource component should just render without a Promise

packages/qwik/src/core/use/use-resource.ts
wmertens: the resource is always a promise, no?

wmertens (commented):
Can you explain more why this is needed? A Resource is supposed to be a placeholder for a Promise, so I don't understand why it should also return a Promise

packages/qwik/src/core/use/use-resource.ts
JerryWu1234: From what I know, it is correct for JavaScript, but it is not correct for TypeScript.
TypeScript will think that the function returns a Promise when async is added before the function.

JerryWu1234: @wmertens

wmertens: you shouldn't add async before the function then

wmertens: aaah ok now I get it, it's when the dev puts an async in the resource functions, ok. Hmm not sure if that's a good idea though - the render functions are just like inline components and we don't really support async inline components either. It works though :)

@varixo thoughts?

JerryWu1234:
@wmertens @Varixo DO we still need this PR, I'm going to close it if it doesn't need

wmertens:
@JerryWu1234 can you merge main and run `pnpm api.update`? We can merge this then.

gioboa:
> @JerryWu1234 can you merge main and run `pnpm api.update`? We can merge this then.

Done ✅

.changeset/polite-singers-sing.md
```diff
@@ -0,0 +1,6 @@
+---
+'@builder.io/qwik-city': patch
+'@builder.io/qwik': patch
+---
+
+follow version 2 to fix this ts error
```
gioboa:
```suggestion
FIX: solve type error when using async _resolved function
```

PR #7427, opened 2025-03-15
chore: toMatchDOM fix and refactor

Varixo:
`toMatchDOM` matcher did not work correctly for comparing attributes. Refactored all matcher code and replaced it with string comparison done by vitest. Now the messages are nicer.


wmertens (approved):
Yes much nicer!

PR #7429, opened 2025-03-15
fix(starters): styled vanilla extract update devDependncies

Shane-Donlon:
# What is it?

- Feature / enhancement
- Bug

# Description
@wmertens 

Hey Wout 
Sorry for the tag, but I think this is your integration, so would appreciate your input here.

Small update:

seems `@vanilla-extract/vite-plugin` was missing from the devDependencies  which would cause `error when starting dev server:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@vanilla-extract/vite-plugin' ` 

I did link the CLI globally and just updating the devDependencies seems to have fixed the issue with the `/styled-flower/` route without any adjustments.

I'll leave it as a draft for now, let me know what you think.

### DIFF
--- a/starters/features/styled-vanilla-extract/package.json
+++ b/starters/features/styled-vanilla-extract/package.json
@@ -24,6 +24,7 @@
   },
   "devDependencies": {
     "@vanilla-extract/css": "^1.12.0",
-    "styled-vanilla-extract": "^0.5.4"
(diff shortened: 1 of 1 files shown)


wmertens (approved):
yes seems ok

LGTM

PR #7430, opened 2025-03-16
fix: removing text node from shared text node

Varixo:
After ssr it, we get ie. "0123" as one text node, then we materialize 4 fragements with one letter in each.
We should get:
```tsx
<>0</><>1</><>2</><>3</>
```
but this also means that every vnode right after materializing will point to the same shared textnode "0123". Of course we need to split it to separate text nodes, and this should be done at time of removing. In this case we did it to late, when virtual parent didn't have connections to siblings.

### DIFF
--- a/.changeset/wicked-pets-chew.md
+++ b/.changeset/wicked-pets-chew.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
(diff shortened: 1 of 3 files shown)


wmertens (approved):
LGTM!

PR #7432, opened 2025-03-16
fix: change client side generated ID to start with build base

Varixo:
Use build base for client side generated ID and convert first char to letter if it is statring from number, because CSS selectors can't start with number


wmertens (approved):
LGTM 🌞

PR #7441, opened 2025-03-20
Feat: filter data that the suffix is serverQrl_rpc 

### LINKED ISSUE #7313: [🐞] qwik-city doesn't check if server$ calls are server$

Qwik City (routing)

### Describe the bug

qwik-city should only allow calling server$ and other qrls if they are generated as server$.

Right now any valid qrl can be called. This is somewhat academic but still a concern.

### Reproduction

n/a

### Steps to reproduce

_No response_

### System Info

```shell
any
```

### Additional Information

_No response_

JerryWu1234:
# What is it?

fixes #7313

By looking at the symbol name you can see if it's a server qrl,

# Description

# Checklist

### DIFF
--- a/packages/qwik-city/src/buildtime/vite/plugin.ts
+++ b/packages/qwik-city/src/buildtime/vite/plugin.ts
@@ -259,6 +259,25 @@ function qwikCityPlugin(userOpts?: QwikCityVitePluginOptions): any {
           }
         }
       }
+
+      // Generate public API file
+      if (qwikPlugin) {
+        const manifest = qwikPlugin.api.getManifest();
+        if (manifest) {
+          const publicApi = Object.fromEntries(
+            Object.entries(manifest.symbols).filter(
+              ([_, symbol]) =>
+                symbol.displayName?.endsWith('serverQrl_rpc') && symbol.ctxKind === 'function'
+            )
+          );
+
+          this.emitFile({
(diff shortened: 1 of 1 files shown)


JerryWu1234:
@wmertens  it's my first time trying to implement this
Maybe it won't fully meet your goal. let me know if there are anything else I should improve

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -998,6 +1006,7 @@ export const SSR_OUT_DIR = 'server';
 const LIB_OUT_DIR = 'lib';
 
 export const Q_MANIFEST_FILENAME = 'q-manifest.json';
+export const Q_SERVER_RPC_CTX_KIND_SYMBOLS_FILENAME = 'q-server-rpc-ctx-kind-symbols.json';
```
wmertens, line 1009: let's call this `q-public-api.json`

wmertens:
Good start, now we need to read this file in qwik-city and only allow known symbols.

Then we should also create a `public-api.md` in the source root that looks like

```md
# Public endpoints

## `server$`

- `/xyz123`:
  - parameters: foo, bar
  - scope: myContext

## `formAction$`

...
```

the dev can then check this in and will know when the API changes.

PR #7444, opened 2025-03-20
fix(optimizer): allow chunking entry points

wmertens:
This hopefully improves insights


packages/qwik/src/optimizer/core/src/test.rs
```diff
@@ -3289,7 +3289,7 @@ export const Local = component$(() => {
 		minify: MinifyMode::Simplify,
 		explicit_extensions: true,
 		mode: EmitMode::Test,
-		manual_chunks: None,
+		// filler to maintain line offsets
```
Varixo, line 3292: why we need to maintain line offsets?

Varixo (approved):
LGTM

packages/qwik/src/optimizer/core/src/test.rs
wmertens: The snapshots contain their line numbers

PR #7446, opened 2025-03-21
fix: inserting new node edge case and creating error overlay

Varixo:
- create error overlay correctly using vnode transformations
- fix inserting node edge case and prevent referencing vnode to self
- add a lot of `insertBefore` tests


wmertens (approved):
LGTM

PR #7448, opened 2025-03-22
fix: don't execute QRLs for elements marked as deleted

Varixo:


Varixo:
The `try - catch` approach won't work here. The problem is we console.error an error anyway, because in vnode_locate we use throwErrorAndStop which is using createAndLogError which is always console logging the error

PR #7452, opened 2025-03-24
fix: custom event names and DOMContentLoaded handling

Varixo:
DOMContentLoaded is a special case for HTML events, because it is case sensitive


gioboa (approved):
:clap: @Varixo

PR #7453, opened 2025-03-24
feat: dynamic modulepreload

wmertens:
This moves the service worker logic into the qrl handling, adding modulepreload link tags to request preloading chunks that will probably be needed soon.

This allows the browser to do its thing way better, and reduces complexity. (it would also allow some preloading in dev, but only for the SSR result, not implemented)

You can easily test the impact by going to the docs site and clicking the search button.

Features:
- automatic no-config preloading based on probabilities, via qrl class
- some mild preloading in dev
- path-to-bundle mapping included in bundle graph
- move all insights related build code into the insights plugin, including amending bundle graph
- server now has manifest built-in, not necessary to pass it (but you can still amend it)

TODO:
- [x] test
- [x] replace qprefetch dispatches with some qwik API call
- [x] push qwik-city path preloads into bundlegraph
- [x] deprecate service workers, make them instead uninstall the current SW and use this
- [x] make this the default, we may not need any strategy any more
- [x] change jslink strategy to use this
- [x] fallback to preload when modulepreload not supported (iPhone 2017, iPad 2017, iPod Touch. All other browser support it for 1.5 years now. https://iosref.com/ios, https://caniuse.com/link-rel-modulepreload 93%)
- [x] implement probability-based loading via bundlegraph
- [x] use probabilities for ssr link provisioning
- [x] debug flag in the prefetch config
- [x] `qwik add service-worker` now that the starter no longer includes it


Varixo:
Is it a good idea to move fetching bundle graph to the head element?
Currently everything is added at the bottom of page, but if we have some promise to await then interactivity will be broken? For qwikloader we have an option to append it at the top, maybe here also should be such as option?

packages/qwik/src/optimizer/src/plugins/bundle-graph.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (!graph[depName]) {
+        // external dependency
+        continue;
+      }
+      clearTransitiveDeps(deps, new Set(), depName);
+    }
+    const internalDynamicImports = bundle.dynamicImports?.filter((d) => graph[d]) || [];
+    // If we have a lot of dynamic imports, we don't know which ones are needed, so we don't add any
+    // This can happen with registry bundles like for routing
+    if (internalDynamicImports.length > 0 && internalDynamicImports.length < 10) {
```
maiieul, line 70: We need to keep the `deps.hasSymbols` logic as explained in https://github.com/QwikDev/qwik/pull/7417#issuecomment-2750991973

> > Ok, in https://github.com/QwikDev/qwik/pull/7453 I just removed all of it and only disallow more than 10 dynamic deps because we're probably not going to guess right. That works fine too it seems.

> We need the hasSymbols guard, otherwise we will prefetch external dynamic deps again (e.g. shiki). 10 sounds like a reasonable safeguard, but I wonder if it's needed at all. The dynamic deps graph seems to stop on its own after a certain number of nodes. There is still a bit of over-prefetching on qwik-ui but I'm quite sure it's because some static deps that have dynamic deps aren't being tree-shaken somehow.

wmertens: but then we won't preload dynamic imports for library components :-/

packages/qwik-city/src/runtime/src/api.md
```diff
@@ -10,8 +10,8 @@ import { CookieOptions } from '@builder.io/qwik-city/middleware/request-handler'
 import { CookieValue } from '@builder.io/qwik-city/middleware/request-handler';
 import { DeferReturn } from '@builder.io/qwik-city/middleware/request-handler';
 import type { EnvGetter } from '@builder.io/qwik-city/middleware/request-handler';
-import { JSXNode } from '@builder.io/qwik';
-import { JSXOutput } from '@builder.io/qwik';
+import { JSXOutput } from 'packages/qwik/dist';
+import { JSXOutput as JSXOutput_2 } from '@builder.io/qwik';
```
Varixo, line 14: is it intended?

packages/qwik/src/optimizer/src/plugins/bundle-graph.ts
maiieul: Hmmm are you sure? I will need to test this. Shouldn't be an issue if libraries use `preserveModules`, but better to support both ofc.

wmertens: I have a better solution now! You can adjust the graph in your vite config now, and I already use it to clear the dynamic imports from qwik-city.
I think preloading shiki or threejs or whatever is good if it is actually needed, so in case it's problematic the dev can optimize manually. See the bundle-graph.unit tests.

maiieul: I think preloading shiki or threejs or similarly "big" libs can be tricky. We can provide a way to preload those, but it should be optional imo. Otherwise that's also a breaking change.

maiieul: So I think we need to exclude external deps by default.

wmertens: How about a setting for maximum preload size?

gioboa (dismissed):
👏👏👏@wmertens

packages/qwik/src/optimizer/src/plugins/bundle-graph.ts
maiieul: Might not be needed at all?

wmertens: yes indeed, we should only care about our own qrl imports, the rest are normally static imports, and when dynamic they are on purpose so we shouldn't interfere.

GrandSchtroumpf:
Is it the same as preload strategy `linkInsert: 'html-append'` and `linkRel: 'modulepreload'` ?
Like it appends <link> into <head> with deps found at SSR ?
If so, to reduce the impact on  I would recommend to delay adding them on idle callback. We measured that delaying when the <link> are included in the DOM improve FCP/LCP metrics.

wmertens:
@GrandSchtroumpf similar, but it also fixes waterfalls after resume. It appends a few modulepreloads and then handles the rest from a queue that it reprioritizes constantly.

when you delay the preloads, won't that increase the total load time?

GrandSchtroumpf:
@wmertens it comes at the price of delaying initial layout as explained here :
https://github.com/QwikDev/qwik-evolution/issues/193#issuecomment-2769876402

So it's a trade-off. 
1. If you need a small amount of large files that you are critical at load time (classic CSR for example) it's better to include it in the head because the delay on layout depends on the amount of requests no matter the size.
2. If you need a large amount of small files that can wait (prefetch everything for offline mode for example) it's better to move it in idle callback.

In the case you describe it's a small amount of small files that are not critical for load time but will be used soon. So not easy to give a definitive answer ah ah.
I would say, since priority of qwik is load time I would recommend to wait idle callback. 

The best, of course, would be to have some kind of threshold of files over which you wait until idle callback. But it might be a lot of work with unknown results.

thejackshelton:
> @wmertens it might. But using link in initial html comes at the price of delaying initial layout as explained here : [QwikDev/qwik-evolution#193 (comment)](https://github.com/QwikDev/qwik-evolution/issues/193#issuecomment-2769876402)
> 
> So it's a trade-off.
> 
> 1. If you need a small amount of large files that you are critical at load time (classic CSR for example) it's better to include it in the head because the delay on layout depends on the amount of requests no matter the size.
> 2. If you need a large amount of small files that can wait (prefetch everything for offline mode for example) it's better to move it in idle callback.
> 
> In the case you describe it's a small amount of small files that are not critical for load time but will be used soon. So not easy to give a definitive answer ah ah. I would say, since priority of qwik is load time I would recommend to wait idle callback.
> 
> The best, of course, would be to have some kind of threshold of files over which you wait until idle callback. But it might be a lot of work with unknown results.

Going off of what @GrandSchtroumpf mentioned here. If the modulepreload links are added during the critical rendering path, it is going to shave off some pagespeed points. 

What we've done to prevent this in our production apps at work is by adding them at the load event during the browser idle periods

https://github.com/kunai-consulting/qwik-design-system/blob/main/apps/docs/src/docs-widgets/module-preload/module-preload.tsx

wmertens:
@thejackshelton looking at performance recordings, the FCP doesn't seem impacted by the modulepreloads even while the preloader is adding linkds, and in any case you can set `maxPreloads` to `2` or even `0` so that the preloader does the loading.

So I think that's ok?

GrandSchtroumpf:
@wmertens how is it going to behaves ?
The dev can specify the amount of links to add at SSR time, the framework add them in order of priority, then it adds remaining links on idle callback ?
For example if there are 10 links to add and dev set `maxPreload` to 3, then 3 links will be in the SSR result and 7 will be appended to the dom on idle callback ?

wmertens:
> @wmertens how is it going to behaves ? The dev can specify the amount of links to add at SSR time, the framework add them in order of priority, then it adds remaining links on idle callback ? For example if there are 10 links to add and dev set `maxPreload` to 3, then 3 links will be in the SSR result and 7 will be appended to the dom on idle callback ?

Yes correct, but it adds them as soon as the preloader loads, and we can make it happen at idle time if you see that that works better. 

Perhaps I should add query parameters to configure the preload for testing. 

In any case I don't think we should wait for the answer to this to merge, we can always refine.

shairez:
I agree, let's merge it now to start the alpha testing. 
@GrandSchtroumpf if you can please run your checks and comparison on the 1.14.0 alpha I'm going to publish in the alpha testers channel, that would be awesome! 

Thank you very much! :pray:

shairez (approved):

PR #7466, opened 2025-03-30
feat: route loaders serialization

Varixo:
Implementation for RFC
https://github.com/QwikDev/qwik-evolution/issues/249


packages/qwik-router/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -222,51 +227,92 @@ export function actionsMiddleware(routeActions: ActionInternal[], routeLoaders:
...
+    loader.__validators,
+    undefined, // data
+    isDev
+  )
+    .then((res) => {
+      if (res.success) {
+        if (isDev) {
+          return measure<Promise<unknown>>(
+            requestEv,
+            loader.__qrl.getSymbol().split('_', 1)[0],
```
wmertens: I don't understand this? If the symbol is `"s_qfpiearnst"`, you'll get `"s"`? Why not use `.getHash()`?

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -418,7 +464,7 @@ export function getPathname(url: URL, trailingSlash: boolean | undefined) {
     }
   }
   // strip internal search params
-  const search = url.search.slice(1).replaceAll(/&?q(action|data|func)=[^&]+/g, '');
+  const search = url.search.slice(1).replaceAll(/&?q(action|data|func|loader)=[^&]+/g, '');
```
wmertens: better to call it `loaders`?

packages/qwik-router/src/runtime/src/use-endpoint.ts
```diff
@@ -9,14 +9,18 @@ export const loadClientData = async (
...
     prefetchSymbols?: boolean;
     isPrefetch?: boolean;
   }
 ) => {
   const pagePathname = url.pathname;
   const pageSearch = url.search;
-  const clientDataPath = getClientDataPath(pagePathname, pageSearch, opts?.action);
+  const clientDataPath = getClientDataPath(pagePathname, pageSearch, {
+    actionId: opts?.action?.id,
+    loaderIds: opts?.loaderIds,
```
wmertens, line 25: For SSG, we must have a fallback when the clientData returns 404, so that we just request the regular data.json and get all the loader data

packages/qwik/src/core/reactive-primitives/impl/async-computed-signal-impl.ts
```diff
@@ -94,6 +98,11 @@ export class AsyncComputedSignalImpl<T>
     return this.$untrackedError$;
   }
 
+  override invalidate() {
```
wmertens, line 101: what is `override`? Can't find it in ecmascript nor typescript docs

packages/qwik/src/core/reactive-primitives/impl/async-computed-signal-impl.ts
```diff
@@ -118,11 +128,12 @@ export class AsyncComputedSignalImpl<T>
           this.untrackedError = null;
         })
         .catch((err) => {
+          this.$promiseValue$ = err;
```
wmertens, line 131: hmm, can't this cause that reading the value gives you the error, without error? Is this because you can't set it to null?

IMHO this class should override `.value` and throw `$untrackedError$` if it exists, and probably leave the `$promiseValue$` as-is?

wmertens: actually, what happens when you read the async computed while it's loading? IMHO it should throw the promise.

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers.ts
Varixo: I don't know it is not my code, its just moved from another place

packages/qwik/src/core/reactive-primitives/impl/async-computed-signal-impl.ts
Varixo: https://www.typescriptlang.org/tsconfig/#noImplicitOverride
here, but don't remember why I used it here, probably to know that it is overwritten from parent

packages/qwik/src/core/reactive-primitives/impl/async-computed-signal-impl.ts
Varixo: the problem is signals should be sync, because it can be used for example in useComputed or other hooks

Varixo: This will be addressed in another PR

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers.ts
Varixo: changed to getHash

PR #7469, opened 2025-03-31
fix: correctly handle initial resource state

Varixo:
Handle initial undefined value for resource. The initial value is undefined because we delegate execution of it to the scheduler


packages/qwik/src/core/use/use-resource.ts
```diff
@@ -204,10 +204,15 @@ function getResourceValueAsPromise<T>(props: ResourceProps<T>): Promise<JSXOutpu
...
-      useBindInvokeContext(props.onRejected)
-    );
+    const value = resource.value;
+    if (value) {
+      return value.then(
+        useBindInvokeContext(props.onResolved),
+        useBindInvokeContext(props.onRejected)
+      );
+    } else {
+      return Promise.resolve(undefined);
```
wmertens, line 214: can you put a comment here explaining what happens next? It seems as if that is the end of the resource loading?

wmertens (approved):
:point_right: LGTM :point_left:

PR #7471, opened 2025-04-01
docs:force remove body overflow-hidden when search closes

Shane-Donlon:
# What is it?

-docs

# Description

Ciao Gio,

`Bug: when you go to qwik.dev and then CLICK search and type "usesignal" and then CLICK on the result, then scrolling doesn't work any more`

This is originating from Qwik UI Modal in Single Page Applications

Package below is adding the overflow hidden - but not removing it, after clicking the link.
`name: body-scroll-lock-upgrade`

The links in the results  are `<Link>` (from Qwik) and not `<a> `

So this will just manually remove the overflow hidden from the body for now.
I'll have a look into fixing it within Qwik UI,  but just adding this pull request so that the search works as expected for now.

# Checklist

- [x ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [x ] I performed a self-review of my own code

### DIFF
--- a/packages/docs/src/components/docsearch/doc-search-modal.tsx
+++ b/packages/docs/src/components/docsearch/doc-search-modal.tsx
@@ -118,6 +118,7 @@ export const DocSearchModal = component$(
 
         return () => {
           document.body.classList.remove('DocSearch--active');
+          document.body.style.overflow = '';
         };
       }
     });


Shane-Donlon:
Just waiting to do a self-review of the preview link then will move this out of draft status

Shane-Donlon:
self-tested works as intended in preview link moving for review

wmertens (approved):
LGTM

PR #7482, opened 2025-04-03
chore(ci): deploy docs whenever docs build worked

wmertens:
hopefully this works - the only way to test is to merge in main


shairez (approved):
LGTM

PR #7495, opened 2025-04-07
chore(README): remove unused badges

wmertens:
these run in the qwikcity-e2e app and vercel is failing, probably due to permissions, and the rest is disabled


gioboa (approved):
That's fair 
Thanks @wmertens

PR #7499, opened 2025-04-07
chore: fix changesets and package.json keywords

Varixo:


gioboa (approved):
Nice catch 👏 
@Varixo

PR #7505, opened 2025-04-09
chore: don't serialize wrapped signal value

Varixo:
Wrapped signals are always based on other values
- if they dont change we dont need value
- if they change we recompute anyway
- if we call `.value` on it - just compute value


wmertens (approved):
Awesome 🪶

PR #7516, opened 2025-04-14
feat(preloader): more sane configuration

wmertens:
5 SSR preloads, 25 parallel preloads and only start after load+idle. That seems to work nicely with the qwik docs


packages/docs/src/entry.ssr.tsx
```diff
@@ -1,10 +1,11 @@
...
+import type { PreloaderOptions, RenderToStreamOptions } from '@builder.io/qwik/server';
+import { renderToStream } from '@builder.io/qwik/server';
 import Root from './root';
 
 // You can pass these as query parameters, as well as `preloadDebug`
 const preloaderSettings = [
-  'maxPreloads',
-  'minProbability',
+  'ssrPreloads',
+  'ssrPreloadProbability',
```
maiieul, line 8: for consistency I would name it `ssrMinPreloadProbability`, or remove the "min" in the the other.

packages/qwik/src/server/prefetch-implementation.ts
```diff
@@ -35,35 +33,30 @@ export function includePreloader(
     }
   }
 
-  const makeLink = (base: string, href: string) => {
-    const linkProps: PropsOf<'link'> = {
-      rel: linkRel!,
+  /** We only use modulepreload links because they behave best. Older browsers have the preloader. */
```
maiieul, line 36: what does this mean? The preloader will be enough for older browsers? how?

gioboa (approved):
👍 LGTM

packages/qwik/src/server/prefetch-implementation.ts
wmertens: No, the small % of older browsers will ignore SSR preloads, and for them the preloader does `preload` instead of `modulepreload` once it runs.

Since the default is now 0 ssr preloads, there is no effect.

gioboa (dismissed):

packages/qwik/src/server/api.md
```diff
@@ -74,6 +71,15 @@ export interface PrefetchStrategy {
     symbolsToPrefetch?: SymbolsToPrefetch;
 }
 
+// @public (undocumented)
+export interface PreloaderOptions {
+    activePreloads?: number;
```
maiieul, line 76: Hmmm I actually prefer `maxSimultaneousPreloads` although it's a bit verbose. "Simultaneous" is so hard to type. Maybe `maxActivePreloads`, or `maxConcurrentPreloads` would be a little easier?

.changeset/fresh-readers-switch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': minor
+---
+
+FEAT: the Qwik preloader can be configured via the `preloader` key on the render options.
```
maiieul, line 5: You should also edit the previous changeset no? Is this one needed?

packages/qwik/src/server/api.md
wmertens: I could accept `maxActivePreloads`, although as the comment says, 100% probable ones have unlimited preloads, so that's a lie, and furthermore the idea is that people hover and read the JSDoc anyway :)

maiieul: Hmmm but this key is for the non-100% probable ones. So maybe `maxBufferedPreloads`?

shairez (approved):
Loved the new API naming, thanks!

PR #7519, opened 2025-04-15
perf: better minify qwikloader

wmertens:
With these changes, the gzip compressed qwikloader is 2.44kb, and brotli is 1.4kb, several hundred bytes profit.

Note that the uncompressed version is actually slightly longer, but it compresses better. The most common case is that the HTML will be served with gzip compression.


gioboa (approved):
That's a good improvement 
👏@wmertens

PR #7523, opened 2025-04-17
fix(docs): render API tables correctly

### LINKED ISSUE #7521: [📖] API docs are somewhat broken
### Suggestion

for example, https://qwik.dev/api/qwik-optimizer/#optimizer

- the tables with call info are not tables but paragraphs
- when you click a definition it jumps there, but using back/forward does not work, it only changes the hash in the URL

wmertens:
- convert into .mdx files so that html is interpreted correctly
- escape non-code blocks so the syntax is correct

fixes #7521


gioboa (approved):
👏@wmertens

gioboa (approved):

PR #7533, opened 2025-04-23
chore: explain cache headers + message on build

wmertens:


tzdesign (approved):
Nice explanation.

I never thought about consistent chunk names, even when translations change.

🚀

gioboa (approved):
👏nice explanation @wmertens 💪

PR #7537, opened 2025-04-24
fix(qwik-city): stop q-data.json cache busting

### LINKED ISSUE #7536: [🐞] qwik-city should not bust cache on q-data.json

Qwik City (routing)

### Describe the bug

I think we should revert #6808.

Qwik-city has a client-side `q-data.json` cache, and it is populated when you hover over a Link that has data prefetch enabled (the default).

When you click the Link, you do an SPA navigation, and it will fetch q-data.json again, as well as clear the client side cache.

Before #6808, that meant that the Cache-Control headers on the `q-data.json` were used, which default to one hour before stale and serving the old data while revalidating.

After #6808, q-data.json will be fetched every time, and it actually blocks the navigation, making SPA annoyingly slow.

I believe the revalidate default is also wrong, because we have no way of knowing when data gets revalidated, the browser doesn't tell us. At best we should fetch q-data normally and then after a few seconds fetch again "only if cached" and if the result differs rerender with the new data.

To achieve fresh q-data, there should be explicit headers turning off the caching for that path's q-data.

### Reproduction

https://qwik.dev/ecosystem

### Steps to reproduce

just go back and forth between the ecosystem items, you'll see the requests always happen.

### System Info

```shell
web
```

### Additional Information

_No response_

wmertens:
Fixes #7536


gioboa (approved):
👏It looks great to me

maiieul (approved):

thejackshelton-kunaico (approved):
LGTM 🚀

starters/apps/playground/src/routes/layout.tsx
```diff
@@ -7,17 +7,6 @@ import Footer from "../components/starter/footer/footer";
 
 import styles from "./styles.css?inline";
 
-export const onGet: RequestHandler = async ({ cacheControl }) => {
-  // Control caching for this request for best performance and to reduce hosting costs:
-  // https://qwik.dev/docs/caching/
-  cacheControl({
```
shairez, line 13: maybe we can leave a comment with a link to the blog post or docs on the matter, something like: "It's a good idea to set your cache for X, here's how to do it..."

So it will be part of the starters

shairez (approved):
Good job @wmertens 

as discussed, leaving my comments as references for the upcoming fixes PR

Thanks!

Varixo (approved):

PR #7538, opened 2025-04-25
fix: inflating text nodes from single shared text node

Varixo:
Fixing insertBefore function for inserting text nodes from shared text node


wmertens (approved):
LGTM

PR #7540, opened 2025-04-25
New Websites for Showcase

denisyilmaz:
Some of our websites we created in recent time with Qwik.


wmertens:
could you also run `pnpm -C packages/docs run build.showcase`?

PR #7551, opened 2025-04-28
chore: better materialization error and dev location fix

Varixo:
Throw an error with better description what is failing
example:
```
Error: Code(Q44): Materialization error: Expected a text or element after
1 + 2 = 3
at location /some/file/location.tsx:123:1
   metadata: {B`6=5|q:type|S}D{B`7=16|q:type|S}D{B`8=11|q:type|S}2,
   value: 50,
   next id: 52
```

This also fixes dev location for CSR rendering

### DIFF
--- a/packages/qwik/src/core/client/vnode-diff.ts
+++ b/packages/qwik/src/core/client/vnode-diff.ts
@@ -99,10 +99,11 @@ import { escapeHTML } from '../shared/utils/character-escaping';
 import { clearAllEffects } from '../reactive-primitives/cleanup';
 import { serializeAttribute } from '../shared/utils/styles';
 import { QError, qError } from '../shared/error/error';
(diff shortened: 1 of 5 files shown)


Varixo:
> Great, but actually the location is unreadable without training. Since it's dev mode, can't we log the actual element so that when you hover it in the console you can see it highlighted on the page?

The problem is qerrors supports only strings. I am thinking if it should be a QError at all, maybe not

shairez:
Thanks Michal!
What about changing "materialization" to "Resuming" or "Resumability" 

reason: we teach about resumability but nobody knows what materialization means (and it's yet another term). I think this will help keeping it simpler.

Plus adding a link to a doc that explains this error in more depth with possible fixes would be nice.

Varixo:
so what do you think about that:
image

wmertens:
closing for now

PR #7552, opened 2025-04-29
feat: allow async operations in useComputed$ hook

Varixo:
This allows async operations in useComputed$ in a predictable way, so deprecating it in v1 is not necessary


wmertens (approved):
🐸

shairez:
see https://github.com/QwikDev/qwik/issues/7560#issuecomment-2856007896

PR #7562, opened 2025-05-02
Support rewrite request (Similarly to redirect)

omerman:
# What is it?

Feature / enhancement

# Description

This PR adds functionality to support URL rewrites. 
At the moment, a user can perform a redirect, which will re-route the request entirely, modifying the url the user sees.
This feature will introduce a new method, `rewrite`, to the event, alongside the already existing `redirect`.
`rewrite` will make an internal redirect, keeping the url of the request intact.

This change addresses qwik-evolution discussion [237](https://github.com/QwikDev/qwik-evolution/discussions/237) by giving developers explicit control over routing behavior while maintaining SEO best practices.

# Checklist

The feature is done and ready for production.
It should be introduced to v1.14, which might take some time.
In the meantime Im keeping this branch up-to-date with main.

You can use this to have it as part of your project:
[pkg-pr](https://github.com/QwikDev/qwik/pull/7562#issuecomment-2852383361)


packages/qwik-city/src/middleware/request-handler/request-event.ts
```diff
@@ -220,6 +221,19 @@ export function createRequestEvent(
       return new RedirectMessage();
     },
 
+    rewrite: (url: string) => {
```
wmertens, line 224: Incidentally, I'd also like us to support location objects for easy parameter changing.

packages/qwik-city/src/middleware/request-handler/types.ts
```diff
@@ -201,6 +202,12 @@ export interface RequestEventCommon<PLATFORM = QwikCityPlatform>
    */
   readonly redirect: (statusCode: RedirectCode, url: string) => RedirectMessage;
 
+  /**
+   * URL to rewrite to. When called, the response will immediately end with the correct rewrite url.
```
wmertens, line 206: The response should stay active, no?

omerman: Yes, I initially taken the phrasing from redirect - I'm still figuring things out

omerman:
@wmertens At the moment I managed to render a white screen 😅, at least without errors.

```
async function runNext(requestEv: RequestEventInternal, resolve: (value: any) => void) {
...
if (e instanceof RedirectMessage) {
      const stream = requestEv.getWritableStream();
      await stream.close();
    } else if (e instanceof RewriteMessage) {
      const stream = requestEv.getWritableStream();
      await stream.close();
    }
...
```

I'm not yet sure where exactly the rewrite can be handled properly.
Do you have ideas?

wmertens:
@omerman when you get a rewrite request, you need to restart processing, so you should basically call the handler again, skipping any initialization tasks, and making sure it can't do an infinite loop (e.g. only allow like 5 redirects, that should be plenty)

omerman:
@wmertens 
> @omerman when you get a rewrite request, you need to restart processing, so you should basically call the handler again, skipping any initialization tasks, and making sure it can't do an infinite loop (e.g. only allow like 5 redirects, that should be plenty)


So far I hopefully managed to understand some of the flow:

we call runQwikCity with a serverRequest event that in dev-time we generate and in prod we get it from all the possible integrations.
We build all the handlers which are phases the request passes through. we then wait for all them to complete. and stream it in the end.

I was thinking the key part is in the runQwikCity implementation:
```

export function runQwikCity<T>(
  serverRequestEv: ServerRequestEvent<T>,
  loadedRoute: LoadedRoute | null,
  requestHandlers: RequestHandler<any>[],
  trailingSlash = true,
  basePathname = '/',
  qwikSerializer: QwikSerializer
): QwikCityRun<T> {
  let resolve: (value: T) => void;
  const responsePromise = new Promise<T>((r) => (resolve = r));
  const requestEv = createRequestEvent(
    serverRequestEv,
    loadedRoute,
    requestHandlers,
    trailingSlash,
    basePathname,
    qwikSerializer,
    resolve!
  );
  return {
    response: responsePromise,
    requestEv,
    completion: asyncStore
      ? asyncStore.run(requestEv, runNext, requestEv, resolve!)
      : runNext(requestEv, resolve!),
  };
}
```

by re-run the flow, do you mean I should support a way for the responsePromise to wait for another requestEv to be created under the same response?

wmertens:
@omerman so the http request object itself cannot be recreated, and should be retained. 

Ideally the whole RequestEvent object is retained, the less work the better. 

After the rewrite, the route modules have to be regenerated and the request handling has to happen again.

omerman:
> @omerman so the http request object itself cannot be recreated, and should be retained.
> 
> Ideally the whole RequestEvent object is retained, the less work the better.
> 
> After the rewrite, the route modules have to be regenerated and the request handling has to happen again.

Do you mean something like this, e.g passing this function to be used somewhere deeper as part of the process, and if rewrite occours rerun the output built route handlers and return the response? Am I getting close 😅 ?

image

By regenerate the route modules you meant building the routeHandlers again, right?

wmertens:
Yes something like that.

Also, very important to add e2e tests. See /e2e

omerman:
@wmertens I managed to make most of it work, I managed to create a working rewrite.
I run into an issue now, that if I attempt navigating(client side) to the same url, but with search params for example, the search bar immediately changes to the internal url I was trying to mask.

E.g:

1. http://localhost:5173/omer >> rewrites to http://localhost:5173/demo/flower (Yey it works!)
2. User clicks a link that it's href is http://localhost:5173/omer?search=test
3. the page renders with the right search param, but the actual url in the address bar changes to http://localhost:5173/demo/flower/?search=test which wasnt the desired result(pathname should of stayed /omer)

Do you have ideas as to where I should look through to fix this?

omerman:
@wmertens It works well now. I will fix up my code making it more readable and better, and will write e2e tests as you asked.
Thank you for all the help 🙏

packages/qwik-city/src/middleware/request-handler/request-event.ts
```diff
@@ -220,6 +229,21 @@ export function createRequestEvent(
...
 
+    rewrite: (_rewriteUrl: string | URL) => {
+      check();
+
+      const rewriteUrl = typeof _rewriteUrl === 'string' ? _rewriteUrl : _rewriteUrl.toString();
+      const fixedURL = rewriteUrl.replace(/([^:])\/{2,}/g, '$1/');
+      if (rewriteUrl !== fixedURL) {
+        console.warn(`Rewrite URL ${rewriteUrl} is invalid, fixing to ${fixedURL}`);
+      }
+      url.pathname = fixedURL;
```
omerman, line 240: // TODO - handle rewrite to urls with search params
      // This is not accurate, because the url might have search params

wmertens: Hmm. Ok, I can see rewriting to a string be like a shortcut that passes on search params, and then passing a URL object should not do anything

wmertens:
Awesome, a day well spent 😊

packages/qwik-city/src/middleware/request-handler/types.ts
omerman: Fixed

packages/qwik-city/src/middleware/request-handler/request-event.ts
omerman: I made the API as flexible as possible. Let me know if you think it should be different.

packages/qwik-city/src/buildtime/vite/dev-server.ts
```diff
@@ -227,10 +228,27 @@ export function ssrDevMiddleware(ctx: BuildContext, server: ViteDevServer) {
             await server.ssrLoadModule('@qwik-serializer');
           const qwikSerializer = { _deserializeData, _serializeData, _verifySerializable };
 
+          const applyRewrite: ApplyRewriteInternal = async (url: URL) => {
```
omerman, line 231: TODO - not sure this is the best name to this function.
it doesnt actually apply the rewrite itself.

omerman:
@wmertens Hey again, I tried to add e2e tests, it failed (which helped find some issues I had).
The thing is, it works perfectly in dev. but in preview it doesnt, and I assume it would not work with any adapter as well.

I will look into it further, but the non-dev run seems quite a bit different from the way it runs with adapters.

Have any insights for me maybe? 🐒

omerman:
@wmertens Ready to go. works both in dev and prod environments.
Added e2e tests.

Let me know if you think something is missing or should be changed.

wmertens:
@omerman awesome work 🚀

I'm guessing that executing the routeloaders got skipped somewhere. I'm not quite sure if we can keep the results of the previous run but I'm inclined to let that happen for efficiency reasons.

omerman:
@wmertens Thank you 🙏  hope I'll crack the remaining issue as well.
I have new info to share on that regard, the issue's root is from a resolveValue i'm using at the head.

image

It would seem that once performing a client navigation, we execute the head code again, which makes perfect sense.
As you mentioned, somehow the routeLoaders didnt run, because the headers were sent (stream got closed)

I did manage to make it work by doing this:
image

I will try to get more clues as to why in this specific case it goes down like this.
Let me know if it rings any bells 🤷

omerman:
@wmertens I've fixed all the issues. it works perfectly now. the flow is fully operational 🎉

packages/qwik-city/src/middleware/request-handler/user-response.ts
```diff
@@ -50,62 +52,85 @@ export function runQwikCity<T>(
...
-          requestEv.send(status, await qwikSerializer._serializeData(e.data, true));
-        } else {
-          const html = getErrorHtml(e.status, e.data);
-          requestEv.html(status, html);
+async function runNext(
+  requestEv: RequestEventInternal,
+  applyRewrite: ApplyRewriteInternal,
+  resolve: (value: any) => void
+) {
+  async function _runNext(
```
omerman, line 70: @wmertens If you have any suggestions as to how to make this function clearer, I'm all ears. It became more complex after introduced with the rewrite handling.

packages/qwik-city/src/middleware/request-handler/types.ts
```diff
@@ -201,6 +202,18 @@ export interface RequestEventCommon<PLATFORM = QwikCityPlatform>
...
+   * URL to rewrite to. When called, the flow will reset to display the given url route.
+   *
+   * URL will remain unchanged in the browser history.
+   *
+   * Note that if the url is a string without a protocol, the rewritten url will inherit the search
+   * params from the canonical url.
+   *
+   * @param url - The url to rewrite to.
+   */
+  readonly rewrite: (url: string | URL) => RewriteMessage;
```
wmertens, line 215: I think this should also accept an object containing partially `pathname`, `search` and/or `params`. That way devs can extend or replace the `params` that are in the request event.

packages/qwik-city/src/middleware/request-handler/types.ts
```diff
@@ -287,6 +300,19 @@ export interface RequestEventBase<PLATFORM = QwikCityPlatform> {
...
+   *
+   * This property was introduced to support the rewrite feature.
+   *
+   * If rewrite is called, the canonicalUrl will remain as the original url, and the url property
+   * will be changed to the rewritten url.
+   *
+   * If rewrite is never called as part of the request, the url property and the canonicalUrl are
+   * considered equal.
+   */
+  readonly canonicalUrl: URL;
```
wmertens, line 314: I think this should be called `originalUrl`. It is not necessarily canonical.

starters/e2e/qwikcity/page.spec.ts
```diff
@@ -108,6 +108,45 @@ function tests() {
...
+        h1: "Product: tshirt",
+        activeHeaderLink: "Products",
+      });
+
+      /***********  Products: shirt (rewrite to /products/tshirt)  ***********/
+      await linkNavigate(
+        ctx,
+        '[data-test-link="products-shirt-rewrite-plain_string_no_trailing_slash"]',
+      );
+      await assertPage(ctx, {
```
wmertens, line 143: also add a test for clearing and adding search parameters

packages/docs/src/routes/docs/(qwikcity)/guides/rewrites/index.mdx
```diff
@@ -0,0 +1,48 @@
...
+```
+
+The `rewrite()` function, which was destructured in the RequestHandler function arguments, takes a pathname string or a URL object.
+
+```tsx
+throw rewrite(`/articles/777/`);
+throw rewrite(new URL(`/articles/777/`, url));
+```
+
+Note: If you provide a relative path, Qwik City will rewrite the URL and apply search params from the original URL.
```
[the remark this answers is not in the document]
omerman: After putting more thought to it, I think it will be too confusing to even support any changes to searchParams.
I think the rewrite should reflect real-life use cases and only accept a pathname.

the appropriate path params of the actual route are derived from the given pathname, so they will change alongside 
image

packages/qwik-city/src/middleware/request-handler/types.ts
omerman: As stated above, the params are derived from the rewrite given pathname, it feels too confusing to have it explicitly stated, rewriting to `articles/7777` while passing params: `{articleId: 1234}` should not be supported to my opinion.

and again, it really feels to confusing to have:

`<url>?a=1&b=2` >>> rewrite to `<different-url>?c=3&d=4`


having in the client-side:
`useLocation() // search params are: {a: 1, b: 2}`

and in the server-side:
`request.query // search params are {c: 3, d: 4}`

it is similar to whats happening with path name, but when using Fullstack frameworks, we always have ways to consume the page's data, with something similar to `routeLoader`. but the search params are always consumed directly from the address bar, so having two sets of it makes me feel uneasy.

starters/e2e/qwikcity/page.spec.ts
omerman: I added a test with search params, making sure they are properly present after a rewrite

omerman:
@wmertens I made all the changes, If after reading my comments you still think rewrite searchParams handling still needs to be dealt with, let me know and I'll make the necessary changes.

itay-explorium (approved):

packages/qwik-city/src/middleware/request-handler/request-event.ts
```diff
@@ -220,6 +244,26 @@ export function createRequestEvent(
...
+        throw new Error('Rewrite does not support absolute urls');
+      }
+
+      const rewriteUrl = new URL(url);
+      rewriteUrl.pathname = pathname;
+
+      // Fix consecutive slashes - e.g //path//to//page -> /path/to/page
+      const fixedPathname = rewriteUrl.pathname.replace(/(^|[^:])\/{2,}/g, '$1/');
+      if (rewriteUrl.pathname !== fixedPathname) {
+        console.warn(`Rewrite URL ${rewriteUrl.pathname} is invalid, fixing to ${fixedPathname}`);
```
wmertens, line 259: there's no point to this warning, you always need to do the .replace anyway so even if you fix it in the code the runtime stays the same. You can replace during the assignment.

Also, the replace can just be `pathname.replace(/\/+/g, '/')`, no? The pathname shouldn't include the protocol?

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -355,23 +360,24 @@ async function pureServerFunction(ev: RequestEvent) {
...
+  const { pathname, search } = originalUrl;
   const isQData = sharedMap.has(IsQData);
   if (!isQData && pathname !== basePathname && !pathname.endsWith('.html')) {
     // only check for slash redirect on pages
     if (trailingSlash) {
       // must have a trailing slash
       if (!pathname.endsWith('/')) {
         // add slash to existing pathname
-        throw ev.redirect(HttpStatus.MovedPermanently, pathname + '/' + url.search);
+        throw ev.redirect(HttpStatus.MovedPermanently, pathname + '/' + search);
```
wmertens, line 372: actually, shouldn't the search come from the current URL? It might have changed, no?

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -528,6 +536,39 @@ export async function handleRedirect(requestEv: RequestEvent) {
...
+
+  const status = requestEv.status();
+  const location = requestEv.headers.get('Rewrite-Location');
+  const isRewrite = status === 200 && location;
+
+  if (isRewrite) {
+    const adaptedLocation = makeQDataPath(location);
+
+    if (adaptedLocation) {
+      requestEv.headers.set('Rewrite-Location', adaptedLocation);
```
wmertens, line 564: Not sure if this is necessary?

packages/qwik-city/src/middleware/request-handler/user-response.ts
```diff
@@ -50,62 +52,94 @@ export function runQwikCity<T>(
...
     }
+
+    return undefined;
+  }
+
+  try {
+    let runResult = await _runNext();
+    if (didRewrite) {
+      rewriteAttempt += 1;
+      const url = new URL(requestEv.headers.get('Rewrite-Location')!);
```
wmertens, line 131: I'm confused, why use the headers as a side-channel? Why not store it on the error object?

wmertens (changes requested):
Hmm, I don't like using the headers as scratch space, and I don't think we should expose rewrites in the response headers.

Also, I can imagine devs wanting to change/add/clear the params, is that not so?

omerman:
> Hmm, I don't like using the headers as scratch space, and I don't think we should expose rewrites in the response headers.
> 
> Also, I can imagine devs wanting to change/add/clear the params, is that not so?

Regarding the headers - I took this pattern from the Redirect flow. But I agree, I'll fix it. TBH it bothered me as well.


Regarding the params: As I mentioned before, I dont think it's relevant, when rewriting to a route, if the route has path params, for example article/[articleId], if I re-write to `article/778` the path params will be derived from it.

in the request event, after we rebuild the route's info
```
    get params() {
      return loadedRoute?.[1] ?? {}; /// We will get {articleId: 778}
    }
```

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
omerman: I took it from the redirect flow as well, and it is necessary actually, but I may be able to do it differently.
It signals the request flow that it should fetch the q-data.json and not the page itself.

I'll see if I can remove the header and make it happen in an alternative way.

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
omerman: Yes, good catch, will fix it

omerman: It would actually be equal(by design), as I said before, maybe you missed it the thread has become very long.
The search param is loosely obtained, unlike path params. For DX reasons I think the search params should always stay as the search params from the URL. as opposed to the path params which are always obtained in the same way from a framework specific api. There is no use case for rewriting search params(that I know of), and it feels much more confusing to support such a thing.

omerman:
@wmertens I fixed and improved the code per your notes, thank you 🙏 
Just so we dont miss each other's comment on this long thread here are the highlights:


https://github.com/QwikDev/qwik/pull/7562#discussion_r2075525648

https://github.com/QwikDev/qwik/pull/7562#issuecomment-2854638499

Aside both comments I did as instructed :)

packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
wmertens: ok fair.

packages/qwik-city/src/middleware/request-handler/user-response.ts
```diff
@@ -50,62 +52,88 @@ export function runQwikCity<T>(
...
+    try {
+      // Run all middlewares
+      await requestEv.next();
+    } catch (e) {
+      if (e instanceof RedirectMessage) {
+        const stream = requestEv.getWritableStream();
+        await stream.close();
+      } else if (e instanceof RewriteMessage) {
+        if (rewriteAttempt > 5) {
+          throw new Error(`Rewrite failed - Max rewrite attempts reached: ${rewriteAttempt - 1}`);
```
wmertens, line 82: Just put it at 50. 5 might happen after years of technical debt but 50 is just an infinite loop. Also, use save precious bytes, just say `Infinite rewrite loop`.

wmertens (changes requested):
Just a tiny nitpick. I think we'll release 1.14 this week and then we can merge this.

packages/qwik-city/src/middleware/request-handler/user-response.ts
omerman: @wmertens Done :)

wmertens:
Can you add this as an experimental feature by adding it to the Experimental flags? Just check the flag and throw if used without it enabled, there's a few examples in the code.

Also, if you would like to have this as a merge commit please squash your commits to be self-contained units of change, otherwise I can merge as a squash commit.

wmertens (approved):
LGTM

PR #7564, opened 2025-05-04
chore: merge main into v2

wmertens:


wmertens:
The CI isn't running any more for some reason, but it works locally so I'll just merge and fixup if needed

PR #7571, opened 2025-05-09
fix(core): prevent server-side usage of browser APIs

JerryWu1234:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/.changeset/strong-rules-rescue.md
+++ b/.changeset/strong-rules-rescue.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': major
+---
+
+fix(core): prevent server-side usage of browser APIs
--- a/packages/qwik/src/core/shared/qrl/qrl-class.ts
+++ b/packages/qwik/src/core/shared/qrl/qrl-class.ts
@@ -1,4 +1,4 @@
-import { isDev } from '@qwik.dev/core/build';
+import { isDev, isServer } from '@qwik.dev/core/build';
 import { assertDefined } from '../error/assert';
 import { QError, qError } from '../error/error';
 import { getPlatform, isServerPlatform } from '../platform/platform';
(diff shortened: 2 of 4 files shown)


.changeset/strong-rules-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix(core): prevent server-side usage of browser APIs
```
wmertens, line 5: it's not preventing, it's explaining the error. So more like `FEAT: When an error occurs during SSR due to using the browser APIs, show an explanation.`

packages/qwik/src/core/shared/qrl/qrl-class.ts
```diff
@@ -113,6 +113,19 @@ export const createQRL = <TYPE>(
...
+            if (
+              e instanceof ReferenceError &&
+              ['window', 'document'].some((v) => e.message.includes(v))
+            ) {
+              e.message =
+                'It seems like you forgot to add "if (isBrowser) {...}" here: ' + e.message;
+              throw e;
+            }
+          }
+        }
```
wmertens, line 128: after here you have to still throw. Best is to move line 125 here so all if cases always throw

packages/qwik/src/core/tests/use-task.spec.tsx
```diff
@@ -810,4 +811,21 @@ describe.each([
...
+          document.body.innerHTML = '';
+        }
+      });
+
+      return <div>1</div>;
+    });
+    try {
+      await render(<Cmp />, { debug });
+    } catch (e: unknown) {
+      expect((e as Error).message.includes('Code(Q51)')).toBeTruthy;
```
wmertens, line 828: I think this needs updating? Also test with a different error, check that it does not add the message but still throws

wmertens (changes requested):

PR #7584, opened 2025-05-12
fix(v2): repl and some small fixes

wmertens:
also adds the names of params and captures to symbols so we can create an API document for server$ and routeAction$ endpoints


Varixo (approved):
LGTM

PR #7601, opened 2025-05-15
feat: qidlevisible event / 'idle-visible' vistask strategy

wmertens:
This makes Link's q-data prefetching less eager, so it doesn't interfere with LCP


packages/qwik/src/core/render/jsx/types/jsx-qwik-events.ts
```diff
@@ -2,6 +2,11 @@ import type { AllEventKeys } from './jsx-qwik-attributes';
 
 /** Emitted by qwik-loader when an element becomes visible. Used by `useVisibleTask$` @public */
 export type QwikVisibleEvent = CustomEvent<IntersectionObserverEntry>;
+/**
+ * Emitted by qwik-loader when an element becomes visible, after document idle. Used by
+ * `useVisibleTask$` @public
+ */
+export type QwikIdlevisibleEvent = CustomEvent<IntersectionObserverEntry>;
```
Varixo, line 9: Shouldn't be `QwikIdleVisibleEvent`?

maiieul:
Before merging this new strategy/API we need to make sure that `VisibleTaskStrategy = "intersection-observer"` can indeed be fired during LCP and therefore slow down LCP.

As discussed it's currently not the case (see screenshot) but might be in the future if we tweak things a little.

image
core is loaded after LCP, and we need core to execute a visible task

wmertens:
converted to draft, we may not need this strategy

wmertens:
closing this for now

PR #7605, opened 2025-05-16
chore(docs): remove sdk-qwik use and copy 404 page

wmertens:
this should remove the pagespeed complaint of unused js


gioboa (approved):
:+1:

PR #7606, opened 2025-05-16
fix(server-functions): only throw SSR guard when 'window' is undefined

### LINKED ISSUE #5874: [🐞] Error: Actions can not be invoked within the server during SSR.

Qwik Runtime

### Describe the bug

When attempting to vitest a component that invokes a `routeAction$` I'm getting the error:

```
Error: Actions can not be invoked within the server during SSR.
Action.run() can only be called on the browser, for example when a user clicks a button, or submits a form.
    at Proxy.routeActionQrl_action_submit_A5bZC7WO00A
```

I thought that vitest would render using the node environment and thus SSR and I could mock out the HTTP API call that my action is ultimately making.  This is not the case though and I can't find any good documentation on how to do this.  

```
  test('Should successfully remove users', async () => {
    const Wrapper = component$(() => {
      const removeDialogRef = useSignal<HTMLElement>();
      const onClose$ = $(onCloseSpy);

      return (
        <QwikCityMockProvider>
          <RemoveUsers
            ref={removeDialogRef}
            selectedUsers={selectedUsers}
            close={onClose$}
          />
        </QwikCityMockProvider>
      );
    });

    const { render, userEvent } = await createDOM();
    await render(<Wrapper />);
    await userEvent('.vi-submit-remove', 'click');
  });
```

Where within the RemoveUsers component it calls `action.submit(data as unknown as FormData)` which ultimately makes an HTTP POST call.

### Reproduction

http://no.repro.url

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (20) arm64 Apple M1 Ultra
    Memory: 212.80 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/.nvm/versions/node/v20.10.0/bin/node
    npm: 10.2.3 - ~/.nvm/versions/node/v20.10.0/bin/npm
    bun: 1.0.7 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 113.0.5672.126
    Safari: 17.3.1
  npmPackages:
    @builder.io/qwik: ^1.4.3 => 1.4.3
    @builder.io/qwik-auth: 0.1.3 => 0.1.3
    @builder.io/qwik-city: ^1.4.3 => 1.4.3
    @builder.io/qwik-react: 0.5.0 => 0.5.0
    undici: 5.28.2 => 5.28.2
    vite: 5.0.12 => 5.0.12
```

### Additional Information

_No response_

LogProphet:
# What is it?

- Bug

# Description

Previously the SSR‐guard in `server-functions.ts` unconditionally threw an error whenever `isServer` was true,
which blocked invocation of `routeAction$` under Vitest+JSDOM (and the QwikCityMockProvider) in user tests.

Now we narrow the guard to:

if (isServer && typeof window === 'undefined') {
  throw …;
}

After this fix:

- True SSR(no 'window') still throws an error as before
- JSDOM/Vitest tests(and browsers) have a global 'window', do not throw the error, and returns a promise

This will unblock testing of actions in JSDOM environments without impacting real SSR safety

Closes #5874

# Checklist

### DIFF
--- a/.changeset/lovely-cycles-pay.md
+++ b/.changeset/lovely-cycles-pay.md
@@ -1,0 +1,22 @@
+---
+'@builder.io/qwik-city': patch
+---
+
+FIX: Bug #5874 - SSR guard: only throw in true SSR (when window is undefined), allowing actions to run in Vitest tests
+
+How this change works:
+
+Previously the SSR‐guard in `server-functions.ts` unconditionally threw whenever `isServer` was true,  
+which blocked invocation of `routeAction$` under Vitest+JSDOM (and the QwikCityMockProvider) in user tests.  
+
+Now we narrow the guard to:
+if (isServer && typeof window === 'undefined') {
+  throw …;
+}
+
+This will ensure:
+
+- True SSR (no window) still throws as before.
+- JSDOM/Vitest tests (and browsers) have a global window, skip the throw, and return a Promise.
+
+This change unblocks testing of actions in JSDOM environments without impacting real SSR safety.
--- a/packages/qwik-city/src/runtime/src/server-functions.ts
+++ b/packages/qwik-city/src/runtime/src/server-functions.ts
@@ -88,7 +88,8 @@ export const routeActionQrl = ((
     });
 
     const submit = $((input: unknown | FormData | SubmitEvent = {}) => {
-      if (isServer) {
+      // only throw in true SSR (no mock context)
+      if (isServer && typeof window === 'undefined') {
         throw new Error(`Actions can not be invoked within the server during SSR.
 Action.run() can only be called on the browser, for example when a user clicks a button, or submits a form.`);
       }


wmertens:
I don't understand - why would isServer be true in this case?

LogProphet:
The original guard in `server-functions.ts` threw an error whenever `isServer` was true. Since Qwik treats the initial, un-hydrated pass under JSDOM as “SSR,” that guard also fired in Vitest tests and blocked actions there.

To resolve this, I refined the condition so we only throw errors in a real server environment (when there is no `window` global):

```ts
if (isServer && typeof window === 'undefined') {
  throw new Error(/* ... */);
}
```

* **Real SSR** (Node, no `window`) still errors as before.
* **JSDOM/Vitest** and **browsers** (where `window` exists) skip the throw and return a `Promise`.

To validate this solution, In a quick Node REPL test(via replit), I simulated both scenarios by defining and deleting `global.window`:

image

Result:
image


With this change, all existing Playwright E2E tests remain green, and Vitest+JSDOM tests now pass without needing further mocks.

Let me know your thoughts and feedback. I'm always happy to have a discussion about this and find better potential solutions!

wmertens:
@LogProphet what I don't understand is that isServer can be true when it's not supposed to be. 

It's a build constant that means the code is meant to be running on the server. So why is code that was meant for server side running client things? 

I think this indicates a bug.

LogProphet:
@wmertens 

It definitely is indicative of a bug. That's why I referenced the bug in my commit and PR description 🙃 #5874

let me clarify where the confusion comes from and why the change is necessary:

1. `isServer` doesnt mean “no DOM right now,” it's defined as: “this is the SSR bundle.”
Qwik (via Vite) compiles two builds: one for SSR and one for the browser. During the initial render—whether on your real server or inside Vitest’s JSDOM—Qwik runs the SSR build, so `isServer === true` even when JSDOM(Vitest) provides a window.

2. That SSR build still needs a way to detect “real” server versus “simulated” server.
Without our extra check, the guard for isServer will throw an error in both cases, blocking actions in any environment that uses the SSR bundle (including tests).


3. The change im making narrows the guard to only trip when there’s truly no DOM:

`if (isServer && typeof window === 'undefined') {
  throw new Error(/* … */);
}`

Real server (Node) has no window → guard still throws.

JSDOM tests (SSR code + fake window via Vitest) skip the throw and allow the action to run.

In short, the patch doesn’t change what counts as “SSR build,” it just adds a DOM–presence check so that running the SSR bundle inside a JSDOM(Vitest) environment no longer trips the guard. This keeps real-SSR protections intact while unblocking our Vitest+JSDOM test suite.

..does that make sense within the context of the linked bug? Let me know if you have any questions, or if you think im off-base here!

wmertens:
aha, I think the problem stems from build-time-replacement `isServer` vs calculated isServer.

`isServer` should only be true if the build is an SSR build, no matter how it was obtained. So you're on the right track but not in the right location.

maiieul:
Hey @Matt-Williams1 I'm closing as we now have vitest browser mode in v2 https://github.com/kunai-consulting/vitest-browser-qwik. Feel free to re-open if you think you found the proper fix.

PR #7607, opened 2025-05-17
perf(preloader): tweak probabilities

wmertens:
- don't decrease the probability of array preloads
- limit ssr preloads to the first 15 qrls on the page
- allow somewhat more ssr preloads
- when preloading at 100%, also preload the entire dynamic graph at 95%


gioboa (dismissed):
💪

maiieul (approved):

packages/qwik/src/server/preload-utils.unit.ts
```diff
@@ -1,5 +1,5 @@
 import { assert, test } from 'vitest';
-import { flattenPrefetchResources } from './prefetch-utils';
+import { flattenPrefetchResources } from './preload-utils';
```
shairez, line 2: I would rename it to "flattenPreloadResources" as well

shairez (approved):
LGTM Thanks!

gioboa (approved):
👏💪

Issue #7600, opened 2025-05-15
[🐞] v2: vite dev mode for a library project fails


JerryWu1234:
### Which component is affected?

Qwik Runtime

### Describe the bug

click content after run pnpm run start

### Reproduction

https://github.com/JerryWu1234/test-demo/tree/master

### System Info

```shell
latest system
```

JerryWu1234:
![Image](https://github.com/user-attachments/assets/d0194a1d-b3e5-4343-a908-277a428a1c55)

wmertens:
so apparently the @qwik-handlers aren't properly handled in dev mode for library projects. TBD

JerryWu1234:
done

PR #7608, opened 2025-05-17
fix(vite):  the @qwik-handlers aren't properly handled in dev mode for library projects

JerryWu1234:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

an http source request will be blocked by middleware in Qwikcore when project is the library mode.
this is because library mode's `plugin` function in `configureServer` will be executed immediately, in contrast, in a normal project that has a router Vite plugin will be deferred.

#7600 

# Checklist


JerryWu1234:
image
image

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -665,14 +665,7 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
...
         // TODO: Removed the "__qwikCityNew" condition in V3
-        const isNEW =
-          (globalThis as any).__qwikRouterNew === true ||
-          (globalThis as any).__qwikCityNew === true;
-        if (isNEW) {
-          return plugin;
-        } else {
-          return plugin();
-        }
+        return plugin;
```
wmertens, line 668: can you look in the commit history to see why this was added?

Are you sure that the function gets executed? Can you add a console.log() inside it to check?

JerryWu1234: If it directly returns a function, that function will be invoked by Vite at the very end

shairez: @JerryWu1234 if the comment says "remove in V3" it means that we deprecate it in V2 and then in V3 we'll remove this functionality as a breaking change. 

But for now, IIRC this was added to prevent backwards compatability issues, are you saying this has no effect?

can we just check for "library mode" instead and return the plugin in that mode specifically?

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -649,8 +649,8 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
       }
 
       if (!qwikViteOpts.csr) {
+        const opts = qwikPlugin.getOptions();
         const plugin = async () => {
-          const opts = qwikPlugin.getOptions();
```
wmertens: I think this should stay like it was, the options can change before this plugin is called

packages/qwik/src/optimizer/src/plugins/vite.ts
```diff
@@ -667,7 +667,9 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
         // TODO: Removed the "__qwikCityNew" condition in V3
         const isNEW =
           (globalThis as any).__qwikRouterNew === true ||
-          (globalThis as any).__qwikCityNew === true;
+          (globalThis as any).__qwikCityNew === true ||
+          server.config.build.lib;
```
wmertens: Please add a comment explaining why this is needed

maiieul:
This is blocking Jerry. We should merge ASAP if good.

shairez:
@wmertens can you please take a look if all good when you have a minute? (thanks :pray: )

wmertens (approved):
I'm not sure I quite understand how this works but LGTM

PR #7609, opened 2025-05-17
chore: minify some class props from core

Varixo:


gioboa (dismissed):

Varixo:
already done in another PR

PR #7613, opened 2025-05-19
perf(qwikloader): put qwikloader in a separate bundle

wmertens:
- move qwikloader to its own eagerly loaded chunk
- automatically unregister events that have no listener attributes
- optimize compressed qwikloader size

We were embedding the qwikloader, which is faster for initial load, but recent experiments with modulepreload show that this isn't necessary.

By putting the qwikloader in its own bundle and using modulepreload + async module script:
- the bundle is discovered during the html download
- the bundle downloads together with the html
- the bundle can be served with brotli compression, which is more costly to do during SSR
- subsequent SSR requests never have to download the 1.5-2kb again
- multiple containers with the same qwik version can all embed the script import tag and it will only be loaded once

This doesn't seem to impact LCP at all.

In the pathological 3G simulation with the fixed 2s lag, the chunk is fetched first, while the html is still downloading, and finishes pretty soon after html finished downloading and while the other resources are also downloading, so the user doesn't really notice anything.

image

compared to before:
image

(the other two yellow preloads you see are the preloader bundle and the qwik core bundle)

PS it would be really cute to find a string to embed in the preloader that would make the current `uD3ipkCl` hash start with `ql_`. Anybody with a lot of compute time to burn? :)


maiieul:
If I'm not mistaken the page can now be visible before the qwikloader is finished downloading, which means some user events might get lost between that time.

For example: 
image
If a user clicks between ~3100ms and ~4400ms on slow 3G the user event won't be registered. It's a little annoying for demos. 

On the other hand you could argue that slow 3G latency is unrealistic and that this change theoretically should improve performance for real users. 

So I'm not sure what I prefer here ^^.

I'd like to test on my Xiaomi on real 2G before we merge this.

wmertens:
would be interesting to see if 103 Early Hints would help, but Safari doesn't implement preload so that's a huge chunk of the mobile userbase for which it won't work.

wmertens:
I tried it on my Android phone as follows:
- go to settings, make sure you have developer options enabled
- search for rate limit, go there, set to 256kb
- go to https://qwikdev-qwikloader-chunk.qwik-8nx.pages.dev/ in incognito mode
- as soon as you see the search button, click it once
- after 25s, it shows the search box

this shows that the qwikloader was loaded fast enough IMHO

maiieul (approved):
Okay so in 2G for me there's only a tiny window where I can click and qwikloader isn't loaded yet. It's more visible on slow 3G throttling because of the fake latency. So users with real world latency could theoretically experience it, although they'd have to click pretty fast. But I believe this should slightly improve the FCP for most apps so I think it's a good change.

shairez:
@wmertens few things - 

1, The code looks great, good job!

2, I think a changeset is needed here

3. can we ship a tiny inline script that just records (remembers) the user events that were emitted until the qwik-loader loads and emits them again? (just to cover all cases)

That way we're safe even if it takes way longer for some reason 


WDYT?

wmertens:
@shairez hmm, the inline script would actually mean repeating much of the qwikloader internals. Our tests show that the qwikloader is available pretty instantly.

Added the changeset.

shairez:
ok cool 
thanks!

shairez (approved):

PR #7616, opened 2025-05-19
Add TS plugin for CSS modules to starter

FDiskas:
# What is it?

- Feature / enhancement

# Description

Satisfaction for developers.
Allow auto compleation in typescript files for css modules class names.

image

# Checklist


wmertens (changes requested):
This looks great but not everyone uses CSS modules.

Can you change your PR to be comments instead?

FDiskas:
> This looks great but not everyone uses CSS modules.
> 
> Can you change your PR to be comments instead? 

I fully understand, but when i create a project using cli, i get it with css modules by default.

FDiskas:
Thanks, good point. Probably even better add this only in boilerplate where css modules are actually used. But I am a bit confused about the repository where it lives

FDiskas:
Thank you for your patience and I apologize for the delay.

PR #7617, opened 2025-05-19
fix: fix up compatibility with a Qwik typescript version

FDiskas:
# What is it?

- Bug

# Description

Currently, it is impossible to install `Integration: i18n $localize (compile time translations)`
```
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: my-qwik-basic-starter@undefined
npm error Found: typescript@5.4.5
```

# Checklist


gioboa (commented):
Thanks for your help @FDiskas 👏

wmertens (changes requested):
Actually, maybe we should remove this integration.

There's https://github.com/wmertens/compiled-i18n which does an extremely similar thing but more integrated, and https://github.com/robisim74/qwik-speak which is more focused on runtime translations.

FDiskas:
> Thanks for your help @FDiskas 👏

Not really because core typescript version is older than angular

gioboa:
> Actually, maybe we should remove this integration.
> 
> There's https://github.com/wmertens/compiled-i18n which does an extremely similar thing but more integrated, and https://github.com/robisim74/qwik-speak which is more focused on runtime translations.

I see, let's create another PR for that.

PR #7618, opened 2025-05-20
perf(preloader): start preloading slightly earlier

wmertens:
I just learned that async scripts can be run earlier than their DOM order, so let's use that.


maiieul (approved):

Varixo (approved):

gioboa (approved):
Nice improvement 👏

PR #7619, opened 2025-05-20
fix(optimizer): support wrapping type-asserted variables and logical expressions in templates

Varixo:
Wrap with `_wrapProp`:
- type assertions: `(count as any).value`
- logical expression: `(count || count2).value`


packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__should_wrap_logical_expression_in_template.snap
```diff
@@ -0,0 +1,68 @@
...
+import { _fnSignal } from "@qwik.dev/core";
+import { _jsxSorted } from "@qwik.dev/core";
+import { useSignal } from "@qwik.dev/core";
+export const test_component_LUXeXe0DQrg = ()=>{
+    const count = useSignal(0);
+    const count2 = useSignal(0);
+    return /*#__PURE__*/ _jsxSorted("div", null, null, _fnSignal((p0, p1)=>(p0 || p1).value, [
+        count,
+        count2
+    ], "(p0||p1).value"), 3, "u6_0");
```
wmertens, line 41: shouldn't this have `(p0,p1)=>` in front?

wmertens (approved):
LGTM

PR #7630, opened 2025-05-22
chore(noSerialize): don't console when bad value

wmertens:
the error is not helpful and it's nicer to just check and ignore


the-zimmermann (approved):
Oh yes!

gioboa (approved):

Varixo (approved):

PR #7631, opened 2025-05-22
fix: ensure components are only rendered when necessary

Varixo:
If we've found the correct component and props are the same don't rerender it


wmertens (approved):
LGTM

gioboa (approved):

PR #7632, opened 2025-05-22
fix: serialize less vnode data

Varixo:


Varixo:
Closing in favor of #7636

PR #7636, opened 2025-05-23
fix: serialize less vnode data

Varixo:
Improve vNode data serialization, simplify the SsrNode class, remove unnecessary vnode preparation logic


wmertens (approved):
LGTM

PR #7637, opened 2025-05-24
fix: don't wrap static objects with signal

Varixo:
Static/const objects should not be wrapped with WrappedSignal, they are static and cannot change


packages/qwik/src/core/reactive-primitives/internal-api.ts
```diff
@@ -57,7 +57,7 @@ export const _wrapProp = <T extends Record<any, any>, P extends keyof T>(...args
     }
   }
   // We need to forward the access to the original object
```
wmertens: this comment is no longer correct, it should be something like "not a reactive object, just access it"

wmertens (approved):
State memory sellers hate this one simple trick! 👑

wmertens (approved):

PR #7643, opened 2025-05-26
docs: fix up cookbook debounce typing

extrordinaire:
# What is it?

# Description

## Summary
This PR addresses a TypeScript error related to the use of readonly tuple types in conjunction with QRL.

## Problem
Using a readonly tuple type A extends readonly unknown[] caused the following TypeScript diagnostic when invoking a QRL:

```txt
Argument of type '[...A]' is not assignable to parameter of type 'QrlArgs<(...args: A) => R>'. [2345]
```

## Root Cause
This appears to be a type compatibility issue between the readonly tuple spread ([...A]) and the expected QrlArgs type in the QRL function signature. It's possible there's an edge case or mismatch in how QrlArgs<(...args: A) => R> resolves in the presence of readonly tuples, or perhaps recent TypeScript updates introduced stricter rules around readonly spreads.

## Solution
The fix involves removing the readonly modifier from the generic tuple constraint:

```diff
- A extends readonly unknown[]
+ A extends unknown[]
```

This ensures compatibility with the QRL's expected argument types while preserving type safety.

### Notes
Further investigation might be warranted to confirm whether this is a TypeScript typing nuance or a limitation in the Qwik QRL typings themselves.

Tested on TypeScript 5.4.5

# Checklist

### DIFF
--- a/packages/docs/src/routes/docs/cookbook/debouncer/index.mdx
+++ b/packages/docs/src/routes/docs/cookbook/debouncer/index.mdx
@@ -31,7 +31,7 @@ The Qwik framework provides unique capabilities for managing state and effects i
 
 <CodeFile src="/src/routes/demo/cookbook/mediaController/index.tsx">
 ```tsx
-export const useDebouncer = <A extends readonly unknown[], R>(
+export const useDebouncer = <A extends unknown[], R>(
   fn: QRL<(...args: A) => R>,
   delay: number,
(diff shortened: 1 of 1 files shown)


gioboa (commented):
Thanks for your help @extrordinaire

wmertens (approved):
LGTM :lizard:

PR #7650, opened 2025-06-01
fix: bundlegraph as regular asset

wmertens:
- make bundlegraph a regular asset and don't assume anything about the path
- add assets list to q-manifest

This fixes issues with i18n and is cleaner


shairez (approved):
great job!

shairez (approved):
great job!

PR #7663, opened 2025-06-08
fix(starters): fix up vercel.json source pattern

cmbartschat:
# What is it?

- Bug

# Description

When deploying to vercel I saw an "[Invalid route source pattern](https://vercel.com/docs/errors/error-list#invalid-route-source-pattern)" error.

This pattern seems to have been modified in https://github.com/QwikDev/qwik/commit/f0388d410524844877c911ae22488e98b4f5858b#diff-c6a948ae10892b24f31a2545d6eb120f1a47f4f338221fa087eff4e7688e34b9R13 which dropped the parenthesis around the `.*`, which seem to be required. After this change I was able to deploy, and see the correct cache headers:

Screenshot showing response headers for a /build/ response from vercel, containing public, max-age=31536000, s-maxage=31536000, immutable


gioboa (approved):
Thanks for your help @cmbartschat

gioboa (dismissed):

PR #7664, opened 2025-06-09
Draft: Qwik plugin check

JerryWu1234:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


packages/eslint-plugin-qwik/tests/scope-use-task/valid-scope-use-return-object.tsx
wmertens, line 1: I'm not sure what this tests? Is there anything in this code that would be flagged with or without the qwik plugin?

packages/eslint-plugin-qwik/src/scope-use-task.ts
```diff
@@ -322,7 +322,15 @@ export const scopeUseTask: Rule.RuleModule = {
...
         if (forbiddenApis.has(node.name)) {
-          if (isIdentifierShadowedByDeclaration(node)) {
+          const isDirectIdentifier =
+            node.parent.type === 'VariableDeclarator' &&
+            (node?.parent?.init as Identifier)?.name === node.name;
+          // node'api usually be invoked as a member expression (e.g., process.env)
+          // or as a direct identifier (e.g., process).
+          if (
+            isIdentifierShadowedByDeclaration(node) ||
+            (node.parent.type !== 'MemberExpression' && node.parent.type !== 'VariableDeclarator')
```
wmertens: can you add a comment explaining why this works? It's very broad

.changeset/loud-mammals-dress.md
```diff
@@ -0,0 +1,5 @@
+---
+'eslint-plugin-qwik': patch
+---
+
+FIX: cover more qwik eslint's edge case
```
wmertens:
```
FIX: eslint-plugin: detect node API usage more accurately
```

wmertens (changes requested):

packages/eslint-plugin-qwik/src/scope-use-task.ts
```diff
@@ -178,29 +157,28 @@ export const scopeUseTask: Rule.RuleModule = {
         currentScopeForSearch = currentScopeForSearch.upper;
       }
 
-      if (!variable) {
-        // Cannot find variable, assume it's not a shadowed global for safety,
-        // though this state implies an undeclared variable (another ESLint rule should catch this).
-        return false;
+      // If we didn't find a variable, it might be a global API or an undeclared variable.
+      if (!GLOBALAPIS.includes(identifierNode.name)) {
+        return true;
```
wmertens: this seems wrong? It will almost always be true, and if it is correct then it should be befor the while loop because it's not using the result of the while loop

packages/eslint-plugin-qwik/src/scope-use-task.ts
```diff
@@ -178,29 +157,28 @@ export const scopeUseTask: Rule.RuleModule = {
...
       }
 
-      if (variable.defs.length === 0) {
+      if (variable?.defs.length === 0) {
```
wmertens: because you removed the !variable test, now you needed to add this. Just add the !varaible test again

packages/eslint-plugin-qwik/tests/scope-use-task/invalid-scope-use-function.tsx
```diff
@@ -4,15 +4,16 @@
 // Expect error: { "messageId": "unsafeApiUsageInCalledFunction" }
 
 import { component$, useTask$ } from '@qwik.dev/core';
-
 export default component$(() => {
+  function child_process() {}
```
wmertens: this is valid code, because it shadows the global child_process, and shouldn't fail the eslint test

packages/eslint-plugin-qwik/tests/scope-use-task/valid-scope-use-return-object.tsx
wmertens, line 1: is this testing for anything?

packages/eslint-plugin-qwik/src/scope-use-task.ts
```diff
@@ -178,29 +162,25 @@ export const scopeUseTask: Rule.RuleModule = {
...
+      if (!variable || variable.defs.length === 0) {
         // No definitions usually means it's an implicit global (e.g., 'process' in Node.js environment).
         // Such a variable is NOT considered "shadowed by a user declaration".
         return false;
       }
 
       // If there are definitions, check if any of them are standard declaration types.
       // This means the identifier refers to a user-declared variable, parameter, function, class, or an import.
-      return variable.defs.some((def) => {
+      return variable?.defs.some((def) => {
```
wmertens: this ? is not needed because you return if empty above it, and then the as boolean is'n't needed either

JerryWu1234:
@wmertens done

packages/eslint-plugin-qwik/tests/scope-use-task/valid-scope-use-return-object.tsx
```diff
@@ -0,0 +1,20 @@
...
+export default component$(() => {
+  const state = useSignal(true);
+  useTask$(({ track }) => {
+    if (isBrowser) {
+      track(() => {
+        if (state.value) {
+          // `path` will be treated as a built-in node API under isBrowser
+          const values = [
+            {
+              path: '1',
```
wmertens, line 12: ah maybe move the comment here

packages/eslint-plugin-qwik/tests/scope-use-task/valid-scope-use-return-object.tsx
```diff
@@ -0,0 +1,20 @@
+import { component$, useTask$, isBrowser, useSignal } from '@qwik.dev/core';
+
+export default component$(() => {
```
wmertens, line 3: maybe add `const process = {cwd: 'hi'}` and use it inside the `isBrowser` `if`

wmertens (changes requested):

PR #7666, opened 2025-06-11
fix: correctly serialize vnode props in production mode

Varixo:
When SsrNode props were equal to EMPTY_PROPS object, then after setting prop they were not updated in vnode data.


wmertens (approved):
LGTM

PR #7668, opened 2025-06-13
fix: resuming a component using styles and a text node

Varixo:
We should skip style node when it is before a text node


wmertens (approved):
LGTM

PR #7672, opened 2025-06-14
fix: reuse the same props instance when props are changing

Varixo:
- reuse the same instance of props when they changed. We need that, because some qrls could use props instance as a capture ref
- compare qrl hashes in chore comparator

### DIFF
--- a/packages/qwik/src/core/client/vnode-diff.ts
+++ b/packages/qwik/src/core/client/vnode-diff.ts
@@ -9,6 +9,7 @@ import {
   directGetPropsProxyProp,
   isJSXNode,
   type Props,
+  type PropsProxy,
 } from '../shared/jsx/jsx-runtime';
 import { Slot } from '../shared/jsx/slot.public';
 import type { JSXNodeInternal, JSXOutput } from '../shared/jsx/types/jsx-node';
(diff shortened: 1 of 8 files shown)


packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1064,8 +1064,21 @@ export const vnode_diff = (
...
         const vNodeProps = vnode_getProp<any>(host, ELEMENT_PROPS, container.$getObjectById$);
-        shouldRender = shouldRender || propsDiffer(jsxProps, vNodeProps);
+        const propsAreDifferent = propsDiffer(jsxProps, vNodeProps);
+        shouldRender = shouldRender || propsAreDifferent;
         if (shouldRender) {
+          if (propsAreDifferent) {
+            if (vNodeProps) {
+              // Reuse the same props instance, qrls can use the current props instance
+              // as a capture ref, so we can't change it.
+              Object.assign(vNodeProps, jsxProps);
```
wmertens: Shouldn't we delete the props that are no longer there?

Varixo: what do you mean?

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1063,16 +1065,40 @@ export const vnode_diff = (
...
+            if (vNodeProps) {
+              // Reuse the same props instance, qrls can use the current props instance
+              // as a capture ref, so we can't change it.
+              // We need to do this directly, because normally we would subscribe to the signals
+              // if any signal is there.
+              if (vNodeProps[_CONST_PROPS]) {
+                Object.assign(vNodeProps[_CONST_PROPS], (jsxProps as PropsProxy)[_CONST_PROPS]);
+              } else {
+                vNodeProps[_CONST_PROPS] = (jsxProps as PropsProxy)[_CONST_PROPS];
+              }
```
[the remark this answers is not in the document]
wmertens: @Varixo that's what I meant, but perhaps it's better to overwrite with null instead of deleting

wmertens (approved):
LGTM

Issue #7675, opened 2025-06-18
[🐞] Vite error overlay causes CPU/GPU to go insane


intellix:
### Which component is affected?

Qwik Runtime

### Describe the bug

The vite-error-overlay has a backdrop-filter which uses blur(20px) and causes my M4 Max to go insane: 

https://github.com/QwikDev/qwik/blob/a96af77cf009eb261f1b018e577a0abf2590d28e/packages/qwik/src/optimizer/src/plugins/vite-error.ts#L22

I checked the Chrome Task Manager and I was using 5GB of RAM and my laptop started getting really hot. I guess it's unnecessary for showing an error and can be replaced with something simpler

### Reproduction

N/A

### Steps to reproduce

N/A

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (14) arm64 Apple M4 Max
    Memory: 3.11 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.0 - ~/.volta/tools/image/node/22.15.0/bin/node
    npm: 10.9.2 - ~/.volta/tools/image/node/22.15.0/bin/npm
    pnpm: 10.6.2 - ~/.volta/bin/pnpm
  Browsers:
    Chrome: 137.0.7151.104
    Safari: 18.5
  npmPackages:
    @builder.io/qwik: ~1.14.1 => 1.14.1 
    @builder.io/qwik-city: ~1.14.1 => 1.14.1 
    @builder.io/sdk-qwik: ^0.20.1 => 0.20.1 
    typescript: ~5.7.2 => 5.7.3 
    undici: ^5.29.0 => 5.29.0 
    vite: ^6.0.0 => 6.3.5
```

### Additional Information

It might be related to the fact that we have so many errors and it's a blur over a blur over a blur etc:

<img width="348" alt="Image" src="https://github.com/user-attachments/assets/e5d91bcd-bfc1-4b54-a2f6-cfcd367feda8" />

PR #7676, opened 2025-06-18
fix(optimizer): remove backdrop-filter from vite-error-overlay for perf

intellix:
# What is it?

- Bug

# Description

Removes the backdrop-filter/blur when an error appears. In the event that you have multiple error overlays, it causes my Super Computer (M4 Max) to absolutely crawl. 

Closes https://github.com/QwikDev/qwik/issues/7675

# Checklist

### DIFF
--- a/.changeset/fancy-nights-chew.md
+++ b/.changeset/fancy-nights-chew.md
@@ -1,5 +1,5 @@
 ---
-'@builder.io/qwik': minor
+'@builder.io/qwik': patch
 ---
 
 Removed backdrop-filter of vite-error-overlay to prevent perf issues with multiple errors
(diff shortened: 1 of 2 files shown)


gioboa (dismissed):
Thanks for your feedback @intellix 
Can you share the before and after screenshots please?

wmertens:
Note that the real bug here is that there should only be one error overlay

intellix:
It was really hard for me to get a before/after because my browser tab is pretty much crashing with the blur on the stacked errors, but I couldn't really see any difference between the before/after

Screenshot 2025-06-18 at 15 09 11
Screenshot 2025-06-18 at 15 13 07

wmertens:
could you quickly change the changeset to `patch` @intellix ?

gioboa (approved):
Thanks for your help @intellix

PR #7677, opened 2025-06-19
WIP: fix: correctly resume vnodes with diacritics as a key

Varixo:


packages/qwik/src/core/client/vnode.ts
```diff
@@ -1549,17 +1549,21 @@ const processVNodeData = (
...
     consume: () => number,
+    getChar: (idx: number) => number,
     nextToConsumeIdx: number
   ) => void
 ) => {
   let nextToConsumeIdx = 0;
   let ch = 0;
   let peekCh = 0;
+  const getChar = (idx: number) => {
+    return idx < vData.length ? vData.charCodeAt(idx) : 0;
```
wmertens, line 1560:
```ts
return vData.charCodeAt(idx)
```

it will return NaN if out of bounds, which is also falsy, so we don't have to check the length.

ianlet:
tested it on our app with the issue and it seems good now

lgtm

PR #7689, opened 2025-06-23
Update index.mdx

SKhajeh:
After testing locally, it seems that the routeLoader$'s callback needs to be async as well.

# What is it?

# Description

Initially I was confused when the on-site provided example didn't work in my browser.
Then tried it locally, didn't work as well(the routeLoader$ still blocked rendering).
Tried it again with both the functions asynced, and it worked as described by the docs.

# Checklist


gioboa (approved):
Thanks, nice catch @SKhajeh

PR #7690, opened 2025-06-24
fix: Change Content-Type header to respect RFC 7231

joaomaridalho:
# What is it?

- Bug

# Description

I recently tried deploying a qwik app behind a nginx proxy with modsecurity. I've found that for example Route Actions were being intercepted by modsecurity after looking at the logs and doing a little bit of research I realized that the rules being triggered were because of the Content-Type header being in the format MIME type , charset. 

Turns out that according to [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5) the correct separator is ;

After looking at the source code I realized it was a simple fix and might help others in the future and make it so that if in use with modsecurity there is no need to disable the multiple rules that it affects

# Checklist

### DIFF
--- a/packages/qwik/package.json
+++ b/packages/qwik/package.json
@@ -45,8 +45,8 @@
       "import": {
         "development": "./dist/core.mjs",
         "production": "./dist/core.prod.mjs",
-        "default": "./dist/core.prod.mjs",
-        "min": "./dist/core.min.mjs"
+        "min": "./dist/core.min.mjs",
+        "default": "./dist/core.prod.mjs"
       },
       "require": {
         "development": "./dist/core.cjs",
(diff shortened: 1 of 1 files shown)


packages/qwik/package.json
wmertens, line 1: this should be undone, unfortunately there's something (not prettier) that sorts the keys and default should stay last

wmertens (changes requested):
Great but we need one change

packages/qwik/package.json
joaomaridalho: It was the pnpm fmt command, but I've reverted that file so it should be fine now

wmertens (approved):
LGTM

gioboa (approved):
Thanks for your help @joaomaridalho 🎉

PR #7691, opened 2025-06-24
chore: set nvmrc to node 22

wmertens:
this is used in the insights build


gioboa (approved):
It looks great to me 👏

PR #7693, opened 2025-06-25
fix: accept requests with HTTP if the origin is HTTPS

### LINKED ISSUE #7688: [🐞] CSRF protection may be too strict

Qwik City (routing)

### Describe the bug

recently, if you use an SSL-terminating proxy, server$ no longer works because qwik-router refuses the origin.

The workaround is to disable csrf https://qwik.dev/docs/deployments/node/

But maybe we should make it so http is accepted when the origin is https?

### Reproduction

https://discord.com/channels/842438759945601056/1385171655294320660

### Steps to reproduce

_No response_

### System Info

```shell
prod deploy with ssl terminating proxy
```

### Additional Information

See csrfCheckMiddleware (qwik-city/lib/middleware/request-handler)

gioboa:
Close #7688

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


packages/qwik-city/src/middleware/request-handler/resolve-request-handlers.ts
```diff
@@ -434,7 +434,13 @@ function csrfCheckMiddleware(requestEv: RequestEvent) {
   if (isForm) {
     const inputOrigin = requestEv.request.headers.get('origin');
     const origin = requestEv.url.origin;
-    const forbidden = inputOrigin !== origin;
+    let forbidden = inputOrigin !== origin;
+
+    // fix https://github.com/QwikDev/qwik/issues/7688
+    if (origin.startsWith('https://') && inputOrigin?.replace('http://', 'https://') === origin) {
```
wmertens:
```ts
    if (forbidden && origin.startsWith('https://') && inputOrigin?.slice(4) === origin.slice(5)) {
```

gioboa: it's a little bit less readable but ok, I'll change it

gioboa:
> Hmm shouldn't we make this configurable? This situation only arises when behind an ssl-terminating proxy

🤔 I would be against expanding a configuration just for such a particular case.

wmertens:
Let's have 2 csrf middlewares so we don't need to pass laxProto

wmertens (approved):
LGTM

Issue #7697, opened 2025-06-29
[🐞] Deno fails to build Qwik app due to absolute `outDir` (`/dist`) path


siguici:
### Which component is affected?

Qwik Runtime

### Describe the bug

When building a Qwik project with **Deno**, the process fails because the Vite `build.outDir` is unexpectedly set to an **absolute path** (`/dist`) instead of a **relative path** (`./dist`).
As a result, Deno attempts to write to the system root directory — which typically requires elevated permissions and therefore triggers a `EACCES: permission denied` error on most systems.

This also blocks the Astro integration from working properly in Deno:
🔗 [https://github.com/QwikDev/astro/issues/266](https://github.com/QwikDev/astro/issues/266)

**Expected behavior**

The build should succeed and generate output inside the local project folder (e.g. `./dist`), just like in Node or Bun environments.

**Actual behavior**

```shell
vite v5.3.5 building for production...

(!) outDir /dist is not inside project root and will not be emptied.
Use --emptyOutDir to override.

✓ 37 modules transformed.
x Build failed in 1.84s
error during build:
EACCES: permission denied, mkdir '/dist'
```

### Reproduction

https://github.com/siguici/qwik-deno

### Steps to reproduce

```shell
deno -A npm:create-qwik empty qwik-deno
cd qwik-deno
deno install --allow-scripts
deno task build
```

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (4) x64 Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz
    Memory: 6.80 GB / 7.72 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 22.17.0 - ~/.volta/bin/node
    npm: 10.9.2 - ~/.volta/bin/npm
    pnpm: 10.11.0 - ~/.local/share/pnpm/pnpm
    bun: 1.2.17 - ~/.bun/bin/bun
    deno: 2.3.5 - ~/.deno/bin/deno

  npmPackages:
    @builder.io/qwik: ^1.14.1 => 1.14.1
    @builder.io/qwik-city: ^1.14.1 => 1.14.1
    typescript: 5.4.5 => 5.4.5
    undici: * => 7.11.0
    vite: 5.3.5 => 5.3.5
```

### Additional Information

[...]

PR #7699, opened 2025-06-30
🐛 fix(#7697): use correct working directory for Deno environment

siguici:
# What is it?

* Bug

# Description

When building a Qwik project using **Deno**, the process fails because the Vite `build.outDir` is resolved to an **absolute path** (`/dist`) rather than a relative one. This happens because the optimizer currently sets `cwd()` to `'/'` by default for platforms other than Node.js and Bun ([source](https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/optimizer/src/platform.ts#L26)).

This causes Deno to attempt writing to the system root directory, which typically fails with a permission error (`EACCES`). This does not occur in Node or Bun environments, where `cwd()` resolves to `process.cwd()`.

In this PR, we fix the issue by using `Deno.cwd()` as the default `cwd()` value when the runtime environment is Deno.

It’s also worth noting that this `cwd()` is used to compute the `rootDir`, which influences Vite’s `outDir` ([see reference](https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/optimizer/src/plugins/plugin.ts#L212-L215)).

### Related Issue

* [https://github.com/QwikDev/qwik/issues/7697](https://github.com/QwikDev/qwik/issues/7697)

# Checklist

* [x] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
* [x] I performed a self-review of my own code
* [x] I added a changeset with `pnpm change`
* [ ] I made corresponding changes to the Qwik docs
* [ ] I added new tests to cover the fix / functionality

### DIFF
--- a/.changeset/dark-experts-scream.md
+++ b/.changeset/dark-experts-scream.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+🐛 Fix(#7697): Use correct working directory for Deno environment
--- a/packages/qwik/src/optimizer/src/platform.ts
+++ b/packages/qwik/src/optimizer/src/platform.ts
@@ -66,6 +66,10 @@ export async function getSystem() {
     sys.path = await sys.dynamicImport('node:path');
     sys.cwd = () => process.cwd();
     sys.os = process.platform;
+  } else if (sysEnv === 'deno') {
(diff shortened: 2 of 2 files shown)


gioboa (approved):
Thanks for your help @siguici 
It looks great to me 🎉

PR #7705, opened 2025-07-04
fix: render SVG attributes with correct namespace

Varixo:
few SVG attributes can have different namespaces


packages/qwik/src/core/client/vnode-namespace.ts
```diff
@@ -293,3 +300,22 @@ interface NewElementNamespaceData {
...
+    case 'xlink:title':
+    case 'xlink:type':
+    case 'xml:base':
+    case 'xml:lang':
+    case 'xml:space':
+      return XML_NS;
+    default:
+      return null;
+  }
+}
```
wmertens: we should benchmark this vs 2 regexes, but this seems to indicate a prepared regex is way faster: https://jsben.ch/XSViz

PR #7706, opened 2025-07-04
fix: adding popstate and scroll event for SPA navigation

Varixo:
For some reason sometimes addEventListener was calling qwik-loader function instead of native one


wmertens:
yeah the problem is the qwikloader polluting `win` in dev, this is ok but it's not the real fix

PR #7707, opened 2025-07-04
fix: don't wrap function calls in signal

Varixo:
Don't wrap a function call in wrapped signal, because it means we need to later serialize it


wmertens (approved):
LGTM

Issue #7718, opened 2025-07-10
[🐞] onSubmit$ form handler not preventing default


richardtorres314:
### Which component is affected?

Qwik Runtime

### Describe the bug

The `onSubmit$` event handler on a `form` is not respecting `preventDefault`. For example, in the form below:

```jsx
<form
  onSubmit$={$((event) => {
    event.preventDefault();
    console.log('default prevented');
  })}
>
  <button type="submit">Hello World</button>
</form>
```

## Expected Results
The form should not be submitted and the console should print "default prevented".

## Actual Results
The form is submitted, the default action on the event is never prevented, and nothing is printed to the console.

### Reproduction

https://stackblitz.com/edit/github-etyhm2js-2mqst9no?file=src%2Froutes%2Findex.tsx

### Steps to reproduce

1. Create an HTML `form` that has an `onSubmit$` handler.
2. Within the form `onSubmit$` handler, define the handler as:

```JavaScript
$((event) => {
  event.preventDefault();
  console.log('default prevented');
}
```
3. Add a `button` of type `submit` within the form.
4. Click the button to submit the form.

### System Info

```shell
System:
  OS: Linux 5.0 undefined
  CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Memory: 0 Bytes / 0 Bytes
  Shell: 1.0 - /bin/jsh
Binaries:
  Node: 18.20.3 - /usr/local/bin/node
  Yarn: 1.22.19 - /usr/local/bin/yarn
  npm: 10.2.3 - /usr/local/bin/npm
  pnpm: 8.15.6 - /usr/local/bin/pnpm
npmPackages:
  @builder.io/qwik: ^1.7.3 => 1.12.0 
  @builder.io/qwik-city: ^1.7.3 => 1.12.0 
  typescript: 5.4.5 => 5.4.5 
  undici: * => 7.2.0 
  vite: ^5.2.10 => 5.4.11
```

maiieul:
Hi @richardtorres314, have you tried using [`sync$`](https://qwik.dev/docs/cookbook/sync-events/#sync-synchronous-events-beta)? This is likely what you're looking for here.

maiieul:
I will close for now as this shouldn't be an issue with `sync$`. Feel free to re-open an issue if `sync$` doesn't work for your use-case.

PR #7723, opened 2025-07-11
chore(eslint): add warning when using async preventDefault() 

Shane-Donlon:
# What is it?

- Developer experience Improvement opportunity 
See also #7718

# Description

eslint rule showing error this is an asynchronous function and does not support preventDefault.
Use preventDefault attributes instead

Hey guys,
Feedback welcome - but thought I'd propose this as an idea.

While the knowledge exists in the documentation, the goal here is to prompt Developers that this is not a bug and to search the docs or to ask within Discord.

This does exclude `sync$((e)=>{e.preventDefault()}` functions, and I did add this as a warning over error just in case there was something else I wasn't considering, and wouldn't result in breaking any builds.

eslint rule for sync function not showing error

# Checklist

- [x ] I performed a self-review of my own code
- [ x] I added new tests to cover the fix / functionality


packages/eslint-plugin-qwik/src/noAsyncPreventDefault.ts
```diff
@@ -0,0 +1,38 @@
...
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Detect preventDefault in $(()=>{}) Async Functions.',
+      recommended: true,
+      url: 'https://qwik.dev/docs/components/events/#preventdefault--stoppropagation',
+    },
+    messages: {
+      noAsyncPreventDefault:
+        'This is an asynchronous function and does not support preventDefault. \nUse preventDefault attributes instead',
```
maiieul, line 13: Can you add a link to the docs? Also I think this should suggest the use of `sync$` too. WDYT?

maiieul (approved):
Really good idea 🔥  Thanks a lot @Shane-Donlon!

packages/eslint-plugin-qwik/src/noAsyncPreventDefault.ts
Shane-Donlon: Thanks Maiieul, I think it's a great idea to include `sync$` also

Is this a more agreeable message?

This is an asynchronous function and does not support preventDefault.
Use sync$ or prevent:default attributes instead.
Documentation: https://qwik.dev/docs/components/events/#preventdefault--stoppropagation
sync$: https://qwik.dev/docs/cookbook/sync-events/


image

Once agreed I'll commit and push up the message to avoid too many commits for editing the message.

Follow-up question:
Is a changeset needed for this?

wmertens: sounds good.

Do note that it's `preventdefault:click` etc, not what you wrote. Furthermore, those props only work if there is also a handler defined, in this case `onClick$`. Perhaps that should be classified as a bug though.

EDIT: actually, preventdefault:x works as long as qwikloader is listening for event x

wmertens (commented):
Looking good, but you also need to do stoppropagation

gioboa:
What's missing here?

Shane-Donlon:
Remaining to-do:

- [x] improve warning message with feedback from Wout above.
- [x] add stop propagation 
- [x] add tests for stop propagation 

I was away last week and forgot to acknowledge the review, but should be able to get around to it  this week. 

> What's missing here?

gioboa (commented):
Thanks @Shane-Donlon for your help.
is this PR still valid?

wmertens:
thansk @Shane-Donlon !

Issue #7720, opened 2025-07-10
[🐞] v2 useTask executed for wrong path


Varixo:
### Which component is affected?

Qwik Runtime

### Describe the bug

After some SPA navigation task is executed for wrong path
Example:
<img width="888" height="892" alt="Image" src="https://github.com/user-attachments/assets/b1fcb72b-58a3-45b4-85a7-0ea3b2145ace" />

Probably related to useLocation() hook and store. Store effects should goes through scheduler and it should fix the issue. 

This is waiting for the scheduler rewrite

### Reproduction

-

### System Info

```shell
-
```

Varixo:
should be solved by #7816

PR #7725, opened 2025-07-13
fix: ensure store updates go through scheduler during SPA navigation

naaa760:
## Hey, I just made a Pull Request!

#7720

# What is it?

- Bug

# Description

Fixes useTask being executed for wrong path after SPA navigation due to store updates not going through scheduler properly. Added untrack() around store mutations in QwikCityProvider to ensure correct scheduling.

- Solution

-  Added untrack() around all store mutations to ensure they go through Qwik's scheduler properly
- Fixed store updates for:

- routeLocation (url, params, isNavigating)
- content (headings, menu)
- documentHead (links, meta, styles, etc.)

Testing:

- The fix can be verified by:
- Creating a component that uses useTask$ with useLocation()
- Performing SPA navigation between routes
- Verifying that the task executes with the correct path

### DIFF
--- a/packages/qwik-city/src/runtime/src/qwik-city-component.tsx
+++ b/packages/qwik-city/src/runtime/src/qwik-city-component.tsx
@@ -392,33 +392,22 @@ export const QwikCityProvider = component$<QwikCityProps>((props) => {
         }
 
         // Update route location
-        if (!isSamePath(trackUrl, prevUrl)) {
-          untrack(() => {
+        untrack(() => {
+          if (!isSamePath(trackUrl, prevUrl)) {
             routeLocation.prevUrl = prevUrl;
-          });
-        }
-
-        untrack(() => {
+          }
           routeLocation.url = trackUrl;
-        });
-        untrack(() => {
           routeLocation.params = { ...params };
-        });
 
-        (routeInternal as any).untrackedValue = { type: navType, dest: trackUrl };
+          // Needs to be done after routeLocation is updated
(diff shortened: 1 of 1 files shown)


packages/qwik-city/src/runtime/src/qwik-city-component.tsx
```diff
@@ -291,7 +292,9 @@ export const QwikCityProvider = component$<QwikCityProps>((props) => {
     }
 
     actionState.value = undefined;
-    routeLocation.isNavigating = true;
+    untrack(() => {
```
wmertens, line 295: Can't you put all the assignments into a single untrack?

nehaaprasad: You are right @wmertens !
I fixed the SPA navigation bug by implementing a single untrack() block for all store mutations, ensuring proper scheduling and preventing unnecessary re-executions during navigation.

Varixo:
Also the #7720 is an issue for v2 only, and this PR is merging to main, so probably not related (on v2 this can be fixed on the scheduler rewrite branch)

wmertens:
So basically, this isn't fixing anything. `untrack()` doesn't actually prevent calling listeners, it only prevents subscribing as a listener.

We'd need a test that currently fails and that passes with this PR.

Varixo:
Thanks @naaa760 for the PR! It was fixed here #7816, so closing it

PR #7728, opened 2025-07-15
e2e(docs) add initial tests

Shane-Donlon:
# What is it?

- E2E Tests

# Description
Hey guys,
Thought I'd get the ball rolling on some playwright tests.

Current status is 124 tests - but majority are only page loading tests.

As such I'm only running them in chrome for now as it's pretty much testing network responses.

Leaving this in draft for now as I'm unsure how exactly this is to be implemented.

# Checklist

- [ x] I added new tests to cover the fix / functionality

## TODO:
There's more to do but some general pointers as to what is remaining.

- [ ] Tutorial e2e tests
- [ ] Qwik Sandbox
- [ ] Blog
- [ ] Theme switcher

### DIFF
--- a/e2e/docs-e2e/.gitignore
+++ b/e2e/docs-e2e/.gitignore
@@ -1,0 +1,7 @@
+
+# Playwright
+node_modules/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
--- a/e2e/docs-e2e/package-lock.json
+++ b/e2e/docs-e2e/package-lock.json
@@ -1,0 +1,97 @@
+{
+  "name": "docs-e2e",
+  "version": "1.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "docs-e2e",
+      "version": "1.0.0",
+      "license": "ISC",
+      "devDependencies": {
+        "@playwright/test": "^1.54.1",
+        "@types/node": "^24.0.13"
+      }
+    },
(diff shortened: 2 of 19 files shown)


wmertens:
This is great, but for me the big one is the sandbox, where we'd want to test that:
- the "app" actually loads
- the html etc tabs works
- changing the code updates the app

Shane-Donlon:
> This is great, but for me the big one is the sandbox, where we'd want to test that:
> 
> * the "app" actually loads
> * the html etc tabs works
> * changing the code updates the app

Thanks & acknowledged.

In terms of the sandbox testing the "hello world" section has been done.
I'll keep plugging away at the other sections too, once the whole sandbox has been covered I'll mark this for review.

image

Shane-Donlon:
_Things to note re: sandbox testing._

- Auto complete sandbox example does not work on the qwik.dev live site.
- I am testing that the page loads, but cannot test the API response etc..
- Will need to create a new issue later on this.
- raised on Discord to chat about it first.

_On the clock page_
- Cannot test paragraph on click 
- Presumably playwright on clicks on buttons links etc.. but I did try to add a test and it would always fail.
- In short cannot test that the below console.log works.

image

_in the routing.spec.ts file,_ 

- I am testing against the URL change only for now.

The page title is inconsistent to test against, will need to log a new issue against this.
But the bones are setup so once that is resolved it's not a big change.


Current Status:
211 tests - all pass using 1 worker, I did have some timing issues running 8 workers locally.

image />

PR #7730, opened 2025-07-15
fix: getting invoke context for loaders in production

Varixo:
We should not use $$ varaibles cross-packages, because they can be minified


wmertens (approved):
LGTM

PR #7731, opened 2025-07-15
feat: useQwikRouter

wmertens:
This allows colocating the headers and root, to make it easier to understand what's going on, and to make it more intuitive to change.

Also, remove the unnecessary qwikRouterConfig prop drilling, and make trailingSlash a build constant.


Varixo (approved):
LGTM

maiieul:
This only affects starters/adapters right? Existing code shouldn't be affected?

wmertens:
@maiieul indeed everything should keep working. The only API change is that the qwikPlugin normalizeOptions now returns a Promise but that's pretty much an internal API, it wasn't even documented.

gioboa (approved):

PR #7746, opened 2025-07-20
fix: use on events for headless components

Varixo:
Correctly handle a case where only primitve value is returned from component

### DIFF
--- a/packages/qwik/src/core/shared/component-execution.ts
+++ b/packages/qwik/src/core/shared/component-execution.ts
@@ -131,70 +131,67 @@ export const executeComponent = (
 };
 
 /**
- * Stores the JSX output of the last execution of the component.
+ * Adds `useOn` events to the JSX output.
  *
- * Component can execute multiple times because:
- *
- * - Component can have multiple tasks
- * - Tasks can track signals
- * - Task A can change signal which causes Task B to rerun.
- *
- * So when executing a component we only care about its last JSX Output.
+ * @param jsx The JSX output to modify.
+ * @param useOnEvents The `useOn` events to add.
+ * @returns The modified JSX output.
  */
-
 function addUseOnEvents(
   jsx: JSXOutput,
   useOnEvents: UseOnMap
 ): ValueOrPromise<JSXNodeInternal<string> | null | JSXOutput> {
-  const jsxElement = findFirstStringJSX(jsx);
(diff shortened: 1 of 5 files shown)


packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -222,50 +233,79 @@ function findFirstStringJSX(jsx: JSXOutput): ValueOrPromise<JSXNodeInternal<stri
...
+      return [placeholder, _jsxSorted(Fragment, null, null, [jsx, placeholder], 0, null)];
     }
 
     if (jsx.children == null) {
-      jsx.children = jsxElement;
-    } else if (Array.isArray(jsx.children)) {
-      jsx.children.push(jsxElement);
+      jsx.children = placeholder;
+    } else if (isArray(jsx.children)) {
+      jsx.children.push(placeholder);
```
wmertens, line 279: Actually if it's a visible task, doesn't putting it at the end make it fire when the last one is visible?

packages/qwik/src/core/shared/utils/types.ts
```diff
@@ -24,6 +24,17 @@ export const isFunction = <T extends (...args: any) => any>(v: unknown): v is T
   return typeof v === 'function';
 };
 
+export const isPrimitive = (v: unknown): v is string | number | boolean | null | undefined => {
```
wmertens: I wonder if it's not faster as "not an object or function"?

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -130,70 +131,74 @@ export const executeComponent = (
...
-                '" using `useOn` hook, ' +
-                'but a node to which you can add an event is not found. ' +
-                'Please make sure that the component has a valid element node. '
-            );
+            continue;
+          }
+        }
+        if (targetElement) {
+          if (targetElement.type === 'script' && key === qVisibleEvent) {
+            eventKey = 'document:onQinit$';
```
wmertens, line 185: Maybe we should also print a warning that the visible changed to init

packages/qwik/src/core/shared/component-execution.ts
Varixo: placeholder uses qinit instead

PR #7752, opened 2025-07-22
fix(router): inline staticPaths and notFounds

wmertens:
Vite was inlining the temporary helper functions for static paths and 404 html.
Now, we just build the helper functions and rewrite the placeholder data inline after the build.


wmertens:
Sadly it's not working, I'm going to approach this differently

PR #7764, opened 2025-07-30
feat: enhance debugging with development hooks in use-computed, use-c…

JerryWu1234:
…ontext, use-id, use-resource, and use-signal

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/packages/qwik/src/core/use/use-computed.ts
+++ b/packages/qwik/src/core/use/use-computed.ts
@@ -9,6 +9,8 @@ import { createComputedSignal } from '../reactive-primitives/signal-api';
 import type { AsyncComputedSignalImpl } from '../reactive-primitives/impl/async-computed-signal-impl';
 import type { SerializerSignalImpl } from '../reactive-primitives/impl/serializer-signal-impl';
 import type { ComputedOptions } from '../reactive-primitives/types';
+import { isDev } from '@qwik.dev/core/build';
+import { HookType, wrapSeq } from './utils/for-devTool';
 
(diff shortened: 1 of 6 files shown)


Varixo:
Why we need to add this? I don't see why we need this in core. Also it will be used during dev mode, so for me its no go. We shouldn't add things to core because of some external libraries/tools.
Instead can't you just export signal classes and compare instances?

JerryWu1234:
> Why we need to add this? I don't see why we need this in core. Also it will be used during dev mode, so for me its no go. We shouldn't add things to core because of some external libraries/tools. Instead can't you just export signal classes and compare instances?


image

I'm going to list all of the user's state. but in some situations it could be listed clearly.
let me give you one example.
`useStyles$` is just a hash, useId is the same too.
I have no idea what it is when it's just a hash, so this way will return object `{useStyle$: hash, useId: number  }`
@Varixo

JerryWu1234:
> > Why we need to add this? I don't see why we need this in core. Also it will be used during dev mode, so for me its no go. We shouldn't add things to core because of some external libraries/tools. Instead can't you just export signal classes and compare instances?
> 
> image
> I'm going to list all of the user's state. but in some situations it could be listed clearly. let me give you one example. `useStyles$` is just a hash, useId is the same too. I have no idea what it is when it's just a hash, so this way will return object `{useStyle$: hash, useId: number }` @Varixo

or May I just export some ? such as useStyle useId ?

PR #7770, opened 2025-08-03
feat(ssr): pass additional `documentHead` via `serverData`

wmertens:
This will mostly be useful for the Vite environment API changes to allow injecting vite dev scripts, but seems useful in general.


packages/qwik-router/src/runtime/src/head.ts
```diff
@@ -18,9 +18,10 @@ export const resolveHead = (
   endpoint: EndpointResponse | ClientPageData,
   routeLocation: RouteLocation,
   contentModules: ContentModule[],
-  locale: string
+  locale: string,
+  fromServer?: DocumentHeadValue
```
Varixo: why name `fromServer`? I would name it somehow better :D

packages/qwik/src/server/types.ts
```diff
@@ -168,7 +168,11 @@ export interface RenderOptions extends SerializeDocumentOptions {
    */
   containerTagName?: string;
   containerAttributes?: Record<string, string>;
-  serverData?: Record<string, any>;
+  /** Metadata that can be retrieved during SSR with `useServerData()`. */
+  serverData?: Record<string, any> & {
+    // note that this is leaking Qwik Router into Qwik but it's just types
+    documentHead?: import('@qwik.dev/router').DocumentHeadValue;
+  };
```
Varixo: can't we do typing on the qwik router side?

packages/qwik-router/src/runtime/src/head.ts
wmertens: right, it's the default values, I'll name it `defaults`

PR #7775, opened 2025-08-04
feat: <DocumentHeadTags />

wmertens:
This component renders the head tags from `head` exports correctly, including the scripts that were added later.

This reduces mistakes and makes the starters cleaner.

Furthermore, the styles and scripts types now allow setting props directly, which is like meta and links.


gioboa (approved):
Thanks @wmertens 
this will improve types and abstract a common component for the end users

PR #7784, opened 2025-08-06
chore: warn when importing core twice

wmertens:
v2 version of #7109 with vendorRoots removal, but not throwing


wmertens:
the reason playwright errored and now the cli tests that both run vite in ssr server mode which loads vite config with uses qwik-router vite plugin, which imports qwik core, and then the preview server runs in-process with its own qwik core copy. So it's actually a bug in our bundling!

gioboa (approved):
💪👏

PR #7796, opened 2025-08-07
chore: remove deprecated qwik-auth 🍉

gioboa:


maiieul:
This feels a bit like a breaking change. Maybe we should merge that to v2 only?

gioboa (commented):
Yep, fair enough, let me close this PR.
I will open a new one against the v2 branch.

PR #7799, opened 2025-08-08
fix(build): make types rolldown-vite compatible

wmertens:
I overrode vite with rolldown-vite and then fixed the types. It doesn't manage to build docs though, but at least the types won't impede progress.


gioboa (approved):
👍

PR #7804, opened 2025-08-10
fix(zod): locale in use for Zod validator messages

knoid:
# What is it?

- Bug

# Description

Runs the Zod validator inside a `withLocale` so error messages use the correct language when configured via the `setErrorMap` method of the zod library.

# Checklist


wmertens (approved):
Actually, I thought that the locale was using async localstorage, which would make this unnecessary, but it's not. I'll make an issue.

LGTM

PR #7809, opened 2025-08-12
fix: reactivity after spreading props

Varixo:
Spread props caused calling get trap for every props property. It means propsproxy will unwrap signals and pass flat value instead. Now we are using new _getVarProps, _getConstProps functions to direct get props object instead of using proxy

### DIFF
--- a/packages/qwik/src/core/tests/component.spec.tsx
+++ b/packages/qwik/src/core/tests/component.spec.tsx
@@ -2365,6 +2365,18 @@ describe.each([
     expect(document.body.innerHTML).toContain('I am the innerHTML content!');
   });
 
+  it('should not throw when props are null', async () => {
+    const Child = component$((props: any) => {
+      props = (globalThis as any).stuff ? (globalThis as any).foo : (globalThis as any).bar;
+      return <div {...props} />;
+    });
+    const Cmp = component$(() => {
+      return <Child />;
+    });
+    const { document } = await render(<Cmp />, { debug });
+    await expect(document.querySelector('div')).toMatchDOM(<div />);
+  });
+
   describe('regression', () => {
     it('#3643', async () => {
       const Issue3643 = component$(() => {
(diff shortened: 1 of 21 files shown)


packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_jsx.snap
```diff
@@ -58,17 +60,15 @@ import { Fragment as _Fragment } from "@qwik.dev/core/jsx-runtime";
 export const Lightweight = (props)=>{
     return /*#__PURE__*/ _jsxSorted("div", null, null, /*#__PURE__*/ _jsxSorted(_Fragment, null, null, [
         /*#__PURE__*/ _jsxSorted("div", null, null, null, 3, null),
-        /*#__PURE__*/ _jsxSplit("button", {
-            ...props
-        }, null, null, 0, null)
+        /*#__PURE__*/ _jsxSplit("button", _getVarProps(props), _getConstProps(props), null, 0, null)
```
wmertens: so, shouldn't this be `{..._getVarProps(props)}`? Or maybe _getVarProps should get a flag to return a shallowly cloned object?
Or we disallow shallowly changing the props object, as an optimization?

packages/qwik/src/optimizer/core/src/transform.rs
```diff
@@ -1525,7 +1500,42 @@ impl<'a> QwikTransform<'a> {
...
 							}
 						}
-						// spread props and other PropOrSpread cases (pretty much impossible)
+						// spread props
+						ast::PropOrSpread::Spread(spread) => {
+							if spread.expr.is_ident() {
+								let (_, var_props_call, const_props_call, _, _) =
+									self.handle_jsx_props_obj_spread(&spread);
+
+								// If this is the first spread and there are more spreads after it,
```
wmertens: `// If there are more spreads after it,`

wmertens (commented):
Looking great, just wondering about making var mutable?

packages/qwik/src/core/tests/component.spec.tsx
```diff
@@ -2365,6 +2365,18 @@ describe.each([
     expect(document.body.innerHTML).toContain('I am the innerHTML content!');
   });
 
+  it('should not throw when props are null', async () => {
+    const Child = component$((props: any) => {
+      props = null;
```
wmertens: this gets optimized into basically `<div {...null} />`. Instead do `props = globalThis.stuff ? globalThis.foo : globalThis.bar;`

wmertens (changes requested):

wmertens (approved):
LGTM

PR #7810, opened 2025-08-12
fix: handle falsy value for dangerouslySetInnerHTML

Varixo:
Don't stringify falsy values for dangerousltSetInnerHTML


wmertens (approved):
LGTM

PR #7818, opened 2025-08-13
perf: remove unregister events in qwikloader

Varixo:
remove a grace period before unregistering events from qwikloader due to perf issues


wmertens (approved):
for context: this was causing a slowdown on new DOM element creation

LGTM

PR #7823, opened 2025-08-14
fix(rolldown): some fixes for rolldown-vite

wmertens:
make it possible to run docs with rolldown-vite

However, editor still breaks on Prism missing from imports


gioboa (approved):
💯 @wmertens you rock

PR #7826, opened 2025-08-15
feat: enhance locale handling with AsyncLocalStorage support for serv…

JerryWu1234:
…er-side requests

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/packages/qwik-router/src/middleware/request-handler/user-response.ts
+++ b/packages/qwik-router/src/middleware/request-handler/user-response.ts
@@ -13,6 +13,7 @@ import {
 } from './request-event';
 import { encoder } from './resolve-request-handlers';
 import type { QwikSerializer, ServerRequestEvent, StatusCodes } from './types';
+import { withLocale } from '@qwik.dev/core';
 // Import separately to avoid duplicate imports in the vite dev server
 import {
   AbortMessage,
@@ -64,8 +65,10 @@ export function runQwikRouter<T>(
     response: responsePromise,
     requestEv,
     completion: asyncStore
-      ? asyncStore.run(requestEv, runNext, requestEv, rebuildRouteInfo, resolve!)
-      : runNext(requestEv, rebuildRouteInfo, resolve!),
+      ? asyncStore.run(requestEv, () =>
+          withLocale(requestEv.locale(), () => runNext(requestEv, rebuildRouteInfo, resolve!))
+        )
+      : withLocale(requestEv.locale(), () => runNext(requestEv, rebuildRouteInfo, resolve!)),
   };
 }
 
--- a/packages/qwik/src/core/use/use-locale.ts
+++ b/packages/qwik/src/core/use/use-locale.ts
@@ -11,6 +11,21 @@ let _locale: string | undefined = undefined;
  * @public
  */
 export function getLocale(defaultLocale?: string): string {
+  // Prefer per-request locale from AsyncLocalStorage if available (server-side)
+  try {
+    const asyncStore = (globalThis as any).qcAsyncRequestStore;
+    const ev = asyncStore?.getStore?.();
+    const evLocale = ev && typeof ev.locale === 'function' ? ev.locale : undefined;
+    if (evLocale) {
+      const l = evLocale();
+      if (l) {
+        return l;
+      }
+    }
+  } catch {
+    // ignore and fallback
+  }
+
   if (_locale === undefined) {
     const ctx = tryGetInvokeContext();
     if (ctx && ctx.$locale$) {
@@ -30,6 +45,24 @@ export function getLocale(defaultLocale?: string): string {
  * @public
  */
 export function withLocale<T>(locale: string, fn: () => T): T {
+  // If running on the server with AsyncLocalStorage, set locale on the current request
+  try {
+    const asyncStore = (globalThis as any).qcAsyncRequestStore;
+    const ev = asyncStore?.getStore?.();
+    const evLocale = ev && typeof ev.locale === 'function' ? ev.locale : undefined;
+    if (evLocale) {
+      const previous = evLocale();
+      try {
+        evLocale(locale);
+        return fn();
+      } finally {
+        evLocale(previous);
+      }
+    }
+  } catch {
+    // ignore and fallback
+  }
+
   const previousLang = _locale;
   try {
     _locale = locale;
@@ -48,5 +81,17 @@ export function withLocale<T>(locale: string, fn: () => T): T {
  * @public
  */
 export function setLocale(locale: string): void {
+  // On the server, prefer setting the locale on the per-request store
+  try {
+    const asyncStore = (globalThis as any).qcAsyncRequestStore;
+    const ev = asyncStore?.getStore?.();
+    const evLocale = ev && typeof ev.locale === 'function' ? ev.locale : undefined;
+    if (evLocale) {
+      evLocale(locale);
+      return;
+    }
+  } catch {
+    // ignore and fallback
+  }
   _locale = locale;
 }


JerryWu1234:
@wmertens does it need a test?.

packages/qwik-router/src/middleware/request-handler/user-response.ts
```diff
@@ -64,8 +65,10 @@ export function runQwikRouter<T>(
     response: responsePromise,
     requestEv,
     completion: asyncStore
-      ? asyncStore.run(requestEv, runNext, requestEv, rebuildRouteInfo, resolve!)
-      : runNext(requestEv, rebuildRouteInfo, resolve!),
+      ? asyncStore.run(requestEv, () =>
```
wmertens: I think you can put the withLocale around the whole thing instead

packages/qwik/src/core/use/use-locale.ts
wmertens, line 1: don't use the qwikcity asyncstore here. Instead make one in the local module (inside `if (isServer)`) and just accept the string that is given and return it like before.

Qwik Core should be usable seperately from Router, so it can't rely on its store.

This module should just make an asyncstore if possible, and then use it to store and return the locale string, or use _locale as before.

wmertens (changes requested):
Good start but I think Core should not be influenced by Router.

packages/qwik-router/src/middleware/request-handler/user-response.ts
```diff
@@ -63,9 +64,11 @@ export function runQwikRouter<T>(
   return {
     response: responsePromise,
     requestEv,
-    completion: asyncStore
-      ? asyncStore.run(requestEv, runNext, requestEv, rebuildRouteInfo, resolve!)
-      : runNext(requestEv, rebuildRouteInfo, resolve!),
+    completion: withLocale(requestEv.locale(), () =>
+      asyncStore
```
wmertens: Please restore the ternary as it was and then put the withLocale around it

packages/qwik-router/src/runtime/src/head.ts
```diff
@@ -37,9 +37,10 @@ export const resolveHead = (
     }
     return data;
   }) as any as ResolveSyncValue;
+  const storeEv = (globalThis as any).qcAsyncRequestStore?.getStore?.();
```
wmertens: You're not actually using storeEv, so just the existence of qcAsyncRequestStore is enough to know that Core has one too

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -1,7 +1,30 @@
...
 
+type LocaleStore = { locale: string | undefined };
+
+type LocaleAsyncStore = import('node:async_hooks').AsyncLocalStorage<LocaleStore>;
+
+let localAsyncStore: LocaleAsyncStore | undefined;
+
+if (isServer) {
+  try {
+    // Lazy import to avoid bundling for non-Node targets
```
wmertens: No, the reason is that it's not always supported.

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -1,7 +1,30 @@
...
+    import('node:async_hooks')
+      .then((module) => {
+        const AsyncLocalStorage = module.AsyncLocalStorage as unknown as new () => LocaleAsyncStore;
```
wmertens: Doesn't it get the correct type?

wmertens (changes requested):

packages/qwik-router/src/middleware/request-handler/user-response.ts
```diff
@@ -63,9 +64,11 @@ export function runQwikRouter<T>(
...
+        ? asyncStore.run(requestEv, () => runNext(requestEv, rebuildRouteInfo, resolve!))
+        : runNext(requestEv, rebuildRouteInfo, resolve!)
```
wmertens: I think this is the most efficient:
```ts
completion: withLocale(requestEv.locale(),
      asyncStore
        ? () => asyncStore.run(requestEv, runNext, requestEv, rebuildRouteInfo, resolve!)
        : () => runNext(requestEv, rebuildRouteInfo, resolve!)
  )
```

packages/qwik-router/src/runtime/src/head.ts
```diff
@@ -37,9 +37,10 @@ export const resolveHead = (
     }
     return data;
   }) as any as ResolveSyncValue;
+  const storeEv = (globalThis as any).qcAsyncRequestStore;
   const headProps: DocumentHeadProps = {
     head,
-    withLocale: (fn) => withLocale(locale, fn),
+    withLocale: (fn) => (storeEv ? fn() : withLocale(locale, fn)),
```
wmertens:
```ts
    withLocale: storeEv ? fn => fn() : withLocale(locale, fn),
```

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -1,7 +1,29 @@
...
+        localAsyncStore = new AsyncLocalStorage();
+      })
+      .catch(() => {
+        // ignore if AsyncLocalStorage is not available
+      });
+  } catch {
+    // ignore and fallback
+  }
```
wmertens: actually I think the try-catch isn't needed, the dynamic import itself shouldn't fail, and the `.catch` will already prevent errors when the module doesn't exist>

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -11,6 +33,17 @@ let _locale: string | undefined = undefined;
...
+  try {
+    const store = localAsyncStore?.getStore?.();
+    const l = store?.locale;
+    if (l) {
+      return l;
+    }
+  } catch {
+    // ignore and fallback
+  }
+
```
wmertens, line 35: doesn't this work too?

```ts
const locale = localAsyncStore?.getStore().locale
if (locale) {
  return locale
}
```

JerryWu1234:
@wmertens done .please check it again ,  thanks

packages/qwik-router/src/runtime/src/head.ts
```diff
@@ -37,9 +37,10 @@ export const resolveHead = (
     }
     return data;
   }) as any as ResolveSyncValue;
+  const storeEv = (globalThis as any).qcAsyncRequestStore;
```
wmertens: can you rename this to `hasAsyncStore` and add a comment `// Qwik Core will also be using the async store if this is present`

Actually, it would be better to export the store from where it is made and to import it.

instead of globalThis, make it `export let asyncRequestStore` and then in the other places it's used just import it. That way, we can be sure it is initialized correctly, and we don't pollute globalThis.

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -11,6 +29,16 @@ let _locale: string | undefined = undefined;
...
+  // Prefer per-request locale from local AsyncLocalStorage if available (server-side)
+  try {
+    const locale = localAsyncStore?.getStore?.()?.locale;
+    if (locale) {
+      return locale;
+    }
+  } catch {
+    // ignore and fallback
+  }
+
```
wmertens:
```ts
if (localAsyncStore) {
  const locale = localAsyncStore.getStore()?.locale
  if (locale) {
    return locale
  }
}
```

don't add try/catch and conditionals when not needed, it has a runtime cost.

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -1,7 +1,25 @@
 import { tryGetInvokeContext } from './use-core';
+import { isServer } from '@qwik.dev/core/build';
 
 let _locale: string | undefined = undefined;
 
+type LocaleStore = { locale: string | undefined };
+
+type LocaleAsyncStore = import('node:async_hooks').AsyncLocalStorage<LocaleStore>;
+
+let localAsyncStore: LocaleAsyncStore | undefined;
```
wmertens:
```ts
import type {AsyncLocalStorage} from 'node:async_hooks'

let asyncStore: AsyncLocalStorage<{locale?: string}> | undefined
```

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -30,6 +58,15 @@ export function getLocale(defaultLocale?: string): string {
...
  */
 export function withLocale<T>(locale: string, fn: () => T): T {
+  // If running on the server with AsyncLocalStorage, set locale for this async context
+  try {
+    if (localAsyncStore?.run) {
+      return localAsyncStore.run({ locale }, fn);
+    }
+  } catch {
+    // ignore and fallback
+  }
```
wmertens:
```ts
if (localAsyncStore) {
  return localAsyncStore.run({ locale }, fn);
}
```

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -48,5 +85,15 @@ export function withLocale<T>(locale: string, fn: () => T): T {
...
+  // On the server, prefer setting the locale on the local per-request store
+  try {
+    const store = localAsyncStore?.getStore?.();
+    if (store) {
+      store.locale = locale;
+      return;
+    }
+  } catch {
+    // ignore and fallback
+  }
```
wmertens: see above

wmertens (changes requested):

packages/qwik-router/src/runtime/src/head.ts
```diff
@@ -13,6 +13,7 @@ import type {
   ActionInternal,
 } from './types';
 import { isPromise } from './utils';
+import { asyncRequestStore } from '../../middleware/request-handler/async-request-store';
```
wmertens: actually, you can do 
```ts
  // Qwik Core will also be using the async store if this is present
import { asyncRequestStore as hasAsyncStore } from '../../middleware/request-handler/async-request-store';
```

and that saves an assignment

packages/qwik-router/src/runtime/src/head.ts
```diff
@@ -37,9 +38,11 @@ export const resolveHead = (
     }
     return data;
   }) as any as ResolveSyncValue;
+  // Qwik Core will also be using the async store if this is present
+  const hasAsyncStore = asyncRequestStore;
```
wmertens: so then this isn't needed

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -30,6 +55,15 @@ export function getLocale(defaultLocale?: string): string {
  * @public
  */
 export function withLocale<T>(locale: string, fn: () => T): T {
+  // If running on the server with AsyncLocalStorage, set locale for this async context
+  try {
```
wmertens: this should not catch!

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -48,5 +82,15 @@ export function withLocale<T>(locale: string, fn: () => T): T {
  * @public
  */
 export function setLocale(locale: string): void {
+  // On the server, prefer setting the locale on the local per-request store
+  try {
```
wmertens: here also, no try/catch, and I think getStore() is always defined?

JerryWu1234:
@wmertens still need a global variable `qcAsyncRequestStore`.
because in server-functions  file can't read `asyncRequestStore`

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -30,6 +55,10 @@ export function getLocale(defaultLocale?: string): string {
  * @public
  */
 export function withLocale<T>(locale: string, fn: () => T): T {
+  if (localAsyncStore) {
+    return localAsyncStore.run({ locale }, fn);
```
wmertens: Just set the locale itself as the store
```ts
    return localAsyncStore.run(locale, fn);
```
There's no reason to create and destructure objects for this.

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -1,7 +1,24 @@
...
+    .then((module) => {
+      const AsyncLocalStorage = module.AsyncLocalStorage as unknown as new () => AsyncLocalStorage<{
+        locale?: string;
+      }>;
+      localAsyncStore = new AsyncLocalStorage();
+    })
+    .catch(() => {
+      // ignore if AsyncLocalStorage is not available
+    });
+}
```
wmertens:
```ts
let localAsyncStore: AsyncLocalStorage<string> | undefined;

if (isServer) {
  import('node:async_hooks')
    .then((module) => {
      localAsyncStore = new module.AsyncLocalStorage();
    })
    .catch(() => {
      // ignore if AsyncLocalStorage is not available
    });
}
```

packages/qwik-router/src/middleware/request-handler/user-response.ts
```diff
@@ -27,12 +29,11 @@ export interface QwikRouterRun<T> {
...
 import('node:async_hooks')
   .then((module) => {
     const AsyncLocalStorage = module.AsyncLocalStorage;
-    asyncStore = new AsyncLocalStorage<RequestEventInternal>();
-    globalThis.qcAsyncRequestStore = asyncStore;
+    const asyncStore = new AsyncLocalStorage<RequestEventInternal>();
+    setAsyncRequestStore(asyncStore);
   })
   .catch((err) => {
     console.warn(
```
wmertens: Why don't you initialize this in async-request-store.ts?

packages/qwik-router/src/runtime/src/server-functions.ts
```diff
@@ -420,7 +421,7 @@ export const serverQrl = <T extends ServerFunction>(
 
       if (isServer) {
         // Running during SSR, we can call the function directly
-        let requestEvent = globalThis.qcAsyncRequestStore?.getStore() as RequestEvent | undefined;
+        let requestEvent = asyncRequestStore?.getStore() as RequestEvent | undefined;
```
wmertens: you say it has to be a global for server-functions.ts, but it's not using the global?

packages/qwik/src/core/use/use-locale.ts
```diff
@@ -48,5 +74,9 @@ export function withLocale<T>(locale: string, fn: () => T): T {
  * @public
  */
 export function setLocale(locale: string): void {
+  if (isServer) {
+    // ignore on server; use withLocale() within a request scope instead
+    return;
+  }
```
wmertens: don't do this, there might still be valid use cases

packages/qwik-router/src/middleware/request-handler/async-request-store.ts
```diff
@@ -0,0 +1,24 @@
+import type { RequestEventInternal } from './request-event';
+import type { AsyncLocalStorage } from 'node:async_hooks';
+
+export type AsyncStore = AsyncLocalStorage<RequestEventInternal>;
+
+// Qwik Core will also be using the async store if this is present
+export const asyncRequestStore: AsyncStore | undefined = globalThis.qcAsyncRequestStore;
```
wmertens: why the global? Where do we need it?

packages/qwik-router/src/middleware/request-handler/async-request-store.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const setAsyncRequestStore = (store: AsyncStore | undefined) => {
+  globalThis.qcAsyncRequestStore = store;
```
wmertens: this won't change the exported value

wmertens (changes requested):
The fact that the current code will not update the async store tells me that we need a test for this. We should have an e2e test that requests the same route twice but with different locales, and with a promise in the middle of the page that gets resolved once both requests are being rendered.

packages/qwik-router/src/middleware/request-handler/async-request-store.ts
```diff
@@ -0,0 +1,19 @@
...
+  .then((module) => {
+    const AsyncLocalStorage = module.AsyncLocalStorage;
+    asyncRequestStore = new AsyncLocalStorage<RequestEventInternal>();
+  })
+  .catch((err) => {
+    console.warn(
+      'AsyncLocalStorage not available, continuing without it. This might impact concurrent server calls.',
+      err
+    );
+  });
```
wmertens: Let's put this inside `if (isServer)`, because this will always fail in the browser and will print a console message even when caught.

JerryWu1234:
@wmertens take a look

gioboa (commented):
Thanks for your commitment ☺️

JerryWu1234:
I have tried a few methods to fix it. doesn't work.
here need @wmertens to take a look.
thanks  it's only for documentation purpose to facilitate tracking

wmertens:
Ok so we had duplication in the router code and that's why the store was undefined. Now we have a way better router build script and everything is working great.

PR #7828, opened 2025-08-15
fix(repl): prismjs

wmertens:
:grimacing:


gioboa (approved):
#7827 is doing the same thing 🤔

wmertens:
@gioboa we both had the same idea at the same time but it was wrong :smile:
I force pushed the correct fix

gioboa (approved):

maiieul (approved):
LGTM

PR #7841, opened 2025-08-19
chore: rename vite.config from `.ts` to `.mts` 🍒

gioboa:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


gioboa (commented):
Vite was complaining that the commonjs build is deprecated and we can't make qwik be ESM only so with mts vite runs as ESM.
but we can do use `type: module` and still support cjs via exports so I will close this PR.

PR #7855, opened 2025-08-23
chore(starters): add sideEffects: false to the lib template 😎

### LINKED ISSUE #7854: library starters package.json should include `"sideEffects": false` field
Otherwise there might be some side effects imports when building a consumer project, which will be added as static imports to the bundle graph, leading to slight over-prefetching of unnecessary bundles for user events.

gioboa:
Close #7854

Otherwise there might be some side effects imports when building a consumer project, which will be added as static imports to the bundle graph, leading to slight over-prefetching of unnecessary bundles for user events.

### DIFF
--- a/starters/apps/library/README.md
+++ b/starters/apps/library/README.md
@@ -48,9 +48,8 @@ npm run build
 
 ## sideEffects: false
 
-This package is configured with "sideEffects": false in its package.json.
-[Full docs here](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)
-This tells bundlers that the module has no side effects when imported.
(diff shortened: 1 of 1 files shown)


wmertens:
Yes great, but the README should include a paraghraph saying that the side-effects are set to false, and that if there are side effects (like a global being set), then it should be removed, possibly only for the file that has it.

That field isn't a part of the npm spec, but it is used by bundlers, and the best documentation is at https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

gioboa:
> Yes great, but the README should include a paraghraph saying that the side-effects are set to false...

[Added](https://github.com/QwikDev/qwik/pull/7855/files#diff-cb63611e80db7618dcb06a502a4300e4bdfd935085c7929875c7572721282bf9R49)

PR #7858, opened 2025-08-23
fix(starter): fix up vscode settings merge 🐼

### LINKED ISSUE #5280: [✨]

Tailwind CSS Integration Overwrites .vscode/settings.json

Preserve Emmet Configuration for .tsx Files and Maintain 'class' and 'for' Attributes Instead of 'className' and 'htmlFor'

Instead of overwriting the entire .vscode/settings.json file, I recommend adding the following configuration at the end of the file:

`"css.lint.unknownAtRules": "ignore"`

By doing this, VS Code can continue to interpret .tsx files as Qwik files, and Emmet functionality will remain unaffected, allowing for a seamless integration of Tailwind CSS without compromising Qwik file interpretation."

### Additional context

_No response_

gioboa:
Close #5280 

base/.vscode/settings.json is a JSON5 file, we need to use json5 lib to parse it.
Right now we have an error because in the Qwik CLI, we're trying to [parse that file](https://github.com/QwikDev/qwik/blob/05d9c8517efc11034dc1b2d0546969f94e3adb65/packages/qwik/src/cli/add/update-files.ts#L117) with JSON.parse; this will fallback to the catch branch and overwrite the entire file.


gioboa:
> Actually, the bug is on our side. The settings file is [JSON5](https://json5.org/), meaning it supports comments, and we should parse it as such. We can use https://www.npmjs.com/package/json5-writer to handle json files in a way that preserves comments.

Fixed

gioboa:
JSON5.parse is working fine but it's stripping the comments :thinking:

gioboa (commented):
JSON5.parse is working fine but it's stripping the comments 🤔
On the other hand it is solving the main issue

gioboa (commented):
with @croct/json5-parser is working better and it's keeping the comments too.

wmertens (dismissed):
LGTM

PR #7863, opened 2025-08-25
test: improve mergeIntegrationDir tests 🍰

gioboa:


packages/qwik/src/cli/add/update-files.unit.ts
```diff
@@ -114,10 +114,18 @@ describe('mergeIntegrationDir', () => {
...
-    )?.content;
-
-    expect(actualGlobalCssContent).toBe('p{color: red}\n\n/* CSS */\n');
+    const tests = {
+      'destDir/subDestDir/apps/subpackage/fake.ts': 'fake file',
+      'destDir/subDestDir/should-stay-in-root.ts': 'fake file',
+      'destDir/subDestDir/package.json': '{"name": "fake"}',
+      'destDir/subDestDir/should-stay/should-also-stay.ts': 'fake file',
+      'destDir/subDestDir/apps/subpackage/.vscode/settings.json':
+        '{ \n          // Comment Foo\n          "css.lint.unknownAtRules": "ignore",\n          "name": "John Doe",\n          "age": 42\n      }\n',
```
wmertens, line 123: Looks like `// Comment` is removed :(

gioboa: yep, it's true. for the second object is not preserving the comments.
That's a limit on the library, but at least we can keep the comment in the first one.

Issue #7542, opened 2025-04-26
[🐞] Qwik React never calls unmount.


omerman:
### Which component is affected?

Qwik React

### Describe the bug

I am trying to migrate from react. but I can see that unmount is never called for any of the components I mount.

I went over the source code,
the `root.unmount()` is never called.
I guess there should be a cleanup param usage via the useTask.
I dont have enough knowledge on react's hydration from ssr to make the change myself, because I'n not sure which edge cases I should go over.

Help please?

P.S thank you so much, qwik is amazing 🙏 

```
//// This is a code snippet from `qwik-react` to show what I believe to be the relevant code
//// See how the `root.unmount()` invocation is missing

    // Task takes cares of updates and partial hydration
    useTask$(async ({ track }) => {
      const trackedProps = track(() => ({ ...props }));
      track(signal);

      if (!isBrowser) {
        return;
      }

      // Update
      if (internalState.value) {
        if (internalState.value.root) {
          internalState.value.root.render(
            main(slotRef.value, scopeId, internalState.value.cmp, trackedProps)
          );
        }
      } else {
        let root: Root | undefined = undefined;
        const Cmp = await reactCmp$.resolve();
        const hostElement = hostRef.value;
        if (hostElement) {
          // hydration
          if (isClientOnly) {
            root = client.createRoot(hostElement);
          } else {
            root = client.flushSync(() => {
              return client.hydrateRoot(
                hostElement,
                mainExactProps(slotRef.value, scopeId, Cmp, hydrationKeys)
              );
            });
          }
          if (isClientOnly || signal.value === false) {
            root.render(main(slotRef.value, scopeId, Cmp, trackedProps));
          }
        }
        internalState.value = noSerialize({
          cmp: Cmp,
          root,
        });
      }
    });
.....
.....
```

### Reproduction

https://stackblitz.com/edit/github-ago3ix4p?file=src%2FReactComponent.tsx,package.json

### Steps to reproduce

Run the stackblitz, the unmount console log is never called, even after the signal's value changes and we no longer render the react component.

### System Info

```shell
System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 74.13 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
[...]

PR #7864, opened 2025-08-25
fix: unmount react root alongside with qwik

sashkashishka:
# What is it?

Bug

# Description

There is a bug, while navigating using SPA like approach, the qwikified react components don't call unmount callbacks specified in `useEffect`s hooks.

There is an open issue that mentions such a problem https://github.com/QwikDev/qwik/issues/7542

In this PR the issue is fixed by calling unmount method in react root entity when the qwik wrapper unmounts.

I haven't added unit tests because I haven't found them. If they are needed, please let me know where to write them.

# Checklist

### DIFF
--- a/.changeset/short-jeans-speak.md
+++ b/.changeset/short-jeans-speak.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+unmount qwikify react root alongside with qwik component
--- a/packages/qwik-react/src/react/qwikify.tsx
+++ b/packages/qwik-react/src/react/qwikify.tsx
@@ -78,6 +78,16 @@ export function qwikifyQrl<PROPS extends Record<any, any>>(
       }
     });
 
+    useTask$(({ track, cleanup }) => {
+      track(signal);
+
+      if (isBrowser) {
+        cleanup(() => {
+          internalState.value?.root?.unmount();
+        });
+      }
+    });
+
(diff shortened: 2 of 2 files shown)


.changeset/short-jeans-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+unmount qwikify react root alongside with qwik component
```
gioboa:
```suggestion
FIX: unmount qwikify react root alongside with qwik component
```

sashkashishka:
@gioboa, sure, I'll make it

wmertens (approved):
LGTM

wmertens:
I would love some tests, but you can add them in a separate PR @sashkashishka.

sashkashishka:
@gioboa, @wmertens, thank you all for approving!
Sorry for hesitating with tests. I can add them within 2 weeks

.changeset/short-jeans-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik': patch
```
shairez, line 2: @gioboa we need to pay attention that this is the right package name.

If the change was in the react package but the changeset was set as the core package, it won't get a release with these changes during the release process

gioboa: I see, thanks for the heads up

PR #7865, opened 2025-08-25
fix: compare URLs without protocols with checkOrigin: lax-proto 

asaharan:
Previously, two CSRF middlewares were added for lax-proto requests: one at the beginning and one at the end. This change replaces them with a single middleware placed at the beginning. Non-lax-proto cases remain unchanged.

# What is it?

- Bug

# Description

fix behaviour of checkOrigin: "lax-proto" in createQwikCity

# Checklist

### DIFF
--- a/.changeset/curvy-glasses-wash.md
+++ b/.changeset/curvy-glasses-wash.md
@@ -2,4 +2,4 @@
 '@builder.io/qwik-city': patch
 ---
 
-fix behaviour of checkOrigin: "lax-proto" in createQwikCity
+FIX: fix behaviour of checkOrigin: "lax-proto" in createQwikCity
--- a/.changeset/curvy-glasses-wash.md
+++ b/.changeset/curvy-glasses-wash.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik-city': patch
+---
+
+fix behaviour of checkOrigin: "lax-proto" in createQwikCity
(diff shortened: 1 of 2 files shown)


gioboa (commented):
Thanks for your help @asaharan 
Is there a specific issue with the actual code?
why are we changing it?

asaharan:
Yes @gioboa , even when I set checkOrigin to lax-proto, I get CSRF error.
ORIGIN=https://saharan.dev
Request is coming from https://saharan.dev
but there is a load balancer(say AWS ALB) in between, so it forwards `x-forwarded-proto: https` but still I get the CSRF error saying saying request domain `https://saharan.dev` doesn't match origin `http://saharan.dev`

`lax-proto` is supposed to handle this case, hence this change.

asaharan:
LGTM. The change in `resolve-request-handlers.ts` correctly selects the CSRF middleware. The change to `qwik-binding-map.ts` appears to be a result of a build script and is not a functional change.

wmertens (dismissed):
LGTM 🌉

asaharan:
@gioboa as csrf is being checked at the very beginning, origin contains http and not https. So, I have compare origin with inputOrigin after removing the protocol(http/https).
Have tested this on my production environment by directly modifying files inside node_modules and it's working fine over there.

.changeset/curvy-glasses-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@builder.io/qwik-city': patch
+---
+
+fix behaviour of checkOrigin: "lax-proto" in createQwikCity
```
gioboa:
```suggestion
FIX: fix behaviour of checkOrigin: "lax-proto" in createQwikCity
```

gioboa (approved):
Thanks for your help @asaharan

PR #7866, opened 2025-08-25
feat: initial backpatching support

thejackshelton:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


packages/qwik/src/core/use/use-core.ts
```diff
@@ -229,6 +229,16 @@ export const trackSignalAndAssignHost = (
...
   }
+
+  if (
+    typeof property === 'string' &&
+    property !== 'vnode' &&
+    property !== '.' &&
+    (container as any).$addSignalAttrMapping$
+  ) {
+    (container as any).$addSignalAttrMapping$(value, host, property, data);
+  }
```
Varixo, line 240: this is true only for ssr? if so we should add isServer check

packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -342,6 +342,50 @@ export const createScheduler = (
...
+              value,
+              subscriptionData?.$scopedStyleIdPrefix$
+            );
+            if (serialized === false) {
+              serialized = null;
+            }
+
+            const vNodeId = (hostNode as ISsrNode).id;
+
+            (container as any).$addBackpatch$?.({
```
Varixo, line 380: probably this is a draft, but we should eliminate `as any` casts

packages/qwik/src/core/use/use-core.ts
Varixo: Also the property thing - we have enums for that :D

packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -342,6 +342,50 @@ export const createScheduler = (
...
+              type: 'attribute',
+              vNodeId,
+              name: attrName,
+              serializedValue: serialized as string | true | null,
+            });
+          }
```
Varixo, line 386: maybe this should be a separate chore and just schedule it as backpatch chore?
or even better, dont do this here, but inside trigger effects. Just pass additional data to the RECOMPUTE_AND_SCHEDULE_EFFECTS chore and then to trigger effects. Then based on that do the logic?

packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -342,6 +342,50 @@ export const createScheduler = (
       const isUpdatable = !!(chore.$host$.flags & SsrNodeFlags.Updatable);
 
       if (!isUpdatable) {
+        if (chore.$type$ === ChoreType.RECOMPUTE_AND_SCHEDULE_EFFECTS) {
+          const effects = chore.$payload$ as Set<any>;
+
+          for (const effect of effects) {
+            if (!Array.isArray(effect) || effect.length < 4) {
```
Varixo, line 349: why `< 4`?

packages/qwik/src/core/shared/scheduler.ts
Varixo: basically the problem I see with this approach is we iterate over effects two times. If we move logic to trigger effects we would do it once

packages/qwik/src/core/shared/scheduler.ts
thejackshelton: I was hacking around with the scheduler and a consumer project. That should probably be removed 😅

packages/qwik/src/core/shared/scheduler.ts
thejackshelton: yep draft. planned on properly typing it once we have a solid approach to the stale state and we accept the tradeoffs

packages/qwik/src/core/use/use-core.ts
thejackshelton: Which case on the enum? I can't seem to find it, and yes only true for SSR I think I ran into an edge case at the time and was hacking. Should add isServer there 👍

packages/qwik/src/core/shared/scheduler.ts
thejackshelton: > maybe this should be a separate chore and just schedule it as backpatch chore? or even better, dont do this here, but inside trigger effects. Just pass additional data to the RECOMPUTE_AND_SCHEDULE_EFFECTS chore and then to trigger effects. Then based on that do the logic?

that is where I originally wanted to do it as well in DM so makes sense to me. would also explain why the data is in the script twice 😅

thejackshelton:
Closing in favor of a new PR

PR #7874, opened 2025-08-27
[Docs] Fix StackBlitz Qwik example app

loganpowell:
I provided a working package.json. I also had to remove the pnpm-lock file.

# What is it?
- Docs

# Description

The stackblitz example (very first example in the docs) is broken.


wmertens:
thank you but this is not the way to do this, we must update the source of qwik.new instead (I don't know where that is)

gioboa (commented):
Thanks for your commitment @loganpowell 
Looking forward to seeing other PRs from you if you like. 👏

PR #7876, opened 2025-08-27
feat: expose loading and error fields of async computed signal

Varixo:
Expose (hidden until now) `loading` and `error` fields of async computed signal

### DIFF
--- a/.changeset/mean-tires-cover.md
+++ b/.changeset/mean-tires-cover.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': minor
+---
+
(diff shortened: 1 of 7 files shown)


gioboa (approved):

wmertens (commented):
I wonder if we really need an error field, let alone a signal for it? Shouldn't errors be never happening?

Varixo:
> I wonder if we really need an error field, let alone a signal for it? Shouldn't errors be never happening?

if you call fetch inside async computed and fetch fails, then I think we should handle this somehow

wmertens (approved):
LGTM

PR #7879, opened 2025-08-28
chore: update rust version

Varixo:


packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_invalid_segment_expr1.snap
```diff
@@ -59,6 +58,8 @@ Some("{\"version\":3,\"sources\":[\"/user/qwik/src/test.tsx\"],\"names\":[],\"ma
 */
 ============================= test.tsx_App_component_ckEPmXZlub0.js (ENTRY POINT)==
 
+import css1 from "./global.css";
```
wmertens, line 61: Surprising but correct? Since it generates errors for the code (intentionally I hope), I suppose it's fine.

wmertens (approved):
LGTM 🍌

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_invalid_segment_expr1.snap
Varixo: yes I think so

PR #7881, opened 2025-08-28
feat: add resolve method for async computed

Varixo:
Async `resolve` function has been added. It should be used to avoid reruning code before `asynccomputed.value` during resolving a promise

### DIFF
--- a/packages/qwik/src/core/qwik.core.api.md
+++ b/packages/qwik/src/core/qwik.core.api.md
@@ -25,7 +25,7 @@ export interface AsyncComputedReadonlySignal<T = unknown> extends ComputedSignal
     error: Error | null;
     loading: boolean;
     // (undocumented)
-    resolve(): Promise<void>;
+    resolve(): Promise<T>;
 }
 
 // @public (undocumented)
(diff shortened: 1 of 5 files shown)


Varixo:
This PR is to discuss as it adds new public API for async computed

packages/qwik/src/core/qwik.core.api.md
```diff
@@ -24,6 +24,8 @@ export type AsyncComputedFn<T> = (ctx: AsyncComputedCtx) => Promise<T>;
 export interface AsyncComputedReadonlySignal<T = unknown> extends ComputedSignal<T> {
     error: Error | null;
     loading: boolean;
+    // (undocumented)
+    resolve(): Promise<void>;
```
wmertens: Let's make it return the value

wmertens:
@Varixo should we also add `.resolved` like in QRLs to get the latest value? An alias for `untrackedValue` basically.

packages/qwik/src/core/qwik.core.api.md
gioboa: Is this still valid?

PR #7886, opened 2025-08-30
fix: finding parent dom element from projected content

Varixo:
Find the correct parent dom element from projected content


wmertens (approved):
LGTM 🦺

packages/qwik/src/core/client/vnode.ts
```diff
@@ -1148,14 +1149,24 @@ export const vnode_insertBefore = (
   }
 };
 
-export const vnode_getDomParent = (vnode: VNode): Element | Text | null => {
-  vnode = vnode_getDomParentVNode(vnode) as VNode;
+export const vnode_getDomParent = (
+  vnode: VNode,
+  includeProjection = true
```
gioboa, line 1154: Nice 💪

PR #7890, opened 2025-08-31
feat(router): make dev more like prod

wmertens:
This brings our Vite plugins more inline with dev mode in the regular Vite ecosystem:

- qwikVite dev mode only does index.html transforms
- qwikRouter dev mode runs the renderer with the node middleware and patches the SSR stream for dev mode while streaming
- fix router build so it doesn't import core
- dev can now hot reload routes
- q-manifest is now read only if and when needed
- ...


packages/qwik-router/src/buildtime/runtime-generation/generate-service-worker.ts
maiieul, line 1: I also changed this in v1. You'll have to merge the conflicts when merging the two branches again.

packages/qwik-router/src/buildtime/vite/dev-middleware.ts
```diff
@@ -0,0 +1,136 @@
...
+    // serve a placeholder service worker, because we can't provide CJS bundles in dev mode
+    // once we support ESM service workers, we can remove this
+    if (req.url === `${server.config.base}service-worker.js`) {
+      // For service worker, we don't need HTML transformation, so use original response
+      res.setHeader('Content-Type', 'text/javascript');
+      res.end(
+        `/* Qwik Router Dev Service Worker */` +
+          `self.addEventListener('install', () => self.skipWaiting());` +
+          `self.addEventListener('activate', (ev) => ev.waitUntil(self.clients.claim()));`
+      );
```
maiieul, line 52: why do we need a service worker here? I thought it was disabled in dev mode and we use modulepreload now

packages/qwik-router/src/buildtime/vite/dev-server.ts
maiieul, line 1: Should we split this file into dev-middleware and html-transform-wrapper without changing the code in another PR first? This way we can track the changes more easily for review but also for potential future debugging.

starters/features/csr/index.html
```diff
@@ -0,0 +1,23 @@
...
+  </head>
+  <body>
+    <div id="root"></div>
+
+    <script type="module">
+      import { render, jsx } from "@qwik.dev/core";
+      import Root from "./src/root";
+
+      /** This renders your application */
+      render(document.querySelector("#root"), jsx(Root));
```
maiieul, line 20: I love this 😍

packages/qwik-router/src/buildtime/vite/dev-middleware.ts
wmertens: this is generic code, for all supported paths that Router builds

wmertens:
for some reason, dev mode fails on the docs docs/ pages, it can't resume properly due to vnode mismatch

gioboa (commented):
I tested the docs with dev env and it's working fine.

wmertens:
@gioboa
I did `pnpm build.full` to be sure, but when I do `pnpm docs.dev` and then visit the playground, I get

```
core.prod.mjs:4183 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at vnode_isElementVNode (core.prod.mjs:4183:24)
    at ensureElementVNode (core.prod.mjs:4218:49)
    at vnode_getVNodeForChildNode (core.prod.mjs:4469:12)
    at vnode_locate (core.prod.mjs:4425:21)
    at newInvokeContextFromTuple (core.prod.mjs:518:25)
    at tryGetInvokeContext (core.prod.mjs:468:60)
    at qrl.resolve (core.prod.mjs:6422:21)
    at inflate (core.prod.mjs:5294:35)
    at DeserializationHandler.get (core.prod.mjs:5164:45)
    at getObjectById (core.prod.mjs:6087:14)
```

Which means that the DOM doesn't look like what Core expected, which must mean that element that have a `:` attribute have been added or removed.

gioboa:
I see, let me double check it

gioboa (approved):
🔥

PR #7900, opened 2025-09-01
feat: v2 backpatch support

thejackshelton:
# What is it?

- Feature / enhancement
- Bug

# Description

This PR implements automatic SSR backpatching, enabling SSR-correct relationships between components whose render order can't be guaranteed. Unlike traditional SSR where once HTML is sent you can't modify it, backpatching automatically corrects attribute relationships after streaming completes.

# Checklist


packages/qwik/src/core/qwik.core.api.md
```diff
@@ -1940,6 +1940,7 @@ export const _waitUntilRendered: (elm: Element) => Promise<void>;
 export function _walkJSX(ssr: SSRContainer, value: JSXOutput, options: {
     currentStyleScoped: string | null;
     parentComponentFrame: ISsrComponentFrame | null;
+    isBackpatching: boolean;
```
wmertens: Looks like this isn't actually used anywhere, can you remove it?

packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -328,28 +325,21 @@ export const createScheduler = (
...
-          chore.$type$
-        )}' chore was scheduled on a host element that has already been streamed to the client.
+        if (
+          chore.$type$ !== ChoreType.NODE_PROP &&
+          chore.$type$ !== ChoreType.RECOMPUTE_AND_SCHEDULE_EFFECTS
+        ) {
+          // We are running on the server.
+          // On server we can't schedule task for a different host!
+          // Server is SSR, and therefore scheduling for anything but the current host
+          // implies that things need to be re-run nad that is not supported because of streaming.
```
wmertens: nad => and

scripts/submodule-backpatch.ts
```diff
@@ -0,0 +1,98 @@
...
+    `exports.QWIK_BACKPATCH_EXECUTOR_DEBUG = QWIK_BACKPATCH_EXECUTOR_DEBUG;`,
+  ];
+  const dtsCode = [
+    `export declare const QWIK_BACKPATCH_EXECUTOR_MINIFIED: string;`,
+    `export declare const QWIK_BACKPATCH_EXECUTOR_DEBUG: string;`,
+  ];
+
+  ensureDir(backpatchDistDir);
+  await writeFile(join(backpatchDistDir, 'index.mjs'), esmCode.join('\n') + '\n');
+  await writeFile(join(backpatchDistDir, 'index.cjs'), cjsCode.join('\n') + '\n');
```
wmertens, line 86: since the backpatcher isn't importing or exporting anything, perhaps only have one cjs script? That way we can still provide SSR to browsers that don't support modules.

wmertens (dismissed):
This is excellent! Some small nitpicks.

Also, this looks like a good start for OOS, but at that point there might be a noticeable delay between DOM content visible on screen and end of the document arriving in the browser.

How should the backpatcher be changed to allow patching while streaming? Maybe emit the backpatcher as soon as it's needed and then emit script tags with patches every 50ms or so?

That is assuming you can change DOM while it's still streaming in :thinking:

packages/qwik/src/core/qwik.core.api.md
thejackshelton: It is being used. We pass it into `processJsxNode`

packages/qwik/src/core/shared/scheduler.ts
thejackshelton: fixed

scripts/submodule-backpatch.ts
thejackshelton: Why not support both? Qwikloader does the same

thejackshelton:
> This is excellent! Some small nitpicks.
> 
> Also, this looks like a good start for OOS, but at that point there might be a noticeable delay between DOM content visible on screen and end of the document arriving in the browser.
> 
> How should the backpatcher be changed to allow patching while streaming? Maybe emit the backpatcher as soon as it's needed and then emit script tags with patches every 50ms or so?
> 
> That is assuming you can change DOM while it's still streaming in 🤔

Yes there would definitely be a noticeable delay in that case. There is only one data script per container rather than mini scoped ones. This is because the executor script runs at the end anyways and you can improve the encoding of the data script a ton, especially with multiple component instances.

I definitely see it as a separate thing. We should be able to see from the scheduler if the consumer clearly intends to be doing OOO Streaming, in that case we can then emit the script tags / wrapper approach as we go.

In Marko 6 they opt-in by providing fallback UI.

scripts/submodule-backpatch.ts
wmertens: the backpatch only runs embedded in the browser, so it's never imported, no?

packages/qwik/src/core/qwik.core.api.md
wmertens: yes but that doesn't use it either

scripts/submodule-backpatch.ts
wmertens: oh I was confused, this is the import that provides the strings to embed. Ok.

packages/qwik/src/core/qwik.core.api.md
thejackshelton: good catch. removed 🙂

.changeset/honest-berries-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+feat: add SSR backpatching (attributes-only) to ensure SSR/CSR parity for signal-driven attributes; limited to attribute updates (not OoO streaming)
```
maiieul, line 5: Can you explain the use-case a little bit better for qwik users? e.g. when is it necessary, how it works, what needs to be done to use it (as an overview ofc)

packages/docs/src/routes/docs/(qwik)/advanced/backpatching/index.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+Backpatching solves a fundamental difference between client and server rendering:
+
+**Client rendering**: Components can render in any order, then establish relationships between each other afterward.
+
+**SSR streaming**: Once HTML is sent to the browser, it's immutable—you can only stream more content forward.
+
+This creates problems for component libraries where elements need to reference each other (like form inputs linking to their labels via `aria-labelledby`). If the input streams before its label, it can't know the label's ID to set the relationship.
+
+Backpatching automatically fixes these relationships by updating attributes after the entire page has streamed, giving you the same flexibility as client-side rendering.
```
maiieul, line 25: Maybe add a sentence to say something along the lines of "as a developer, you don't need to do anything special. Qwik will automatically detect if it needs to backpatch a certain attribute when necessary".

packages/docs/src/routes/docs/(qwik)/advanced/backpatching/index.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+> Note: This is not Out-of-Order Streaming. It only corrects already-sent attributes without delaying the stream.
+
+### Example
```
maiieul, line 29: Is this "Example without backpatching"?

thejackshelton: no

packages/docs/src/routes/docs/(qwik)/advanced/backpatching/index.mdx
thejackshelton: yeah I can add that in

.changeset/honest-berries-knock.md
thejackshelton: hm... I could try from a high level, although that is a lot to put in a changelog

maiieul: Suggestion:

feat: add SSR backpatching (attributes-only) to allow parent components to know some information about their children during SSR: like how many children they have, or if a certain child contains an attribute or not... This was already achievable in CSR, and is now possible in SSR too.

packages/docs/src/routes/docs/(qwik)/advanced/backpatching/index.mdx
maiieul: Then I don't get it. It seemed to me the example shows the problem before the PR. After the PR it will just work right? Better to make that more explicit imo.

gioboa (approved):

PR #7903, opened 2025-09-02
feat(core): idle-visible for useVisibleTask$

wmertens:
not sure if this is a good idea


devcaeg:
Would this have the same effect as this current code?

```ts
useOnDocument(
	"qidle",
	$(() => {
		// Code
	}),
);
```

wmertens:
@devcaeg no, it's a combo of visible and idle, meaning it waits until the element is visible after the page went idle.

I'm thinking that this is better for the core web vitals, since it will run less code before idle.

I'm thinking this is generally what people want when running useVisibleTask$, except when they want to animate something as soon as it's visible in the first screenful.

devcaeg:
Ah, I see, wait until it is visible once the website is idle.

I think that's an excellent idea, the real question is whether this should be the new default value or not.

maiieul:
Yeah I'm not convinced tbh. 

First I doubt this improves LCP at all. From what I understand, LCP is usually delayed by network speed, not CPU cluttering. If your LCP is delayed because of your CPU I doubt you can do much on your phone 😄. In a lot of cases, core+qwik-city finish preloading after LCP, but if you have a big fat image, they could finish preloading before, then execute your visible task once it is preloaded, and then maybe only clutter the network a little more through a fetch request that clutters the network a tiny little more? The execution of the visible task is unlikely slowing down LCP even in this case.

Second it's quite confusing and hard to explain to newcomers what is "document-idle", vs "idle-visible". It's a least one more concept to explain. I wouldn't introduce anything new unless it solves a well-measured problem.

wmertens:
yeah makes sense. I'll leave the branch up in case we find a use case :)

PR #7912, opened 2025-09-04
fix: fix up no-empty-object-type error 🦄

gioboa:
I started a new project, added the Vercel adapter and I had this issue. 
This PR will remove empty object to prevent the eslint error.

image

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/.changeset/bitter-loops-train.md
+++ b/.changeset/bitter-loops-train.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+FIX: The Deno integration now builds successfully with version v2.4.3 and above.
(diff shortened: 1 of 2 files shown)


e2e/adapters-e2e/src/entry.express.tsx
```diff
@@ -7,7 +7,7 @@
  * - https://qwik.dev/docs/deployments/node/
  *
  */
-import { createQwikCity, type PlatformNode } from '@builder.io/qwik-city/middleware/node';
```
gioboa: I'm wondering if we can remove all these types too.

maiieul: I think if you remove the the type then it will complain in consumer projects. I tried removing it recently during an update. But you can use a `type QwikCityPlatform = ...` to fix it.

maiieul (changes requested):

PR #7916, opened 2025-09-05
refactor(playground)

wmertens:
This is a big refactor of the playground:

- the service worker is only used for proxying, so it never needs to change
- instead each qwik version gets its own bundling webworker
- this means we can use ESM for everything
- the code is also way easier to debug

Now, we are only a small step away from:
- using the repl for every example in the docs!
- running qwik-city / qwik-router in the playground!

You can now also go to `/repl/[replId]/` in a separate tab :)


gioboa (approved):

gioboa (commented):
We are reintroducing prismjs, is it better that shiki?

--- UPDATE ---

I saw your comment about bad performance of shiki 👍

PR #7785, opened 2025-08-06
fix: Alt + click not working on windows

LazyClicks:
# What is it?

- Feature / enhancement
- Bug
- Docs / tests / types / typos
- Infra

# Description

# Checklist

- [ ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [ ] I performed a self-review of my own code
- [ ] I added a changeset with `pnpm change`
- [ ] I made corresponding changes to the Qwik docs
- [ ] I added new tests to cover the fix / functionality

PR #7921, opened 2025-09-07
Fix Alt+Click "open in editor" whe a  custom `srcDir` is defined.

jails:
# What is it?

- Bug

# Description

Problem with the Alt+Click "open in editor" feature when using custom `srcDir` option. see https://github.com/QwikDev/qwik/pull/7785#issuecomment-3239670388 for more detail related to the issue.

# Checklist

### DIFF
--- a/packages/qwik/src/optimizer/src/plugins/click-to-component.html
+++ b/packages/qwik/src/optimizer/src/plugins/click-to-component.html
@@ -60,7 +60,7 @@
   (function () {
     const inspectAttribute = 'data-qwik-inspector';
     const hotKeys = globalThis.qwikdevtools.hotKeys;
-    const srcDir = globalThis.qwikdevtools.srcDir;
+    const rootDir = globalThis.qwikdevtools.rootDir;
     document.querySelector('#qwik-inspector-info-popup').textContent =
       `Click-to-Source: ${hotKeys.join(' + ')}`;
     console.debug(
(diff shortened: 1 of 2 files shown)


gioboa (commented):
It looks amazing to me.
We are simply a lot the code. Thanks.
I will test the package generated from the PR on Linux and Mac too.

wmertens:
I think #7923 is a better approach, do you agree?

gioboa:
> I think #7923 is a better approach, do you agree?

This PR is good because it's solving the specific problem, on the other hand your PR @wmertens is rewriting the logic of this functionality and it's a big improvement too.

wmertens:
I found a way to make the other PR not have to patch launch-editor, so now I think it's a no-brainer to use the other PR instead of this one.

gioboa:
Thanks for your amazing help @jails 
We decided to solve the issue by changing the implementation and using the same mechanism Vite dev mode already uses.
So that's why we are closing this PR in favor of #7923.
Can you help us and test if it's solving the issue with a custom srcDir please? 🙏 
Thanks.

gioboa (commented):
Thanks for your amazing help @jails
We decided to solve the issue by changing the implementation and using the same mechanism Vite dev mode already uses.
So that's why we are closing this PR in favor of #7923.
Can you help us and test if it's solving the issue with a custom srcDir please? :pray:
Thanks.

PR #7922, opened 2025-09-08
fix(repl): small things

wmertens:
- url wasn't updating
- bundle worker not starting wasn't noticed


gioboa (approved):
LGTM 👏
Thanks for your commitment 💯

PR #7923, opened 2025-09-08
fix(click to component): use vite to resolve path

wmertens:
This way, there's never a problem with windows or whatever, we just use the same mechanism Vite dev mode already uses.

### DIFF
--- a/packages/insights/package.json
+++ b/packages/insights/package.json
@@ -70,6 +70,6 @@
     "test.unit": "vitest",
     "test.watch": "vitest --watch"
   },
-  "type": "module",
-  "sideEffects": false
+  "sideEffects": false,
+  "type": "module"
 }
--- a/packages/qwik-city/package.json
+++ b/packages/qwik-city/package.json
@@ -187,7 +187,7 @@
   "scripts": {
     "build": "cd src/runtime && vite build --mode lib"
   },
+  "sideEffects": false,
   "type": "module",
-  "types": "./lib/index.d.ts",
-  "sideEffects": false
+  "types": "./lib/index.d.ts"
 }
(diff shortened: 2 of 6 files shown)


packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -445,6 +445,36 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
...
+              const launchEditor = (await import('launch-editor')).default;
+              launchEditor(path);
+            } catch (e: any) {
+              console.error(`Failed to open editor: ${e.message}`);
+            }
+          }
+        }
+      }
+      return { id: `\0editor` };
+      // Some dummy content
```
gioboa: let's remove this :)

packages/qwik/src/optimizer/src/plugins/plugin.ts
```diff
@@ -577,6 +607,10 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
     id: string,
     loadOpts?: Parameters<Extract<Plugin['load'], Function>>[1]
   ): Promise<Rollup.LoadResult> => {
+    if (id === '\0editor') {
+      // Some dummy content
```
gioboa: and this too :)

pnpm-workspace.yaml
```diff
@@ -1,28 +1,28 @@
...
+  typescript: 5.4.5
+  vfile: 6.0.3
 
 patchedDependencies:
-  'density-clustering@1.3.0': 'patches/density-clustering@1.3.0.patch'
-  'domino@2.1.6': 'patches/domino@2.1.6.patch'
+  density-clustering@1.3.0: patches/density-clustering@1.3.0.patch
+  domino@2.1.6: patches/domino@2.1.6.patch
 
-engineStrict: true
```
gioboa, line 26: why we are removing these property?

wmertens: it's moved automatically by pnpm itself

PR #7935, opened 2025-09-10
chore(repl): use rolldown for bundling

wmertens:
This is 4x faster than rollup :exploding_head:

we need to check in various browsers if everything still works, right now a problem is that insights doesn't work any more because it needs CORS headers from the insights server, and CORS needs to be turned on for SharedArrayBuffer to work.


gioboa (approved):
🔥👏@wmertens

PR #7937, opened 2025-09-10
fix: resuming nested container in shadow root

Varixo:
Pass first child of shadow root's as container. In other case the tree walker returns the same container element for nextNode function.

### DIFF
--- a/packages/docs/src/routes/playground/parser/html/index.tsx
+++ b/packages/docs/src/routes/playground/parser/html/index.tsx
@@ -46,7 +46,7 @@ export default component$(() => {
 
       // Try to get Qwik container if available
       try {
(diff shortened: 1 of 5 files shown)


packages/docs/src/routes/playground/parser/html/index.tsx
```diff
@@ -46,7 +46,7 @@ export default component$(() => {
 
       // Try to get Qwik container if available
       try {
-        const container = _getDomContainer(doc.documentElement);
+        const container = _getDomContainer(doc.body.firstElementChild!);
```
Varixo, line 49: its just docs parser fix, not related

gioboa (approved):
Looking forward to trying this fix 👏💪
You rock @Varixo

PR #7941, opened 2025-09-11
chore(ci): switch to macos-13 for e2e

wmertens:
We'll see if that's faster


.github/workflows/ci.yml
```diff
@@ -743,7 +743,8 @@ jobs:
       matrix:
         settings:
           - host: ubuntu-latest
-          - host: macos-latest
+          # Use the older macos for x86 testing + faster
+          - host: macos-13
```
gioboa, line 747: I tried it few days ago, it's not working. But let's see 🙈

PR #7943, opened 2025-09-12
fix: resuming shadow dom container with multiple root children

Varixo:
Pass shadowroot to the treewalker and walkContainer to correctly resume container with additional nodes before or after container inside shadowroot.


starters/e2e/e2e.backpatching.e2e.ts
```diff
@@ -1,6 +1,6 @@
 import { test, expect } from "@playwright/test";
 
-test.describe.only("backpatching", () => {
```
gioboa, line 3: Nice catch 😅

PR #7949, opened 2025-09-12
fix(docs): svg images need extensions

wmertens:


gioboa:
The preview is working fine and production as well 💯

PR #7957, opened 2025-09-16
fix(qwik-city): force exit on SSG hang

wmertens:
Implement a workaround to address the issue where the SSG process hangs after completion by forcibly exiting the process after a delay.


gioboa (approved):
LGTM 👍

PR #7959, opened 2025-09-16
docs: prioritize doc search results 📚

### LINKED ISSUE #6429: [📖] Algolia search often points to `/api/qwik` instead of the corresponding route in the docs (e.g. search for useResource)
### Suggestion

image

As you can see, it shows the API results first. I think it should be the other way around.

gioboa:
Close #6429

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/.changeset/fifty-dots-wonder.md
+++ b/.changeset/fifty-dots-wonder.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik-city': patch
+---
+
+Fix: SSG sometimes hangs after completion, now we forcibly exit the SSG process when this happens.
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/docsearch/doc-search-modal.tsx
```diff
@@ -81,7 +81,7 @@ export const DocSearchModal = component$(
       })
         .then(({ collections }) => {
           state.status = 'idle';
-          state.collections = collections.map((c) => ({
+          state.collections = collections.reverse().map((c) => ({
```
gioboa, line 84: Reverse the order is enough to achieve the expected result.

maiieul: Fair enough 😄 👍

Issue #7868, opened 2025-08-26
[🐞] V2 Router: head() override order is wrong


devcaeg:
### Which component is affected?

Qwik Runtime

### Describe the bug

When using Qwik v2 with nested layouts, defining Head with title and description in a deeper layout does not override the values defined in a higher layout. The browser keeps the outer layout’s title/description. This seems to contradict the docs for nested layouts & Head.

https://qwik.dev/docs/pages/#nested-layouts-and-head

The docs imply head content from nearer (deeper) layouts/pages should take precedence when merged.

**Expected behavior**

On a route rendered by a deeper layout, the title and description from that deeper layout’s <Head> should be applied (or at least take precedence over the outer layout’s values).

### Reproduction

https://stackblitz.com/edit/github-5yxcyvac-o3sjnsqu?file=src%2Froutes%2Flayout.tsx

### System Info

```shell
System:
    OS: macOS 26.0
    CPU: (8) arm64 Apple M2
    Memory: 113.97 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.18.0
    npm: 10.9.3
    pnpm: 10.15.0
    bun: 1.2.21
  Browsers:
    Chrome: 139.0.7258.139
    Edge: 139.0.3405.111
    Safari: 26.0
  npmPackages:
    @qwik.dev/core: https://pkg.pr.new/QwikDev/qwik/@qwik.dev/core@9b03622 => 2.0.0-beta.7 
    @qwik.dev/router: https://pkg.pr.new/QwikDev/qwik/@qwik.dev/router@9b03622 => 2.0.0-beta.7 
    typescript: 5.9.2
    vite: 7.1.3
```

Varixo:
@devcaeg do you have reproduction?

devcaeg:
@Varixo @wmertens 

I have added a reproduction.

As you can see on the home page, the meta title is “Title default,” and on the “/test” page, the meta title should be “Test 2,” but it still says “Title default.”

wmertens:
actually, this also happens in v1, and it's intended. The idea is that earlier head() calls can wrap nested ones.

https://qwik.dev/docs/pages/#nested-layouts-and-head

I suppose we could also have a middleware approach but that would require a RFC.

PR #7970, opened 2025-09-18
feat(router): [breaking] change head merge order

wmertens:
The current way of merging head objects is surprising, like #7868.

This PR changes it to merge objects in the unsurprising order, but keep merging functions in the original order.

### DIFF
--- a/.changeset/some-emus-fly.md
+++ b/.changeset/some-emus-fly.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': major
+---
+
+Breaking: The order of head export merging has been slightly. Plain objects now override outer ones. Functions still are run inner-first.
--- a/packages/docs/src/routes/docs/(qwikrouter)/pages/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikrouter)/pages/index.mdx
@@ -36,7 +36,7 @@ export default component$(() => {
 
 ## `head` export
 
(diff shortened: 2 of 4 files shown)


Varixo (approved):

.changeset/some-emus-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/router': major
+---
+
+Breaking: The order of head export merging has been slightly. Plain objects now override outer ones. Functions still are run inner-first.
```
shairez, line 5: maybe add :boom: in front of the BREAKING

packages/docs/src/routes/docs/(qwikrouter)/pages/index.mdx
```diff
@@ -121,11 +121,30 @@ export const head: DocumentHead = ({resolveValue, params}) => {
...
+export const head: DocumentHead = ({ head }) => {
+  return {
+    title: `MySite - ${head.title}`,
+  };
+};
+```
+
+So first all plain object `head` exports are merged, and then the function `head` exports are called in reverse order.
+
+Furthermore, if two values in arrays (like `meta` or `links`) have the same `key`, the last one wins. This allows you to override specific meta tags.
```
shairez: this is related to objects not functions right? 
if so I think it should come before the paragraph about functions, to make sure people won't think it's related to functions.

wmertens: It applies to both. Let me think about how to word it better

shairez: cool thanks

PR #7975, opened 2025-09-19
fix(docs): redirects

wmertens:
Hopefully this works, to test first


wmertens:
gaah it detected it as markdown and converted `*` to `_`

PR #7991, opened 2025-09-23
fix: resuming app with non-qwik elements inside

Varixo:
Skip non-qwik elements while resuming


packages/qwik/src/core/client/vnode-namespace.ts
```diff
@@ -173,7 +194,8 @@ function vnode_cloneElementWithNamespace(
 
       // Descend into children
       // We need first get the first child, if any
-      const vFirstChild = vnode_getFirstChild(vCursor);
+      // vFirstChild = vnode_getFirstChild(vCursor);
```
wmertens: shouldn't we remove this?

PR #7997, opened 2025-09-24
perf: more perf improvements

Varixo:
- use replaceChildren for truncating nodes
- better algorithm for diffing keyed nodes

### DIFF
--- a/.changeset/tasty-glasses-bet.md
+++ b/.changeset/tasty-glasses-bet.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: memory leak for reactive attributes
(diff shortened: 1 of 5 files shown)


packages/qwik/src/core/client/vnode.ts
```diff
@@ -968,6 +969,15 @@ export const vnode_applyJournal = (journal: VNodeJournal) => {
           idx++;
         }
         break;
+      case VNodeJournalOpCode.RemoveAll:
+        const removeAllParent = journal[idx++] as Element;
+        if (removeAllParent.replaceChildren) {
+          removeAllParent.replaceChildren();
+        } else {
+          // fallback if replaceChildren is not supported
+          removeAllParent.textContent = '';
```
wmertens, line 978: We should probably count the children first in the fallback, not deleting if there are none

wmertens (approved):
LGTM 🥬

PR #8005, opened 2025-09-28
chore: enable types in lint

wmertens:
We missed quite a few promises due to these rules not being enabled.

However, they take a long time to run, so let's see where we can partially enable them


maiieul:
Closing as stale.

PR #8008, opened 2025-09-29
feat(core): allow qwikloader inlining

wmertens:


gioboa (approved):
LGTM

packages/qwik/src/server/render.ts
```diff
@@ -136,7 +151,22 @@ export async function renderToStream(
...
+    const qwikLoaderScript = getQwikLoaderScript({
+      debug: opts.debug,
+    });
+    beforeContent.unshift(
+      jsx('script', {
+        id: 'qwikloader',
+        // Qwik only works when modules work
+        type: 'module',
+        // Don't block HTML parsing
+        async: true,
```
maiieul, line 168: It seems async does nothing for inline scripts https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script?utm_source=chatgpt.com#attributes

wmertens: for classic scripts but it's ok for module scripts

maiieul: This is an inline script. MDN says it's useless.

> Warning: This attribute must not be used if the src attribute is absent (i.e., for inline scripts) for classic scripts, in this case it would have no effect.

chatgpt confirms. Are we both dumb? 😄

maiieul:
@wmertens @Varixo better fix CI in a separate PR, but cool 👍. Thanks for doing that.

packages/qwik/src/server/render.ts
maiieul: Okay I'll make a separate PR about that since CI is green. it's not harmful anyways.

PR #8028, opened 2025-10-04
fix: scheduling previously blocked chore

Varixo:
Always add unblocked chore to chores queue on vnode. This is needed to correctly find blocking chore later.


thejackshelton:
Giving a heads up that when I clicked the deployment preview for this PR, when I clicked the Search button at the top right, the entire page became inert

Varixo:
> Giving a heads up that when I clicked the deployment preview for this PR, when I clicked the Search button at the top right, the entire page became inert

should be fixed

PR #8039, opened 2025-10-07
chore: improve CSR starter 🍒

gioboa:
CSR or SPA is more understandable then browser-only IMHO.

### DIFF
--- a/starters/features/csr/index.html
+++ b/starters/features/csr/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Qwik browser-only starter</title>
+    <title>Qwik SPA starter</title>
 
     <script async type="module" src="@qwik.dev/core/qwikloader.js">
       // ^ This is the QwikLoader, it's required for Qwik to work
--- a/starters/features/csr/package.json
+++ b/starters/features/csr/package.json
@@ -1,7 +1,7 @@
 {
-  "description": "Use Qwik in browser-only mode",
+  "description": "Use Qwik in SPA mode (client side rendering)",
   "__qwik__": {
-    "displayName": "Integration: Qwik in browser-only mode",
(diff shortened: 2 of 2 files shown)


starters/features/csr/vite.config.mts
```diff
@@ -5,9 +5,22 @@
...
 
-export default defineConfig((): UserConfig => {
+export default defineConfig((config): UserConfig => {
   return {
-    plugins: [qwikVite({ csr: true }), tsconfigPaths({ root: "." })],
+    plugins: [
+      qwikVite({ csr: true }),
+      tsconfigPaths({ root: "." }),
+      config.mode === "development" &&
+        viteStaticCopy({
```
gioboa, line 16: This is needed to resolve 
`<script async type="module" src="@qwik.dev/core/qwikloader.js">` in dev mode.

maiieul: why is it needed in dev mode?

gioboa: without this fix, in dev mode, is not resolving and serving the file correctly.

maiieul: What about using the `qwikloader: "inline"` prop we just brought back in v1.17.0?

starters/features/csr/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Qwik browser-only starter</title>
+    <title>Qwik SPA starter</title>
```
wmertens, line 6: Hmmm - SPA implies multiple routes, no? Without qwik-router there's no routes...
Technically, it's the CSR starter.

starters/features/csr/vite.config.mts
wmertens: Hmm that seems wrong, vite should be parsing the index.html and resolving the qwikloader.js into a module.
Isn't it working? I thought I tested it.

PR #8041, opened 2025-10-07
Load core

GrandSchtroumpf:
# What is it?
- Docs 

# Description
Add documentation about waking up core

# Checklist

### DIFF
--- a/packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
+++ b/packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
@@ -1,6 +1,7 @@
 # Core
 
 In order to be super fast at load time Qwik only load the framework itself (let's call it "core") when it needs it.
+
 If you want to squeeze the last bit of loadtime performance, you can delay its execution until you really need it.
 
 ⚠️ Delaying core execution should **not** affect your UX or devX, this is an advance performance trick ⚠️
@@ -17,12 +18,8 @@ useVisibleTask$(() => {
 
 // Requires core on requestIdleCallback
 useVisibleTask$(
-  () => {
-    console.log('Hello core');
-  },
-  {
-    strategy: 'document-idle',
-  }
+  () => console.log('Hello core'),
+  { strategy: 'document-idle' }
 );
 ```
 
@@ -32,7 +29,7 @@ In **some** cases you can replace `useVisibleTask$` with either `useOn` or `useT
 
 Replace `useVisibleTask$` with `useOn('qvisible')` / `useOn('qidle')` if
 
-- You only need to trigger a callback on init
+- You only need to trigger a callback once
 - The code must run on the client
 
 `useOn`, `useOnDocument` & `useOnWindow` execute core if they use a variable from the component scope :
@@ -43,6 +40,7 @@ const globalId = 'global-id';
 const Component = component$(() => {
   const ref = useSignal();
   const id = useId();
+
   // Executes core at load time
   useOn('qidle', $(() => console.log(ref)));
   // Executes core at load time
@@ -139,4 +137,4 @@ useTas$(({ cleanup }) => {
 });
 ```
 
-As we can see, this is a LOT of work, and a not a great dev experience !
+As we can see, this is a LOT of work, and it's not a great dev experience !
--- a/packages/docs/src/routes/docs/(qwik)/advanced/core/index.md
+++ b/packages/docs/src/routes/docs/(qwik)/advanced/core/index.md
@@ -1,0 +1,142 @@
+# Core
+
+In order to be super fast at load time Qwik only load the framework itself (let's call it "core") when it needs it.
+If you want to squeeze the last bit of loadtime performance, you can delay its execution until you really need it.
+
+⚠️ Delaying core execution should **not** affect your UX or devX, this is an advance performance trick ⚠️
+
+## useVisibleTask$
+
+`useVisibleTask$` will **always** execute core before it's callback is called.
+
+```jsx
+// Requires core when component is visible in the viewport
+useVisibleTask$(() => {
+  console.log('Hello core');
+});
+
+// Requires core on requestIdleCallback
+useVisibleTask$(
+  () => {
+    console.log('Hello core');
+  },
+  {
+    strategy: 'document-idle',
+  }
+);
+```
+
+In **some** cases you can replace `useVisibleTask$` with either `useOn` or `useTask$`
+
+## useOn
+
+Replace `useVisibleTask$` with `useOn('qvisible')` / `useOn('qidle')` if
+
+- You only need to trigger a callback on init
+- The code must run on the client
+
+`useOn`, `useOnDocument` & `useOnWindow` execute core if they use a variable from the component scope :
+
+```jsx
+import { libId } from 'library';
+const globalId = 'global-id';
+const Component = component$(() => {
+  const ref = useSignal();
+  const id = useId();
+  // Executes core at load time
+  useOn('qidle', $(() => console.log(ref)));
+  // Executes core at load time
+  useOn('qidle', $(() => console.log(id)));
+  // Do not execute core at load time
+  useOn('qidle', $(() => console.log(globalId)));
+  // Do not execute core at load time
+  useOn('qidle', $(() => console.log(libId)));
+  // Do not execute core at load time
+  useOn('qidle', $(() => console.log('id')));
+
+  return (
+    <p ref={ref}></p>
+    <p id={id}></p>
+    <p id={globalId}></p>
+    <p id={libId}></p>
+    <p id="id"></p>
+  )
+})
+```
+
+## useTask$
+
+Replace `useVisibleTask$` with `useTask$` if
+
+- You need to listen on state changes
+- The code can execute on the server
+
+```jsx
+const Component = component$(() => {
+  const search = useSignal();
+  // Do not execute core at load time
+  useTask$(({ track }) => {
+    track(search);
+    console.log(search);
+  });
+  return <input bind:value={search} type="search" />;
+});
+```
+
+## useOn + useTask$
+
+A classic usecase of `useVisibleTask$` is to start listening on browser specific event:
+
+```jsx
+const isMobile = useSignal(false);
+useVisibleTask$(({ cleanup }) => {
+  const query = window.matchMedia('(max-width: 400px)');
+  const handler = (event) => {
+    isMobile.value = event.matches;
+  };
+  query.addEventListener('change', handler);
+  cleanup(() => query.removeEventListener('change', handler));
+});
+```
+
+In this case we actually need core when the handler is triggered. Here is how to delay core execution :
+
+```jsx
+const isMobile = useSignal(false);
+// On idle, start listening on the event
+useOn(
+  'qidle',
+  $(() => {
+    const query = window.matchMedia('(max-width: 400px)');
+    // Store mediaQuery & handler to cleanup later
+    globalThis['media-query:(max-width: 400px)'] = {
+      query,
+      handler: (event) => {
+        // Forward the event to the document
+        const copy = new e.constructor('media-query:(max-width: 400px)', event);
+        document.dispatchEvent(copy);
+      },
+    };
+    query.addEventListener('change', globalThis['media-query:(max-width: 400px)'].handler);
+  })
+);
+
+// useOnDocument execute core when it actually needs it
+useOnDocument(
+  'media-query:(max-width: 400px)',
+  $((event) => {
+    isMobile.value = event.matches;
+  })
+);
+
+// useTask$ is used to cleanup event listeners
+useTas$(({ cleanup }) => {
+  cleanup(() => {
+    if (!globalThis['media-query:(max-width: 400px)']) return;
+    const { query, handler } = globalThis['media-query:(max-width: 400px)'];
+    query.removeEventListener('(max-width: 400px)', handler);
+  });
+});
+```
+
+As we can see, this is a LOT of work, and a not a great dev experience !


packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+// useOnDocument execute core when it actually needs it
+useOnDocument(
+  'media-query:(max-width: 400px)',
+  $((event) => {
+    isMobile.value = event.matches;
+  })
+);
+
+// useTask$ is used to cleanup event listeners
+useTas$(({ cleanup }) => {
```
gioboa: Typo here 👀

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+In order to be super fast at load time Qwik only load the framework itself (let's call it "core") when it needs it.
+
+If you want to squeeze the last bit of loadtime performance, you can delay its execution until you really need it.
+
+⚠️ Delaying core execution should **not** affect your UX or devX, this is an advance performance trick ⚠️
+
+## useVisibleTask$
+
+`useVisibleTask$` will **always** execute core before it's callback is called.
```
wmertens: `its`

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+const globalId = 'global-id';
+const Component = component$(() => {
+  const ref = useSignal();
+  const id = useId();
+
+  // Executes core at load time
+  useOn('qidle', $(() => console.log(ref)));
+  // Executes core at load time
+  useOn('qidle', $(() => console.log(id)));
+  // Do not execute core at load time
```
wmertens: `Does not`

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+Replace `useVisibleTask$` with `useTask$` if
+
+- You need to listen on state changes
+- The code can execute on the server
+
+```jsx
+const Component = component$(() => {
+  const search = useSignal();
+  // Do not execute core at load time
```
wmertens: How about ``Does not execute until `search` changes``?

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+  useTask$(({ track }) => {
+    track(search);
+    console.log(search);
```
wmertens: `search.value` or maybe `console.log(track(search))`

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+  });
+  return <input bind:value={search} type="search" />;
+});
+```
+
+## useOn + useTask$
+
+A classic usecase of `useVisibleTask$` is to start listening on browser specific event:
```
wmertens: I think a nicer pattern for this is to have a `sync$` qrl listening on the browser event, and then dispatching a custom event if the conditions are right. Then you can just put `onMyCustomEvent$={...}` somewhere.

GrandSchtroumpf: I updated the code with `sync` and I store directly the cleanup method on the document to make it more readable.
Let me know if you want more changes :)

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
+# Core
+
+In order to be super fast at load time Qwik only load the framework itself (let's call it "core") when it needs it.
+
+If you want to squeeze the last bit of loadtime performance, you can delay its execution until you really need it.
+
+⚠️ Delaying core execution should **not** affect your UX or devX, this is an advance performance trick ⚠️
```
maiieul: Looking at the profiler on a small repro I have, it seems core is "woken up" eagerly anyways, likely because of modulepreload.

In any case, slow network delays are often orders of magnitude higher than slow CPU delays. On calibrated low tier CPU slowdown, there's some heavier js execution right after core is preloaded, but it's "just ~70ms (on a low tier device!), whereas on 3G the network delays can be counted in seconds.

So it's fine that core is woken up eagerly. In fact we preload it eagerly because it's 99% likely to be needed. Otherwise why use Qwik in the first place? 😄

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+  // Executes core at load time
+  useOn('qidle', $(() => console.log(ref)));
+  // Executes core at load time
+  useOn('qidle', $(() => console.log(id)));
+  // Does not execute core at load time
+  useOn('qidle', $(() => console.log(globalId)));
+  // Does not execute core at load time
+  useOn('qidle', $(() => console.log(libId)));
+  // Does not execute core at load time
+  useOn('qidle', $(() => console.log('id')));
```
maiieul: If core indeed executes eagerly then this is not true. Also with a current bug in the optizer output this is probably not deterministic. I would remove this section, it's not really actionable atm.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
maiieul, line 1: Let's not create new routes unless absolutely necessary. More routes means more redirects eventually and is annoying to maintain. I would put this in `/guides/best-practices/`.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+In order to be super fast at load time Qwik only load the framework itself (let's call it "core") when it needs it.
+
+If you want to squeeze the last bit of loadtime performance, you can delay its execution until you really need it.
+
+⚠️ Delaying core execution should **not** affect your UX or devX, this is an advance performance trick ⚠️
+
+## useVisibleTask$
+
+`useVisibleTask$` will **always** execute core before its callback is called.
```
maiieul: So the problem with visible tasks is not that it wakes up core, but that if you have too many of them, you will have to preload all their bundles eagerly with 100% probability, preventing any user event from getting higher priority in the preload queue.

Some visible tasks are legitimate and must execute ASAP, but some could just be a useTask or useOn indeed.

I would put the emphasis on the fact that visible tasks tend to create preload traffic jams or something like that.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+// Requires core on requestIdleCallback
+useVisibleTask$(
+  () => console.log('Hello core'),
+  { strategy: 'document-idle' }
+);
+```
+
+In **some** cases you can replace `useVisibleTask$` with either `useOn` or `useTask$`
+
+## useOn
```
maiieul: This should be a ###

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+    <p ref={ref}></p>
+    <p id={id}></p>
+    <p id={globalId}></p>
+    <p id={libId}></p>
+    <p id="id"></p>
+  )
+})
+```
+
+## useTask$
```
maiieul: This should be a ###

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+  // Does not execute until `search` changes
+  useTask$(({ track }) => {
+    track(search);
+    console.log(search.value);
+  });
+  return <input bind:value={search} type="search" />;
+});
+```
+
+## useOn + useTask$
```
maiieul: This should be a ###

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
```diff
@@ -0,0 +1,140 @@
...
+useTask$(({ cleanup }) => {
+  cleanup(() => {
+    if (!document['cleanup:media-query:(max-width: 400px)']) return;
+    document['cleanup:media-query:(max-width: 400px)']();
+    delete document['cleanup:media-query:(max-width: 400px)'];
+  });
+});
+```
+
+As we can see, this is a LOT of work, and it's not a great dev experience !
```
maiieul: Indeed. So what's the message here? That we should avoid this pattern? If so, we should remove this section. I remember an RFC to add media queries to useOnDocument.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
GrandSchtroumpf: I'm not sure to fully understand. There is a difference between preloading and executing.
- Preloading with modulepreload will fetch & compile in another thread
- Executing will fetch, compile & execute in the main thread

It's all a question of priority. If you don't execute core, you give room for the browser to do something else.
In any case core will always preload because we'll need it at some point.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
GrandSchtroumpf: I want to explain what it requires to not execute core at load time. There are some valid cases for that though, if you're building a library to start listening on event from the browser like the example. There is no reason to execute core just to start an event listener.

The promise of Qwik is to optimize network & CPU usage at scale. A good part of the dev who use Qwik are looking to squeeze performance as much as possible, and delaying core execution is one tool. The devX is terrible, but I think the doc should mention how to achieve it.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
maiieul: The important question is: what is actually leading to delays?

1) Network delays
2) blocking the main thread

If you run a minimal qwik app with just a counter, you'll see that even without clicking the counter, there is a main thread task coming from the core bundle right after it is preloaded.

But the most important thing here, is that the "blocking the main thread" is actually very little time. Network delays are 10x if not 100x bigger.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
maiieul: So we can add this temporarily until the the media queries can be used with `useOnDocument`. It will be much simpler. Can you add a link to the RFC here? Once the PR is merged we should also create an issue to remove it once the RFC is completed.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
GrandSchtroumpf: I just did the tests, core is always preloaded :
- no hook: no execution at load time
- `useOn('qvisible', $(() => console.log('id'))`: no execution at load time
- `useOnDocument('qidle', $(() => console.log('id'))`: module is evaluated (5µs on my computer)
- `useOn('qvisible', $(() => console.log(state))`: core is executed (2ms on my computer)
- `useVisibleTask$(() => console.log('id'))`: core is executed in 3 different tasks (2.5ms in total)

It's small numbers because it's on my machine & on a super small project, but it creates a small traffic on main thread, and if a dev wants to reduce it as much as possible I think he should know how to.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
GrandSchtroumpf: media query is just an example, any event listener that happen only on the browser could leverage this pattern :
- navigation.addEventListener("navigate", ...);
- navigator.connection.addEventListener('change', ...);
- navigator.devicePosture.addEventListener("change", ...);
...

They won't all fit into `useOnDocument`. Also, I didn't find the RFC you mentioned.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
GrandSchtroumpf: Here is an actionable use case: 
Let's say a dev read the doc and see that if he can he should replace `useVisibleTask$` with `useOn('qvisible')`. So he takes the effort to write this : 
```jsx
const ref = useState();
useOn('qvisible', $(() => ref.value.focus()));
```
The effort would have been for nothing, because `ref` is in the scope of the component. With this doc, he now knows that this would not execute core : 
```jsx
useOn('qvisible', $(() => document.getElementById('id').focus()))
```
If hardcoding the id is no effort for him, this is a quick win :)

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
GrandSchtroumpf: `useOn('qvisible')` won't create the traffic jam `useVisibleTask$` is doing ?

maiieul: Hmmm it should create the same traffic jam.

packages/docs/src/routes/docs/(qwik)/advanced/core/index.mdx
wmertens: The rules are as follows:

Waking up means executing the core module and deserializing the required state.

The only way core can wake up is via an event that is processed by qwikloader. Qwikloader runs the qrl, which in turn imports core to schedule or to retrieve captured scope.

Sync qrls never wake up core.

Regular qrls that capture scope always wake up core, for deserializing.

Tasks always wake up core, for scheduling.

UseOn* registers a qrl handler, no direct impact on core.

packages/docs/src/routes/docs/(qwikcity)/guides/best-practices/index.mdx
maiieul, line 22: Don't you want to add your beautiful avatar here? 😉

GrandSchtroumpf: No all good ^^ let's merge it if we can :)

PR #8043, opened 2025-10-08
chore(docs): Move origin doc

wmertens:
the previous place was the aute API docs.

I also indented the ToC so that it's easier to see what section goes where


gioboa (approved):
It sounds good to me 🙏

PR #8050, opened 2025-10-10
docs: light mode dark mode

JerryWu1234:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

Feature: Add System-Sync ("Auto") Theme to Docs

# Checklist

### DIFF
--- a/packages/docs/src/components/theme-toggle/load-theme.js
+++ b/packages/docs/src/components/theme-toggle/load-theme.js
@@ -1,18 +1,16 @@
-(() => {
-  try {
-    const getItem = localStorage.getItem('theme-preference');
-    const el = document.firstElementChild;
-    if (!el) {
-      throw new Error('documentElement not found');
-    }
-
-    if (getItem === 'light' || getItem === 'dark') {
-      el.setAttribute('data-theme', getItem);
-    } else {
-      const isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-      el.setAttribute('data-theme', isDark ? 'dark' : 'light');
-    }
-  } catch (err) {
-    console.error(err);
+try {
+  const getItem = localStorage.getItem('theme-preference');
+  const el = document.firstElementChild;
+  if (!el) {
+    throw new Error('documentElement not found');
   }
-})();
+
+  if (getItem === 'light' || getItem === 'dark') {
+    el.setAttribute('data-theme', getItem);
+  } else {
+    const isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
+    el.setAttribute('data-theme', isDark ? 'dark' : 'light');
+  }
+} catch (err) {
+  console.error(err);
+}


JerryWu1234:
> Thanks for your help @JerryWu1234 with this change the top right icons are not visible anymore. 🤔
> 
> Image

https://qwik.dev/
please look at the official website

JerryWu1234:
> I see, can you fix this issue as well please?

done

packages/docs/src/components/code-sandbox/index.css
```diff
@@ -17,6 +17,26 @@
   background-color: #161616;
 }
 
+@media (prefers-color-scheme: dark) {
```
wmertens: I think all these dark mode changes should only look at data-theme.

packages/docs/src/components/theme-toggle/theme-script.tsx
```diff
@@ -3,11 +3,10 @@ export const themeStorageKey = 'theme-preference';
 export const ThemeScript = () => {
   const themeScript = `
         try {
-          document.firstElementChild
-              .setAttribute('data-theme',
-                  localStorage.getItem('${themeStorageKey}') ??
-                  (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light')
-              );
+          const getItem = localStorage.getItem('${themeStorageKey}')
+          if(getItem === 'light' || getItem === 'dark'){
```
wmertens: Let's instead make this a media query listener.
Every time the media query changes, it's because the user changed the theme. 

When that happens, check local storage (every time). If it's not (empty or auto), then use that for data-theme, otherwise use dark/light depending on the query. 

This runs independently from the theme toggle, which also changes data-theme but sets local storage. So they don't interfere with each other

packages/docs/src/components/theme-toggle/theme-script.tsx
wmertens, line 1: It would be best to move this file to the theme toggle component folder

packages/docs/src/components/theme-toggle/theme-toggle.css
```diff
@@ -1,33 +1,17 @@
...
+  height: 21px;
 }
 
-.theme-toggle > svg {
-  inline-size: 100%;
-  block-size: 100%;
-  stroke-linecap: round;
+html[data-theme='light'] .themeIcon.light,
+html[data-theme='dark'] .themeIcon.dark,
+html:not([data-theme]) .themeIcon.auto {
```
wmertens: Data-theme should always hold the current theme even if it's auto. That way the css is much simpler. 

So for auto mode we need a different prop. Maybe `data-theme-name`? Then all three icons can check that. And `data-theme` basically only ever contains "dark" (unless we add more themes).

packages/docs/src/components/theme-toggle/theme-toggle.tsx
```diff
@@ -1,10 +1,32 @@
...
 import { themeStorageKey } from '../router-head/theme-script';
 import themeToggle from './theme-toggle.css?inline';
-import { GlobalStore } from '../../context';
+import { SunIcon } from './Sun';
+import { MoonIcon } from './Moon';
+import { BrillianceIcon } from './Brilliance';
+export type ThemePreference = 'dark' | 'light' | 'auto';
+export const setPreference = (theme: ThemePreference) => {
+  if (theme === 'auto') {
+    document.firstElementChild?.removeAttribute('data-theme');
```
wmertens: Here it should pick the media query answer

packages/docs/src/components/theme-toggle/theme-toggle.tsx
```diff
@@ -1,10 +1,32 @@
...
+  let theme;
+  try {
+    theme = localStorage.getItem(themeStorageKey);
+  } catch {
+    //
+  }
+  return (theme as ThemePreference) || 'auto';
+};
 
 export const colorSchemeChangeListener = (onColorSchemeChange: (isDark: boolean) => void) => {
```
wmertens: So this should live in the head script

packages/docs/src/routes/api/api.css
```diff
@@ -60,6 +60,8 @@
 [data-theme="dark"] .docs.api table {
   box-shadow: 0 0 30px 0 rgba(80,80,80,0.1);
 }
+
```
wmertens: How did this extra space get here? 

We should make sure prettier also formats these files

JerryWu1234:
@wmertens done

wmertens (approved):
LGTM

gioboa (commented):
Thanks for this improvement 👏

PR #8064, opened 2025-10-14
v2: make client outDir sane

wmertens:
**BREAKING**

When you pass `base` to vite, Qwik will also add `base` to the client output.

E.g. `base: "/hi/there/"` will put the client files under `"dist/hi/there"`.

This cannot be tweaked and most of the time it is not what you want.

This PR reverts that behavior and lets the dev add `base` themselves if they like.

This PR also allows passing a q-manifest by path.

### DIFF
--- a/.changeset/clear-shoes-cheat.md
+++ b/.changeset/clear-shoes-cheat.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': minor
+---
+
+FEAT: `qwikVite` now accepts `ssr.manifestInputPath` for when the `q-manifest.json` file from the client build is at an unexpected location.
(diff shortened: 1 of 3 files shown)


.changeset/clear-shoes-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': minor
+---
+
+FEAT: `qwikVite` now accepts `ssr.manifestInputPath` for when the `q-manifest.json` file from the client build is at an unexpected location.
```
shairez, line 5: if we have use cases then it would be great to document one or two examples here.
if not, maybe this should be private for now, until we have a use case

PR #8067, opened 2025-10-15
chore: remove .cjs builds

wmertens:
Maintaining the .cjs builds is extra work and it looks like there is no real reason to do it.


maiieul:
Chatgpt approves 👍 
image

gioboa:
> Chatgpt approves 👍

:rofl:

wmertens:
close in favor of #8103

PR #8076, opened 2025-10-20
v2: filter tree walker to only traverse qwik nodes

thejackshelton:
In dev mode specifically vite (or something else) will inject extra nodes that the tree walker is currently counting, which is offsetting the node positions.

This means it currently backpatches the wrong node in latest (dev mode).

I found that we can discriminate the qwik nodes by filtering for nodes with the `:` attribute in the DOM in the backpatch executor.


packages/docs/src/routes/api/qwik-testing/index.mdx
```diff
@@ -436,6 +436,7 @@ export declare function ssrRenderToDom(
     debug?: boolean;
     raw?: boolean;
     qwikLoader?: boolean;
+    injection?: (document: Document) => void;
```
Varixo: maybe we should call it like 'onBeforeResume' or something like that?

packages/qwik/src/backpatch-executor.ts
```diff
@@ -18,15 +18,21 @@ if (executorScript) {
       const data = JSON.parse(script.textContent || '[]');
       const walker = document.createTreeWalker(container, NodeFilter.SHOW_ELEMENT);
       let currentNode: Node | null = walker.currentNode;
-      let currentNodeIdx = 0;
+      let currentNodeIdx = (currentNode as Element).hasAttribute(':') ? 0 : -1;
```
Varixo, line 21: you could assign elementprototype.hasAttribute to variable and use in both places? maybe it will take less bytes

thejackshelton-kunaico: Like this?

>       const hasAttr = Element.prototype.hasAttribute;
      let currentNode: Node | null = walker.currentNode;
      let currentNodeIdx = hasAttr.call(currentNode, ':') ? 0 : -1;

packages/docs/src/routes/api/qwik-testing/index.mdx
thejackshelton-kunaico: changed

packages/qwik/src/backpatch-executor.ts
Varixo: yeah

thejackshelton: so turns out with Element it is larger. 314b without and 321b with. Also it was annoying to get tests to pass with Element because domino I don't think understands the Element prototype 😅

gioboa (approved):
👏LGTM

PR #8104, opened 2025-10-30
Add RequestEvent clients getter

DustinJSilk:
# What is it?

- Feature / enhancement

# Description

Interfacing with API clients in loaders and server$ functions is a common use case for REST, gRPC, and GraphQL servers.
To share clients on the RequestEvent, the only possible solution is to store it in the sharedMap which makes using them cumbersome:

```ts
export const useLoader = routeLoader$(async (ev) => {
  const clients = ev.sharedMap.get("qwik-api-clients");
  const result = await clients.foo.say({ sentence: "Success!" });
});
```

It would be better to have a DX similar to this:

```ts
export const useLoader = routeLoader$(async (ev) => {
  const result = await ev.clients.foo.say({ sentence: "Success!" });
});
```

To achieve this, we would need to add a getter to the RequestEvent object which returns`ev.sharedMap.get("qwik-api-clients")`.

This way, it's easy for users to extend the RequestEvent with their API clients:

```ts
// Register clients by storing them in the sharedMap
export function registerApiClients(ev: RequestEventBase) {
  ev.sharedMap.set("qwik-api-clients", someClientsObject);
}

// Extend the RequestEventBase to add the client types
declare module "@builder.io/qwik-city" {
  interface RequestEventBase {
    clients: SomeClients
  }
}
``` 

# Checklist

### DIFF
--- a/packages/qwik-city/src/middleware/request-handler/request-event.ts
+++ b/packages/qwik-city/src/middleware/request-handler/request-event.ts
@@ -318,6 +318,10 @@ export function createRequestEvent(
       }
       return writableStream;
     },
+
+    get clients() {
+      return this.sharedMap.get('qwik-api-clients');
+    },
   };
   return Object.freeze(requestEv);
 }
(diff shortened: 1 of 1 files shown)


DustinJSilk:
For discussion in Discord

wmertens:
Like we discussed, I think it would be better not to seal the object

PR #8107, opened 2025-10-31
feat: introduce deferUpdates option for useTask$

Varixo:
Add new option for useTask$ `blockRender`. It helps managing animations and view transitions

### DIFF
--- a/packages/docs/src/routes/api/qwik/api.json
+++ b/packages/docs/src/routes/api/qwik/api.json
@@ -2304,7 +2304,7 @@
         }
       ],
       "kind": "Interface",
-      "content": "```typescript\nexport interface TaskOptions \n```\n\n\n<table><thead><tr><th>\n\nProperty\n\n\n</th><th>\n\nModifiers\n\n\n</th><th>\n\nType\n\n\n</th><th>\n\nDescription\n\n\n</th></tr></thead>\n<tbody><tr><td>\n\n[deferUpdates?](#)\n\n\n</td><td>\n\n\n</td><td>\n\nboolean\n\n\n</td><td>\n\n_(Optional)_ If true, the task will block the rendering of the component until it is complete.\n\n\n</td></tr>\n</tbody></table>",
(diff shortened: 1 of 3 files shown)


wmertens:
now that I read it again, `blockRender` sounds like rendering blocks, not blocking render :)

Maybe just `blocking`?

wmertens (commented):
three thumbs up, LGTM except maybe the option name

gnemanja:
Maybe it would be nice to include a description which says that useTask will always block rendering when executed on server. This option is only for browser

Varixo:
> now that I read it again, `blockRender` sounds like rendering blocks, not blocking render :)
> 
> Maybe just `blocking`?

changed to deferUpdates after talking to Jack

Varixo:
We need to talk about this change if it should be default to match v1 behavior. 
Imo the v1 behavior is not intuitive and we could change it.

wmertens:
We decided that the v1 blocking behavior should stand and we need to enable non-blocking via an option.

The alternative is creating a Promise but not returning it, but that doesn't feel great so the option is still useful

packages/docs/src/routes/docs/(qwik)/core/tasks/index.mdx
```diff
@@ -110,7 +135,7 @@ Additionally, this task can be reactive and will re-execute when **tracked** [st
 
 > If `useTask$()` does not track any state, it will run **exactly once**, either in the server **or** in the browser (**not both**), depending where the component is initially rendered. Effectively behaving like an "on-mount" hook.
 
-`useTask$()` will block the rendering of the component until after its async callback resolves, in other words, tasks execute sequentially even if they are asynchronous. (Only one task executes at a time).
+`useTask$()` will block the rendering of the component until after its async callback resolves. Tasks execute sequentially even if they are asynchronous (only one task executes at a time within a component). Subsequent re-executions (when tracking state changes) run asynchronously by default and don't block rendering unless `deferUpdates: true` is set.
```
gioboa: Is this description correct? 🤔 The previous descriptions are saying the opposite.

Varixo: fixed

wmertens:
I'd like to rename the option to something that needs to be set to `true` instead of `false`.
`concurrent` would work but it doesn't explain that it's for render only, it still goes synchronous in the component render

PR #8115, opened 2025-11-03
fix: sort backpatch entries before emitting

thejackshelton:
Sorts backpatch entries by element index in emitPatchDataIfNeeded() so the executor can use a single-pass TreeWalker. This simplifies the executor and improves performance.
The executor now assumes patches are in document order, so it can advance through the DOM once instead of searching for each element.

It also fixes a bug where out of order DOM patches skip over existing patches.


PatrickJS (approved):

gioboa (approved):
👍

thejackshelton:
I'm going to close since I think this was a misunderstanding of the problem, which is caused by something else 🤔 . It could be they are in order and something else in my codebase specifically is causing the issue.

PR #8119, opened 2025-11-03
fix: dont block qrl chores

Varixo:
We shouldnt block qrl run and resolve in the scheduler


wmertens:
Great!
followup: just remove those chore types and do the retry in _run

PR #8124, opened 2025-11-04
fix: don't emit script before qwik style element

Varixo:
Emitting script before qwik style is confusing resuming


wmertens (approved):
LGTM

PR #8126, opened 2025-11-05
feat: async computed changes

Varixo:
- first time throw promise to resolve
- the next times return stale value and update value when promise resolves
- rename `resolve` to `promise`

### DIFF
--- a/.changeset/deep-places-allow.md
+++ b/.changeset/deep-places-allow.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': minor
+---
+
+feat: rename .resolve() to .promise()
--- a/.changeset/petite-experts-press.md
+++ b/.changeset/petite-experts-press.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': minor
+---
+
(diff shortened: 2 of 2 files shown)


wmertens:
After discussing this on Discord, how about:

- renaming to `useAsync$`
- add `initial: unknown | () => unknown` option like for useSignal. If it is passed, it won't throw when read while not initialized
- add `eagerCleanup` option to run cleanup in the next tick if there are no more subscribers
- add `poll(ms: number)` callback that will compute again after `ms` time, if there are subscribers
- and/or add `poll: number` option that does the same
- do not make the result read-only, so that you can write new values like an observable:
  ```tsx
  const result = useAsync(({cleanup}) => {
    const channel = await makeChannel()
    // here we write directly to value, updating the subscribers. This also clears loading and error
    channel.on('data', d => result.value = d)
    // this throws the error when you read from value
    channel.on('error', e => result.error = e)
    cleanup(() => channel.close())
    return 'PENDING'
  }, {eagerCleanup: true})
  ```

I think that all these combine together to a very powerful primitive, that allows fetching data, auto-updating, calculation offload etc.

PR #8131, opened 2025-11-06
fix: serializing reused qrl

Varixo:
Don't override the QRL's `$capture$`


wmertens (approved):
Shouldn't we remove `$capture$` altogether?

LGTM

Varixo:
> Shouldn't we remove `$capture$` altogether?

Don't know. Can it be useful?

PR #8133, opened 2025-11-06
wip remove $capture$ from qrl

wmertens:
this is an optimization

it has failing unit tests


wmertens:
actually, it's probably better to lazily decode state when needed

PR #8135, opened 2025-11-07
feat: make RequestEvent readonly instead of frozen

DustinJSilk:
# What is it?

- Feature
- types

# Description

Switch from freezing RequestEvents to using readonly so that users can extend the RequestEvent while keeping the event safe from malpractice internally.

# Checklist

### DIFF
--- a/.changeset/shaggy-items-send.md
+++ b/.changeset/shaggy-items-send.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik-city': patch
+---
+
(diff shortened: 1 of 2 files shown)


wmertens (approved):
LGTM

wmertens:
We'll discuss this in next week's core meeting but looks good

wmertens:
Merging this. We still recommend using sharedMap, but at least you can change the object if you want to.

PR #8137, opened 2025-11-08
wip: journal per chain of rerenders

wmertens:
this attempts to fix the problem where we flush between removing a node and adding its replacement


wmertens:
closing in favor of #8143 and the future solution https://github.com/QwikDev/qwik-evolution/discussions/323

PR #8142, opened 2025-11-08
implementing bind_store to bind store values

MarcWeber:
# What is it?

- Feature / enhancement

# Description

Allowing to bind stores
```
          <input bind_store:value={state.nested.x.a}></input>.
```

No more updates to documentation done yet because no idea whether you like it.
If you do I will put in the work. Tested store.name and store.nested.nested.value (not includedin the patch but works). The code was written by Gemini ;) So maybe it can be improved a lot. Just didn't have the time to read through > 1000 lines of code. So treat this as proof of concept that the feature can be implemented easily. Maybe you can help me tell me where a good place would be to add the feature to the test suite. Cause partytown eventually should be minimal to stay as readable as possible ? Adding it to docs would be trivial replacing "stores not supported" by 2 lines of sample code:

```
const state = useStore({ nested: { nested: { value: ""} } });
return ( <input bind_store:value={state.nested.nested.value}></input> )
```

But there is also the tutorial maybe more to think about ...

### DIFF
--- a/packages/qwik/src/optimizer/core/src/transform.rs
+++ b/packages/qwik/src/optimizer/core/src/transform.rs
@@ -1167,7 +1167,13 @@ impl<'a> QwikTransform<'a> {
 				raw: None,
 			}))),
 		};
-		(dynamic_props, mutable, immutable_props, children, flags)
+		(
+			dynamic_props,
+			mutable,
+			immutable_props,
+			children,
+			flags.into(),
+		)
 	}
 
 	#[allow(clippy::cognitive_complexity)]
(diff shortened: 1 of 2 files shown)


wmertens:
As discussed on Discord, this would break standard JS interpretations of what is going on, and that is not "the Qwik way". When you pass `foo.bar` into a prop you don't expect the receiver to be able to change `bar`.

So I think the only way to do this is to pass both the store and the propname, like `bind:store={foo}` and `bind:value="bar"`. That way, the runtime sees that the value is not a Signal but a string, and it knows to read `bind:store` to get the object to change.

Also, this is made a lot easier in the `build/v2` branch, where the translation happens in jsx-node.ts, not in the optimizer.

MarcWeber:
I created a babel based transformer which runs before qwik to get the same
magic.
changes
- vite.config.ts -> add it
- pnpm
  pnpm add -D @babel/core
  pnpm add -D @babel/plugin-syntax-jsx
  pnpm add -D @babel/preset-typescript
  probably because vite is using rust now ?
  No idea how to add custom transformer yet
- src/qwik-bind-store.d.ts (adds the typing so that the bind_store:value/checkbox attributes are known to ts

The .js files I just put in top level directory for testing


https://github.com/MarcWeber/ts-u-qwik-bind-store 
Now the code has a home and can be referenced somewhere else.

wmertens:
Ah, yes that's definitely a good approach to get what you want. Babel isn't the fastest but it will work, and `qwikVite()` aims to be a team player so you can transform the code any way you like before it hits the optimizer.

thejackshelton:
@MarcWeber this is similar to what I do with vite + oxc-parser. Modifying the source code before the optimizer runs.

https://github.com/kunai-consulting/qwik-design-system/blob/v2-migration/libs/tools/vite/as-child.ts

We have a rewrite of the qwik optimizer in the works that uses the rust part of oxc, but it's also exposed in typescript and is roughly 20-30x faster than babel.

MarcWeber:
@thejackshelton : Its not easy to understand how your repo is different from QuickUI.

 To be honest I'd be interested in how Lisp would change the game. Like why have TS/JS/Rust in the first place ? .. But I don't have answers yet. Cause Lisp in theory can generate code which can be compiled to the speed difference for specific cases might turn much smaller. The problem with Rust is .. now you want my or your babel layer ? What do you do ? Keep slow JS ? create a module in webassebly which you load in addition to the optimizer ? Do you integrate a scripting engine (which one rhai or how they are all called into Rust) ? You the problem is the diversification and the many options and the boxes and choosing what is best. => Design paralysis . help get me out  .. :-P  And there are more problems such as backward compatible APIs I don't know how to cope with traditional languages.. Lisp could just dump code .. and you could load it .. and you can have a path back to and just put the old versions which were installed on the server into your repo even. I am tired of creating more solutions which cause problems which need to be solved (and learned). Speed is not my problem. My pages will have no more than up to 20 max 50 pages or so. Even if you use Rust for the optimizer (cause 10x faster) .. how about the backend code ? Its still JS right ? Want it XX faster, too ? And with lisp I think maybe the backend code can be optimized and compiled down to ocaml like so the let's reimplement react (or whatever) in ocaml story is gone, too. Not that speed is all. I guess being good enough is ok. So this all is an open question to me. Just dreaming about a simpler future .. sry. Also al l the vite stuff is about putting everything in a directory. What if I want to listen to 50 domains .. and only have minor differences ? Yeah could be reading and walking all the detais maybe there is a solution .. and ..  So while everything is fine works .. I also imagine some constraints and limitations.

PR #8143, opened 2025-11-09
fix: skip apply journal during vnode diff

Varixo:
Skip applying journal during diffing or component execution to prevent flickering

### DIFF
--- a/packages/qwik/src/core/shared/scheduler.ts
+++ b/packages/qwik/src/core/shared/scheduler.ts
@@ -194,6 +194,8 @@ export const createScheduler = (
   let currentTime = performance.now();
   const nextTick = createNextTick(drainChoreQueue);
   let flushTimerId: number | null = null;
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/shared/scheduler.ts
```diff
@@ -289,6 +296,7 @@ export const createScheduler = (
       $blockedChores$: null,
       $startTime$: undefined,
       $endTime$: undefined,
+      $blocksJournalFlush$: shouldBlockJournalFlush,
```
wmertens: isn't it better to calculate from type this every time? Here we're adding a minimum of 8 bytes to each chore.

We could also make Chore a class and use a getter.

Varixo: hmm yeah, but anyway I am waiting for dev's response if this helps

PR #8144, opened 2025-11-09
feat: memoize qrl symbols in qwikloader

Varixo:


wmertens:
Not sure if we should put it in a global - qwikloader can maintain its own? 

Also, if we just memoize the chunk imports and then get the symbol from those, we maintain a smaller object.

Issue #8151, opened 2025-11-13
[🐞] V1 build.assetsDir causes public files to not work


intellix:
### Which component is affected?

Qwik City (routing)

### Describe the bug

We're sharing a domain with our old angular app and the new qwik one and setting build.assetsDIr to "q" allows us to whitelist an entire directory in the load balancer to go to qwik to serve it.

With this set, I'm noticing that various files within our public directory aren't being handled as a static file, like robots.txt or manifest.json. A quick fix could be to move manifest.json into /q/ post build but Google expects robots.txt to be served from the root, so I need to fix it.

After building and serving our node entrypoint, I'm noticing that there's a isStaticFile check, which has a generated array of staticFiles that were detected during build. These have doubled up /q/q/ paths but everything still works because of the cheap /q/build check beforehand. Example: 

<img width="744" height="742" alt="Image" src="https://github.com/user-attachments/assets/746b1b2d-2e91-45c0-851f-787939aed8cf" />

I've looked around and it seems that the pathname is being prefixed with assetsDir here: https://github.com/QwikDev/qwik/blame/38429a16069f5cf03986cb82f636d78680fdf389/packages/qwik-city/src/adapters/shared/vite/index.ts#L172

Removing this prefixing of assetDir seems to fix it for us, proof (with the assetsDir correctly added to the build JS files):

<img width="1654" height="216" alt="Image" src="https://github.com/user-attachments/assets/0da29caf-eb90-497e-a639-4ff4b16fc574" />

The git blame says this was added by @gioboa via https://github.com/QwikDev/qwik/pull/6588 1 year ago. 

Removing this prefixing of assetsDir works both with/without a build.assetsDir being set for me but I don't want to regress whatever it was added for. Can you confirm that removing that line is ok @gioboa? I'll do a PR and patch it locally so I can move forward.

Varixo:
Solved by #8817, if not please reopen or create new issue

PR #8152, opened 2025-11-13
fix(qwik-city): remove doubling of assetDir on postBuild

intellix:
# What is it?

- Bug

# Description

Removed doubling of assetsDir during bundle close when delivering staticPaths list: #8151 

All info in the linked issue but I'll repeat a little how this is broken.

**Vite config:**
```ts
build: {
  assetsDir: 'q',
},
```

Now if you look at the list of generated staticPaths, you'll see these:
- /q/robots.txt
- /q/q/build/abc-123.js
- /q/q/build/def-456.js

# Checklist

### DIFF
--- a/.changeset/deep-bushes-live.md
+++ b/.changeset/deep-bushes-live.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik-city': minor
+---
+
+remove double assetsDir on bundle close of staticPaths
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ external
 lib
 tsdoc-metadata.json
 packages/qwik/bindings
+e2e/vite-e2e/output
 
 # IDE and local environment
 .idea
--- a/e2e/adapters-e2e/public/robots.txt
+++ b/e2e/adapters-e2e/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: /
--- a/e2e/adapters-e2e/src/components/click-me/assets/circle.svg
+++ b/e2e/adapters-e2e/src/components/click-me/assets/circle.svg
@@ -1,0 +1,3 @@
+<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
+  <circle cx="50" cy="50" r="50" />
+</svg>
--- a/e2e/adapters-e2e/src/components/click-me/click-me.tsx
+++ b/e2e/adapters-e2e/src/components/click-me/click-me.tsx
@@ -1,10 +1,12 @@
 import { component$, useSignal } from '@builder.io/qwik';
+import circleSvg from './assets/circle.svg?url';
 
 // We need to extract the component to see the bug on 1.5.7
 export default component$(() => {
   const isOpenSig = useSignal(false);
   return (
     <>
+      <img src={circleSvg} />
       <button
         onClick$={() => {
           return (isOpenSig.value = !isOpenSig.value);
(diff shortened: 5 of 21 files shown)


packages/qwik-city/src/buildtime/vite/plugin.ts
```diff
@@ -277,10 +277,9 @@ function qwikCityPlugin(userOpts?: QwikCityVitePluginOptions): any {
           const clientOutDir = qwikPlugin!.api.getClientOutDir();
 
           if (outDir && clientOutDir) {
-            const assetsDir = qwikPlugin!.api.getAssetsDir();
             const { staticPathsCode, notFoundPathsCode } = await postBuild(
               clientOutDir,
-              assetsDir ? join(api.getBasePathname(), assetsDir) : api.getBasePathname(),
+              api.getBasePathname(),
```
wmertens: @thejackshelton thoughts? Does Astro need this?

intellix:
I am noticing one thing locally and that's sounds imported via ?url suffix aren't working correctly but I think that's a Vite thing, even on the latest. I'll test with/without this fix as well just to be sure.

The mp3 is loaded relative to the component like so:
```
import click from './click.mp3?url';
```

On a local serve, it loads correctly. On preview/build, the file is sent to **/q/assets/AbC-dEf123-click.mp3** but the code tries to load it without the /q/ prefix.

intellix:
The mp3 issue isn't fixed by removing my patch. I guess it's unrelated as the assetsDir in my PR is only used for generating the staticPaths. I'm not sure if it's a vite or qwik thing but I'll investigate in a separate reproduction

packages/qwik-city/src/buildtime/vite/plugin.ts
intellix: I'm not even sure if this is the correct place it's happening. There's 3x vite plugins and I'm so confused

intellix: it was wrong, I've fixed it and tested more

wmertens (changes requested):
can you add a check to one or more of the /e2e tests that verifies that the staticpaths are correct?

Ideally, have the e2e tests build multiple times, with combinations of assetDir and base

intellix:
ok, I'm adding some e2e tests that verify a mixture of the vite configs and including everything that's causing/caused us problems to ensure it's fixed or at least exposed

intellix:
ok so this is basically testing and fixing a few things related to build.assetsDir. Today in v1 we've got this in our vite.config:
```ts
    build: {
      assetsDir: 'q',
      rollupOptions: {
        output: isSsrBuild
          ? {
              // qwik's optimizer skips assetsDir for SSR builds
              assetFileNames: 'q/assets/[hash]-[name].[ext]',
            }
          : {},
      },
    },
```

And it's testing 4x scenarios:
- vite with none of the below
- vite with a base
- vite with a base + build.assetsDir
- vite with just a build.assetsDir

I'm testing:
- the generated @qwik-city-static-paths is correct and not doubled
- that a public/robots.txt gets copied across properly
- a relative referenced SVG with ?url gets put into assets correctly and referenced properly in SSR (my assetFileNames was fixing that before)

I believe e2e worked, but something timed out and I don't think I can click retry.

I believe v2 also doesn't respect the assetsDir so similar tests could be used there?

maiieul:
Closing as stale. We removed `assetsDir` support in v2. Use `assetFileNames` and `chunkFileNames` until `assetsDir` is supported again 🙏 .

PR #8156, opened 2025-11-14
feat: reactive var props

Varixo:
Make props proxy more reactive by adding effects for var props like a store.

### DIFF
--- a/.changeset/dull-insects-slide.md
+++ b/.changeset/dull-insects-slide.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': minor
+---
+
+feat: make props more reactive for var props


packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1456,15 +1412,68 @@ function getComponentHash(vNode: VNode | null, getObject: (id: string) => any):
...
+  host: VirtualVNode,
+  jsxProps: PropsProxy,
+  vNodeProps: PropsProxy | null,
+  container: ClientContainer
+): boolean {
+  let shouldRender = false;
+  let propsAreDifferent = false;
+  if (vNodeProps) {
+    const effects = vNodeProps[_PROPS_HANDLER].$effects$;
+    const constPropsDifferent = handleChangedProps(
```
wmertens, line 1448: let's only do this if the key is different

packages/qwik/src/core/reactive-primitives/impl/store.ts
```diff
@@ -276,7 +277,7 @@ export function addStoreEffect(
   // to unsubscribe from. So we need to store the reference from the effect back
   // to this signal.
   ensureContainsBackRef(effectSubscription, target);
-  addQrlToSerializationCtx(effectSubscription, store.$container$);
+  addQrlToSerializationCtx(effectSubscription, '$container$' in store ? store.$container$ : null);
```
wmertens: add a comment here to say it's to avoid subscribing?

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -302,7 +302,12 @@ export async function serialize(serializationContext: SerializationContext): Pro
   const writeObjectValue = (value: {}) => {
     if (isPropsProxy(value)) {
       const owner = value[_OWNER];
-      output(TypeIds.PropsProxy, [_serializationWeakRef(owner), owner.varProps, owner.constProps]);
+      output(TypeIds.PropsProxy, [
+        _serializationWeakRef(owner),
```
wmertens, line 306: I think we should take this out, and instead add a `_serializationWeakRef(propsProxy)` to jsxNode, and do propsProxy.owner = jsxNode in inflate of jsxNode

packages/qwik/src/core/client/vnode-diff.ts
Varixo: but to know if key is changed we need to run it anyway

packages/qwik/src/core/shared/serdes/serialize.ts
Varixo: you cant have varprops and const props without owner, so how we can inflate propsproxy then?

wmertens (approved):
LGTM

packages/qwik/src/core/shared/serdes/serialize.ts
wmertens: hmm ok to fix this we have to move ownership of the props to propsproxy, as soon as it is created. But we shouldn't create it for every jsxnode so we must support both paths.

I think that can go in another PR.

PR #8160, opened 2025-11-15
fix: serialize correctly null or undefined value for signals

Varixo:
Correctly serialize null or undefined value.
Remove only undefined values during serialization.
Make default undefined value for effects and back refs instead of null

### DIFF
--- a/packages/qwik/src/core/shared/serdes/serdes.unit.ts
+++ b/packages/qwik/src/core/shared/serdes/serdes.unit.ts
@@ -16,7 +16,11 @@ import {
   isSignal,
 } from '../../reactive-primitives/signal.public';
 import { SubscriptionData } from '../../reactive-primitives/subscription-data';
-import { StoreFlags } from '../../reactive-primitives/types';
+import {
+  EffectProperty,
+  StoreFlags,
+  type EffectSubscription,
+} from '../../reactive-primitives/types';
 import { createResourceReturn } from '../../use/use-resource';
 import { Task } from '../../use/use-task';
 import { QError } from '../error/error';
@@ -350,7 +354,7 @@ describe('shared-serialization', () => {
           {number} 0
           {number} 0
           RootRef 1
-          Constant null
+          Constant undefined
           Object [
             {string} "shared"
             {number} 2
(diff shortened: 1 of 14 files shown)


packages/qwik/src/core/shared/serdes/serdes.unit.ts
```diff
@@ -471,18 +484,20 @@ describe('shared-serialization', () => {
         "
         0 ComputedSignal [
           RootRef 4
+          Constant undefined
```
wmertens: Strange that this undefined is in the array, shouldn't it be filtered out?

Varixo: Value should be always serialized for computed. If not, invalid flag (and symbol invalid) is set during resuming and some edge cases are not working

Varixo: Actually wait, maybe this is a bug

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -434,12 +440,25 @@ export async function serialize(serializationContext: SerializationContext): Pro
...
+             */
+            keepUndefined = true;
+          }
         }
-        output(isAsync ? TypeIds.AsyncComputedSignal : TypeIds.ComputedSignal, out);
+        output(isAsync ? TypeIds.AsyncComputedSignal : TypeIds.ComputedSignal, out, keepUndefined);
       } else {
-        output(TypeIds.Signal, [value.$untrackedValue$, ...(value.$effects$ || [])]);
+        const v = value.$untrackedValue$;
+        const keepUndefined = v === undefined;
```
wmertens, line 460: I don't think this is needed, [][0] is undefined

Varixo: it is needed, but you are right, its bugged. It should keep only value if effects are undefined. I pushed some changes

wmertens: I don't get it, if you don't keep undefined, it will be an empty array and it won't create the effects set?

Varixo: its for value, not effects

wmertens: oh right, missing value means invalid, not undefined

wmertens (approved):
LGTM

PR #8164, opened 2025-11-17
fix: execute cleanup cb for all component tree while calling dispose.cleanup method returned by render fn

sashkashishka:
# What is it?

- Bug

# Description

We have a use case when we render a qwik container client side only (while SPA navigation). We faced with the problem, when leaving the page and calling [dispose](https://qwik.dev/api/qwik/#renderresult) function, previously rendered components don't call cleanup function in useTasks at all. After some investigation we found that it only affects child components rendered in a Slot. So everything above the slot calls cleanup, but below it doesn't.

Here is an example:

```js
const Child = component$(() => {
  useVisibleTask$(async ({ cleanup }) => {
    console.log('mount Child');
    cleanup(() => console.log('unmount Child')); // 3
  });

  return (
    <div style={{ background: 'red', color: 'white' }}>
      Child component
    </div>
  );
});

const Provider = component$(() => {
  useTask$(({ cleanup }) => {
    console.log('mount Provider');
    cleanup(() => console.log('unmount Provider')); // 2
  });

  return (
    <Slot />
  );
});


const Root = component$(() => {
  useTask$(({ cleanup }) => {
    console.log('mount root');
    cleanup(() => console.log('unmount')); // 1
  });

  return (
    <Provider>
      <Child />
    </Provider>
  );
});

render(document.getElementById('root'), <Root />)
```

So when we dispose the component, it will only call 1 and 2 cleanup function, but not 3.

But actually, it should call all three cleanups.

# Checklist

Besided, on my local machine I faced with the failing e2e tests (it turned out that my laptop is too fast and test assertions was conducted before the js manages to render the content). Please, check out if the provided fixes works for you. if not, I'll revert them. Also, found some bugs with rerender test cases and fixed them (there were assertions to check whether the counter was incremented, but it hasn't managed to do that and the assertion looked like `expect(0).toBe(0)` instead of checking whether the increment has worked `expect(1).toBe(1)`. you can find it [here](https://github.com/sashkashishka/qwik/blob/c3e45961f165a6263e92f1701a283889a3f007c1/starters/e2e/e2e.style.spec.ts#L23)


wmertens:
@sashkashishka very nice! Can you check if this problem also exists in v2?

.changeset/bumpy-cycles-carry.md
```diff
@@ -0,0 +1,9 @@
+---
+'eslint-plugin-qwik': patch
```
gioboa, line 2: Thanks for your help @sashkashishka 
Some packages are not affected by this change though. 🤔

sashkashishka: only `eslint`? looks like these one are using qwik as dependency. I'll open a new PR with fix

```
'create-qwik': patch
'@builder.io/qwik-react': patch
'@builder.io/qwik-city': patch
'@builder.io/qwik': patch
```

sashkashishka:
> @sashkashishka very nice! Can you check if this problem also exists in v2?

sure, but at the end of the week

PR #8170, opened 2025-11-18
fix(qwikloader): don't trigger document and window events for normal events

Varixo:
Don't trigger document and window events for normal events


packages/qwik/src/core/tests/render-api.spec.tsx
```diff
@@ -521,7 +521,7 @@ describe('render api', () => {
           qwikLoader: 'module',
         });
         expect(result.html).toContain(
-          '(window.qwikevents||(window.qwikevents=[])).push("focus", "-my---custom", "click", "dblclick", "another-custom", "blur")'
+          '(window.qwikevents||(window.qwikevents=[])).push(":focus", ":-my---custom", ":click", ":dblclick", ":another-custom", ":blur")'
```
wmertens: Let's add document and windows events here too

packages/qwik/src/qwikloader.ts
```diff
@@ -207,11 +209,13 @@ const camelToKebab = (str: string) => str.replace(/([A-Z-])/g, (a) => '-' + a.to
...
  * @param ev - Browser event.
  */
-const processDocumentEvent = async (ev: Event) => {
+const processDocumentEvent = async (ev: Event, scope: string) => {
   // eslint-disable-next-line prefer-const
   let type = camelToKebab(ev.type);
   let element = ev.target as Element | null;
-  broadcast('-document', ev, type);
+  if (scope === documentPrefix) {
+    broadcast(documentPrefix, ev, type);
```
wmertens, line 217: Shouldn't it return after this? The element events are handled in another listener

wmertens (changes requested):

PR #8171, opened 2025-11-21
feat: used optimized method for setting changed attribute value

Varixo:
Skip some unnecessary work during setting attribute for vnode diff


wmertens (approved):
LGTM

PR #8181, opened 2025-11-25
feat: cursor based scheduling

Varixo:


packages/qwik/src/core/client/dom-container.ts
```diff
@@ -160,7 +137,24 @@ export class DomContainer extends _SharedContainer implements IClientContainer {
     }
   }
 
-  $setRawState$(id: number, vParent: ElementVNode | VirtualVNode): void {
+  /**
+   * The first time we render we need to hoist the styles. (Meaning we need to move all styles from
```
maiieul, line 141: Doesn't that break the benefits of html streaming? If you hoist the styles to the top, then all styles need to be downloaded before the html continues rendering. Whereas if only the styles that are needed for the top navbar are loaded then it can be rendered right away.

wmertens: What this does it take all the inline styles that were emitted by `use*Styles$` and move them into head, so that they don't interfere with DOM handling.
This is the way it is in v1 as well, and it should not have any effect on page loading.

wmertens (approved):
LGTM

shairez:
You're amazing @Varixo ! :heart:

PR #8188, opened 2025-11-28
fix: removing children from var props

Varixo:
We need to remove children prop instead of setting undefined, because then ownKeys trap needs to return unique keys


gioboa (approved):
LGTM 💪

PR #8193, opened 2025-12-04
feat(api): add QRLInternal type and update documentation

JerryWu1234:
Introduced the QRLInternal type alias in the API reference, enhancing the documentation with its definition and references. Updated internal exports to include QRLInternal, ensuring proper type visibility across the codebase.

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/packages/qwik/src/core/shared/qrl/qrl-class.ts
+++ b/packages/qwik/src/core/shared/qrl/qrl-class.ts
@@ -56,6 +56,7 @@ export type QRLInternalMethods<TYPE> = {
   $setContainer$(containerEl: Element | undefined): Element | undefined;
 };
 
+/** @internal */
 export type QRLInternal<TYPE = unknown> = QRL<TYPE> & QRLInternalMethods<TYPE>;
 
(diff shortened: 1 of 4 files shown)


wmertens (commented):
This really is an internal type, and because of the `$x$` the names get shortened too.
What do you want to access exactly?

JerryWu1234:
> This really is an internal type, and because of the `$x$` the names get shortened too.
> 
> What do you want to access exactly?

Devtool has to access the $x$ property. $chunk, $Qrl

packages/qwik/src/core/shared/qrl/qrl-class.ts
```diff
@@ -56,6 +56,7 @@ export type QRLInternalMethods<TYPE> = {
   $setContainer$(containerEl: Element | undefined): Element | undefined;
 };
 
+/** @public */
```
Varixo: its internal, so it should be `@internal`

JerryWu1234: done

PR #8103, opened 2025-10-30
chore: remove CJS/UMD builds

JerryWu1234:
…e. Remove CJS generation configuration one module at a time, which makes it easier to find bugs.

# What is it?

- Feature / enhancement
- Bug
- Docs / tests / types / typos
- Infra

# Description

# Checklist

- [ ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [ ] I performed a self-review of my own code
- [ ] I added a changeset with `pnpm change`
- [ ] I made corresponding changes to the Qwik docs
- [ ] I added new tests to cover the fix / functionality

PR #8196, opened 2025-12-09
fix(create-qwik): add missing execa dependency

JerryWu1234:
After PR #8103 migrated from CJS to ESM, the execa package was marked as external in the esbuild config but was not declared as a dependency in create-qwik's package.json.

This caused 'Cannot find package execa' error when running: npx create-qwik@2.0.0-beta.15

Fixes the runtime error by adding execa to dependencies.

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist

### DIFF
--- a/packages/create-qwik/package.json
+++ b/packages/create-qwik/package.json
@@ -5,6 +5,9 @@
   "author": "Qwik Team",
   "bin": "./create-qwik.mjs",
   "bugs": "https://github.com/QwikDev/qwik/issues",
+  "dependencies": {
+    "execa": "9.6.0"
+  },
   "devDependencies": {
     "@clack/prompts": "0.11.0",
     "@types/yargs": "17.0.33",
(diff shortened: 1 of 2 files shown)


JerryWu1234:
image

wmertens (changes requested):
The only place execa is used is in https://github.com/QwikDev/qwik/blob/build/v2/packages/qwik/src/cli/utils/run-build-command.ts

I think the correct fix is to mark execa as noExternal for qwik/create-qwik so that it is tree shaken and less deps need to be installed before creating a new qwik app.

JerryWu1234:
> The only place execa is used is in https://github.com/QwikDev/qwik/blob/build/v2/packages/qwik/src/cli/utils/run-build-command.ts
> 
> I think the correct fix is to mark execa as noExternal for qwik/create-qwik so that it is tree shaken and less deps need to be installed before creating a new qwik app.

Due to the CJS dynamic require() in execa's dependency chain, it cannot be bundled into an ESM bundle.
The only solution is:
Keep execa in the external list (don't bundle it)
Declare it as a dependency in package.json
This way, npm/pnpm will automatically install execa and its dependencies when users install create-qwik.

PR #8203, opened 2025-12-13
docs: add guide for finding leftover console.log statements

### LINKED ISSUE #7661: [📖] Please provide a section about what is your recommendation to find useless console.logs
### Suggestion

Using `console.log` seems to be common. Yet it has a downside. Sometimes it's left unintentionally in the code.
When the codebase becomes bigger, then a terminal is littered with logs, but the file and line of code that made that log are not apparent.

One technique that we use is to override `console.log` to include the stack trace too:

```
    const originalLog = console.log
    console.log = props => {
        const stack = new Error()
            .stack
            .split("\n")
            .filter(i => !/^.*(\.vite|keycloak|\.main\.jsx).*/.test(i))
        originalLog(stack, props)
    }
```

Try `console.log("something")` and you will see a good stack helping you find the place of the left log.

This works when we put it inside a component. But when we want to put it inside the `entry.express` or `entry.dev` or `entry.preview`, then it has no effect.

Please add a section to the documentation explaining your official way of finding useless `console.log` statements.

KyeongJooni:
Fixes #7661

# What is it?

# Description

Added a section to the debugging guide explaining how to find leftover console.log statements using stack traces. Includes examples for both client-side components and server-side entry files, plus alternative approaches with ESLint and IDE search.

# Checklist

### DIFF
--- a/packages/docs/src/routes/docs/(qwikcity)/guides/debugging/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikcity)/guides/debugging/index.mdx
@@ -5,7 +5,7 @@ contributors:
   - gioboa
   - KyeongJooni
 updated_at: '2025-12-13T00:00:00Z'
-created_at: '2025-12-13T00:00:00Z'
+created_at: '2025-08-08T00:00:00Z'
 ---
 # Debugging
 
@@ -15,6 +15,7 @@ During development, you can quickly access the source code of any element on a w
 click on whatever you're curious about, and Qwik will immediately open your source code right on top of that element.
 Seriously, any component, link, header - you name it, you can peek at its code.
 
+Qwik tries to guess what code editor you like to use.
 If it guesses wrong, no biggie! Just give Qwik a hint by setting the `LAUNCH_EDITOR` environment variable.
 For example, if you're on a Mac and use VS Code, you'd type `export LAUNCH_EDITOR=code` in your terminal.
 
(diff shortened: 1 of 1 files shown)


packages/docs/src/routes/docs/(qwikcity)/guides/debugging/index.mdx
```diff
@@ -19,3 +20,79 @@ If it guesses wrong, no biggie! Just give Qwik a hint by setting the `LAUNCH_EDI
...
+
+### Using stack traces in client-side components
+
+You can override `console.log` to include stack trace information in your component code. For example, in your root component or a layout component:
+
+```tsx
+// root.tsx or layout.tsx
+export default component$(() => {
+  // Override console.log to include stack traces
+  useVisibleTask$(() => {
```
gioboa, line 65: I personally don't like this approach with visible tasks because you need to remember and remove them too.eslint seems to be the best way to tackle the problem.

KyeongJooni: Thanks for the feedback @gioboa! 
Thanks for the feedback! I've moved ESLint to the top as the primary recommendation. I kept the stack trace approach at the bottom since the original issue mentioned it can be helpful for identifying existing logs in codebases that are already cluttered. Added a note to remove it after debugging. Happy to adjust further if needed!

PR #8218, opened 2025-12-31
fix: apply journal in correct order for some edge cases

Varixo:
In case when we mark vnode as dirty we need to check first if we already have parent cursor to merge


gioboa (approved):
👏 thanks @Varixo

PR #8219, opened 2026-01-01
feat(perf): hoist wrapped signal function

Varixo:
Hoist wrapped signal function. This is very usefull for loops. In that case we have the same signal function instance, so less memory and possibility to compare instances during diffing


wmertens (commented):
LGTM 🛞

PR #8222, opened 2026-01-03
feat(optimizer): move event conversion logic to the optimizer

Varixo:
Move event conversion logic to the optimizer for better performance.


wmertens (approved):
LGTM

PR #8223, opened 2026-01-06
feat(perf): qrl loop optimizations

Varixo:
This PR adds loops optimizations:
- add q:p or q:ps for every vnode which contains also event listener
- use q:p or q:ps instead of capture ref variable
- hoist event listener qrls inside loops


wmertens:
Such a great change!

Things to check:
- nested maps (at first they can deopt and later support an array of arguments)
- multiple handlers on the same tag (should just work, but let's be sure)

I think we should rename the `q:row` to `q:p` (shorter for serialization and more generic), and later on support `q:ps` for an array of params to pass to the handler.

Then the next step is to move all lexical scope out of handler QRLs and into `q:ps`, so that handler QRLs themselves have no lexical scope and can be hoisted to the top of the module. Different PR.

wmertens (approved):
LGTM

Issue #7754, opened 2025-07-25
[🐞] [v2] enabling SSG in the dev-server stops images from emitting


wmertens:
### Which component is affected?

Qwik Runtime

### Describe the bug

When adding the node adapter to the starters testing dev-server, the frontpage image tests fail, because the images generated by vite-imagetools are no longer under dist/assets

### Reproduction

https://github.com/QwikDev/qwik/blob/aad9ef68913a8db3e939b121cb6bcae8ee205c8f/starters/dev-server.ts#L153-L172

### Steps to reproduce

uncomment the lines and run `pnpm serve`, then go to the qwikrouter-test and notice the broken images.

Not sure if this happens in v1 as well.

### System Info

```shell
v2
```

gioboa:
I'll take this one, what's the expected behaviour?

wmertens:
@gioboa just run `pnpm serve` on build/v2 as-is and go to that page, you'll see the colored squares and in the console you'll see the assets/*webp files listed.

Then uncomment the two blocks and run `pnpm serve` again, the assets are no longer there.

Note that I already verified that the files are emitted with `this.emitFile` by vite-imagetools, so it's something else.

wmertens:
still happening :(

PR #8225, opened 2026-01-07
chore: some fixes

wmertens:
includes doubleclick fix


wmertens:
the reason the tests fail is because vite-imagetools emits two different sets of images in client and ssr builds (different hashes). Still investigating.

Maybe https://github.com/QwikDev/qwik/issues/7754 is the same issue

wmertens:
created https://github.com/JonasKruckenberg/imagetools/issues/852 for the imagetools issue

PR #8245, opened 2026-01-15
fix: bind:value and onInput$ on the same element part2

Varixo:
The part2 of #8240

### DIFF
--- a/packages/qwik/src/core/shared/jsx/jsx-internal.ts
+++ b/packages/qwik/src/core/shared/jsx/jsx-internal.ts
@@ -1,4 +1,3 @@
-import { untrack } from '../../use/use-core';
 import type { OnRenderFn } from '../component.public';
 import { createQRL } from '../qrl/qrl-class';
 import type { QRLInternal } from '../qrl/qrl-class';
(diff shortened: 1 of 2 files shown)


packages/qwik/src/core/shared/jsx/jsx-internal.ts
```diff
@@ -72,6 +72,8 @@ export const _jsxSplit = <T extends string | FunctionComponent<any>>(
 ): JSXNodeInternal<T> => {
   return untrack(() => {
```
wmertens: Didn't we decide we don't need the untrack?

packages/qwik/src/core/shared/jsx/jsx-internal.ts
```diff
@@ -89,54 +95,128 @@ export const _jsxSplit = <T extends string | FunctionComponent<any>>(
...
           const attr = jsxEventToHtmlAttribute(k);
           if (attr) {
-            if (!constProps || !_hasOwnProperty.call(constProps, k)) {
+            if (!varPropsCopied) {
+              varProps = { ...varProps };
+              varPropsCopied = true;
+            }
+            // Check if the converted attribute already exists in constProps
+            if (constProps && _hasOwnProperty.call(constProps, attr)) {
+              // Merge into constProps since it already has this event
```
wmertens: Actually, since it's a varprop it should move the constprop out of const and into var.

In fact, by doing
```tsx
const Foo = component$(() => {
  const sig = useSignal('hi')
  const doIt = Math.random()>0.5
  const Cmp = doIt ? 'input' : 'input'
  return <Cmp onInput$={() => console.log('hi')} bind:value={doIt ? sig : undefined} />
})
```
you will get `on:input` that is sometimes const and sometimes var :-/

So the problem is that because you specify both, the intention is to have both.

However, with `{...spread}` before, the const props should override the spread, and with it after, the props are already var, but also should be overridden.

Our problem is that `onX` and `on:x` and `bind:y` are aliases and we don't have strict rules to handle them.

I think the only sane thing to do is to not allow mixing. This means that the last one wins.

So:
```tsx
<div onInput$={1} on:input={2} /> // => const {"on:input": 2}
<div {...foo} onInput$={1} /> // => const {"on:input": 1} always, never in var
<div bind:value={maybe && sig} onInput$={3} /> // => var {value: sig.value} const {on:input=3} always
```

In that last case maybe we should show a warning in the optimizer and in jsxsplit.

So we need to read the props in object order, and when encountering an alias for prop x:
- if varProps and x is in constProps, ignore
- if not copying yet, recreate the props until that point (keep track of index)
- put the the real prop x into the copied props instead of the alias, and from then on keep copying into the new props.

That way, aliases can be overridden by object order.

wmertens: and since the only way to get constProps is when the optimizer sets them, we know that de-aliasing is already done, so we only need to process varProps

Varixo: I think its not intuitive. 
Users dont know that bind:value is covnerted to onInput and imo it should be merged with existing one

packages/qwik/src/core/shared/jsx/jsx-internal.ts
```diff
@@ -89,54 +97,74 @@ export const _jsxSplit = <T extends string | FunctionComponent<any>>(
...
-      }
-      if (constProps) {
-        if (_hasOwnProperty.call(constProps, BIND_CHECKED)) {
-          const value = constProps[BIND_CHECKED];
-          delete constProps[BIND_CHECKED];
-          if (value) {
-            constProps.checked = value;
-            constProps['on:input'] = createQRL(null, '_chk', _chk, null, null, [value]);
+
+        // Handle bind:* - only in varProps (constProps are transformed by Rust)
```
wmertens: Actually, bind: should not be transformed in Rust, because it needs to merge with on:input after all props are processed

Varixo: I think we currently need to be backward compatible, so the logic must stay as is for now - until changes in the future

packages/qwik/src/core/shared/jsx/jsx-split.unit.ts
wmertens, line 1: we also need test cases for
```ts
_jsxSplit('button', {onClick$: 1, on:click: 2}) // 2
_jsxSplit('button', {"bind:value": doBind && sig}, {onInput$: 2}) // [_val, 2] or 2 depending on doBind
```

PR #8253, opened 2026-01-19
Fix bun add command in documentation

mrbuk:
Fix command to add the qwik dependency with bun. Current command `bun run qwik add bun` is not producing any sensible output: 

```
% bun run qwik add bun
error: Script not found "qwik"
```

Instead using `bun add qwik` adds the dependency:

```
 % bun add qwik
bun add v1.3.6 (d530ed99)

installed qwik@1.0.57

1 package installed [296.00ms]
```
and creates a `package.json`

```
% cat package.json
{
  "dependencies": {
    "qwik": "^1.0.57"
  }
}%
```

As an enhancement - not implemented with this PR - one could consider using `bun create qwik` instead for the docs as this provides a way to create a full demo app that can be tested afterwards with `bun start`

# What is it?

# Description

# Checklist

- [X] I made corresponding changes to the Qwik docs

### DIFF
--- a/packages/docs/src/routes/docs/deployments/bun/index.mdx
+++ b/packages/docs/src/routes/docs/deployments/bun/index.mdx
@@ -25,7 +25,7 @@ To integrate the `bun` adapter, use the `add` command:
 - For the [integrated HTTP server](https://bun.sh/docs/api/http):
 
 ```shell
-bun run qwik add bun
+bun add qwik
 ```
 
 ## Production build


mrbuk:
@gioboa I think you have missed that there are two parts:

1) there is an error in the current file. If you run the command as defined in [index.mdx](https://github.com/QwikDev/qwik/blob/main/packages/docs/src/routes/docs/deployments/bun/index.mdx?plain=1#L28) you will get an error because this is wrong. This is what was changed

2) There was an "enhancement" question additionally, not directly related to the doc change. You seem to have closed this based on the fact that the you don't want the enhancement, which is fine but there is still an error addressed by the change that otherwise remains.

gioboa:
Thanks for your response @mrbuk 
`bun run qwik add bun` is adding the bun adapter to an already created project.

mrbuk:
Thanks for clarifying @gioboa. I have misunderstood the docs then. I was not aware that they implied that `qwik` is already installed in that case.

gioboa:
No problem at all, that's fine ☺️thanks for your commitment and support @mrbuk

PR #8254, opened 2026-01-21
fix: computed signal memory leak when reusing effect subscriber in loop

Varixo:
Clear back refs for effect subscriber when reusing in computed signal computation

### DIFF
--- a/packages/qwik/src/core/reactive-primitives/impl/computed-signal-impl.ts
+++ b/packages/qwik/src/core/reactive-primitives/impl/computed-signal-impl.ts
@@ -12,6 +12,7 @@ import { SignalImpl } from './signal-impl';
 import type { QRLInternal } from '../../shared/qrl/qrl-class';
 import { _EFFECT_BACK_REF, type BackRef } from '../backref';
 import { isDev } from '@qwik.dev/core/build';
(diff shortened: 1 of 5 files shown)


gioboa (approved):
LGTM 🚀

packages/qwik/src/core/reactive-primitives/impl/computed-signal-impl.ts
```diff
@@ -90,7 +90,12 @@ export class ComputedSignalImpl<T, S extends QRLInternal = ComputeQRL<T>>
 
     const ctx = tryGetInvokeContext();
     const previousEffectSubscription = ctx?.$effectSubscriber$;
-    ctx && (ctx.$effectSubscriber$ = getSubscriber(this, EffectProperty.VNODE));
+    if (ctx) {
+      const effectSubscriber = getSubscriber(this, EffectProperty.VNODE);
+      // clear existing back refs
+      effectSubscriber.backRef?.clear();
```
wmertens: - I think the backrefs can be cleared as soon as the signal is marked dirty, because nothing else can make it dirtier
- But most importantly, when clearing backrefs, each signal must also have this subscriber removed

PR #8263, opened 2026-01-24
fix: rendering var prop svg attributes with correct namespaces

Varixo:
Set var props attributes the same way as we do it for const props. We need to sometimes set namespace for svg's.

### DIFF
--- a/packages/qwik/src/core/shared/utils/attribute.ts
+++ b/packages/qwik/src/core/shared/utils/attribute.ts
@@ -1,0 +1,43 @@
+import { getAttributeNamespace } from '../../client/vnode-namespace';
+
+let _setAttribute: typeof Element.prototype.setAttribute | null = null;
+
+const fastSetAttribute = (target: Element, name: string, value: string): void => {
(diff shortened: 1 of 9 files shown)


jesperp:
Can confirm that this fixed my issue of using xlink namespace. Thanks for such a swift resolution 🙏🏽

packages/qwik/src/core/shared/utils/attribute.ts
```diff
@@ -0,0 +1,29 @@
...
+  attrName: string,
+  attrValue: any,
+  isSvg: boolean
+): void {
+  if (attrValue != null) {
+    if (isSvg) {
+      // only svg elements can have namespace attributes
+      const namespace = getAttributeNamespace(attrName);
+      if (namespace) {
+        element.setAttributeNS(namespace, attrName, attrValue);
```
wmertens: Why not have a fast version of this too?

Varixo: done

PR #8264, opened 2026-01-25
useTask fix

gnemanja:
These changes made by Claude fixed a couple of issues in our app.

First is the one you guys also solved today where async useTask changing an array wouldn't render it on ssr.

Other issues are mostly complicated to explain and vary. 
For example, on navigation, some component would be rendered twice, once in an incomplete state, the other would be as expected, fully rendered. No errors would be thrown.
When the same page is opened again (by navigating away from it and again to it) it would render only once, normally.

I think most of the issues are coming from using Slots.
Qwik team's changes from few days ago regarding cursor and projections did help a bit, but still some issues remain without this PR's changes.

Hopefully this helps!

### DIFF
--- a/packages/qwik/src/core/use/use-task.ts
+++ b/packages/qwik/src/core/use/use-task.ts
@@ -12,7 +12,7 @@ import { isPromise, maybeThen, safeCall } from '../shared/utils/promises';
 import { type ValueOrPromise } from '../shared/utils/types';
 import { ChoreBits } from '../shared/vnode/enums/chore-bits.enum';
 import { markVNodeDirty } from '../shared/vnode/vnode-dirty';
-import { newInvokeContext } from './use-core';
+import { invoke, newInvokeContext } from './use-core';
 import { useLexicalScope } from './use-lexical-scope.public';
 import type { ResourceReturnInternal } from './use-resource';
 import { useSequentialScope } from './use-sequential-scope';
@@ -173,6 +173,11 @@ export const useTaskQrl = (qrl: QRL<TaskFn>, opts?: TaskOptions): void => {
   const { $waitOn$: waitOn } = iCtx;
   const result = maybeThen(waitOn, () => runTask(task, container, iCtx.$hostElement$));
   if (isPromise(result)) {
+    // For render-blocking tasks, throw the Promise to trigger component retry
+    // This ensures the component re-renders after async tasks complete
+    if (task.$flags$ & TaskFlags.RENDER_BLOCKING) {
+      throw result;
+    }
     iCtx.$waitOn$ = result;
   }
 };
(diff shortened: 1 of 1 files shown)


Varixo:
Hey, thanks for this but bug reproduction is needed!

gnemanja:
@Varixo yeah, I understand
Spent hours trying, didn't manage to reproduce in isolation

packages/qwik/src/core/use/use-task.ts
```diff
@@ -192,8 +197,9 @@ export const runTask = (
   const [cleanup] = cleanupFn(task, (reason: unknown) => container.handleError(reason, host));
 
   const taskApi: TaskCtx = { track, cleanup };
-  return safeCall(
-    () => taskFn(taskApi),
+  const result = safeCall(
+    // Wrap task execution in invoke context to preserve context across async boundaries
+    () => invoke(iCtx, () => taskFn(taskApi)),
```
wmertens, line 202: This is not needed because ictx is already bound in getFn.

If this is needed then there's actually an issue somewhere

packages/qwik/src/core/use/use-task.ts
```diff
@@ -204,6 +210,13 @@ export const runTask = (
       }
     }
   );
+
+  // If task is async, wrap the Promise continuation to maintain invoke context
+  if (isPromise(result)) {
+    return result.then((value) => invoke(iCtx, () => value));
```
wmertens, line 216: This is also not needed because value is not a function

packages/qwik/src/core/use/use-task.ts
```diff
@@ -173,6 +173,11 @@ export const useTaskQrl = (qrl: QRL<TaskFn>, opts?: TaskOptions): void => {
   const { $waitOn$: waitOn } = iCtx;
   const result = maybeThen(waitOn, () => runTask(task, container, iCtx.$hostElement$));
   if (isPromise(result)) {
+    // For render-blocking tasks, throw the Promise to trigger component retry
+    // This ensures the component re-renders after async tasks complete
+    if (task.$flags$ & TaskFlags.RENDER_BLOCKING) {
+      throw result;
```
wmertens, line 179: This might be the fix of your issues but I'd prefer to fix it without throwing. Possibly the waiting is in the wrong order? 

We need a test that fails without this

gnemanja: I'll try to reproduce it :/

Varixo: yeah we removed that behavior. Are you sure nightly v2 version didnt fix your issue?

gnemanja: This fix was done for beta 17. But I copy pasted it in the build/v2 today and it also worked. But have in mind that some of the things this fixes were fixed by last changes on build/v2 (so now this is less useful, but still relevant)

gnemanja:
@wmertens @Varixo 
Okey I managed to reproduce some version of it :)
https://github.com/QwikDev/qwik/issues/8266

Please let me know if you plan to work on this since its a big blocker to upgrade to new version :)

Varixo:
Closing, the #8270 and #8267 fixed it

PR #8277, opened 2026-01-28
fix: wait for paused cursors before container drain completes

wmertens:
- use async qrls in tests
- wait for paused cursors before container drain completes


Varixo (approved):
LGTM

PR #8286, opened 2026-02-01
fix: over-cleaning store subscriptions

Varixo:
We shouldnt cleanup all effects and hostelement every time for wrappedsignal


wmertens (approved):
LGTM

PR #8292, opened 2026-02-03
refactor(core): qwikloader and handlers

wmertens:
- support adding containers at runtime by dispatching the qidle/qinit events and subscribing qvisible
- rename the internal handler attributes `on:click` etc to `e:click` (element), `d:click` (document) etc
- allow running v1 qwikloader when you need to support a v1 container. The qwikloaders will not interfere with eachother
- code improvments

### DIFF
--- a/packages/qwik/src/core/qwik.core.api.md
+++ b/packages/qwik/src/core/qwik.core.api.md
@@ -1954,7 +1954,7 @@ export function _vnode_toString(this: _VNode | null, depth?: number, offset?: st
 // @internal
 export const enum _VNodeFlags {
     // (undocumented)
-    Cursor = 64,
+    Cursor = 256,
     // (undocumented)
     Deleted = 32,
     // (undocumented)
@@ -1964,19 +1964,23 @@ export const enum _VNodeFlags {
     // (undocumented)
     ELEMENT_OR_VIRTUAL_MASK = 3,
     // (undocumented)
+    HasIterationItems = 64,
+    // (undocumented)
     Inflated = 8,
     // (undocumented)
     INFLATED_TYPE_MASK = 15,
     // (undocumented)
-    NAMESPACE_MASK = 384,
+    InflatedIterationItems = 128,
     // (undocumented)
-    NEGATED_NAMESPACE_MASK = -385,
+    NAMESPACE_MASK = 1536,
+    // (undocumented)
+    NEGATED_NAMESPACE_MASK = -1537,
     // (undocumented)
     NS_html = 0,
     // (undocumented)
-    NS_math = 256,
+    NS_math = 1024,
     // (undocumented)
-    NS_svg = 128,
+    NS_svg = 512,
     // (undocumented)
     Resolved = 16,
     // (undocumented)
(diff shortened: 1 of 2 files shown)


packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -792,46 +786,29 @@ function expectNoMoreTextNodes(diffContext: DiffContext) {
...
-            registerQwikLoaderEvent(diffContext, loaderScopedEvent);
-          }
-        }
+        value = registerEventHandlers(
+          key,
+          value,
+          element,
+          diffContext.vNewNode as ElementVNode,
+          diffContext
+        );
```
Varixo: the registerEventHandlers function is not returning anything, am I missing somthing?

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -447,36 +449,58 @@ export const vnode_ensureElementKeyInflated = (vnode: ElementVNode) => {
...
+        const value = attr.value;
+        const scopedKebabName = key.slice(2);
+        const qrls = value.split('|');
+        const handlers = qrls.map((qrl) => {
+          const handler = parseQRL(qrl);
+          handler.$container$ = container;
+          // These QRLs are mostly _run and _task and don't need wrapping with retryOnPromise
+          return handler;
+        });
+        (element._qDispatch ||= {})[scopedKebabName] = handlers;
```
Varixo: maybe we can move it to separate function for clarity?

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -447,36 +449,58 @@ export const vnode_ensureElementKeyInflated = (vnode: ElementVNode) => {
...
           vnode_setProp(elementVNode, 'value', element.value);
         }
-      } else if (!key.startsWith(EventNameHtmlScope.on)) {
+      } else {
         const value = attr.value;
         vnode_setProp(elementVNode, key, value);
       }
     }
+    // In tests we might not have a container
+    if (container) {
```
Varixo: if its only for tests we should add `import.meta.env.TEST` condition or true always

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -447,36 +449,58 @@ export const vnode_ensureElementKeyInflated = (vnode: ElementVNode) => {
...
+      // We need to deserialize single iteration items here, once, because they can be strings
+      vnode_getProp<unknown>(elementVNode, ITERATION_ITEM_SINGLE, container.$getObjectById$);
```
Varixo: cant we do it when running qrl?

packages/qwik/src/core/client/vnode-diff.ts
wmertens: oh right, Copilot ;)

packages/qwik/src/core/client/vnode-utils.ts
wmertens: no, because once it's serialized it could be a string and then we can't know if it needs deserializing.
But we could add a flag to the VNode, checking if it has iteration items, and also if they were deserialized already :thinking:

wmertens: added 2 flags, one for "has them" and one for "did inflate"

PR #8297, opened 2026-02-04
v2: useAsyncComputed$ => useAsync$

wmertens:
this rename to useAsync$ is because it's mostly used to fetch information, not calculate things.

This PR also includes some performance improvements to useConstant() and untrack() use.

There are some useful options planned too but not implemented yet.


.changeset/free-wasps-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': major
+---
+
+BREAKING (if you used previous betas): `useAsyncComputed` is renamed to `useAsync`. This reflects the many uses for it, not just computing.
```
Varixo: missing $ at the end of names

packages/qwik/src/core/shared/error/error.ts
```diff
@@ -90,7 +90,6 @@ export const enum QError {
   materializeVNodeDataError = 27,
   cannotCoerceSignal = 28,
   computedNotSync = 29,
-  computedReadOnly = 30,
```
Varixo, line 93: we should leave it as UNUSED_30 = 30

PR #8301, opened 2026-02-04
feat(useAsync$): implement RFC 324

wmertens:
This implements the [`useAsync$` RFC](https://github.com/QwikDev/qwik-evolution/issues/324)

- deprecates `useResource$` and removes it from docs
- adds `interval` to useAsync$ options to automatically rerun every x ms
- you can change `signal.interval`, even from within the compute function (but you will have to have a [self reference](https://github.com/QwikDev/qwik-evolution/issues/293#issuecomment-3854467524)), and it will do the right thing
- when you set pollMs during SSR, it will automatically resume on the client, as part of qidle
- refactors the async signal class to simplify
- the default serialization strategy for `useAsync$` is now 'always' instead of 'never', because it is likely to be expensive to get
- `concurrency`, `abort` and `timeout` support

BREAKING
- the `.promise()` method on `useAsync$` now returns a `Promise<void>` instead of `Promise<T>`, to avoid having to put `.catch()` on every call and to promote using the reactive `result.value` and `result.error` properties for handling async results and errors.

### DIFF
--- a/packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
+++ b/packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
@@ -150,6 +150,7 @@ export class AsyncSignalImpl<T>
   }
 
   get untrackedLoading() {
+    // reading `.loading` means someone is interested in the result, so we should trigger the computation. The alternative is eager computation or imperative calls to invalidate; this seems nicer.
     this.$computeIfNeeded$();
     // During SSR there's no such thing as loading state, we must render complete results
     if ((import.meta.env.TEST ? isServerPlatform() : isServer) && this.$current$?.$promise$) {
@@ -217,12 +218,13 @@ export class AsyncSignalImpl<T>
       if (!this.$hasSubscribers$()) {
         this.abort();
       }
-    }, 0);
+    });
   }
 
   /** Returns a promise resolves when the signal finished computing. */
   async promise(): Promise<void> {
     this.$computeIfNeeded$();
+    // Wait for the current computation to finish, but if we became invalid while running, we need to wait for the new computation instead. So we loop until we are no longer invalid
     while (this.$current$?.$promise$) {
       await this.$current$?.$promise$;
     }
--- a/packages/qwik/src/core/shared/serdes/serialize.ts
+++ b/packages/qwik/src/core/shared/serdes/serialize.ts
@@ -409,20 +409,12 @@ export async function serialize(serializationContext: SerializationContext): Pro
           value.$flags$ & SerializationSignalFlags.SERIALIZATION_STRATEGY_NEVER;
         const isInvalid = value.$flags$ & SignalFlags.INVALID;
         const isSkippable = fastSkipSerialize(value.$untrackedValue$);
-        const interval =
-          value instanceof AsyncSignalImpl && value.$interval$ > 0 ? value.$interval$ : undefined;
-        const concurrency =
-          value instanceof AsyncSignalImpl && value.$concurrency$ !== 1
-            ? value.$concurrency$
-            : undefined;
-        const timeout =
-          value instanceof AsyncSignalImpl && value.$timeoutMs$ !== 0
-            ? value.$timeoutMs$
-            : undefined;
+        const isAsync = value instanceof AsyncSignalImpl;
+        const interval = isAsync && value.$interval$ > 0 ? value.$interval$ : undefined;
+        const concurrency = isAsync && value.$concurrency$ !== 1 ? value.$concurrency$ : undefined;
+        const timeout = isAsync && value.$timeoutMs$ !== 0 ? value.$timeoutMs$ : undefined;
         const eagerCleanup =
-          value instanceof AsyncSignalImpl && value.$flags$ & AsyncSignalFlags.EAGER_CLEANUP
-            ? true
-            : undefined;
+          isAsync && value.$flags$ & AsyncSignalFlags.EAGER_CLEANUP ? true : undefined;
 
         if (isInvalid || isSkippable) {
           v = NEEDS_COMPUTATION;
@@ -437,7 +429,6 @@ export async function serialize(serializationContext: SerializationContext): Pro
           filterEffectBackRefs(value[_EFFECT_BACK_REF]),
           value.$effects$,
         ];
-        const isAsync = value instanceof AsyncSignalImpl;
         if (isAsync) {
           // After SSR, the signal is never loading, so no need to send it
           out.push(value.$loadingEffects$, value.$errorEffects$, value.$untrackedError$);
--- a/packages/qwik/src/core/tests/use-async.spec.tsx
+++ b/packages/qwik/src/core/tests/use-async.spec.tsx
@@ -417,7 +417,7 @@ describe.each([
       await trigger(container.element, 'button', 'click');
       // on server after resuming cleanup is not called yet
       // on client it is called as usual
-      // so now the log is equal
+      // so from this point the log is equal for ssr and client
       expect((globalThis as any).log).toEqual(['cleanup']);
       await trigger(container.element, 'button', 'click'); //show
       await trigger(container.element, 'button', 'click'); //hide
--- a/packages/qwik/src/server/ssr-container.ts
+++ b/packages/qwik/src/server/ssr-container.ts
@@ -901,21 +901,27 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
     if (!this.serializationCtx.$roots$.length) {
       return;
     }
-    const attrs: string[] = ['type', 'qwik/state'];
 
-    // Add q-d:qidle attribute if there are signals to eagerly resume
-    if (this.serializationCtx.$eagerResume$.size > 0) {
-      const qrl = createQRL(null, '_res', _res, null, [...this.serializationCtx.$eagerResume$]);
-      const qrlStr = qrlToString(this.serializationCtx, qrl);
-      attrs.push('q-d:qidle', qrlStr);
-      // Add 'd:qidle' to event names set
-      this.serializationCtx.$eventNames$.add('d:qidle');
-    }
+    const attrs = this.stateScriptAttrs();
 
     this.openElement('script', null, attrs);
     return maybeThen(this.serializationCtx.$serialize$(), () => {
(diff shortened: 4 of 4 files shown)


gioboa (approved):

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
```diff
@@ -94,7 +100,27 @@ export class AsyncSignalImpl<T> extends ComputedSignalImpl<T, AsyncQRL<T>> imple
...
+      this.$scheduleNextPoll$();
+    }
+  }
+
   override invalidate() {
+    // clear the poll timeout before invalidating
+    if (this.$pollTimeoutId$ !== undefined) {
+      clearTimeout(this.$pollTimeoutId$);
+      this.$pollTimeoutId$ = undefined;
+    }
```
Varixo, line 199: maybe separate helper function? the code is duplicated now

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
```diff
@@ -226,4 +217,23 @@ export class AsyncSignalImpl<T> extends ComputedSignalImpl<T, AsyncQRL<T>> imple
     }
     return didChange;
   }
+
+  private $scheduleNextPoll$() {
+    if (
+      (isBrowser || (import.meta.env.TEST && isDomContainer(this.$container$))) &&
```
Varixo: in other places for test I check isServerPlatform, I wonder if we want to keep the same approach across the codebase

packages/qwik/src/server/ssr-container.ts
```diff
@@ -898,7 +901,18 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
...
+    const attrs: string[] = ['type', 'qwik/state'];
+
+    // Add q-d:qidle attribute if there are signals to eagerly resume
+    if (this.serializationCtx.$eagerResume$.size > 0) {
+      const qrl = createQRL(null, '_res', _res, null, [...this.serializationCtx.$eagerResume$]);
+      const qrlStr = qrlToString(this.serializationCtx, qrl);
+      attrs.push('q-d:qidle', qrlStr);
+      // Add 'd:qidle' to event names set
+      this.serializationCtx.$eventNames$.add('d:qidle');
+    }
```
Varixo: hmm isnt better to add qidle in a hook like visible task?

wmertens: We only know what signals to resume after SSR completes, so we have no place to put the hook

packages/qwik/handlers.mjs
```diff
@@ -6,4 +6,4 @@
  *
  * Make sure that these handlers are listed in manifest.ts
  */
-export { _chk, _run, _task, _val } from '@qwik.dev/core';
+export { _chk, _res, _run, _task, _val } from '@qwik.dev/core';
```
wmertens:
```suggestion
export { _chk, _rsc, _res, _run, _task, _val } from '@qwik.dev/core';
```

.changeset/light-eels-lead.md
```diff
@@ -0,0 +1,6 @@
+---
+'@qwik.dev/core': major
```
Varixo, line 2: I think its not major change? its just minor?

.changeset/rich-peas-invite.md
```diff
@@ -0,0 +1,12 @@
...
+  In-flight invocations will update the signal value only if they complete before a newer invocation completes. For example, if you have a search input that triggers a new `useAsync$` execution on every keystroke, results will show in the correct order.
+
+- FEAT: `useAsync$()` now has an `abort()` method, which aborts the current computation and runs cleanups if needed. This allows you to cancel long-running tasks when they are no longer needed, such as when a component unmounts or when a new computation starts. The compute function needs to use the `abortSignal` provided to handle aborts gracefully.
+  When a new computation starts, the previous computation will be aborted via the abortSignal. This allows you to prevent unnecessary work and ensure that only the latest computation is active. For example, if you have a search input that triggers a new `useAsync$` execution on every keystroke, the previous search will be aborted when a new one starts, ensuring that only the latest search is performed.
```
Varixo, line 12: I think every feat should be in separate file

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
```diff
@@ -71,16 +150,18 @@ export class AsyncSignalImpl<T> extends ComputedSignalImpl<T, AsyncQRL<T>> imple
   }
 
   get untrackedLoading() {
+    this.$computeIfNeeded$();
```
Varixo, line 154: maybe we should add comment why we start computation here. At first glance it looks like a bug

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
```diff
@@ -94,136 +175,245 @@ export class AsyncSignalImpl<T> extends ComputedSignalImpl<T, AsyncQRL<T>> imple
...
+  }
+
+  /** Invalidates the signal, causing it to re-compute its value. */
+  override async invalidate() {
+    this.$flags$ |= SignalFlags.INVALID;
+    this.$clearNextPoll$();
+    if (this.$effects$?.size || this.$loadingEffects$?.size || this.$errorEffects$?.size) {
+      // compute in next microtask
+      await true;
+      this.$computeIfNeeded$();
```
Varixo, line 198: why not use queueMicrotask?

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
```diff
@@ -94,136 +175,245 @@ export class AsyncSignalImpl<T> extends ComputedSignalImpl<T, AsyncQRL<T>> imple
...
+      return;
+    }
+    if (!(import.meta.env.TEST ? !isServerPlatform() : isBrowser)) {
+      return;
+    }
+    setTimeout(() => {
+      if (!this.$hasSubscribers$()) {
+        this.abort();
+      }
+    }, 0);
```
Varixo: 0 is default, no?

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
```diff
@@ -94,136 +175,245 @@ export class AsyncSignalImpl<T> extends ComputedSignalImpl<T, AsyncQRL<T>> imple
...
+  }
+
+  /** Returns a promise resolves when the signal finished computing. */
+  async promise(): Promise<void> {
+    this.$computeIfNeeded$();
+    while (this.$current$?.$promise$) {
```
Varixo, line 228: can we add comment, why while loop is needed here?

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -424,13 +409,27 @@ export async function serialize(serializationContext: SerializationContext): Pro
           value.$flags$ & SerializationSignalFlags.SERIALIZATION_STRATEGY_NEVER;
         const isInvalid = value.$flags$ & SignalFlags.INVALID;
         const isSkippable = fastSkipSerialize(value.$untrackedValue$);
-
-        if (shouldAlwaysSerialize) {
+        const interval =
+          value instanceof AsyncSignalImpl && value.$interval$ > 0 ? value.$interval$ : undefined;
```
Varixo: its better to check instance once

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -440,20 +439,21 @@ export async function serialize(serializationContext: SerializationContext): Pro
         ];
         const isAsync = value instanceof AsyncSignalImpl;
```
Varixo: we already do that here, so we can just move this higher

packages/qwik/src/core/tests/use-async.spec.tsx
```diff
@@ -358,9 +417,8 @@ describe.each([
       await trigger(container.element, 'button', 'click');
       // on server after resuming cleanup is not called yet
       // on client it is called as usual
-      expect((globalThis as any).log).toEqual(
-        render === ssrRenderToDom ? ['cleanup'] : ['cleanup']
-      );
+      // so now the log is equal
```
Varixo: the comment makes no sense to me now

packages/qwik/src/core/use/utils/destroyable.ts
```diff
@@ -1,16 +1,14 @@
...
-  if (destroy) {
-    destroyable.$destroy$ = null;
+  if (destroyable.$destroy$) {
     try {
-      destroy();
+      destroyable.$destroy$();
     } catch (err) {
       logError(err);
     }
+    destroyable.$destroy$ = null;
```
Varixo, line 12: shouldnt be in finally block?

packages/qwik/src/server/ssr-container.ts
Varixo: maybe extract it to separate function for clarity?

.changeset/light-eels-lead.md
wmertens: it's a breaking API change vs previous betas so I made it major, still ends up being 2 so that's ok

.changeset/rich-peas-invite.md
wmertens: it's ok - formatted like this the changelog looks correct, and we often adjust manually anyway

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
wmertens: I couldn't remember the name, and now that I think about it, doesn't that use extra memory for creating the callback?

packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
wmertens: turns out the default is `1` but it has the same effect :) I removed it

packages/qwik/src/core/use/utils/destroyable.ts
wmertens: both code paths end up here

PR #8316, opened 2026-02-09
ci: add AI-powered issue triage via OpenCode

maiieul:
# What is it?

- Infra

# Description

Adds a GitHub Action that auto-labels new issues using Kimi K2.5 via OpenCode Zen. Applies type, component, and status labels. Skips accounts younger than 30 days. Chains with existing labeling-issues.yml for auto-comments on needs-reproduction/missing-info.


wmertens (approved):
LGTM

PR #8330, opened 2026-02-10
fix(ci): add missing permissions for OpenCode triage action

maiieul:
# What is it?
- Bug

# Description
OpenCode needs `contents: write` and `pull-requests: write` permissions to apply labels. Without them it falls back to commenting with the suggested labels instead.


gioboa (approved):
🤞

PR #8340, opened 2026-02-10
fix(cli): bundle execa instead of leaving it as external dependency

### LINKED ISSUE #8311: [🐞] v2: cli - `execa` is not in the dependencies

Starters / CLI

### Describe the bug

Version : 2.0.0-beta.21

`dist/cli.mjs` (`packages/qwik/src/cli/utils/utils.ts`) does `import { execa } from "execa"`   but it's not in the dependencies of @qwik.dev/core

### Reproduction

https://github.com/QwikDev/qwik/blob/c008e88d39ac28a53a291104cc8bdc1c77adc559/packages/qwik/package.json

### Steps to reproduce

_No response_

### System Info

```shell
/
```

### Additional Information

_No response_

maiieul:
# What is it?
- Bug

# Description
Remove `execa` from the esbuild external list so it gets inlined into the CLI bundle, and add a `createRequire` banner for CJS interop. Fixes #8311.

### DIFF
--- a/scripts/submodule-cli.ts
+++ b/scripts/submodule-cli.ts
@@ -16,8 +16,13 @@ export async function submoduleCli(config: BuildConfig) {
     target: nodeTarget,
     sourcemap: false,
     bundle: true,
-    banner: { js: getBanner('@qwik.dev/core/cli', config.distVersion) },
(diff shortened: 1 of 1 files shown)


gioboa:
Can you add the changeset please?

PR #8353, opened 2026-02-14
perf(ssr): improve SSR performance

Varixo:
- faster attributes processing
- optimize write stream function
- optimize flush/start/end block commands
- remove snapshotresult
- ssr jsx stack walking
- improve closeElement function
- using class for serialization instead of outer and inner functions


wmertens (approved):
LGTM

PR #8360, opened 2026-02-16
fix(build): ensure that bindings map is correct

wmertens:
currently qwik is always using the wasm build


wmertens:
I verified it to be correct now

PR #8361, opened 2026-02-16
fix: serializing falsy event handlers

Varixo:
Correctly serialize event handlers with falsy value

### DIFF
--- a/packages/qwik/src/core/ssr/ssr-events.ts
+++ b/packages/qwik/src/core/ssr/ssr-events.ts
@@ -49,8 +49,6 @@ export function setEvent(
   } else if (isQrl(qrls)) {
     value = getQrlString(qrls);
     addQwikEventToSerializationContext(serializationCtx, key, qrls);
-  } else if (typeof qrls === 'string') {
-    value = qrls;
   }
 
   return value;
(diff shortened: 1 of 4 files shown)


packages/qwik/src/server/ssr-container.ts
```diff
@@ -1050,8 +1050,12 @@ class SSRContainer extends _SharedContainer implements ISSRContainer {
       if (this.qlInclude !== QwikLoaderInclude.Done) {
         this.emitQwikLoaderInline();
       }
-      // emit the used events so the loader can subscribe to them
-      this.emitQwikEvents(Array.from(this.serializationCtx.$eventNames$, (s) => JSON.stringify(s)));
+      if (this.renderOptions.qwikLoader !== 'never') {
```
wmertens: I think we want to show this always because you could have multiple containers on a page and only one would add the qwikloader

packages/qwik/src/core/ssr/ssr-events.ts
```diff
@@ -49,6 +49,8 @@ export function setEvent(
   } else if (isQrl(qrls)) {
     value = getQrlString(qrls);
     addQwikEventToSerializationContext(serializationCtx, key, qrls);
+  } else if (typeof qrls === 'string') {
```
wmertens: Could you add a comment explaining when this is possible?

Varixo: actually I removed that, it was for q-d:idle, but now we can serialize it in write attrs instead of custom function

wmertens (approved):
LGTM

PR #8368, opened 2026-02-17
switch to npm:rolldown-vite@7.3.1

makinghappen:
# What is it?

- Feature / enhancement

# Description

Rolldown improves build speed by 25%.

# Checklist

- [ x] I added a changeset with `pnpm change`

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "syncpack": {
       "versionGroups": [
         {
-          "label": "Be lenient in vite versions for prod. v4 is broken, v5 is good",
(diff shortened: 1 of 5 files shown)


wmertens:
@makinghappen you should do this with the build/v2 branch, but even then, vite 8 will have rolldown, isn't there a beta we can use?

makinghappen:
yes, we would better use beta, I have found that rolldown-vite@7.3.1 has some degradation, while 8.0.0-beta.14 works well, I will create a new PR in v2 branch

PR #8379, opened 2026-02-23
feat: Vite 8 / rolldown support

maiieul:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


packages/qwik-router/src/buildtime/vite/get-route-imports.ts
```diff
@@ -44,6 +44,9 @@ function isBundlePartOfRoute(bundle: QwikBundle, routeAndLayoutPaths: string[])
   }
   for (const bundleOrigin of bundle.origins) {
     const originPath = removeExtension(bundleOrigin);
-    return routeAndLayoutPaths.some((path) => path.endsWith(originPath));
+    if (routeAndLayoutPaths.some((path) => path.endsWith(originPath))) {
+      return true;
+    }
```
maiieul: This now properly adds the route path to the bundle graph

packages/qwik/src/core/preloader/queue.ts
```diff
@@ -203,8 +203,7 @@ export const adjustProbabilities = (
        * too.
        */
       let newInverseProbability: number;
-      if (probability === 1 || (probability >= 0.99 && depsCount < 100)) {
-        depsCount++;
+      if (probability === 1 || probability >= 0.99) {
```
maiieul: Removed the depsCount check. In case the route was >100 bundles to load, it would then take forever instead of just a bit more time. The proper fix is to fallback to MPA when it takes too long.

packages/qwik-vite/src/plugins/bundle-graph.ts
```diff
@@ -141,13 +125,9 @@ export function convertManifestToBundleGraph(
     const bundle = graph[bundleName];
     // external dependencies are not included in `graph`
     const deps = new Set(bundle.imports!);
-    for (const depName of deps) {
-      clearTransitiveDeps(deps, depName);
-    }
     const dynDeps = new Set(bundle.dynamicImports!);
     const depProbability = new Map<string, number>();
     for (const depName of dynDeps) {
-      clearTransitiveDeps(dynDeps, depName);
```
maiieul, line 150: Removed clearTransitiveDeps as it would remove deps that are actually needed in the graph when using rolldown for preloading to work

packages/qwik-vite/src/plugins/bundle-graph.unit.ts
```diff
@@ -51,30 +51,30 @@ describe('convertManifestToBundleGraph', () => {
       2,
       'static-dep.js', // 2
       -7,
-      5,
-      // doesn't list 13 because it's also statically imported by dynamic-dep.js
-      'dynamic-dep.js', // 5
+      13,
```
maiieul, line 54: Now lists 13 as expected. The graph doesn't seem to be significantly bigger.

packages/qwik-vite/src/plugins/plugin.ts
```diff
@@ -89,8 +90,11 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
 
   const clientResults = new Map<string, TransformOutput>();
   const clientTransformedOutputs = new Map<string, [TransformModule, string]>();
+  const clientSegments = new Map<string, SegmentAnalysis>();
```
maiieul, line 105: Must create our own Map because of a bug in Rolldown where `meta` doesn't get passed to `codeSplitting`

packages/qwik-vite/src/plugins/plugin.ts
```diff
@@ -383,26 +387,21 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
...
+    if (sys.env === 'node') {
+      const fs: typeof import('fs') = await sys.dynamicImport('node:fs');
+      if (!fs.existsSync(opts.rootDir)) {
+        throw new Error(`Qwik rootDir "${opts.rootDir}" not found.`);
+      }
+      if (typeof opts.srcDir === 'string' && !fs.existsSync(opts.srcDir)) {
+        throw new Error(`Qwik srcDir "${opts.srcDir}" not found.`);
+      }
+      for (const [_, input] of Object.entries(opts.input || {})) {
+        const normalizedInput = normalizePath(sys.path.resolve(opts.rootDir, input));
```
maiieul: normalizing to avoid issues with `src/...` paths in user land

packages/qwik-vite/src/plugins/vite.unit.ts
maiieul, line 1: Basically running the tests twice: once for Rollup and once for Rolldown.

packages/qwik-vite/src/manifest.ts
```diff
@@ -529,6 +529,9 @@ export function generateManifestFromBundles(
       if (!manifest.qwikLoader && modulePaths.some((m) => qwikLoaderRegex.test(m))) {
         manifest.qwikLoader = bundleFileName;
       }
+      if (!qwikHandlersName && modulePaths.some((m) => handlersRegex.test(m))) {
```
maiieul, line 549: @wmertens bringing this back as it's actually needed with Rolldown 🤷‍♂️

scripts/submodule-qwikloader.ts
```diff
@@ -75,12 +75,8 @@ export const minifyClientScript = async (content: string) => {
 export const getLoaderJsonString = async (config: BuildConfig, name: string) => {
   const filePath = join(config.distQwikPkgDir, name);
   const content = await readFile(filePath, 'utf-8');
-  // Remove vite comments and leading/trailing whitespace
```
maiieul, line 71: Is now handled by Rolldown.

starters/adapters/aws-lambda/adapters/aws-lambda/vite.config.ts
```diff
@@ -2,11 +2,12 @@ import { nodeServerAdapter } from "@qwik.dev/router/adapters/node-server/vite";
 import { extendConfig } from "@qwik.dev/router/vite";
 import { builtinModules } from "module";
 import baseConfig from "../../vite.config.ts";
+
 export default extendConfig(baseConfig, () => {
   return {
     ssr: {
       // This configuration will bundle all dependencies, except the node builtins (path, fs, etc.)
-      external: builtinModules,
+      external: [...builtinModules],
```
maiieul, line 10: builtinModules is readonly

packages/docs/package.json
```diff
@@ -29,7 +29,8 @@
     "@shikijs/transformers": "3.21.0",
     "@shikijs/types": "3.21.0",
     "@supabase/supabase-js": "2.53.0",
-    "@tailwindcss/vite": "4.1.18",
+    "@tailwindcss/postcss": "4.2.2",
```
thejackshelton: why postcss instead of tailwind vite here?

packages/qwik-vite/src/plugins/plugin.ts
```diff
@@ -449,6 +451,67 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
         : opts.target === 'ssr' || opts.target === 'test';
   };
 
+  const setCachedSegment = (id: string, segment: SegmentAnalysis | null) => {
```
thejackshelton, line 583: Assuming this is not final version, but could we simplify the path handling from lines 454 to 514. 

We ran into a lot of path handling bugs with similar changes in the optimizer previously

packages/qwik-vite/src/plugins/plugin.ts
```diff
@@ -1091,10 +1170,40 @@ export const manifest = ${serverManifest ? JSON.stringify(serverManifest) : 'glo
         return 'qwik-loader';
       }
     }
+    return mergeRelatedSegments(id, ctx);
+  };
+
+  const codeSplitting: CodeSplittingOptions = {
```
thejackshelton, line 1391: There is so much regex here that I think it would be valuable to use magic-regexp, this is what I use for all bundler tooling

https://regexp.dev/guide/examples

AI is also really good at getting this right over regex, and it's more human readable

packages/qwik-vite/src/plugins/rollup.ts
```diff
@@ -238,10 +264,52 @@ export async function normalizeRollupOutputOptionsObject(
...
+  return outputOpts;
+}
+
+/*
+ * The Vite 8 types override the Rollup types to use the Rolldown ones.
+ * This is why manualChunks appears as deprecated.
+ * Using the Rollup types directly leads to type errors in normalizeSharedOutputOptions because the functions are about the same but the types signatures have been changed in Rolldown.
+ */
+
+export async function normalizeRollupOutputObject(
```
thejackshelton, line 293: Are we going to rename this file to rolldown.ts? 

Worst case scenario we can make qwik v2 vite 8 only if it removes a lot of the complexity

packages/qwik-vite/src/plugins/vite.ts
```diff
@@ -239,6 +244,26 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
...
+        },
+      };
+      const rollupBuildOptions: RollupOptions = {
+        ...sharedBuildOptions,
+        /**
+         * This is a workaround to have predictable chunk hashes between builds. It doesn't seem to
+         * impact the build time.
+         * https://github.com/QwikDev/qwik/issues/7226#issuecomment-2647122505
+         */
+        maxParallelFileOps: 1,
```
thejackshelton, line 295: Are we sure this is still the case with rolldown?

packages/qwik-vite/src/plugins/vite.ts
```diff
@@ -308,21 +324,39 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
         },
       };
 
+      if (this.meta.rolldownVersion) {
```
thejackshelton, line 358: The ternaries here are a bit hard to read, and we're doing assignment anyways. Could we move these to regular if statements

packages/docs/package.json
maiieul: It's a peer dep that should be installed actually

packages/qwik-vite/src/plugins/plugin.ts
maiieul: Agreed regex can be hard. We had those there already though. magic-regex is a refactor we could do in a separate PR.

packages/qwik-vite/src/plugins/rollup.ts
maiieul: Will do in a follow up. We should support rollup until it's not supported by Vite anymore. They just dropped support for vite 5. When they drop support for 7 we can drop Rollup support. Some teams/projects might depend on it still.

packages/qwik-vite/src/plugins/vite.ts
maiieul: It doesn't exist in rolldown. This is why I separated the configs for rolldown and rollup.

packages/qwik-vite/src/plugins/vite.ts
maiieul: hmmm it's fine no?

packages/qwik-vite/src/plugins/plugin.ts
maiieul: Shouldn't be a problem but I simplified it and it still works 👍

packages/qwik-vite/src/plugins/plugin.ts
```diff
@@ -879,8 +939,10 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
       // This can happen in the repl when the plugin is re-initialized
       // and possibly in other places
       // NOTE: this should be Promise.all to avoid deadlocks
-      await Promise.all([...deps.values()].map((id) => ctx.load({ id })));
-
+      // This seems to work fine for Rolldown so we ignore it there
+      if (!ctx.meta?.rolldownVersion) {
```
maiieul: They explicitly mention not to await this.load() in the rolldown docs: https://rolldown.rs/reference/Interface.TransformPluginContext#load

> Deadlocks caused by awaiting this.load in cyclic dependencies

> While it is safe to use this.load in a [resolveId](https://rolldown.rs/reference/Interface.Plugin#resolveid) hook, you should be very careful when awaiting it in a [load](https://rolldown.rs/reference/Interface.Plugin#load) or [transform](https://rolldown.rs/reference/Interface.Plugin#transform) hook. If there are cyclic dependencies in the module graph, this can easily lead to a deadlock, so any plugin needs to manually take care to avoid waiting for this.load inside the [load](https://rolldown.rs/reference/Interface.Plugin#load) or [transform](https://rolldown.rs/reference/Interface.Plugin#transform) of the any module that is in a cycle with the loaded module.

packages/docs/vite.config.ts
```diff
@@ -181,6 +181,13 @@ export default defineConfig(() => {
       conditions: ['browser', 'worker', 'import', 'default'],
     },
     ssr: {
+      external: [
+        'prismjs',
+        'prismjs/components/prism-clike',
+        'prismjs/components/prism-jsx',
+        'prismjs/components/prism-markup',
+        'prismjs/components/prism-tsx',
+      ],
```
wmertens: these need to be internal or else be in `dependencies`

packages/optimizer/core/src/snapshots/qwik_core__test__example_parsed_inlined_qrls.snap
```diff
@@ -81,7 +81,7 @@ q_s_odz7dfdfdM.s(STYLES);
 q_s_odzdfdfdM.s(STYLES);
 
 
-Some("{\"version\":3,\"sources\":[\"/user/qwik/src/test.tsx\"],\"names\":[],\"mappings\":\";;;AACA,SAAS,YAAY,EAAc,QAAQ,EAAa,eAAe,QAAQ,iBAAiB;;;;;;;kBAmBtE;IACpB,MAAM,CAAC,MAAM,GAAG;IAChB,OAAO,MAAM,KAAK;AACnB;kBApBoD;IACxD;IACA;IAEA,MAAM,QAAQ,SAAS;QACtB,OAAO;IACR;IACA,OAAO,WAAW,GAAG,WAAK,mBACf;QACT,WAAW,GAAG,WAAK,iBACR;YACT;sBACA;SACA;QAEF,WAAW,GAAG,WAAI,iBACP,WAAW,GAAG,WAAI;YAC3B,WAAQ;;;iBAME;KAGZ;AAEH;AA5BA,OAAO,MAAM,MAAM,WAAW,GAAG,8BA4BA;AAEjC,OAAO,MAAM,SAAS,uBAAuB;iBA7BtB;gBACA\"}")
+Some("{\"version\":3,\"sources\":[\"/user/qwik/src/test.tsx\"],\"names\":[],\"mappings\":\";;;AACA,SAAS,YAAY,EAAc,QAAQ,EAAa,eAAe,QAAQ,iBAAiB;;;;;;;kBAmBtE;IACpB,MAAM,CAAC,MAAM,GAAG;IAChB,OAAO,MAAM,KAAK;AACnB;kBApBoD;IACxD;IACA;IAEA,MAAM,QAAQ,SAAS;QACtB,OAAO;IACR;IACA,OAAO,WAAW,GAAG,WAAK,mBACf;QACT,WAAW,GAAG,WAAK,iBACR;YACT;sBACA;SACA;QAEF,WAAW,GAAG,WAAI,iBACP,WAAW,GAAG,WAAI;YAC3B,WAAQ,oBAG6C;gBACpD;aACA;iBACS;KAGZ;AAEH;AA5BA,OAAO,MAAM,MAAM,WAAW,GAAG,8BA4BA;AAEjC,OAAO,MAAM,SAAS,uBAAuB;iBA7BtB;gBACA\"}")
```
wmertens: weird, do you know why the sourcemaps change but not the result?

packages/qwik-vite/src/plugins/bundle-graph.ts
wmertens: that doesn't make sense, the deps should be followed recursively and then result in the same deps.

Please put this back and make sure preload follows the dep graph.

packages/qwik-vite/src/plugins/bundle-graph.unit.ts
wmertens: disagree, 13 should be imported by dynamic-dep

packages/qwik-vite/src/plugins/plugin.ts
wmertens: BTW these PR comments should be real comments in the code

packages/qwik-vite/src/plugins/plugin.ts
wmertens: ok this means that we have to move the load to resolveId :thinking:

packages/qwik-vite/src/plugins/vite.ts
wmertens: the reason is probably the `await this.load`, it seems that load order influences minification and thus hashes.

Let's cache loads in `resolveId` and use them sync in load(). Hopefully that makes minification stable and we can remove this setting.

wmertens (changes requested):
great but some reservations

packages/qwik-vite/src/plugins/bundle-graph.ts
maiieul: It doesn't work with Rolldown's bundling. I found the issue through the debugger. the function removed the parent dep while it was valid. Removing this logic really doesn't seem to impact the size of bundlegraph much.

packages/qwik-vite/src/plugins/bundle-graph.unit.ts
maiieul: It is still imported by dynamic-dep

packages/optimizer/core/src/snapshots/qwik_core__test__example_parsed_inlined_qrls.snap
maiieul: Because Rolldown gives a different input to the optimizer

packages/qwik-vite/src/plugins/plugin.ts
maiieul: Right I'll add more context

packages/qwik-vite/src/plugins/plugin.ts
maiieul: Why do we need to load() here? Everything seems to work fine without it (at least when using Rolldown)

packages/docs/vite.config.ts
maiieul: I'll have a look

packages/qwik-vite/src/plugins/plugin.ts
wmertens: It's when reloading the server in dev mode hmm

packages/docs/vite.config.ts
maiieul: Seems to work without adding those here now. Maybe since the docs refactor 🤷‍♂️

packages/qwik-vite/src/plugins/plugin.ts
```diff
@@ -402,7 +402,11 @@ export function createQwikPlugin(optimizerOptions: OptimizerOptions = {}) {
         throw new Error(`Qwik srcDir "${opts.srcDir}" not found.`);
       }
       for (const [_, input] of Object.entries(opts.input || {})) {
-        const resolved = await resolver(input);
+        const normalizedInput =
+          sys.path.isAbsolute(input) || input.startsWith('@') || input.startsWith('\0')
```
thejackshelton: This is very hacky. Shouldn't rolldown be resolving ts aliases?

packages/optimizer/core/src/transform.rs
```diff
@@ -78,6 +78,10 @@ pub struct SegmentData {
 	pub hash: Atom,
 	pub need_transform: bool,
 	pub migrated_root_vars: Vec<ast::ModuleItem>,
+	/// When the bundler inlines capture variables, the explicit captures array
```
thejackshelton: https://github.com/QwikDev/qwik/pull/8493

Heads up you no longer need any of the transform.rs changes / can remove it. I fixed this exact issue finding it separately some time ago and it is merged in build/v2

packages/qwik-vite/src/plugins/plugin.ts
maiieul: It's a bug in rolldown where they don't know how to process  `src/entry.preview.tsx`

briancarbone:
`Uncaught ReferenceError: __VITE_PRELOAD__ is not defined` in the console.

Edit: Apparently, the Qwik Optimizer removes the `vite:build-import-analysis` plugin which was split into two plugins... `native:import-analysis-build` should also be removed.

Here? https://github.com/QwikDev/qwik/blob/rolldown-vite-support/packages/qwik-vite/src/plugins/vite.ts#L485

`const names = ['vite:build-import-analysis', 'native:import-analysis-build'];`

briancarbone:
Additionally, in Vite 8, the regex for `__QWIK_ROUTER_STATIC_PATHS_ARRAY__` needs to match backticks.

It should be:
```
/([`'"])__QWIK_ROUTER_(STATIC_PATHS|NOT_FOUND)_ARRAY__\1/g
```

https://github.com/QwikDev/qwik/blob/rolldown-vite-support/packages/qwik-router/src/adapters/shared/vite/post-build.ts#L121

I've tested both this change and the one above; both work.

briancarbone:
There's also a handful of warnings for `INVALID_ANNOTATION`; doesn't affect the behavior of the server but pollutes the build log. You should see those for any build using rolldown.

For example:

```
[INVALID_ANNOTATION] A comment "/* @__PURE__ */" in "./node_modules/@qwik.dev/core/dist/core.prod.mjs" contains an annotation that Rolldown cannot interpret due to the position of the comment.
      │
 9965 │ const Pa = /* @__PURE__ */ "" + (t => {
      │            ───────┬───────
      │                   ╰───────── comment ignored due to position
```

briancarbone:
@maiieul PR here: https://github.com/QwikDev/qwik/pull/8669

briancarbone:
Thanks. The backtick matching is also required in order for `isStaticPath` to work, otherwise the regex fails to match. Rolldown rewrites the strings as templates:

```
'__QWIK_ROUTER_STATIC_PATHS_ARRAY__'
'__QWIK_ROUTER_NOT_FOUND_ARRAY__'
```

becomes

```
`__QWIK_ROUTER_STATIC_PATHS_ARRAY__`
`__QWIK_ROUTER_NOT_FOUND_ARRAY__`
```

but the current regex is only matching for `'` and `"`.

@maiieul

maiieul:
thanks @briancarbone I had not seen your comments 😄. Addressed now :)

maiieul:
@briancarbone, for the `/* @__PURE__ */` warnings, do you get that a lot? (what percentage?). It's likely an optimizer fix needed so I'd rather do that in a follow-up.

briancarbone:
@maiieul Here's all of the invalid annotations; there's not too many: https://gist.github.com/briancarbone/8a880779c8d2eb3aff8ec043d4283c62

Also, I'm getting a deadlock on the server build since updating from 816cb57 to 15fe095.

```
✓ Built client modules

> lint
> eslint "src/**/*.ts*"


> build.server
> qwik check-client src dist && vite build -c adapters/fastly-compute/vite.config.ts


      ............
    .::: :--------:.
   .::::  .:-------:.
  .:::::.   .:-------.
  ::::::.     .:------.
 ::::::.        :-----:
 ::::::.       .:-----.
  :::::::.     .-----.
   ::::::::..   ---:.
    .:::::::::. :-:.
     ..::::::::::::
             ...::::


vite v8.0.14 building ssr environment for production...
Read client manifest from /dist/q-manifest.json
```

Deadlock occurs at reading the client manifest. Reverting to 816cb57 fixed it.

maiieul:
thanks @briancarbone I'm on it :)

maiieul:
@briancarbone all issues should be resolved now. Let me know if you find anything else!

maiieul:
Closing as superseded by #8909

PR #8384, opened 2026-02-24
fix: bind:value on components and with empty var props

Varixo:
- bind:value on components should use jsxSplit to correctly transform it in runtime
- bind:value should be transformed from const props to var props even if there is no var props

### DIFF
--- a/packages/qwik/src/optimizer/core/src/transform.rs
+++ b/packages/qwik/src/optimizer/core/src/transform.rs
@@ -1252,6 +1252,22 @@ impl<'a> QwikTransform<'a> {
 		key == &*BIND_VALUE || key == &*BIND_CHECKED
 	}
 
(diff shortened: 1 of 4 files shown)


packages/qwik/src/optimizer/core/src/transform.rs
```diff
@@ -1566,7 +1582,12 @@ impl<'a> QwikTransform<'a> {
 					.count();
 
 				let has_spread_props = spread_props_count > 0;
-				let should_runtime_sort = has_spread_props;
+
+				// Check if we have bind:* props on a component (is_fn = true)
+				// Components with bind props need _jsxSplit for runtime handling
```
wmertens, line 1587: ... In the edge case when fn is actually a string component at runtime

PR #8385, opened 2026-02-25
fix(qwik-city): support Deno as package manager for production builds

### LINKED ISSUE #8364: [🐞] Deno as package manager: production build fails

Qwik Runtime

## Description

When building a Qwik app for production **under Deno** (e.g. with `deno task build`), three issues prevent a working build out of the box.

### 1. `qwik build` CLI fails under Deno

The `qwik build` command fails with:

```
TypeError: (0 , import_node_events11.addAbortListener) is not a function
```

This comes from `node:events.addAbortListener` used by `execa` in [`packages/qwik/src/cli/utils/run-build-command.ts`](https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/cli/utils/run-build-command.ts). Deno doesn't implement this Node API yet, so this isn't strictly a Qwik issue.

**Workaround:** Use `deno task build.client` + `deno task build.server` directly (the autogenerated scripts). These work because `qwik check-client` and `vite build` don't hit this code path. This leads to issues 2 and 3 below.

### 2. SSG static generation is a no-op stub

[`packages/qwik-city/src/static/deno/index.ts`](https://github.com/QwikDev/qwik/blob/main/packages/qwik-city/src/static/deno/index.ts) exits immediately:

```ts
async function generate(_opts) {
  console.error(`Deno not implemented`);
  Deno.exit(1);
}
```

The Node implementation (`static/node`) works fine under Deno's Node compatibility layer — Deno supports `node:fs`, `node:path`, `node:os`, and `undici`.

**Suggested fix:** Delegate to the Node implementation:

```ts
export { generate } from "../node/index";
```

### 3. Client manifest not passed to server build

In [`packages/qwik/src/optimizer/src/plugins/vite.ts`](https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/optimizer/src/plugins/vite.ts), the temp file mechanism that passes the client manifest to the server build is gated on:

```ts
if (sys.env === 'node' || sys.env === 'bun') {
```

Since `getEnv()` returns `"deno"` when `typeof Deno !== "undefined"`, this block is skipped. The result:

- Client build never writes the manifest temp file
- Server build never reads it
- `@qwik-client-manifest` resolves to `null`
- Server logs `Missing client manifest` and QRL serialization fails with `Code(31)` errors

**Suggested fix:** Include `"deno"`:

```ts
if (sys.env === 'node' || sys.env === 'bun' || sys.env === 'deno') {
```

The `node:fs` and `node:os` APIs used in that block work under Deno's Node compat layer.

## Reproduction

<https://github.com/ianlet/qwik-deno-build-issues-repro>

Scaffolded with `deno run -A npm:create-qwik@latest empty` + `deno task qwik add deno` on Qwik 1.19.0. See the repo README for step-by-step reproduction.

## Workaround

We patch both files post-install with a Deno script before building, and use direct `vite build` commands instead of `qwik build`. Happy to submit a PR if the approach looks good.

### Reproduction

https://github.com/ianlet/qwik-deno-build-issues-repro

### Steps to reproduce

_No response_

### System Info

```shell
- Qwik / Qwik City: 1.19.0
- Deno: 2.6.9
- Vite: 7.3.1
```

### Additional Information

_No response_

ianlet:
## Summary

Fixes #8364. When using Deno to manage dependencies and run builds (`deno task build.client` + `deno task build.server`), two issues prevented a working production build:

1. **Client manifest not passed to server build** — The Vite plugin's temp file mechanism was gated on `node`/`bun` only, so Deno builds got null manifests and `Code(31)` errors.
2. **SSG stubbed out** — The Deno SSG entry called `Deno.exit(1)` instead of delegating to the Node implementation, which works under Deno's Node compatibility layer.

### Changes

- Extract `hasNodeCompat()` predicate to include `deno` in all 4 env gates in the Vite plugin (`vite.ts`)
- Route Deno to the Node SSG module at runtime, same as Bun (`static/index.ts`)
- Remove dead `static/deno/` stub and its build step (`scripts/qwik-city.ts`)

## Test plan

- [x] Unit tests pass (`pnpm test.unit`) — 771 tests, including 2 new Deno-specific tests
- [x] `pnpm build --qwik --qwikcity --dev` succeeds
- [x] Validated end-to-end with a Deno project: client build writes manifest, server build reads it, SSG runs successfully — all without patches

### DIFF
--- a/packages/qwik/src/core/platform/platform.ts
+++ b/packages/qwik/src/core/platform/platform.ts
@@ -1,6 +1,11 @@
 // keep this import from qwik/build so the cjs build works
 import { isServer } from '@builder.io/qwik/build';
-import { qError, QError_qrlMissingChunk, QError_qrlMissingContainer } from '../error/error';
+import {
+  qError,
+  QError_dynamicImportFailed,
+  QError_qrlMissingChunk,
+  QError_qrlMissingContainer,
+} from '../error/error';
 import { getSymbolHash } from '../qrl/qrl-class';
 import type { QwikElement } from '../render/dom/virtual-element';
 import { qDynamicPlatform } from '../util/qdev';
@@ -16,6 +21,8 @@ export const createPlatform = (): CorePlatform => {
         if (regSym) {
           return regSym;
         }
+        // we never lazy import on the server
+        throw qError(QError_dynamicImportFailed, symbolName);
       }
       if (!url) {
         throw qError(QError_qrlMissingChunk, symbolName);
--- a/packages/qwik/src/core/platform/platform.unit.ts
+++ b/packages/qwik/src/core/platform/platform.unit.ts
@@ -1,0 +1,141 @@
+import { describe, test, expect, beforeEach, afterEach, vi } from 'vitest';
+import { createPlatform, setPlatform } from './platform';
+import { getSymbolHash } from '../qrl/qrl-class';
+
+describe('core platform', () => {
+  beforeEach(() => {
+    // Initialize a fresh Map for each test to avoid pollution
+    (globalThis as any).__qwik_reg_symbols = new Map<string, any>();
+  });
+
+  afterEach(() => {
+    // Clean up global state
+    delete (globalThis as any).__qwik_reg_symbols;
+  });
+
+  describe('importSymbol - server mode', () => {
+    test('returns registered symbol without importing', async () => {
+      const platform = createPlatform();
+
+      // Register a mock symbol
+      const symbolName = 'myComponent_abc123';
+      const hash = getSymbolHash(symbolName);
+      const mockFunction = () => 'mock component';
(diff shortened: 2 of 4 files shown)


promptless-for-oss:
@ianlet ✅ Promptless drafted a docs update: https://github.com/QwikDev/qwik/pull/8386

shairez:
@ianlet do you know what this promptless app is by any chance?
It started commenting on our PRs after this PR was created

ianlet:
@shairez nope, no idea. I wrote to you about it in private.

wmertens (commented):
This looks good to me, but in v2 this code has seen some changes and I won't try to port it.

So I'm ok with merging if you first create a PR to add this for v2 @ianlet ?

ianlet:
@wmertens will do 🫡

PR #8387, opened 2026-02-25
fix(eslint valid scope): check only value of signals

wmertens:
this changes the eslint plugin so it only checks the .value of signals for valid lexical scoping


promptless-for-oss:
@wmertens ℹ️ Promptless reviewed this request and did not find a docs change to propose.

PR #8397, opened 2026-03-02
chore: reduce client bundle size

Varixo:
~ 0.5kB less for core.min.mjs

### DIFF
--- a/starters/apps/base/package.json
+++ b/starters/apps/base/package.json
@@ -18,10 +18,10 @@
   "devDependencies": {
     "@qwik.dev/core": "latest",
     "@qwik.dev/router": "latest",
-    "@eslint/js": "9.39.2",
+    "@eslint/js": "latest",
     "@types/node": "latest",
-    "typescript-eslint": "8.52.0",
-    "eslint": "9.39.2",
+    "typescript-eslint": "latest",
+    "eslint": "latest",
     "eslint-plugin-qwik": "latest",
     "globals": "latest",
     "prettier": "latest",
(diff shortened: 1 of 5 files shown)


scripts/submodule-core.ts
```diff
@@ -12,20 +12,30 @@ import {
...
+ * mangling and keep $...$ names in sync across both bundles.
  */
-export function submoduleCore(config: BuildConfig) {
+export async function submoduleCore(config: BuildConfig): Promise<object | undefined> {
   if (config.dev) {
-    return submoduleCoreDev(config);
+    await submoduleCoreDev(config);
+    return undefined;
   }
   return submoduleCoreProd(config);
```
wmertens, line 35: why undefind|object?

scripts/submodule-server.ts
```diff
@@ -1,17 +1,23 @@
...
  *
  * This is submodule for helping to generate server-side rendered pages, along with providing
  * utilities for prerendering and unit testing.
+ *
+ * @param nameCache - Terser nameCache from the core build. When provided (prod builds), server.mjs
+ *   gets a property-mangling pass using the same $...$ mappings so that cross-bundle calls between
+ *   core and server keep property names in sync.
  */
-export async function submoduleServer(config: BuildConfig) {
+export async function submoduleServer(config: BuildConfig, nameCache?: object) {
```
wmertens, line 20: I think we need this on preloader too

starters/apps/base/package.json
```diff
@@ -18,10 +18,10 @@
   "devDependencies": {
     "@qwik.dev/core": "latest",
     "@qwik.dev/router": "latest",
-    "@eslint/js": "latest",
+    "@eslint/js": "9.39.2",
     "@types/node": "latest",
-    "typescript-eslint": "latest",
-    "eslint": "latest",
+    "typescript-eslint": "8.52.0",
+    "eslint": "9.39.2",
```
wmertens: is this still needed on build/v2?

scripts/submodule-core.ts
Varixo: from coredev we dont have namecache to mangle, so it is undefined

PR #8402, opened 2026-03-04
fix(qwik-city): prevent EBADF during sitemap generation

mcoelho80:
During Qwik City SSG runs on FreeBSD we observed intermittent failures:

```
EBADF: bad file descriptor, write
```

Investigation showed the error originates from the sitemap generation logic. The current implementation flushes the sitemap buffer using repeated `fs.promises.appendFile` calls while SSG workers are still active.

Each `appendFile` results in an `open → write → close` cycle. Under concurrent worker activity this produces frequent descriptor churn, which on FreeBSD can lead to writes being attempted on already closed file descriptors.

This patch replaces the repeated `appendFile` calls with a persistent `fs.createWriteStream` that:

* opens the sitemap file once
* performs incremental writes
* closes the stream during shutdown

This approach avoids repeated descriptor allocation and prevents the EBADF condition observed during SSG runs on FreeBSD.

Benefits:

* prevents EBADF failures during SSG on FreeBSD
* eliminates repeated open/close cycles
* reduces filesystem overhead
* preserves the existing sitemap generation behavior

---

# What is it?

* Bug

---

# Description

Fixes a race condition in sitemap generation during Qwik City SSG.

When the sitemap buffer exceeds the flush threshold, the code uses `fs.promises.appendFile` to append entries. Because SSG rendering runs concurrently across multiple workers, this results in many rapid open/write/close cycles on the same file.

On FreeBSD this can trigger `EBADF` errors when writes occur on file descriptors that have already been closed.

The fix replaces `appendFile` with a persistent `WriteStream`, ensuring the sitemap file is opened once and written incrementally until SSG shutdown.

---

# Checklist

* [x] My code follows the developer guidelines of this project
* [x] I performed a self-review of my own code
* [ ] I added a changeset with `pnpm change`
* [ ] I made corresponding changes to the Qwik docs
* [ ] I added new tests to cover the fix / functionality

### DIFF
--- a/packages/qwik-city/src/static/node/node-main.ts
+++ b/packages/qwik-city/src/static/node/node-main.ts
@@ -18,6 +18,7 @@ export async function createNodeMainProcess(sys: System, opts: StaticGenerateOpt
   const ssgWorkers: StaticGeneratorWorker[] = [];
   const sitemapBuffer: string[] = [];
   let sitemapPromise: Promise<any> | null = null;
+  let sitemapStream: fs.WriteStream | null = null;
 
   opts = { ...opts };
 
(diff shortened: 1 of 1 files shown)


wmertens:
@mcoelho80 looking great, can you fix the linting issues?

wmertens:
@mcoelho80 ?

mcoelho80:
> @mcoelho80 looking great, can you fix the linting issues?

Check PR #8403

Issue #7803, opened 2025-08-10
[🐞] [v1->v2] builder dependencies need overriding


wmertens:
### Which component is affected?

Qwik Runtime

### Describe the bug

When migrating from v1 to v2, qwik libraries will still want the @builder.io dependencies.

This isn't a problem for the build, because the optimizer replaces those imports with @qwik.dev. However, typescript won't have types because it's looking in the wrong place.

### Steps to reproduce

- take a v1 project with a qwik lib dependency like e.g. @auth/qwik
- run `npx qwik migrate-v2`
- the types for @auth/qwik will be `any`

To fix direct dependencies of dependencies: add overrides
```
	"overrides": {
		"@builder.io/qwik": "npm:@qwik.dev/core",
		"@builder.io/qwik-city": "npm:@qwik.dev/router",
		"@builder.io/qwik-react": "npm:@qwik.dev/react",
		"@builder.io/eslint-plugin-qwik": "npm:@qwik.dev/eslint-plugin-qwik",
	},
```

To fix peer dependencies: add aliased dependencies so that they are used as peers when needed

in package.json:
```
	"devDependencies": {
		"@builder.io/qwik": "npm:@qwik.dev/core",
		"@builder.io/qwik-city": "npm:@qwik.dev/router",
	},
```
Other packages could be added too but they don't normally get used as peer dependencies and they would get installed.

dmitry-stepanenko:
@wmertens I'm not sure we should do any actions here.

1. Overrides cannot simply look like `"@builder.io/qwik": "npm:@qwik.dev/core",`, in order to not have any mess we should match them to the same version as the corresponding @qwik.dev package .  This can get our of control once people will just migrate their repo themselves and forget to update the override. 
2. Adding old names in `devDependencies` has the same problem as above. On top of that this installs the second version of the @qwik.dev dependency as an old name and it gives no guarantee APIs will actually match between v1 and v2.

I've played around with adding both of them and it feels just as a workaround. It really feels like those packages should simply release a new version on their side to switch to our v2 packages once we release them as stable

wmertens:
1. Hmm, in my experience it just took whatever the @qwik.dev/core package was that was installed?
2. We are the ones guaranteeing the API compatibility between the old and new names :)

But indeed you're right, a better option is to switch to qwik.dev for releases now and to have the builder packages re-export them.
We would basically change the package.json files to be what they are in v2, and add extra packages dir for the builder packages.

PR #8408, opened 2026-03-05
v2: merge main into v2

wmertens:


wmertens:
@jantimon I couldn't add your css url fix because the code is different now, can you check if it's also a problem on v2?

jantimon:
@wmertens I guess that happens now in `packages/qwik-router/src/buildtime/vite/dev-middleware.ts`?

I will try - is there already any npm alpha of v2?

wmertens:
@jantimon yes you can just run `npx qwik migrate-v2` and it should upgrade your project to v2. You may need to do #7803 if you use qwik libraries.

PR #8413, opened 2026-03-08
React: allow Qwik components inside React

wmertens:
`reactify$()` to make it easy to gradually migrate from React to Qwik

### DIFF
--- a/packages/qwik/src/core/client/dom-container.ts
+++ b/packages/qwik/src/core/client/dom-container.ts
@@ -227,7 +227,7 @@ export class DomContainer extends _SharedContainer implements IClientContainer {
   }
 
   getParentHost(host: VNode): VNode | null {
-    let vNode: VNode | null = host.parent || host.slotParent;
+    let vNode: VNode | null = host.parent;
     while (vNode) {
       if (vnode_isVirtualVNode(vNode)) {
         if (vnode_getProp(vNode, OnRenderProp, null) !== null) {
--- a/packages/qwik/src/core/shared/projection/external-projection.ts
+++ b/packages/qwik/src/core/shared/projection/external-projection.ts
@@ -22,7 +22,7 @@ import type { QRL } from '../qrl/qrl.public';
  * VNode is stored as a projection on the parent, and a low-priority cursor is added so the cursor
  * walker will process it.
  *
- * Use `_slotReady` to set the DOM target element before the cursor fires.
+ * Use `_setProjectionTarget` to set the DOM target element before the cursor fires.
  *
  * @internal
  */
@@ -35,7 +35,7 @@ export function _addProjection(
 ): VirtualVNode {
   const vnode = vnode_newVirtual();
   vnode_setProp(vnode, QSlot, slotName);
-  vnode.slotParent = parentVNode;
+  vnode.parent = parentVNode;
   vnode_setProp(parentVNode, slotName, vnode);
   vnode_setProp(vnode, OnRenderProp, componentQRL);
   vnode_setProp(vnode, ELEMENT_PROPS, props);
@@ -52,7 +52,7 @@ export function _addProjection(
  *
  * @internal
  */
-export function _slotReady(vnode: VirtualVNode, targetElement: Element): void {
+export function _setProjectionTarget(vnode: VirtualVNode, targetElement: Element): void {
   vnode_setProp(vnode, QTargetElement, targetElement);
   vnode.flags |= VNodeFlags.HasTargetElement;
 }
--- a/packages/qwik-react/src/react/reactify.ts
+++ b/packages/qwik-react/src/react/reactify.ts
@@ -2,7 +2,7 @@ import { Component, createElement, createRef } from 'react';
 import { implicit$FirstArg, type QRL } from '@qwik.dev/core';
 import {
   _addProjection,
-  _slotReady,
+  _setProjectionTarget,
   _updateProjectionProps,
   _removeProjection,
 } from '@qwik.dev/core/internal';
@@ -78,10 +78,10 @@ export function reactifyQrl(qwikCompQrl: QRL<any>): any {
         // Use pendingProps if React updated props before the QRL resolved
         const reactProps = getReactProps(this.pendingProps || this.props);
         this.vnode = _addProjection(container, parentVNode, qwikCompQrl, reactProps, this.slotName);
-        // Clear the marker comment right before _slotReady so the div is empty
+        // Clear the marker comment right before _setProjectionTarget so the div is empty
         // when Qwik's cursor walker processes it on the next microtask.
         div.replaceChildren();
-        _slotReady(this.vnode, div);
+        _setProjectionTarget(this.vnode, div);
       });
     }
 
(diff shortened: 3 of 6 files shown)


packages/qwik/src/core/client/dom-container.ts
```diff
@@ -228,7 +228,7 @@ export class DomContainer extends _SharedContainer implements IClientContainer {
   }
 
   getParentHost(host: VNode): VNode | null {
-    let vNode: VNode | null = host.parent;
+    let vNode: VNode | null = host.parent || host.slotParent;
```
Varixo: why it is needed? can cause unexpected bugs

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -1064,8 +1065,16 @@ export const vnode_insertVirtualBefore = (
...
+  const targetEl =
+    !parentIsElement && parent.flags & VNodeFlags.HasTargetElement
+      ? (parent.props?.[QTargetElement] as Element)
+      : null;
+  const domParentVNode = targetEl
+    ? null
+    : parentIsElement
+      ? parent
+      : vnode_getDomParentVNode(parent, false);
+  const parentNode = targetEl || domParentVNode?.node;
```
Varixo, line 1076: basically this function needs as low as possible additional checks, not sure how it influences the benchmark. It shouldnt, but we should check it before merge

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -1297,8 +1306,13 @@ export const vnode_insertBefore = (
...
-  vnode = vnode_getDomParentVNode(vnode, includeProjection) as VNode;
-  return (vnode && (vnode as ElementVNode).node) as Element | null;
+  while (vnode && !vnode_isElementVNode(vnode)) {
+    if (vnode.flags & VNodeFlags.HasTargetElement) {
+      return vnode.props?.[QTargetElement] as Element;
+    }
+    vnode = (vnode.parent || (includeProjection ? vnode.slotParent : null))!;
+  }
+  return vnode ? (vnode as ElementVNode).node : null;
 };
```
Varixo, line 1315: what happened with vnode_getDomParentVNode? both functions should work the same way

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -1578,12 +1599,25 @@ export const fastNextSibling = (node: Node | null): Node | null => {
...
 ): Node | null {
+  const isSingleValue = typeof commentValue === 'string';
   while (node) {
-    if (node.nodeValue?.startsWith(commentValue)) {
+    const nodeValue = node.nodeValue;
+    let isMatch;
+    if (isSingleValue) {
+      isMatch = nodeValue?.startsWith(commentValue as string);
+    } else {
+      for (let i = 0; i < (commentValue as string[]).length; i++) {
```
Varixo: isnt the .length executed for every iteration? maybe extract it outside a loop

packages/qwik/src/core/client/vnode-utils.ts
```diff
@@ -1611,6 +1648,18 @@ const fastFirstChild = (node: Node | null): Node | null => {
...
+  // Handle q:ignore as first child (e.g. qwikify$ Host with reactify$ projections).
+  // Navigate depth-first to the first q:container-island and return its first element.
+  if (
+    node &&
+    fastNodeType(node) === /* Node.COMMENT_NODE */ 8 &&
+    node.nodeValue?.startsWith(QIgnore)
+  ) {
+    if (!_fastNextSibling) {
+      _fastNextSibling = fastGetter<typeof _fastNextSibling>(node, 'nextSibling')!;
+    }
```
Varixo, line 1660: I think we have somewhere that code already? cant we reuse it?

packages/qwik/src/core/shared/projection/external-projection.ts
```diff
@@ -0,0 +1,94 @@
...
+  slotName: string
+): VirtualVNode {
+  const vnode = vnode_newVirtual();
+  vnode_setProp(vnode, QSlot, slotName);
+  vnode.slotParent = parentVNode;
+  vnode_setProp(parentVNode, slotName, vnode);
+  vnode_setProp(vnode, OnRenderProp, componentQRL);
+  vnode_setProp(vnode, ELEMENT_PROPS, props);
+  vnode.dirty = ChoreBits.COMPONENT;
+  addCursor(container, vnode, 1); // low priority
```
Varixo, line 43: I think all prio are 0, so actually it is the highest prio

packages/qwik/src/core/shared/projection/external-projection.ts
```diff
@@ -0,0 +1,94 @@
...
+    const targetEl = vnode_getProp<Element>(vnode, QTargetElement, null);
+    if (targetEl) {
+      targetEl.replaceChildren();
+    }
+    vnode_setProp(vnode, QTargetElement, null);
+    vnode.flags &= ~VNodeFlags.HasTargetElement;
+  }
+
+  // Mark as deleted
+  vnode.flags |= VNodeFlags.Deleted;
```
Varixo: shouldnt we cleanup it? its not standard projection so maybe it should be removed?

packages/qwik/src/core/client/dom-container.ts
wmertens: Without it, resuming doesn't work because the projected reactified qwik components don't have a parent.

And they don't have a parent because they are added after the qwikify$ parent rendered.

Maybe they should get a parent assigned on resume?

packages/qwik/src/core/client/vnode-utils.ts
wmertens: So this is used to keep track of where the reactified vnode should be rendered, the target is a React-managed div and there is no vnode associated with it and so it can't call vnode_getDomParentVNode.

Should we create an elementvnode for the React div instead, and make that the parent?

packages/qwik/src/core/client/vnode-utils.ts
wmertens: it depends on if `fastFirstChild` or `fastNextSibling` are called first. Not sure how to reuse it here hmm.

packages/qwik/src/core/shared/projection/external-projection.ts
wmertens: no, lower number is higher priority. -1 is high prio

packages/qwik/src/core/client/dom-container.ts
Varixo: I think we talked about it? I think we should connect parent instead

packages/qwik/src/core/client/vnode-utils.ts
Varixo: Hmm I dont like it but ok

packages/qwik/src/core/client/vnode-utils.ts
Varixo: I think you can just call the fast functions. Maybe something needs extracting

packages/qwik/src/core/client/dom-container.ts
wmertens: yes I did the change :+1:

PR #8414, opened 2026-03-09
feat(router): 404 during dev, eTag, caching, SSG fixes

wmertens:
Made routes + menus into a trie structure instead of an array and much more

- 404.tsx now works during dev
- pages can put eTag in head
- pages can opt in to in-memory caching (50 pages only, configurable) by putting cacheKey in head
- SSG now doesn't hang, and can be run again after the build
- added `routeConfig` to combine head and caching


wmertens:
actually let's move eTag and cacheKey into head(), and document them.

packages/qwik-router/src/middleware/request-handler/etag.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  ssrCache.set(key, html);
+}
+
+/**
+ * Clear the in-memory SSR cache. Call after deployments or data changes.
+ *
+ * @public
+ */
+export function clearSsrCache(): void {
```
intellix: maybe it would be beneficial to allow deleting an individual key?

wmertens: good idea, done

packages/qwik-router/src/buildtime/types.ts
```diff
@@ -1,8 +1,35 @@
...
+ */
+export interface BuildTrieNode {
+  /** Parameter name (for `_W` and `_A` nodes) */
+  _P?: string;
+  /** Prefix for infix params (e.g. `pre` for `pre[slug]post`) — only on `_W` nodes */
+  _0?: string;
+  /** Suffix for infix params (e.g. `post` for `pre[slug]post`) — only on `_W` nodes */
+  _9?: string;
+  /** Rewrite target path (set by rewriteRoutes), e.g. '/about/' or '/blog/_W/' */
+  _G?: string;
```
Varixo, line 18: why these weird names?

packages/qwik-router/src/runtime/src/types.ts
```diff
@@ -250,46 +278,66 @@ export interface ContentHeading {
...
+  /** The parameter name when this node is reached via `_W` or `_A` from the parent */
+  _P?: string;
+  /** Prefix for infix params (e.g. "pre" for `pre[slug]post`) — only on `_W` nodes */
+  _0?: string;
+  /** Suffix for infix params (e.g. "post" for `pre[slug]post`) — only on `_W` nodes */
+  _9?: string;
+  /** Group (pathless layout) nodes merged into this level, sorted by group name */
+  _M?: RouteData[];
+  /** Menu loader for this subtree (from menu.md). Runtime uses nearest ancestor during traversal. */
+  _N?: MenuModuleLoader;
```
Varixo, line 318: and here too

starters/e2e/qwikrouter/page.e2e.ts
```diff
@@ -148,7 +148,7 @@ function tests() {
         400,
       );
       await assertPage(ctx, {
-        title: "400 Bad Request",
+        title: "Error - Qwik",
```
Varixo: so from where is this title?

packages/qwik-router/src/buildtime/types.ts
wmertens: they get embedded in the client code so they need to be short, and they live next to the route names so they must be escaped with `_`

starters/e2e/qwikrouter/page.e2e.ts
wmertens: from routerhead and src/routes/error.tsx
```tsx
  const title = head.title ? `${head.title} - Qwik` : `Qwik`;
```

wmertens: combination of router-head.tsx and the title from error.tsx. I added a status in it now as well

Issue #4769, opened 2023-07-10
[🐞] Math.round is not serializing into q:func


mhevery:
### Which component is affected?

Qwik Optimizer (rust)

### Describe the bug

```typescript
import { component$, useSignal } from '@builder.io/qwik';

export default component$(() => {
  const count = useSignal(123);
  return (
    <>
      Count: {Math.round(count.value + 1)}
      <button onClick$={() => (count.value += 0.5)}>+1</button>
    </>
  );
});
```

The above should serialize into:
```html
<script q:func="qwik/json">
  document.currentScript.qFuncs=[
    (p0,)=>Math.round(p0.value)
  ]
</script>
```

But it does not. The result is that the component is being downloaded and re-rendered using the JSX path rather than the signal-dom path.

### Reproduction

https://stackblitz.com/edit/qwik-starter-v96ske?file=src%2Froutes%2Findex.tsx

### System Info

```shell
n/a
```

gioboa:
@Varixo @wmertens is this case addressed with Qwik v2?

Varixo:
> @Varixo @wmertens is this case addressed with Qwik v2?

I don't think so :/

JerryWu1234:
Maybe I can try this one

JerryWu1234:
I dug into this problem and discovered this bug in Rust. I struggle with Rust, so it takes me a long time.

gioboa:
I tried with v2 and this is still unresolved.

JerryWu1234:
@wmertens @Varixo 

need to discuss this further

wmertens:
So the reason is that the optimizer doesn't know that Math has no side effects and is safe to move. 

I propose that the optimizer can assume that all globals are safe to move and it gets a new option to list unsafe globals.

wmertens:
[Playground link](https://qwikdev-build-v2.qwik-8nx.pages.dev/playground/#f=Q0o0RmZYE40OovalsFUCCQpICW6LMFXD0MgY6BhcYeIMUm%2BlUO2bWJKhBy7kNcBG6AEzILBo1VYw1KyFhwC4aYKokashlqNqsFUw0DPVrLXTNoS3ZdCCRdN6NCkPvaQMAA), look at the client bundles

basically, moving globals into wrappedFn should be done. I think it's always safe.

JerryWu1234:
@wmertens  Can I close this ?

wmertens:
[...]

PR #8420, opened 2026-03-12
feat: Enhance optimizer to use `_fnSignal` for derived signal express…

JerryWu1234:
…ions and improve constant detection for safe global calls.

# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

#4769 

# Checklist


packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_derived_signals_children.snap
```diff
@@ -103,19 +102,19 @@ const App_component_ckEPmXZlub0 = ()=>{
         /*#__PURE__*/ _jsxSorted("div", null, null, dep, 3, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, dep.thing, 3, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, dep.thing + 'stuff', 3, null),
-        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing, 1, null),
-        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing.thing, 1, null),
-        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing.thing + 'stuff', 1, null),
+        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing, 3, null),
+        /*#__PURE__*/ _jsxSorted("div", null, null, _wrapProp(globalThing, "thing"), 3, null),
```
wmertens, line 112: ok

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_derived_signals_children.snap
```diff
@@ -103,19 +102,19 @@ const App_component_ckEPmXZlub0 = ()=>{
         /*#__PURE__*/ _jsxSorted("div", null, null, dep, 3, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, dep.thing, 3, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, dep.thing + 'stuff', 3, null),
```
wmertens, line 110: I think these can all be made into _fnSignal

packages/qwik/src/optimizer/core/src/snapshots/qwik_core__test__example_derived_signals_children.snap
```diff
@@ -103,19 +102,19 @@ const App_component_ckEPmXZlub0 = ()=>{
...
-        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing.thing, 1, null),
-        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing.thing + 'stuff', 1, null),
+        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing, 3, null),
+        /*#__PURE__*/ _jsxSorted("div", null, null, _wrapProp(globalThing, "thing"), 3, null),
+        /*#__PURE__*/ _jsxSorted("div", null, null, globalThing.thing + 'stuff', 3, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, signal.value(), 1, null),
-        /*#__PURE__*/ _jsxSorted("div", null, null, signal.value + unknown(), 1, null),
+        /*#__PURE__*/ _jsxSorted("div", null, null, signal.value + unknown(), 3, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, mutable(signal), 1, null),
         /*#__PURE__*/ _jsxSorted("div", null, null, signal.value + dep, 1, null)
```
wmertens, line 117: these can also be turned into _fnSignal I think

wmertens:
I'm not sure what breaks if we just wrap all expressions into fnSignal, but I can't think of anything :thinking: 
@Varixo ?

Varixo:
> I'm not sure what breaks if we just wrap all expressions into fnSignal, but I can't think of anything 🤔 @Varixo ?

We shouldnt wrap everything, because it means more vnode data and state to serialize, but idk if it breaks something

wmertens:
@Varixo so we should wrap only things that aren't referenced in the component, right?

For example, if a component uses foo.bar in the render function and in the jsx it also uses foo.bar, then we don't need to wrap

but if it uses untrack(() => foo.bar) then we should wrap

Varixo:
> @Varixo so we should wrap only things that aren't referenced in the component, right?
> 
> For example, if a component uses foo.bar in the render function and in the jsx it also uses foo.bar, then we don't need to wrap
> 
> but if it uses untrack(() => foo.bar) then we should wrap

Fnsignal is jsx only, no?

JerryWu1234:
@wmertens I tried few times. I'm not good at the Rust. I got some helps from AI but it modified a lot of e2e files. so the rest of function need your help because my current ability can't cover it I think

wmertens:
Ok, I now remembered the problem:

When you use a global or an import in an expression, those can change when you re-run the component. So they have to be in varProps, not constProps.

If they are in constProps, they will not update even when the component rerenders.

When wrapping an expression with _fnSignal, it will only update if there is a signal  in the expression that updates.

The problem with `Math.round(val)` is that `Math.random()` is variable. So even putting all Math things in const is a problem.

So, currently the code is the most correct unfortunately. We could make a special case out of Math but not Math.random and Date but not Date.now or Date(), and so on.

I think we should not do this change after all, sorry :-/

Thoughts @JerryWu1234 @Varixo ?

Varixo:
> Ok, I now remembered the problem:
> 
> When you use a global or an import in an expression, those can change when you re-run the component. So they have to be in varProps, not constProps.
> 
> If they are in constProps, they will not update even when the component rerenders.
> 
> When wrapping an expression with _fnSignal, it will only update if there is a signal in the expression that updates.
> 
> The problem with `Math.round(val)` is that `Math.random()` is variable. So even putting all Math things in const is a problem.
> 
> So, currently the code is the most correct unfortunately. We could make a special case out of Math but not Math.random and Date but not Date.now or Date(), and so on.
> 
> I think we should not do this change after all, sorry :-/
> 
> Thoughts @JerryWu1234 @Varixo ?

yeah, seems like we shouldnt merge that

wmertens:
thanks for your efforts @JerryWu1234 !

PR #8421, opened 2026-03-12
Real, instant, HMR :tada:

wmertens:
Adds event handlers to all components that get triggered when their parent file changes and makes them rerender.
This allows real hot module reload, without resuming :)

The slight disadvantage is that all components now send their state during development (because now they can always rerender on the client). You can disable HMR and fall back to full page reloads by setting `{devTools: {hmr: false}}` in the `qwikVite()` plugin configuration.


termermc:
Big ups to you @wmertens!

PR #8445, opened 2026-03-19
fix: separate ssg bundle graph

Varixo:
- Fix an SSG hang where worker-side request handling modules could be bundled back through run-ssg.js, causing builds to stall after Starting Qwik Router SSG....
- Fix static route reconstruction in orchestrator.ts so basePathname is not prefixed twice.
- Avoid eagerly loading non-page route modules during static route discovery, preventing route collection from blocking on unrelated layout/module imports.
- Introduce dedicated virtual worker import ids, resolved by the router build, so the SSG worker graph stays isolated from run-ssg.js.
- Add regression coverage that asserts worker chunks do not import the run-ssg entry back, plus targeted SSG unit coverage for the refactored worker path.


wmertens (commented):
I like it but I'm still not convinced we need to duplicate the requestevent code :-/

Varixo:
> I like it but I'm still not convinced we need to duplicate the requestevent code :-/

is it duplicated? it creates different objects, but the code is shared, no?

PR #8446, opened 2026-03-19
fix(router): sync history push for webkit engine

Varixo:
Webkit can decide to ignore pushed history if it was not done during user interaction. Thats why we should do it in sync with user interaction.
We push the history in sync and update it later after loading all necessary data.


gioboa (approved):
LGTM Thanks @Varixo

PR #8456, opened 2026-03-20
Remove macOS x86_64 runner from CI build-bindings matrix

Varixo:
### Motivation
- Remove the last x86 macOS standard runner (`macos-13`, `x86_64-apple-darwin`) from the `build-bindings` matrix because that runner is no longer available as a standard GitHub-hosted runner.

### Description
- Deleted the `- host: macos-13` / `target: x86_64-apple-darwin` matrix entry in `.github/workflows/ci.yml` under the `build-bindings` job.

### Testing
- No automated tests were run as part of this change since it only modifies the GitHub Actions workflow configuration.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_69bd7b18cd9c8331a56e197f8e1d5c28)

### DIFF
--- a/packages/qwik-city/src/utils/fs.ts
+++ b/packages/qwik-city/src/utils/fs.ts
@@ -152,7 +152,7 @@ export function isIndexModule(extlessName: string) {
 }
 
 export function isPluginModule(extlessName: string) {
-  return /^plugin(|@[^.]+)$/.test(extlessName);
+  return /^plugin(|@.+)$/.test(extlessName) && !/\.(test|unit|spec)(\.[jt]s)?$/.test(extlessName);
 }
 
 export function isLayoutModule(extlessName: string) {


gioboa (dismissed):

packages/docs/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "incremental": true,
     "target": "es2022",
-    "module": "es2022",
+    "module": "esnext",
```
gioboa, line 5: Is this change correct?

Varixo: hmm localy without it I have errors for with { type: 'json' };. on v2 branch we already have that, so I think its fine?

gioboa: Thanks

packages/docs/src/repl/ui/repl-version.unit.ts
```diff
@@ -1,5 +1,9 @@
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
+vi.mock('../bundler/bundled', () => ({
```
wmertens, line 3: Why is this needed? Wasn't it working before?

packages/qwik-city/src/utils/fs.ts
```diff
@@ -148,15 +148,15 @@ const MARKDOWN_EXTS: { [type: string]: boolean } = {
 };
 
 export function isIndexModule(extlessName: string) {
-  return /^index(|!|@.+)$/.test(extlessName);
+  return /^index(|!|@[^.]+)$/.test(extlessName);
```
wmertens: I don't think this matters, and some people may already have a period in their names?

packages/docs/src/repl/ui/repl-version.unit.ts
Varixo: yes it was throwing error
```
Error: Failed to resolve entry for package "csstype". The package may have incorrect main/module/exports specified in its package.json.
  Plugin: vite:import-analysis
  File: C:/Users/Michal/Documents/git/qwik/packages/qwik/dist/core.d.ts?raw-source:1:23
  1  |  import * as CSS_2 from "csstype";
     |                          ^
  2  |  import { isBrowser } from "./build";
  3  |  import { isDev } from "./build";
  ```
explanation:  
1. repl-version.ts imports `bundled` from `../bundler/bundled`
2. bundled.tsx imports `core.d.ts?raw-source` from the built `@builder.io/qwik` package
3. `core.d.ts` has `import * as CSS_2 from "csstype"` at the top
4. Vite tries to resolve `csstype` as a real module for the `?raw-source` transform, but `csstype` is a types-only package with no `main`/`module`/`exports` field that Vite can load

File: packages/qwik-city/src/utils/fs.ts [hunk 4]
```diff
@@ -152,7 +152,7 @@ export function isIndexModule(extlessName: string) {
 }
 
 export function isPluginModule(extlessName: string) {
-  return /^plugin(|@.+)$/.test(extlessName);
+  return /^plugin(|@[^.]+)$/.test(extlessName);
```
wmertens: why is this needed?

Varixo: to ignore files with additional dots like .unit.ts. We could also just ignore ".unit" or ".spec", wdyt?

wmertens: ok let's specifically ignore `/\.(test|unit|spec)\.[jt]s/`

PR #8466, opened 2026-03-22
feat: add $destroy$ method and qDestroy to DomContainer

thejackshelton:
## `$destroy()` method and `qDestroy` for DomContainer

  Adds a teardown API for Qwik containers, enabling meta-framework integrations (like Astro's `ClientRouter`) to properly destroy containers during client-side page transitions.

  ### Before
  The `@qwik.dev/astro` integration required a massive inline script manually nulling Qwik internals (`qContainer`, `qVnodeData`, `qVNodeRefs`, `document.qVNodeData`), suppressing errors, and
  re-dispatching lifecycle events. Brittle, unmaintainable, broke on internal changes.

  ### After
  ```js
  document.addEventListener("astro:after-preparation", function () {
    document.querySelectorAll("[q\\:container]").forEach(function (e) {
      if (e.qDestroy) e.qDestroy();
    });
  });
  ```
  
  
  Details

  - `$destroy$()` on DomContainer: nulls vNodeLocate, clears state data in-place, overrides `$getObjectById$`, removes element properties and q:container attribute, clears document.qVNodeData
  - qDestroy on the DOM element: minification-safe bridge to `$destroy$()` for inline scripts that can't import from @qwik.dev/core

### DIFF
--- a/.changeset/tiny-mammals-allow.md
+++ b/.changeset/tiny-mammals-allow.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: sibling node duplication when toggling elements with dangerouslySetInnerHTML
--- a/packages/docs/src/components/code-block/code-block.css
+++ b/packages/docs/src/components/code-block/code-block.css
@@ -49,7 +49,8 @@ pre[class*='language-'] {
   word-spacing: normal;
   word-break: normal;
   word-wrap: normal;
-  line-height: 1.5;
+  font-size: 0.9em;
+  line-height: 1.28;
   tab-size: 4;
   hyphens: none;
 }
--- a/packages/docs/src/components/footer/footer.tsx
+++ b/packages/docs/src/components/footer/footer.tsx
@@ -52,19 +52,19 @@ const socialLinks = [
 
 export const Footer = component$(() => {
   return (
-    <footer class="bg-violet-0 px-20 pt-32 pb-10">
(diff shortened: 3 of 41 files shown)


wmertens:
waiting for @Varixo approval before merge. I think it's a good idea.

Varixo:
We also have cleanup function returned from dom-render

```ts
  return {
    cleanup: () => {
      /**
       * This can lead to cleaning up projection vnodes via the journal, but since we're cleaning up
       * they don't matter so we ignore the journal
       */
      cleanup(container, [], container.rootVNode);
    },
  };
  ```
I think we should call cleanup in the $destroy$ and maybe use destroy here instead

**Comment by thejackshelton-kunaico** [comment]
> We also have cleanup function returned from dom-render
> 
> ```ts
>   return {
>     cleanup: () => {
>       /**
>        * This can lead to cleaning up projection vnodes via the journal, but since we're cleaning up
>        * they don't matter so we ignore the journal
>        */
>       cleanup(container, [], container.rootVNode);
>     },
>   };
> ```
> 
> I think we should call cleanup in the d e s t r o y and maybe use destroy here instead

Doesn't matter to me as long as it can be called in a way like this:

```tsx
document.addEventListener("astro:after-preparation", function () {
  document.querySelectorAll("[q\\:container]").forEach(function (e) {
    if (e.qDestroy) e.qDestroy();
  });
});
```

Where it's a method on the container that can be called in an inline script

PR #8467, opened 2026-03-22
fix(ssr): remove backref serialization

wmertens:
We don't need it

### DIFF
--- a/packages/qwik/src/core/shared/serdes/serialize.ts
+++ b/packages/qwik/src/core/shared/serdes/serialize.ts
@@ -27,19 +27,6 @@ import { isPropsProxy } from '../jsx/props-proxy';
 import { Slot } from '../jsx/slot.public';
 import type { QRLInternal } from '../qrl/qrl-class';
 import { isQrl } from '../qrl/qrl-utils';
-import { _OWNER, _PROPS_HANDLER, _UNINITIALIZED } from '../utils/constants';
-import { EMPTY_ARRAY, EMPTY_OBJ } from '../utils/flyweight';
-import { ELEMENT_ID, ELEMENT_PROPS, QBackRefs } from '../utils/markers';
-import { isPromise, maybeThen } from '../utils/promises';
-import { fastSkipSerialize, SerializerSymbol } from './verify';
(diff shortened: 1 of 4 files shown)


packages/qwik/src/core/tests/component.spec.tsx
```diff
@@ -242,19 +242,19 @@ describe.each([
           <Component ssr-required>
             <div>
               <span>Component 1</span>
-              <Signal ssr-required>1</Signal>
```
Varixo, line 245: this is wrong, it should be serialized, or we will create it every time we diff

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -510,22 +509,12 @@ export class Serializer {
...
-            // add all back refs to the roots
-            for (let i = 0; i < childVNodeData.length; i++) {
-              const value = childVNodeData[i];
-              if (isSsrAttrs(value)) {
-                const backRefs = tryGetBackRefs(value);
-                if (backRefs) {
-                  this.$serializationContext$.$addRoot$(backRefs);
-                }
-              }
-            }
```
Varixo, line 528: This was correct, the thing to remove is ensureContainsBackRef for ssr

packages/qwik/src/core/tests/component.spec.tsx
wmertens: but it's not reactive, so this is an edge case. If it were reactive, it would ssr as a signal and include the data in the state.

I think it's better to create it on first diff than to send the state.

packages/qwik/src/core/shared/serdes/serialize.ts
wmertens: I don't understand why this is correct. Events are the only entry points to the app, and only the state they can access should be serialized, and everything useful will be discovered from forward state discovery.

packages/qwik/src/core/tests/component.spec.tsx
Varixo: I dont think so, if you have input inside non reactive virtual then it will reset. It has to stay

PR #8469, opened 2026-03-23
chore(core): move optimizer binaries into separate NPM package

wmertens:
This:

- improves repo structure
- speeds up package installs


packages/qwik-vite/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@qwik.dev/qwik-vite",
```
Varixo, line 2: isnt better like `@qwik.dev/vite`?

wmertens: I didn't want to make it confusing with the vite plugin in Router

PR #8481, opened 2026-03-25
fix(core): tasks await their cleanup

### LINKED ISSUE #7985: [🐞] [v2] task callback should wait for async cleanups to resolve

Qwik Runtime

### Describe the bug

Currently task callback doesn't wait for cleanup operation to resolves, this can be an issue if you have async cleanup.
In my case, the cleanup sends a message to a webworker to terminate current process before creating a new instance. But there is a race condition now making it fails in some cases.
```js
useVisibleTask$(({ task, cleanup }) => {
  track(change);
  worker.createInstance();
  cleanup(async () => {
    await worker.terminate();
  })
})
```

### Reproduction

https://qwik.dev/playground/#v=2.0.0-beta.10&f=Q0o0JuYW2BKNDqL2BTPB5RrFKQkSLODSHJhPbRF2aBgAHaaAqJg1qiHVrA64RMkrLVCohRuiAJHSgBoD1qigAKxflJwTc3JAtYISsJwD1XoFwKIKWO0lpQJrt1SgNMgoIL%2B0QAmsBbmloA7Tqw41D2qvBqQQQ3gBBDCqGw1gcQ1SgKgqQSI6CoYGBppQ89Ctg7sFZiEoL4FIpAwHaXMhmhrVEGdAPa4HLItKU7W1a%2B2qUURq4Y210Qw6VDMoAA

### Steps to reproduce

_No response_

### System Info

```shell
.
```

### Additional Information

_No response_

wmertens:
fixes #7985

### DIFF
--- a/packages/qwik/src/core/tests/use-task.spec.tsx
+++ b/packages/qwik/src/core/tests/use-task.spec.tsx
@@ -1134,109 +1134,101 @@ describe.each([
       expect((error as Error).message).toBe('HANDLE ME');
     }
   });
-});
-
-describe('domRender: useTask async cleanup', () => {
   it('should await async cleanup before rerun', async () => {
-    const log: string[] = [];
-    const Counter = component$(() => {
-      const count = useSignal(0);
-      useTask$(({ track, cleanup }) => {
-        const current = track(() => count.value);
-        log.push(`task:${current}`);
-        cleanup(async () => {
-          log.push(`cleanup:${current}:start`);
-          await delay(10);
-          log.push(`cleanup:${current}:end`);
+    vi.useFakeTimers();
+    try {
+      (globalThis as any).log = [];
+      const isSsr = render === ssrRenderToDom;
+      const Counter = component$(() => {
+        const count = useSignal(0);
+        useTask$(({ track, cleanup }) => {
+          const current = track(count);
+          (globalThis as any).log.push(`task:${current}`);
+          cleanup(async () => {
+            (globalThis as any).log.push(`cleanup:${current}:start`);
+            await delay(100);
+            (globalThis as any).log.push(`cleanup:${current}:end`);
+          });
         });
+        return <button onClick$={() => count.value++}>{count.value}</button>;
       });
-      return <button onClick$={() => count.value++}>{count.value}</button>;
-    });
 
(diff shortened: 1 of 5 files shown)


packages/qwik/src/core/tests/use-task.spec.tsx
```diff
@@ -1129,3 +1135,108 @@ describe.each([
...
+
+    await vi.advanceTimersByTimeAsync(1);
+    await triggerPromise;
+    expect(log).toEqual(['task:0', 'cleanup:0:start', 'cleanup:0:end', 'task:1']);
+
+    vi.useRealTimers();
+  });
+});
+
+describe('ssrRenderToDom: useTask async cleanup', () => {
```
Varixo: why separate tests? the tests above should run for both renderers

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -1127,3 +1127,33 @@ describe.each([
     });
   });
 });
+
+describe('domRender: useVisibleTask async cleanup', () => {
```
Varixo: the same here

packages/qwik/src/core/use/utils/destroyable.ts
```diff
@@ -12,3 +15,44 @@ export const cleanupDestroyable = (destroyable: Destroyable) => {
     destroyable.$destroy$ = null;
   }
 };
+
+export const cleanupAsyncDestroyable = (
```
Varixo, line 19: do we need separate function? do we need the above function?

packages/qwik/src/core/use/use-task.ts
```diff
@@ -185,34 +185,63 @@ export const runTask = (
...
+          handleError(err);
+        }
       }
-    }
-  );
+    );
+  });
+
+  if (isPromise(result)) {
+    taskPromise = Promise.resolve(result).finally(() => {
```
Varixo: if its already a promise then why wrap it in another promise?

packages/qwik/src/core/use/utils/destroyable.ts
wmertens: we're using cleanupDestroyable elsewhere, where we don't await async cleanups

PR #8482, opened 2026-03-25
feat(optimizer): automatic loop optimization

Varixo:
Replace .map with Each component in most common cases


Varixo:
closing it for now, we decieded to dont do that

PR #8493, opened 2026-03-26
fix: preserve useTask$ captures when bundled with non-identifier expressions

thejackshelton:
After ping ponging a bunch with Opus 4.6 and Codex, and checking QDS manually with a link we added the following:                                                                                                                                                   
                                                                                                                                                                                                                    
  When rolldown bundles library code, it can inline hook calls like `useSignal(true)` directly into the captures array of `inlinedQrl`. The optimizer's capture extraction only recognized `Ident` expressions,       
  silently dropping call expressions and shifting all subsequent capture indices. This caused `_captures[N]` mismatches at runtime, crashing SSR with `Cannot read properties of undefined`.                          
                                                                                                                                                                                                                      
  ### Example                                                                                                                                                                                                         
                                                           
  Source code:                                                                                                                                                                                                        
  ```tsx                                                   
  const isOpen = useSignal(false);             
  const isInitialLoadSig = useSignal(true);
                                                                                                                                                                                                                      
  useTask$(({ track }) => {                                                                                                                                                                                           
    track(() => isOpen.value);                                                                                                                                                                                        
    if (!isInitialLoadSig.value) props.onChange$?.(isOpen.value);                                                                                                                                                     
    isInitialLoadSig.value = false;                                                                                                                                                                                   
  });                                          
```                 
                                                                                                                                                                                                                    
  Rolldown bundles this into:                                                                                                                                                                                         
```tsx
  inlinedQrl(fn, "symbol", [isOpen, useSignal(true), props])
  //                         ^0     ^1               ^2     
```                                                                                                                                                          
                                                                                                                                                                                                                      
  The optimizer dropped `useSignal(true)` (non-ident), emitting `.w([isOpen, props])`. This is only 2 captures. The function body still read `_captures[2]` for props, which was now `undefined`.                                    
                                                                                                                                                                                                                      
  Changes                                                                                                                                                                                                             
                                                                                                                                                                                                                      
  - Added Captures enum (Auto / Explicit) to distinguish source-level captures from pre-built ones                                                                                                                      
  - Treated explicit captures from inlinedQrl as opaque, emitting them verbatim
  - Centralized capture emission in emit_captures()                                                                                                                                                                    
  - Added regression test with AST assertion


packages/optimizer/core/src/test.rs
```diff
@@ -7036,3 +7036,44 @@ impl TestInput {
...
+		.chars()
+		.filter(|c| !c.is_whitespace())
+		.collect();
+
+	assert!(
+		compact_code.contains(".w([left,true,right])"),
+		"expected transformed capture array [left, true, right] to be preserved, got:\n{}",
+		entry_module.code
+	);
+}
```
Varixo, line 7079: instead can we make it snapshot to see output?

thejackshelton: moved to snapshot test

thejackshelton:
> hmm, the useSignal(true) inside captures doesnt look right, hooks have to be executed directly in component body
> 
> inlinedQrl(fn, "symbol", [isOpen, useSignal(true), props])

Yep, but this is what rolldown injects. The deeper fix would be ensuring rolldown doesn't inline hook calls into capture arrays.

You would have to add `treeshake: false` in the rolldown config to fix it, which disables treeshaking entirely.

packages/optimizer/core/src/test.rs
wmertens: actually it should be both a snapshot and a code test. Snapshots are great for quickly understanding what it does, but they're very easy to update and miss a subtle wrong change.

thejackshelton:
> Looks great, but can you restore the code test while keeping the snapshot test?

Added

PR #8495, opened 2026-03-26
feat(core): add Temporal serialization support

FloezeTv:
# What is it?

- Feature / enhancement

# Description

This PR adds serialization-support for the types in [`Temporal`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Temporal).
See https://github.com/QwikDev/qwik-evolution/discussions/343.
It adds a type-stub and checks for availability of the `Temporal`-type, as it is not yet supported by all browsers (though support is growing).

The tests are only executed if `Temporal` exists / is supported (either by a polyfill or natively by the runtime).

# Checklist

### DIFF
--- a/packages/qwik/src/core/shared/serdes/can-serialize.ts
+++ b/packages/qwik/src/core/shared/serdes/can-serialize.ts
@@ -18,6 +18,7 @@ import { isSerializerObj } from '../../reactive-primitives/utils';
 const getKeyVal = <T>(value: T, key: keyof T) => value[key];
 
 export const canSerialize = (value: unknown, seen: WeakSet<any> = new WeakSet()): boolean => {
+  const hasTemporal = typeof Temporal !== 'undefined';
   if (
     value == null ||
     typeof value === 'string' ||
@@ -71,21 +72,21 @@ export const canSerialize = (value: unknown, seen: WeakSet<any> = new WeakSet())
       return true;
     } else if (value instanceof Date) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.Duration) {
+    } else if (hasTemporal && value instanceof Temporal.Duration) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.Instant) {
+    } else if (hasTemporal && value instanceof Temporal.Instant) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.PlainDate) {
+    } else if (hasTemporal && value instanceof Temporal.PlainDate) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.PlainDateTime) {
+    } else if (hasTemporal && value instanceof Temporal.PlainDateTime) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.PlainMonthDay) {
+    } else if (hasTemporal && value instanceof Temporal.PlainMonthDay) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.PlainTime) {
+    } else if (hasTemporal && value instanceof Temporal.PlainTime) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.PlainYearMonth) {
+    } else if (hasTemporal && value instanceof Temporal.PlainYearMonth) {
       return true;
-    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.ZonedDateTime) {
+    } else if (hasTemporal && value instanceof Temporal.ZonedDateTime) {
       return true;
     } else if (value instanceof RegExp) {
       return true;
--- a/packages/qwik/src/core/shared/serdes/serialize.ts
+++ b/packages/qwik/src/core/shared/serdes/serialize.ts
@@ -67,7 +67,8 @@ export class Serializer {
   private $parent$: SeenRef | undefined;
   private $qrlMap$ = new Map<string, QRLInternal>();
   private $writer$: StreamWriter;
-  private $temporalDefined$: boolean = typeof Temporal !== 'undefined';
+  /** We need to determine this at runtime because polyfills may not be loaded a module load time */
+  private $hasTemporal$ = typeof Temporal !== 'undefined';
 
   constructor(public $serializationContext$: SerializationContext) {
     this.$writer$ = $serializationContext$.$writer$;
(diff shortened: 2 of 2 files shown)


packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -486,6 +486,8 @@ export class Serializer {
       this.output(TypeIds.URL, value.href);
     } else if (value instanceof Date) {
       this.output(TypeIds.Date, Number.isNaN(value.valueOf()) ? '' : value.valueOf());
+    } else if (isTemporal(value)) {
+      this.output(TypeIds.Temporal, [getTemporalType(value), value.toJSON()]);
```
wmertens: this adds about 8-9 bytes of overhead per Temporal value.

Can you instead encode the entire value as a single string? That will also avoid trying to deduplicate a bunch of temporal values.

Is there no canonical string representation that encodes the type and data all at once?
If not, maybe do `${getTemporalType(value)},${value.toJSON()}`, and consider using the numeric value instead of the toJSON?

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -770,6 +772,47 @@ function isObjectLiteral(obj: unknown): obj is object {
...
+  }
+  if (value instanceof Temporal.ZonedDateTime) {
+    return TemporalType.ZonedDateTime;
+  }
+
+  // Not a temporal-type
+};
+
+export const isTemporal = (value: unknown): value is __TemporalStub<unknown> =>
+  getTemporalType(value) !== undefined;
```
wmertens: this seems slow. It should be sufficient that Temporal exists and `value instanceof Temporal`

FloezeTv: That won't work, as `Temporal` is not a class, only the types in `Temporal` (`Temporal.Instant`, `Temporal.Duration`, ...) are.

The alternative would have been to add a `TypeId` for each type in `Temporal`, though that would still require an `instanceof` check for every type.

packages/qwik/src/core/shared/serdes/serialize.ts
FloezeTv: As far as I know, there is no canonical string that encodes type information, and there is no function that would accept it to get back the original type.

I could either add the type as the first char of the string (like you suggested) or split it into different `TypeId`s (one for each type in `Temporal`).
Let me know what you think would be the better solution.

leibale: Not trying to "backseat game", but I think that having a type per temporal class is the way to go:

1. It solves https://github.com/QwikDev/qwik/pull/8495#discussion_r3000166145 "for free"
2. `__TemporalStub` can be removed
3. At some point in the future the `Temporal` namespace might get a new type which won't be similar to the other types, which would require a different type anyway.

The only negative I can think of is having "too many types" in the `TypeIds` enum, but that doesn't feel like a big deal compared to the upside.

FloezeTv: Yeah, good point.
I originally tried to reduce the new types in `TypeIds`, but I guess that should indeed not be that big of a problem.
I have updated the PR accordingly, though I kept `__TemporalStub` to reduce duplication.

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -486,6 +486,22 @@ export class Serializer {
       this.output(TypeIds.URL, value.href);
     } else if (value instanceof Date) {
       this.output(TypeIds.Date, Number.isNaN(value.valueOf()) ? '' : value.valueOf());
+    } else if (typeof Temporal !== 'undefined' && value instanceof Temporal.Duration) {
```
Varixo: typeof Temporal !== 'undefined'
this is always false or always true, right? we can check this once, maybe should be defined somewhere in constructor or module scoped

FloezeTv: Yeah, makes sense.
From my testing, setting it module scoped (in `constants.ts`) only works in `preview`-mode when the polyfill is set up in `root.tsx` (i.e., it does not work in the dev server or when the polyfill is only set up in an `index.tsx`).
I have changed it to a variable in the constructor.

packages/qwik/src/core/shared/serdes/allocate.ts
```diff
@@ -196,3 +220,11 @@ export function retrieveVNodeOrDocument(
...
     : container.element?.ownerDocument;
 }
+
+/** Throws an error if `Temporal` does not exist */
+const assertTemporalExists = () => {
+  if (typeof Temporal === 'undefined') {
+    // Temporal is not supported => polyfill missing
+    throw qError(QError.deserializeErrorMissingType, ['Temporal']);
+  }
+};
```
wmertens: I'm not sure this is needed. Most of the users will have it, and this is adding a bunch of overhead for the few that don't.

If you instead just `return Temporal.ZonedDateTime.from(value as string)`, that'll error as `Uncaught ReferenceError: Temporal is not defined`, which is plenty of info for the dev and equally useless to users as Error Q42.

FloezeTv: Ok, changed.

packages/qwik/src/core/shared/serdes/can-serialize.ts
```diff
@@ -71,6 +71,24 @@ export const canSerialize = (value: unknown, seen: WeakSet<any> = new WeakSet())
       return true;
     } else if (value instanceof Date) {
       return true;
+    } else if (typeof Temporal !== 'undefined') {
```
FloezeTv: I think this will skip serializing all later types when `Temporal` exists, as then this `else if` branch will always be taken, even if the `value` is not one of the `Temporal`-types, though I cannot verify currently, as I'm not home.

wmertens: ah yes you're right hmm

wmertens: ok fixed with a module scoped test. The tests were passing because node doesn't have Temporal

FloezeTv: From my testing in `serialize`, module-scoped tests did always not work when polyfilling `Temporal`.
Checking once at the start of the method would work though.

wmertens: makes sense!

PR #8496, opened 2026-03-27
fix(runtime): throw error when functions are passed as JSX children

### LINKED ISSUE #8425: [🐞] v2: inlined function returning jsx does not render

Qwik Runtime

### Describe the bug

```tsx
import { component$ } from '@qwik.dev/core';

  return (
    <div>
      <h1>Hello from Qwik!</h1>
      {() => <div>hiiii</div>} // <-- does not render
    </div>
  );
});
```

### Reproduction

https://qwikdev-build-v2.qwik-8nx.pages.dev/playground/#f=Q040FgQSDT2bIgoK1RAzwFoyMoEA0gKpxWip4E56hpZIjStSkh6KT0cTHi0SHgA

### Steps to reproduce

_No response_

### System Info

```shell
...
```

### Additional Information

In v1 we had an SSR error instead https://qwik.dev/playground/#f=Q040FgQSDT2bIgoK1RAzwFoyMoEA0gKpxWipjCa9oZn0AA

restareaByWeezy:
# What is it?

- Bug

# Description

Closes #8425

Right now if you do something like `{() => <div>hi</div>}` inside a component, it just... renders nothing. No error, no warning in the console (unless you're in dev mode). Pretty confusing when you're staring at your code wondering where your content went.

In v1 this actually threw an SSR error, which was way more helpful. The v2 `processData()` in both SSR and DOM paths has a catch-all `else` that logs a warning and moves on, so functions just get silently swallowed.

The fix adds a `typeof node === 'function'` check before that catch-all in both `render-ssr.ts` and `render-dom.ts`. Now you get a clear error telling you to either call the function (`{fn()}`) or wrap it with `component$()`.

Dev mode already catches this at JSX creation time via `validateJSXNode`, but that's behind a `qDev` guard and gets stripped in prod. This ensures the error shows up regardless.

## One thing worth discussing

This technically turns a silent no-op into a runtime error. If someone has `{myFn}` in production code today, it renders nothing but doesn't crash. After this change, it would crash.

In practice the impact should be minimal -- dev mode and ESLint both already flag this pattern, so it's unlikely to survive into prod unnoticed. But if you'd prefer a softer approach (like `logError` instead of `throw`), happy to adjust.

# Checklist

### DIFF
--- a/packages/qwik/src/core/render/dom/render-dom.ts
+++ b/packages/qwik/src/core/render/dom/render-dom.ts
@@ -176,6 +176,12 @@ export const processData = (
     return node.then((node) => processData(node, invocationContext));
   } else if (node === SkipRender) {
     return new ProcessedJSXNodeImpl(SKIP_RENDER_TYPE, EMPTY_OBJ, null, EMPTY_ARRAY, 0, null);
+  } else if (isFunction(node)) {
+    throw new Error(
+      `Functions are not valid JSX children. ` +
+        `You might have passed a function instead of calling it: use {fn()} instead of {fn}. ` +
+        `If this is a component, wrap it with component$().`
+    );
   } else {
     logWarn('A unsupported value was passed to the JSX, skipping render. Value:', node);
     return undefined;
--- a/packages/qwik/src/core/render/dom/render.unit.tsx
+++ b/packages/qwik/src/core/render/dom/render.unit.tsx
@@ -44,6 +44,16 @@ test('should only render string/number', async () => {
   await expectRendered(fixture, '<div>string123</div>');
 });
 
+test('should throw error when function is passed as JSX child', async () => {
+  const fixture = new ElementFixture();
+  try {
+    await render(fixture.host, <div>{() => <span>hello</span>}</div>);
+    assert.fail('Expected an error to be thrown');
+  } catch (e: any) {
+    assert.match(e.message, /is not an accepted value|Functions are not valid JSX children/);
+  }
+});
+
 test('should serialize events correctly', async () => {
   const fixture = new ElementFixture();
   await render(
--- a/packages/qwik/src/core/render/ssr/render-ssr.ts
+++ b/packages/qwik/src/core/render/ssr/render-ssr.ts
@@ -923,6 +923,12 @@ const processData = (
   } else if (isPromise(node)) {
     stream.write(FLUSH_COMMENT);
     return node.then((node) => processData(node, rCtx, ssrCtx, stream, flags, beforeClose));
+  } else if (isFunction(node)) {
+    throw new Error(
+      `Functions are not valid JSX children. ` +
+        `You might have passed a function instead of calling it: use {fn()} instead of {fn}. ` +
+        `If this is a component, wrap it with component$().`
+    );
   } else {
     logWarn('A unsupported value was passed to the JSX, skipping render. Value:', node);
     return;
--- a/.changeset/error-function-jsx-children.md
+++ b/.changeset/error-function-jsx-children.md
@@ -1,0 +1,5 @@
+---
+'@builder.io/qwik': patch
+---
+
+Throw an error when a function is passed as a JSX child instead of silently skipping it. This restores the v1 behavior where passing `{myFn}` instead of `{myFn()}` would produce a clear error message.
--- a/packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
+++ b/packages/qwik/src/core/render/ssr/render-ssr.unit.tsx
@@ -1968,6 +1968,11 @@ export const HtmlContext = component$(() => {
   return <Slot />;
 });
 
+test('should throw error when function is passed as JSX child', async () => {
+  const fn = () => jsx('div', { children: 'hello' });
+  await throws(() => testSSR(jsx('body', { children: [fn] }), ''));
+});
+
 async function testSSR(
   node: JSXOutput,
   expected: string | string[],


packages/qwik/src/core/render/dom/render-dom.ts
```diff
@@ -176,6 +176,12 @@ export const processData = (
     return node.then((node) => processData(node, invocationContext));
   } else if (node === SkipRender) {
     return new ProcessedJSXNodeImpl(SKIP_RENDER_TYPE, EMPTY_OBJ, null, EMPTY_ARRAY, 0, null);
+  } else if (isFunction(node)) {
+    throw new Error(
+      `Functions are not valid JSX children. ` +
```
wmertens, line 181: This isn't quite true: functions are valid JSX children, but only when the parent is an Inline Component that will use `children()` somehow.

so the message should be something like `JSX Function values cannot be rendered`.

Furthermore, make sure that the error output is sufficient to find where it is defined. Maybe throwing is enough, but maybe you need to find the file from the dev data, or `console.error` the node or something like that.

packages/qwik/src/core/render/dom/render.unit.tsx
```diff
@@ -44,6 +44,16 @@ test('should only render string/number', async () => {
   await expectRendered(fixture, '<div>string123</div>');
 });
 
+test('should throw error when function is passed as JSX child', async () => {
+  const fixture = new ElementFixture();
+  try {
+    await render(fixture.host, <div>{() => <span>hello</span>}</div>);
+    assert.fail('Expected an error to be thrown');
+  } catch (e: any) {
+    assert.match(e.message, /is not an accepted value|Functions are not valid JSX children/);
```
wmertens, line 53: here, ensure that it's easy to find where the wrong definition was.

packages/qwik/src/core/render/ssr/render-ssr.ts
```diff
@@ -923,6 +923,12 @@ const processData = (
   } else if (isPromise(node)) {
     stream.write(FLUSH_COMMENT);
     return node.then((node) => processData(node, rCtx, ssrCtx, stream, flags, beforeClose));
+  } else if (isFunction(node)) {
+    throw new Error(
+      `Functions are not valid JSX children. ` +
```
wmertens, line 928: same as DOM

wmertens (changes requested):

Varixo:
Thanks for the PR, but I think we should try to render the example from the issue as inline component.
Btw this is v1 code, so this PR doesnt solve the issue

restareaByWeezy:
@Varixo thanks for the feedback. Just to clarify the direction -- when you say "render as inline component", do you mean the runtime should automatically invoke `{() => <div/>}` and render its output (like how `<Func />` works), or that we should guide users to write it as `<Func />` and throw a clear error when a raw function is passed as a child?

Varixo:
> @Varixo thanks for the feedback. Just to clarify the direction -- when you say "render as inline component", do you mean the runtime should automatically invoke `{() => <div/>}` and render its output (like how `<Func />` works), or that we should guide users to write it as `<Func />` and throw a clear error when a raw function is passed as a child?

if you go to the playground and open segments for this example, you can see this:
```ts
export const app_component_QbmWc3TlyqA = ()=>{
    return /*#__PURE__*/ _jsxSorted("div", null, null, [
        /*#__PURE__*/ _jsxSorted("h1", null, null, "Hello from Qwik!", 3, null, {
            fileName: "../app.tsx",
            lineNumber: 6,
            columnNumber: 7
        }),
        ()=>/*#__PURE__*/ _jsxSorted("div", null, null, "hiiii", 3, "ga_0", {
                fileName: "../app.tsx",
                lineNumber: 7,
                columnNumber: 14
            })
    ], 3, "ga_1", {
        fileName: "../app.tsx",
        lineNumber: 5,
        columnNumber: 5
    });
```
This means the optimizer is wrong, because it should be wrapped like this:
```ts
export const app_component_QbmWc3TlyqA = ()=>{
    return /*#__PURE__*/ _jsxSorted("div", null, null, [
        /*#__PURE__*/ _jsxSorted("h1", null, null, "Hello from Qwik!", 3, null, {
            fileName: "../app.tsx",
            lineNumber: 6,
            columnNumber: 7
        }),
        /*#__PURE__*/ _jsxSorted(()=>/*#__PURE__*/ _jsxSorted("div", null, null, "hiiii", 3, "ga_0", {
                fileName: "../app.tsx",
                lineNumber: 7,
                columnNumber: 14
            }), null, null, null, 3, "ga_1", {
            fileName: "../app.tsx",
            lineNumber: 9,
            columnNumber: 7
        })
    ], 3, "ga_1", {
        fileName: "../app.tsx",
        lineNumber: 5,
        columnNumber: 5
    });
```
(this is just an example so lineNumber, and columnNumber will be different)

And if the optimizer will do this, then everything should be fine without js changes

but if we want to fix it in v2 then this PR should be closed and you need to do changes againt build/v2 branch.

wmertens:
@Varixo so you think we should treat `{fn}` as an inline component and render it without props? I don't really like that, what is the usecase?

I think `{() => {...}}` is just a kind of typo and it should error

Varixo:
> @Varixo so you think we should treat `{fn}` as an inline component and render it without props? I don't really like that, what is the usecase?
> 
> I think `{() => {...}}` is just a kind of typo and it should error

then probably the optimizer should throw it to not do it runtime

maiieul:
I'm thinking `{() => {...}}` in jsx is just valid jsx so we should support it? Use-case is just for devs to be able to write inline components within QRL components without having to create a separate function. 

Closing for now as stale.

PR #8501, opened 2026-03-28
q-data.json: split per loader

wmertens:
Big sanity improvement, some breaking changes:
Route loaders now each get their own cacheable .json path, and SPA navigation no longer needs to contact the server on every click.

Breaking:
- Routeloaders are now Async Signals, and the `loader.value.failed` in-band data is no longer available. Instead, read from `loader.error`.
- Routeloaders no longer resolve action results. The reason is that routeloaders are supposed to be cacheable and having them differ based on actions is wrong.
  There are several other ways to use action results, so this isn't limiting.

The net result of this change:
- less server traffic
- no incorrect q-data.json caches
- saner API


.changeset/green-experts-turn.md
```diff
@@ -0,0 +1,3 @@
+'@qwik.dev/router': minor
+
+Refactor route loaders to be backed by shared async signals across SSR, client refresh, and action invalidation.
```
Varixo: format here is wrong

packages/qwik-router/src/buildtime/runtime-generation/generate-routes.ts
```diff
@@ -314,6 +317,40 @@ function serializeBuildTrie(
     }
   }
 
+  // Emit _R: routeLoader$ hashes for this node.
+  // In dev mode (loadersByFile populated after invalidation), emit directly.
+  // In build mode, emit placeholder string for renderChunk replacement.
+  {
+    const routeFiles = node._files
+      .filter((f) => f.type === 'route' || f.type === 'layout')
+      .map((f) => f.filePath);
```
Varixo, line 326: how much performance matters here?

packages/qwik-router/src/middleware/request-handler/handlers/action-handler.ts
```diff
@@ -0,0 +1,215 @@
...
+      if (err instanceof ServerError) {
+        actionError = err;
+      } else if (err instanceof Error) {
+        console.error('Action error:', err);
+        actionError = new ServerError(500, 'Internal Server Error');
+      } else {
+        // RedirectMessage, AbortMessage, etc. — re-throw for middleware
+        throw err;
+      }
+    }
```
Varixo: maybe wrap with something like executeAction?

packages/qwik-router/src/middleware/request-handler/handlers/action-handler.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    requestEv.sharedMap.set('@actionResult', actionError ?? actionData);
+
+    // Build response envelope: d=data, e=error, s=status, h=hashes, l=loaders
+    const responseData: Record<string, unknown> = {
+      ...(actionError ? { e: actionError } : { d: actionData }),
+      s: actionError ? actionError.status : requestEv.status(),
+    };
```
Varixo: This is like a new response format?

packages/qwik-router/src/middleware/request-handler/handlers/action-handler.ts
```diff
@@ -0,0 +1,215 @@
...
+    if (e instanceof Error && qrl.dev) {
+      (e as any).loc = qrl.dev;
+    }
+    throw e;
+  }
+}
+
+function now() {
+  return typeof performance !== 'undefined' ? performance.now() : 0;
+}
```
Varixo: this should be moved to some utils

packages/qwik-router/src/middleware/request-handler/handlers/loader-handler.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    // ETag support: for string/function eTags, check If-None-Match BEFORE running the loader
+    if (loader.__eTag && loader.__eTag !== true) {
+      const eTag = resolvePreETag(loader.__eTag, requestEv);
+      if (eTag && checkETagMatch(requestEv, eTag)) {
+        return;
+      }
+    }
+
+    const responseData = await getRouteLoaderResponse(loader.__qrl, loader.__validators, requestEv);
```
Varixo, line 45: getRouteLoaderResponse doesnt tell me a lot. I see what it returns, maybe better name it like executeLoader?

packages/qwik-router/src/middleware/request-handler/handlers/loader-handler.ts
```diff
@@ -0,0 +1,204 @@
...
+ * Early handler that wraps `next()` for JSON API requests (q-loader and q-action).
+ *
+ * For `IsQLoader` requests, it also rewrites the URL using the `X-Qwik-fullpath` header so that
+ * downstream middleware sees the real page URL.
+ *
+ * By calling `await next()` inside a try/catch, middleware redirects and errors are captured and
+ * returned as JSON envelopes instead of HTTP redirects/error pages. This keeps SPA navigation
+ * intact on the client.
+ */
+export function jsonRequestWrapper(): RequestHandler {
```
Varixo: is it related to route loader? I see it is using loader and action, so maybe should be in separate file?

packages/qwik-router/src/middleware/request-handler/request-path.ts
```diff
@@ -1,19 +1,53 @@
...
-      pathname.length - QDATA_JSON.length + (globalThis.__NO_TRAILING_SLASH__ ? 0 : 1);
-    pathname = pathname.slice(0, trimEnd);
-    if (pathname === '') {
-      pathname = '/';
+export function recognizeRequest(pathname: string): {
+  type: typeof IsQLoader;
+  trimLength: number;
+  data: { loaderId?: string; manifestHash?: string } | null;
+} | null {
+  // Quick length check for common cases
```
Varixo: maybe some interface as return type for nicer code?

packages/qwik-router/src/middleware/request-handler/request-event-core.ts
```diff
@@ -75,12 +66,26 @@ export function createRequestEventWithDeps(
...
+    const trimmed = url.pathname.slice(0, url.pathname.length - recognized.trimLength);
+    url.pathname = trimmed || '/';
+    if (!globalThis.__NO_TRAILING_SLASH__ && !url.pathname.endsWith('/')) {
+      url.pathname += '/';
+    }
+    sharedMap.set(recognized.type, true);
+    if (recognized.data?.loaderId) {
+      sharedMap.set(deps.QLoaderId, recognized.data.loaderId);
+    }
+  }
```
Varixo, line 79: there is already very similar logic in `trimInternalPathname`

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
```diff
@@ -239,40 +238,69 @@ export function createResolveRequestHandlers(deps: ResolveRequestHandlersDeps) {
...
+              }
+            } catch (err) {
+              if (err instanceof deps.ServerError) {
+                actionResult = err;
+              } else if (err instanceof Error) {
+                console.error('Action error:', err);
+                actionResult = new deps.ServerError(500, 'Internal Server Error');
+              } else {
+                // RedirectMessage, AbortMessage, etc. — re-throw for middleware
+                throw err;
```
Varixo: this is the same logic like in action-handler? if we move the logic there to something like `executeAction` then we can reuse it here

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers.unit.ts
```diff
@@ -46,18 +46,7 @@ function createMockRequestEvent(url = 'http://localhost:3000/test', trailingSlas
...
-    it('should remove q-data.json', () => {
-      globalThis.__NO_TRAILING_SLASH__ = false;
-      expect(getPathname(new URL('http://server/path/q-data.json?foo=bar#hash'))).toBe(
-        '/path/?foo=bar#hash'
-      );
-      globalThis.__NO_TRAILING_SLASH__ = true;
-      expect(getPathname(new URL('http://server/path/q-data.json?foo=bar#hash'))).toBe(
-        '/path?foo=bar#hash'
-      );
-    });
```
Varixo, line 58: probably we should add similar test for the new qaction or qloader?

packages/qwik-router/src/middleware/request-handler/static-paths.ts
```diff
@@ -21,12 +21,11 @@ export function isStaticPath(method: string, url: URL) {
...
     return true;
   }
-  if (p.endsWith('/q-data.json')) {
-    const pWithoutQdata = p.replace(/\/q-data.json$/, '');
-    if (staticPaths.has(pWithoutQdata + '/')) {
-      return true;
-    }
-    if (staticPaths.has(pWithoutQdata)) {
+  // Per-loader files: q-loader-{id}.{hash}.json
+  const loaderMatch = p.match(/\/q-loader-[^.]+\.[^.]+\.json$/);
```
Varixo: I see this regexp for sure second time :D

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+      }
+      if (response.e) {
+        // Error — throw so AsyncSignal enters error state
+        throw response.e;
+      }
+      return response.d;
+    },
+
+    {
+      serializationStrategy: loader.__serializationStrategy,
```
Varixo, line 364: here is missing `|| DEFAULT_LOADERS_SERIALIZATION_STRATEGY`, right?

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+ *
+ * They are represented by an AsyncSignal.
+ */
+
+const REQUEST_ROUTE_LOADER_STATE = '@routeLoaderState';
+const REQUEST_LOADER_PATHS_STORE = '@loaderPathsStore';
+const REQUEST_ROUTE_LOADERS = '@routeLoaders';
+
+/** Header name sent by client to tell the server the actual page URL for loader requests. */
+export const FULLPATH_HEADER = 'X-Qwik-fullpath';
```
Varixo, line 55: we need to make sure we dont create any vulnerability here, like chaning header content or something like that

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+    return undefined;
+  }
+  // Ensure the route path includes the base path (root trie loaders get '/' but
+  // need the full base path for fetching)
+  let resolvedPath = routePath;
+  const basePath = opts?.basePath ?? '/';
+  if (basePath !== '/' && !resolvedPath.startsWith(basePath)) {
+    resolvedPath = basePath + resolvedPath.slice(1);
+  }
+  const pathBase = resolvedPath.endsWith('/') ? resolvedPath : resolvedPath + '/';
```
Varixo: I feel like I see this logic n time already

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+  const pageUrl = opts?.pageUrl;
+  const search = pageUrl?.search ?? '';
+  // TODO allowList search params + compat flag that allows search params
+  const url = `${pathBase}q-loader-${loaderId}.${manifestHash}.json${search}`;
```
Varixo: maybe we should have util like get loader name, its already the same in few places

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+      interval: getLoaderInterval(loader.__expires, loader.__poll),
+    }
+  );
+};
+
+/** Build a sorted, stable search string from only the allowed param names. */
+const filterSearchParams = (params: URLSearchParams, allowed: string[]): string => {
+  const filtered = new URLSearchParams();
+  for (const name of allowed) {
+    for (const value of params.getAll(name)) {
```
Varixo: for of can be replaced with for i

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+}
+
+/** Update the loader paths store on client-side navigation. */
+export const updateRouteLoaderCtx = (
+  ctx: RouteLoaderCtx,
+  loaderPaths: Record<string, string> | undefined,
+  pageUrl: URL
+) => {
+  ctx.pageUrl = pageUrl;
+  for (const key of Object.keys(ctx.loaderPaths)) {
```
Varixo: Object.keys create an array, we should avoid this, better use for in

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+    if (!loaderPaths?.[key]) {
+      ctx.loaderPaths[key] = undefined;
+    }
+  }
+  // Add/update paths
+  if (loaderPaths) {
+    for (const [key, value] of Object.entries(loaderPaths)) {
```
Varixo: Object.entries also create an array

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+      ctx.loaderPaths[key] = value;
+    }
+  }
+};
+
+export const getModuleRouteLoaders = (mods: readonly (RouteModule | undefined)[]) => {
+  const routeLoaders: LoaderInternal[] = [];
+  const seen = new Set<string>();
+  for (const mod of mods) {
```
Varixo: for i can be used

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+    if (!mod) {
+      continue;
+    }
+    for (const value of Object.values(mod)) {
```
Varixo: if mod is not a map or set then we shoud use for in

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+  return (state[loader.__id] ||= createRouteLoaderSignal(loader, routeLoaderCtx));
+};
+
+export const ensureRouteLoaderSignals = (
+  mods: readonly (RouteModule | undefined)[],
+  state: Record<string, AsyncSignal<unknown>>,
+  routeLoaderCtx: RouteLoaderCtx
+) => {
+  const loaders = getModuleRouteLoaders(mods);
+  for (const loader of loaders) {
```
Varixo: for i

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+function verifySerializable(data: any, qrl: QRL) {
+  try {
+    _verifySerializable(data, undefined);
+  } catch (error: any) {
+    if (error instanceof Error && qrl.dev) {
+      (error as any).loc = qrl.dev;
+    }
+    throw error;
+  }
+}
```
Varixo, line 768: this should be exported as util probably

packages/qwik-router/src/runtime/src/route-loaders.ts
```diff
@@ -0,0 +1,559 @@
...
+  loader.__eTag = eTag;
+  loader.__search = search;
+  Object.freeze(loader);
+  return loader;
+}) as LoaderConstructorQRL;
+
+/** @public */
+export const routeLoader$: LoaderConstructor = /*#__PURE__*/ implicit$FirstArg(routeLoaderQrl);
+
+async function runValidators(
```
Varixo, line 737: similar function in `action-handler.ts`

packages/qwik-router/src/runtime/src/routing.ts
```diff
@@ -224,12 +226,23 @@ function collectNodeMeta(
...
+  matchedPathname = '/'
 ) {
   for (const g of groups) {
     if (g._L) {
       layouts.push(g._L);
     }
+    if (g._R && loaderHashes) {
+      loaderHashes.push(...g._R);
+      if (loaderPathsByHash) {
+        for (const hash of g._R) {
```
Varixo: avoid for of

packages/qwik-router/src/runtime/src/routing.ts
```diff
@@ -243,6 +256,14 @@ function collectNodeMeta(
   if (node._L) {
     layouts.push(node._L);
   }
+  if (node._R && loaderHashes) {
+    loaderHashes.push(...node._R);
+    if (loaderPathsByHash) {
+      for (const hash of node._R) {
```
Varixo: and here too. I wont make more comments about for of, maybe we should enable the eslint rule for qwik router too

packages/qwik-router/src/runtime/src/types.ts
```diff
@@ -455,6 +450,8 @@ export interface QwikRouterEnvData {
   params: PathParams;
   response: EndpointResponse;
   loadedRoute: LoadedRoute;
+  routeLoaderCtx: import('./route-loaders').RouteLoaderCtx;
```
Varixo: we we cant import this type normally?

packages/qwik-router/src/runtime/src/utils.unit.ts
```diff
@@ -63,29 +62,6 @@ import {
...
-  { pathname: '/about', search: '?foo=bar', expect: '/about/q-data.json?foo=bar' },
-  { pathname: '/about/', search: '?foo=bar', expect: '/about/q-data.json?foo=bar' },
-  { pathname: '/about/', search: '?foo=bar&baz=qux', expect: '/about/q-data.json?foo=bar&baz=qux' },
-].forEach((t) => {
-  test(`getClientEndpointUrl("${t.pathname}", "${t.search}")`, () => {
-    const endpointPath = getClientDataPath(t.pathname, t.search);
-    assert.equal(endpointPath, t.expect);
-  });
-});
-
```
Varixo, line 88: should we move these tests?

Varixo (commented):
Great work! Added some comments/questions

packages/qwik-router/src/buildtime/runtime-generation/generate-routes.ts
wmertens: not much, it's build time, so a few ms won't matter

packages/qwik-router/src/middleware/request-handler/handlers/action-handler.ts
wmertens: yes, {d, e, s, r} {data, error, status, redirect}
which means we can send data together with redirect if we want to (like for which routeloaders to invalidate, but not needed right now I think)

packages/qwik-router/src/runtime/src/route-loaders.ts
wmertens: it's normalized by the option processing, always present

wmertens:
there's still an issue with http://localhost:3300/qwikrouter-test/loader-redirect/, on firefox it does not remove the old index when redirecting

packages/qwik-router/src/buildtime/vite/plugin.ts
```diff
@@ -61,6 +72,107 @@ export function qwikRouter(userOpts?: QwikRouterVitePluginOptions): PluginOption
...
+      }
+    }
+    if (hashes.length > 0) {
+      return `_R: ${JSON.stringify(hashes)},`;
+    }
+    // Trailing commas inside object literals are legal, so removing a mid-object entry
+    // (and the trailing comma it emitted with) leaves the surrounding trie literal valid.
+    return '';
+  });
+}
```
Varixo, line 99: this looks weird, is it safe to do that? why we need it?

packages/qwik-router/src/runtime/src/document-head-tags-component.tsx
```diff
@@ -25,21 +25,23 @@ import type { DocumentHeadValue } from './types';
...
-  let head = useDocumentHead();
-  if (props) {
-    head = { ...head, ...props };
-  }
+  const head = useDocumentHead();
+  const title = props.title ?? head.title;
+  const meta = props.meta ?? head.meta;
+  const links = props.links ?? head.links;
+  const styles = props.styles ?? head.styles;
+  const scripts = props.scripts ?? head.scripts;
```
Varixo: this will create a lot of component level subscriptions

packages/qwik-router/src/runtime/src/qwik-router-component.tsx
```diff
@@ -157,43 +212,21 @@ export const useQwikRouter = (props?: QwikRouterProps) => {
...
-    const loadersSerializationObject: Record<string, unknown> = {};
-    for (const [k, v] of Object.entries(obj)) {
-      loadersSerializationObject[k] = getSerializationStrategy(k) === 'always' ? v : _UNINITIALIZED;
+  const navResolver: { r?: () => void; p?: Promise<void> } = {};
+  // deep: true so that changes to loaderPaths and page path/search properties are tracked by
+  // AsyncSignal QRLs.
+  const routeLoaderCtx = useStore(env.routeLoaderCtx);
+  // Create AsyncSignals whose QRL closures capture the store proxy for client-side reactivity.
+  // Then set .value from middleware-computed loader values (inert, non-reactive data).
+  const loaderState = useStore<Record<string, AsyncSignal<unknown>>>({}, { deep: false });
```
Varixo: why useStore? with deep false its just Record?

packages/qwik/src/core/shared/cursor/chore-execution.ts
```diff
@@ -198,13 +198,21 @@ export function executeComponentChore(
...
     OnRenderProp
   );
 
+  if (componentQRL && !componentQRL.resolved) {
+    /**
+     * If the component QRL is not resolved, we need to wait for it to resolve before we can execute
+     * it. We just return the promise and we'll be called again, possibly with updated props, which
+     * is why we don't clear the dirty bit yet.
+     */
+    return componentQRL.resolve() as any as Promise<void>;
```
Varixo: do we need this ugly cast?

Varixo (commented):
Great job!

packages/qwik-router/src/buildtime/vite/plugin.ts
wmertens: yes, we only know what the routeLoaders are after we do the build

packages/qwik-router/src/runtime/src/document-head-tags-component.tsx
wmertens: It should, shouldn't it?

wmertens: ok reverted, tests still pass

PR #8503, opened 2026-03-29
feat(router): small features and fixes

wmertens:
bugs:
- prevent crashing due to container missing during navigation
- don't append `/` to paths that are not known
- remove `qwikRouterConfig` from router creation, it's entirely internally managed

features:
- async signal now accepts negative interval to just invalidate but not recalc
- documentHead now includes the manifesthash

### DIFF
--- a/packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
+++ b/packages/qwik/src/core/reactive-primitives/impl/async-signal-impl.ts
@@ -294,7 +294,7 @@ export class AsyncSignalImpl<T>
   set interval(value: number) {
     this.$clearNextPoll$();
     this.$interval$ = value;
-    if (this.$interval$ > 0 && this.$effects$?.size) {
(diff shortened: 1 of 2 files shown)


packages/qwik/src/core/reactive-primitives/signal.public.ts
```diff
@@ -103,8 +103,11 @@ export interface AsyncSignal<T = unknown> extends ComputedSignal<T> {
    */
   error: Error | undefined;
   /**
-   * Poll interval in ms. Writable and immediately effective when the signal has consumers. If set
-   * to `0`, polling stops.
+   * Staleness/poll interval in ms. Writable and immediately effective.
+   *
+   * - **Positive**: Poll — re-compute after this many ms when subscribers exist.
+   * - **Negative**: Stale-only — mark stale after `|interval|` ms, no auto-recompute.
```
Varixo, line 109: we should probably add test for this

PR #8510, opened 2026-03-31
fix(preloader): yield preload scheduling work

Varixo:
This adds the preloader scheduling so browser do time-slicing the real DOM insertion.
It also keeps SSR preload queue generation synchronous, so server-rendered preload hints are still produced in the same turn.

### DIFF
--- a/scripts/validate-benchmark-memory.ts
+++ b/scripts/validate-benchmark-memory.ts
@@ -310,12 +310,17 @@ async function bundleMemoryScenarios(): Promise<string> {
         name: 'qwik-memory-stubs',
         setup(buildApi) {
           buildApi.onResolve({ filter: /^@qwik\.dev\/core$/ }, () => ({
-            path: resolve(process.cwd(), 'packages/qwik/dist/core.mjs'),
-            external: true,
+            path: resolve(process.cwd(), 'packages/qwik/src/core/index.ts'),
           }));
           buildApi.onResolve({ filter: /^@qwik\.dev\/core\/internal$/ }, () => ({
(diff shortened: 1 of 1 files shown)


packages/qwik/src/core/preloader/queue.ts
```diff
@@ -1,25 +1,49 @@
...
+import { config, doc, loadStart, rel, yieldInterval } from './constants';
+import type { BundleImport, BundleImports, ImportProbability } from './types';
 import {
   BundleImportState_Loaded,
   BundleImportState_None,
   BundleImportState_Preload,
   BundleImportState_Queued,
 } from './types';
 import type { QwikSymbolEvent } from '../shared/jsx/types/jsx-qwik-events';
+import { createMacroTask } from '../shared/platform/next-tick';
```
Varixo, line 12: for now import from core, we can copy that when inline is needed

packages/qwik/src/core/preloader/queue.ts
```diff
@@ -102,15 +141,145 @@ export const trigger = () => {
...
+  // Keep existing work on the stack hot and append new roots behind it.
+  adjustmentStack.unshift({
+    $bundle$: bundle,
+    $inverseProbability$: inverseProbability,
+    $seen$: seen,
+    $context$: context,
+  });
 };
 
+const processAdjustmentFrame = () => {
```
Varixo, line 161: this is moved from below, not a new code

maiieul:
LGTM! Can you make another PR for v1? Since it's critical code I'd like that we keep the code as close as possible between the two versions.

PR #8515, opened 2026-04-02
chore: improve view transition time out error message

maiieul:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


maiieul:
I think it's important in prod too and 99% of the time the devs behind the app will play with prod with disableCache enabled. A good a error message will save them time trying to debug something that is not really a bug.

Varixo:
> I think it's important in prod too and 99% of the time the devs behind the app will play with prod with disableCache enabled. A good a error message will save them time trying to debug something that is not really a bug.



> I think it's important in prod too and 99% of the time the devs behind the app will play with prod with disableCache enabled. A good a error message will save them time trying to debug something that is not really a bug.

Maybe better short message where to find explanation? Like docs link or something like that?

maiieul:
Ok I made it shorter.

PR #8519, opened 2026-04-03
docs: style fixes

Varixo:


packages/docs/src/components/action/action.tsx
```diff
@@ -27,7 +29,12 @@ type LinkProps = PropsOf<'a'> & {
 
 const getButtonClasses = (variant: ButtonVariant) => ({
   faceClass: button[variant],
-  shadowClass: variant === 'primary' ? button.primaryShadow : button.secondaryShadow,
+  shadowClass:
```
thejackshelton-kunaico, line 32: maybe better to have regular if here?

packages/docs/src/components/header/desktop-header.tsx
```diff
@@ -88,31 +89,37 @@ export const DesktopHeader = component$(() => {
...
+      hidden.value = false;
+    }
+    lastScrollY.value = y;
+  });
+
+  useOnDocument('scroll', toggleNavbar);
+
+  useVisibleTask$(
+    () => {
+      toggleNavbar();
```
thejackshelton-kunaico, line 124: Why do we need this?

packages/docs/src/global.css
```diff
@@ -76,49 +86,49 @@
 
   /* Text — Headings */
   --text-h1: 96px;
-  --text-h1--line-height: 116.667%;
+  --text-h1--line-height: 112px;
```
thejackshelton-kunaico: Is this directly from design system? That is what the percentages were from

Varixo: Yes,  px is the same

packages/docs/src/components/header/desktop-header.tsx
Varixo: After scrolled refresh the navbar was wrong

packages/docs/src/global.css
thejackshelton-kunaico: I'd rather have % because if the font size gets changed the percentage scales with it. Now you would have to change both

packages/docs/src/components/action/action.tsx
Varixo: I am going to merge this, we can fix it later

PR #8540, opened 2026-04-08
fix(qwikloader): preserve synchronous event modifiers and async event order

Varixo:
- prevent default and stop propagation should be called in sync way
- event handlers should be executed in event calling order


wmertens (approved):
LGTM

PR #8544, opened 2026-04-09
fix(router): SPA routing issue with back/forward navigation after manual refresh

Varixo:
Fix browser  back/forward navigation after manual refresh.
Before this change, if a page was manually refreshed and the user then used browser history navigation, the URL could change without rerendering the page, or the router could fall back to an extra document reload during pre-resume recovery.

### DIFF
--- a/.changeset/long-cooks-joke.md
+++ b/.changeset/long-cooks-joke.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: fix an SPA routing bug where using browser back/forward after a manual refresh could change the URL without rerendering the page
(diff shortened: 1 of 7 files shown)


wmertens (approved):
LGTM! 🎩

jordanw66:
Is it necessary to remove the check for `scrollRestoration === "manual"`? That gates `spaInit` and prevents it from running if the page never was SPA. Without it, `spaInit` will run on every page unless the user explicitly specifies `noSPA` config.

Also is there a reason to set `h.scrollRestoration = 'manual';`? The only time `_qRouterScroll` will exist is when `scrollRestoration` is already set to manual which happens here:
https://github.com/QwikDev/qwik/blob/4250f52e16c9dc2deb93c8a93366970f515e31c6/packages/qwik-router/src/runtime/src/qwik-router-component.tsx#L616

Some things have changed since I made this stuff so I could be wrong but I think this is still relevant.

PR #8554, opened 2026-04-13
fix(routing): prefix roots should point to root

### LINKED ISSUE #8504: [🐞] v2: `rewriteRoutes` root paths behavior change in v2 beta 30

Qwik Runtime

### Describe the bug

Continuation of: https://github.com/QwikDev/qwik/issues/8471#issuecomment-4141689340

Up until v2.0.0-beta.28, [rewriteRoutes](https://qwikdev-build-v2.qwik-8nx.pages.dev/docs/routing/#rewrite-routes) did not just rewrite the routes inside the defined paths, but also the root path of the prefix.

Take this simple configuration for example:

```ts
{
  rewriteRoutes: [{
    prefix: "en",
    "paths": {
      "contact": "contact"
    }
  }, {
    prefix: "de",
    "paths": {
      "contact": "kontakt"
    }
  }]
}
```

This previously not only created the explicit rewrites of the defined page:

- `/en/contact/` -> `routes/contact/index.tsx`
- `/de/kontakt/` -> `routes/contact/index.tsx`

but also implicitly rewrote the root path of the prefix:

- `/en/` -> `routes/index.tsx`
- `/de/` -> `routes/index.tsx`

With v2.0.0-beta.30, the implicit rewrite of the root path of the prefix was removed. 

The proposed way forward is to use dynamic routes. I could not get simple dynamic routes like `[lang]` to work ([reproduction in the "dynamic" branch](https://github.com/schoero/qwik-rewrite-routes-repro/tree/dynamic)) but I could get them to work with a "catchAll" route like `[...lang]`.

Either way I see a few issues with that approach:

- It requires special handling for only the root paths via a dynamic route. All other localized paths are handled via the `rewriteRoutes` configuration and do not belong in the dynamic route.

```diff
  routes/
+  ├─ [...lang]/
+  │  └─ index.tsx
   ├─ contact/
   │  └─ index.tsx
-  └─ index.tsx
```

- It also catches other paths like `/doesnotexist/`, forcing the user to manually return a 404 for actually non-existing paths.
- If the simple dynamic route `[lang]` would work, it would require the user to either still keep a duplicate `routes/index.tsx` file to catch the `/` route and a dynamic `[lang]` route for the localized root paths.

As I understand, the whole point of the `rewriteRoutes` feature is to avoid having to duplicate routes for different locales and not having to use dynamic routes for that. The new behavior defeats that purpose.

### Proposed solution

While I agree that the previous implicit (and I think undocumented) behavior may not be ideal, I think it would still be worth it to have an option to rewrite the root path of the prefix as well, without having to use a dynamic route for that. 
Maybe a new option in the `rewriteRoutes` configuration like `rewriteRoot` or the ability to rewrite the "" route.

```ts
{
  rewriteRoutes: [{
    prefix: "en",
    // new option
    rewriteRoot: true,
    "paths": {
      // ability to rewrite the "" route
      "": "",
      // alternatively with "/"
      "/": "/",
      "contact": "contact"
    }
  }]
}
```

### Reproduction

https://github.com/schoero/qwik-rewrite-routes-repro

### Steps to reproduce

Reproduction repository: https://github.com/schoero/qwik-rewrite-routes-repro

This contains two branches:
- The `dynamic` branch shows the non-working setup with a dynamic `[lang]` route.
- The `catchAll` branch shows the working, but complicated setup with a catchAll `[...lang]` route.

### System Info

```shell
System:
    OS: macOS 26.4
    CPU: (18) arm64 Apple M5 Pro
    Memory: 2.59 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 25.8.1 - /Users/schoero/.volta/tools/image/node/25.8.1/bin/node
    npm: 11.11.0 - /Users/schoero/.volta/tools/image/node/25.8.1/bin/npm
  Browsers:
    Chrome: 147.0.7727.24
    Firefox: 148.0.2
    Safari: 26.4
  npmPackages:
    typescript: 5.9.3 => 5.9.3 
    vite: 7.3.1 => 7.3.1
```

### Additional Information

_No response_

wmertens:
fixes #8504

### DIFF
--- a/packages/qwik-router/src/buildtime/runtime-generation/generate-routes.ts
+++ b/packages/qwik-router/src/buildtime/runtime-generation/generate-routes.ts
@@ -198,7 +198,7 @@ function serializeBuildTrie(
   }
 
   // _G rewrite target
-  if (node._G) {
+  if (node._G != null) {
     lines.push(`${nextIndent}_G: ${JSON.stringify(node._G)},`);
   }
 
--- a/packages/qwik-router/src/runtime/src/routing.ts
+++ b/packages/qwik-router/src/runtime/src/routing.ts
@@ -164,7 +164,7 @@ function resolveLoaders(
   node: RouteData,
   gatheredLayouts: ModuleLoader[]
 ): ModuleLoader[] | undefined {
-  if (node._G) {
+  if (node._G != null) {
     // Rewrite: re-walk from root using _G's keys
     const keys = (node._G as string).split('/').filter((p) => p.length > 0);
     const target = walkTrieKeys(root, keys);
@@ -176,7 +176,7 @@ function resolveLoaders(
 
     // If the target node doesn't have _I directly, check group children
     // (e.g., root index inside (common) group, or routes inside pathless groups)
-    if (!targetNode._I && !targetNode._G) {
+    if (!targetNode._I && targetNode._G == null) {
       const indexResult = findIndexNode(targetNode);
       if (indexResult) {
         for (const g of indexResult.groups) {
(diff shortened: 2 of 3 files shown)


Varixo (approved):
LGTM

PR #8557, opened 2026-04-14
fix: honor node streaming backpressure

Varixo:
This PR makes Qwik SSR honor downstream backpressure during Node streaming. The core renderer now waits for pending async flushes before emitting the next chunk, and the Node middleware translates ServerResponse.write() / drain semantics into an awaitable stream signal instead of fire-and-forget writes.

It also adds a regression test in render-api.spec.tsx to ensure a second chunk is not emitted while the first async write is still pending, plus Node middleware regression coverage so handled errors are not rethrown after headers were already sent.

Performance testing shows time drops from ~250ms to ~40ms for response

### DIFF
--- a/packages/qwik-router/src/middleware/node/http.unit.ts
+++ b/packages/qwik-router/src/middleware/node/http.unit.ts
@@ -1,5 +1,16 @@
-import { assert, test } from 'vitest';
-import { normalizeUrl } from './http';
+import {
+  getPlatform,
+  setPlatform,
+  type JSXOutput,
+  type StreamWriter,
+} from '@qwik.dev/core/internal';
+import { renderToStream } from '@qwik.dev/core/server';
+import { EventEmitter } from 'node:events';
+import { createServer, type IncomingMessage, type ServerResponse } from 'node:http';
+import { connect } from 'node:net';
+import { performance } from 'node:perf_hooks';
+import { assert, describe, expect, test, vi } from 'vitest';
+import { fromNodeHttp, normalizeUrl } from './http';
 
 [
   {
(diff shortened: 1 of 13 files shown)


packages/qwik/src/core/shared/serdes/serialization-context.ts
```diff
@@ -251,7 +251,9 @@ export const createSerializationContext = (
   if (!writer) {
     const buffer: string[] = [];
     writer = {
-      write: (text: string) => buffer.push(text),
+      write: (text: string) => {
+        buffer.push(text);
```
wmertens, line 255: interesting, what does this do/prevent? Seems worthy of a comment.

Varixo: if I remember correctly its just type thing

packages/qwik-router/src/middleware/node/http.unit.ts
```diff
@@ -42,3 +53,156 @@ import { normalizeUrl } from './http';
...
+      new Headers([['Content-Type', 'text/html; charset=utf-8']]),
+      { headers: () => [] } as any,
+      () => {},
+      undefined as any
+    );
+    const writer = writableStream.getWriter();
+
+    const writePromise = writer.write(new Uint8Array([1, 2, 3]));
+    await Promise.resolve();
+    expect(res.listenerCount('drain')).toBe(0);
```
wmertens, line 91: could you add a comment explaining when the drain listeners are attached?

wmertens (approved):
LGTM! But I'd like to see the comment?

PR #8559, opened 2026-04-15
fix(qwik): guard undefined vNode in scheduleTask and scheduleEffects

juslintek:
# What is it?

- Bug fix

# Description

Fixes a runtime crash during hydration where `markVNodeDirty` receives an `undefined` vNode, causing:

```
TypeError: Cannot read properties of undefined (reading 'dirty')
  at markVNodeDirty (@qwik.dev/core/dist/core.mjs:6727:29)
  at String.scheduleTask (@qwik.dev/core/dist/core.mjs:8788:5)
```

This crash makes the entire application non-functional — no page renders, no login, all routes are dead.

## Root Cause

`scheduleTask` (in `use-task.ts`) and `scheduleEffects` (in `reactive-primitives/utils.ts`) pass `task.$el$` / `consumer.$el$` to `markVNodeDirty` without checking if the host element is defined.

`task.$el$` can be `undefined` when:
- The container is destroyed during async dispatch (the qwikloader's `dispatch` is now async, and if a navigation/destruction happens during an `await`, the container's `$stateData$` is truncated, causing subsequent deserialization to return `undefined` for VNode references)
- The serialized Task array has fewer elements than expected during inflation

Other callers of `markVNodeDirty` already have this guard — for example, `wrapped-signal-impl.ts` checks `if (this.$container$ && this.$hostElement$)` before calling `markVNodeDirty`. This PR adds the same defensive pattern to the two unguarded call sites.

### `packages/qwik/src/core/use/use-task.ts`
- Added null check for `task?.$el$` in `scheduleTask` before calling `markVNodeDirty`

### `packages/qwik/src/core/reactive-primitives/utils.ts`
- Added null check for `consumer.$el$` in the `isTask(consumer)` branch of `scheduleEffects` before calling `markVNodeDirty`

### `packages/qwik/src/core/use/use-task.unit.ts`
- Added three test cases: two verifying `scheduleTask` does not throw when `task.$el$` or `_captures[0]` is `undefined`, and one positive test verifying `markVNodeDirty` IS called when `task.$el$` is defined
- Tests use module-level `vi.mock` with per-test `Object.defineProperty` to avoid the Vitest hoisting trap

### `.changeset/gentle-tasks-guard.md`
- Patch changeset for `@qwik.dev/core`

# Docs

No user-facing documentation changes needed — `scheduleTask`, `scheduleEffects`, and `markVNodeDirty` are internal APIs not exposed in the public docs or GLOSSARY.

# Checklist

- [x] I made corresponding changes to the Qwik docs (N/A — internal APIs only)

### DIFF
--- a/.changeset/gentle-tasks-guard.md
+++ b/.changeset/gentle-tasks-guard.md
@@ -1,0 +1,7 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: guard undefined vNode in scheduleTask and scheduleEffects
+
+Prevent `TypeError: Cannot read properties of undefined (reading 'dirty')` crash in `markVNodeDirty` when `task.$el$` or `consumer.$el$` is undefined during async event dispatch.
--- a/packages/qwik/src/core/reactive-primitives/utils.ts
+++ b/packages/qwik/src/core/reactive-primitives/utils.ts
@@ -95,6 +95,10 @@ export const scheduleEffects = (
       const property = effectSubscription.property;
       isDev && assertDefined(container, 'Container must be defined.');
       if (isTask(consumer)) {
+        if (!consumer.$el$) {
+          // Host element not available (container may have been destroyed)
+          return;
+        }
         consumer.$flags$ |= TaskFlags.DIRTY;
         markVNodeDirty(container!, consumer.$el$, ChoreBits.TASKS);
       } else if (consumer instanceof SignalImpl) {
--- a/packages/qwik/src/core/use/use-task.ts
+++ b/packages/qwik/src/core/use/use-task.ts
@@ -270,6 +270,11 @@ export function scheduleTask(this: string, _event: Event, element: Element) {
     setCaptures(deserializeCaptures(container, this));
   }
   const task = _captures![0] as Task;
+  if (!task?.$el$) {
+    // Task or its host element was not properly deserialized
+    // (e.g., container destroyed during async dispatch)
+    return;
+  }
   task.$flags$ |= TaskFlags.DIRTY;
   markVNodeDirty(container, task.$el$, ChoreBits.TASKS);
 }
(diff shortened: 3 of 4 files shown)


Varixo (changes requested):
Hello! Thank you for the PR. 
Unfortunately for me it looks more like workaround instead of real solution. Basically task should have always host element. If there is a situation when it doesn't then its already a bug imo. Also the test case should show real example when useTask$ will have no host element.

Additionally the rolldown support branch is a bit behind the v2 branch, so first we need to check if this is fixed.

juslintek:
> Hello! Thank you for the PR. Unfortunately for me it looks more like workaround instead of real solution. Basically task should have always host element. If there is a situation when it doesn't then its already a bug imo. Also the test case should show real example when useTask$ will have no host element.
> 
> Additionally the rolldown support branch is a bit behind the v2 branch, so first we need to check if this is fixed.

Hi, thanks for review. Should I attempt a rebase of parent branch with this fix in a separate branch to validate whether it is working or not?

Regarding test I will create real example of my current project that I am working with this version of qwik.

Varixo:
> > Hello! Thank you for the PR. Unfortunately for me it looks more like workaround instead of real solution. Basically task should have always host element. If there is a situation when it doesn't then its already a bug imo. Also the test case should show real example when useTask$ will have no host element.
> > Additionally the rolldown support branch is a bit behind the v2 branch, so first we need to check if this is fixed.
> 
> Hi, thanks for review. Should I attempt a rebase of parent branch with this fix in a separate branch to validate whether it is working or not?
> 
> Regarding test I will create real example of my current project that I am working with this version of qwik.

The bug shouldn't be related to rolldown support, so you can just fix it on the build/v2 branch. Thank you for your work!

Varixo:
I think the error should be already fixed, closing it

PR #8560, opened 2026-04-15
Fixes

wmertens:
- AsyncSignal: don't trigger .loading for sync functions
- optimizer: don't change `let counter` to `const counter`
- cursors: actually do the 60fps yield instead of per cursor


Varixo (approved):
LGTM, but look at ai review for  the optimizer

wmertens:
@copilot review

PR #8562, opened 2026-04-15
feat: Suspense

wmertens:
A component that lets you show a loader while a subtree is waiting for promises.

```tsx
        <Suspense fallback={<span>Loading...</span>} timeout={10}>
          <SlowChild />
        </Suspense>
```


packages/qwik/src/core/shared/jsx/suspense.public.tsx
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Client-only: ms to wait while the children cursor is paused on a Promise before switching to
+   * `fallback`. If omitted, the fallback is never shown.
+   */
+  timeout?: number;
+  /**
+   * Rendered with `{ error }` when a descendant throws. Must be a `component$` so the prop is
+   * serializable. Uses the nearest `useErrorBoundary` boundary internally.
+   */
+  Error?: FunctionComponent<SuspenseErrorProps>;
```
maiieul: As copilot flagged this should be `error` not `Error` no?

packages/qwik/src/core/client/dom-container.ts
```diff
@@ -183,17 +183,20 @@ export class DomContainer extends _SharedContainer implements IClientContainer {
   }
 
   handleError(err: any, host: VNode | null): void {
+    if (host) {
+      if (err && err instanceof Error) {
+        if (!('hostElement' in err)) {
+          (err as any)['hostElement'] = String(host);
```
Varixo: String(host) doesnt make sense for prod

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,214 @@
...
+    vCurrent?: VNode | null,
+    vEnd?: VNode | null,
+    forceCreationMode?: boolean
+  ) => void;
+  drainAsyncQueue: (diffContext: DiffContext) => ValueOrPromise<void>;
+  cleanupDiffContext: (diffContext: DiffContext) => void;
+}
+
+export function isSuspenseBoundaryVNode(vNode: VNode | null): vNode is VirtualVNode {
+  return !!vNode && vnode_isVirtualVNode(vNode) && !!(vNode.flags & VNodeFlags.SuspenseBoundary);
```
Varixo: SuspenseBoundary is always virtual, no?

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,214 @@
...
+  if (timer) {
+    clearTimeout(timer);
+  }
+  vnode_setProp(host, QSuspenseTimer, null);
+  vnode_setProp(host, QSuspensePending, 0);
+  vnode_setProp(host, QSuspenseState, 'pending' as SuspenseState);
+}
+
+export function suspenseContentChanged(oldProps: PropsProxy | null, newProps: PropsProxy): boolean {
+  return !oldProps || oldProps.children !== newProps.children;
```
Varixo, line 95: I think this is always false, jsx is generated every time new.
This can be bad, because then we need to serialize children later (after ssr cursors)

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,214 @@
...
+  props: PropsProxy,
+  queueContentRender: boolean,
+  suspensePriority: number,
+  diffFns: SuspenseDiffFns
+) {
+  const contentRoot = ensureSuspenseContentRoot(diffContext, host);
+  const state = vnode_getProp<SuspenseState>(host, QSuspenseState, null) ?? 'pending';
+
+  updateSuspenseContentRootStyle(diffContext, contentRoot, state === 'fallback');
+
```
Varixo: maybe "pending", "fallback" should be enum

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1552,6 +1619,69 @@ function expectVirtual(diffContext: DiffContext, type: VirtualType, jsxKey: stri
...
+
+  vnode_setProp(host, ELEMENT_PROPS, jsxProps);
+  vnode_setProp(host, QSuspenseS, '');
+  vnode_setProp(host, QSuspenseTimeout, directGetPropsProxyProp(jsxNode, 'timeout') ?? 200);
+
+  if (!wasExistingSuspense) {
+    resetSuspenseState(host);
+  } else if (shouldRenderContent) {
+    resetSuspenseState(host);
+  }
```
Varixo: if (!wasExistingSuspense || shouldRenderContent) {
    resetSuspenseState(host);
  }

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1930,6 +2060,20 @@ export function cleanup(
...
+      if (type & VNodeFlags.SuspenseBoundary) {
+        const timer = vnode_getProp<ReturnType<typeof setTimeout>>(vCursor, QSuspenseTimer, null);
+        if (timer) {
+          clearTimeout(timer);
+          vnode_setProp(vCursor, QSuspenseTimer, null);
+        }
+        if (container.$suspenseCount$ > 0) {
+          container.$suspenseCount$--;
+        }
+        vCursor.flags &= ~VNodeFlags.SuspenseBoundary;
```
Varixo: maybe move it to function like "cleanupSuspense"

packages/qwik/src/core/shared/jsx/suspense-internal.ts
```diff
@@ -0,0 +1,109 @@
...
+const toNumber = (value: unknown): number | null => {
+  if (typeof value === 'number') {
+    return value;
+  }
+  if (typeof value === 'string' && value !== '') {
+    const parsed = parseInt(value, 10);
+    return Number.isNaN(parsed) ? null : parsed;
+  }
+  return null;
+};
```
Varixo: We have somewhere similar function I think

packages/qwik/src/core/shared/cursor/chore-execution.ts
```diff
@@ -171,6 +171,16 @@ export function executeNodeDiff(
...
+  if (vNode.flags & VNodeFlags.SuspenseBoundary) {
+    return vnode_diff_suspense_boundary(
+      container as ClientContainer,
+      journal,
+      vNode as VirtualVNode,
+      cursor,
+      null
+    );
+  }
+
```
Varixo, line 185: this feels weird. We have Suspense check in normal diffing, but then we need separate function to call. Seems a bit overcomplicated. Cant we just call vnode_diff?

packages/qwik/src/core/qwik.core.api.md
```diff
@@ -2061,9 +2074,11 @@ export const enum _VNodeFlags {
     // (undocumented)
     Resolved = 16,// http://www.w3.org/1999/xhtml
     // (undocumented)
-    Text = 4,// http://www.w3.org/2000/svg
+    SuspenseBoundary = 4096,// http://www.w3.org/2000/svg
+    // (undocumented)
+    Text = 4,// http://www.w3.org/1998/Math/MathML
```
Varixo, line 2080: the comments here a bit off, not sure why hmm

Varixo (commented):
Good job! Added some comments

packages/qwik/src/core/qwik.core.api.md
wmertens: right I don't understand why either, it doesn't seem to come from something done directly

packages/qwik/src/core/client/suspense-diff.ts
wmertens: I was thinking that it might be rerunning without any change for some reason and then the children would be ===

How will this cause serialization? It's just a shortcut to not do a cursor, and if children aren't serialized, it will just run the cursor and see nothing needs doing?

Varixo: propsproxy children needs to be serialized in order to compare with oldProps, no?

wmertens: yes, but we simply don't do that, it will be falsy, and then it will walk the deferred subtree, which will not be dirty, and there's no issue?

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,257 @@
...
+  return null;
+}
+
+function scheduleSuspenseContentRender(
+  diffContext: DiffContext,
+  host: VirtualVNode,
+  contentRoot: ElementVNode,
+  children: JSXChildren,
+  suspensePriority: number,
+  bootstrap: boolean
```
Varixo, line 139: bootstrap is always true. It is called only from one place. Or I am wrong?

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,257 @@
...
+    state === SuspenseState.Fallback && !showStale
+  );
+
+  diffSuspenseFallbackRange(
+    diffContext,
+    host,
+    contentRoot,
+    getSuspenseFallback(state, props),
+    diffFns
+  );
```
Varixo, line 216: not quite understand why we need this function that calls the syncSuspenseBoundary again

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,257 @@
...
+
+function scheduleSuspenseContentRender(
+  diffContext: DiffContext,
+  host: VirtualVNode,
+  contentRoot: ElementVNode,
+  children: JSXChildren,
+  suspensePriority: number,
+  bootstrap: boolean
+) {
+  vnode_setProp(contentRoot, NODE_DIFF_DATA_KEY, children);
```
Varixo, line 141: there is helper function for this

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,257 @@
...
+    onSuspensePause(host, diffContext.$container$);
+  }
+  if (!isCursor(contentRoot)) {
+    addCursor(diffContext.$container$, contentRoot, suspensePriority, host, bootstrap);
+  } else if (bootstrap) {
+    const cursorData = getCursorData(contentRoot);
+    if (cursorData) {
+      cursorData.$suspenseBootstrap$ = true;
+    }
+  }
```
Varixo, line 153: why we cant just call markVNodeDirty function with NODE_DIFF + additional code for it?

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,257 @@
...
+  if (isSuspenseContentRoot(currentFirstChild)) {
+    return currentFirstChild;
+  }
+  const doc = import.meta.env.TEST ? diffContext.$container$.document : document;
+  const contentRoot = vnode_newElement(doc.createElement(SUSPENSE_TAG), SUSPENSE_TAG);
+  vnode_insertElementBefore(diffContext.$journal$, host, contentRoot, currentFirstChild);
+  return contentRoot;
+}
+
+function updateSuspenseContentRootStyle(
```
Varixo, line 109: can you add comment why we want hide/show contentRoot?

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,257 @@
...
+  diff: (
+    diffContext: DiffContext,
+    jsxNode: JSXChildren,
+    vStartNode: VNode,
+    vCurrent?: VNode | null,
+    vEnd?: VNode | null,
+    forceCreationMode?: boolean
+  ) => void;
+  drainAsyncQueue: (diffContext: DiffContext) => ValueOrPromise<void>;
+  cleanupDiffContext: (diffContext: DiffContext) => void;
```
Varixo, line 60: why we cant import and use these functions directly?

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1552,6 +1589,71 @@ function expectVirtual(diffContext: DiffContext, type: VirtualType, jsxKey: stri
...
+      )
+    ) {
+      host = createSuspenseHost(diffContext, jsxKey);
+    } else {
+      host = (diffContext.$vNewNode$ || diffContext.$vCurrent$) as VirtualVNode | null;
+      if (!isSuspenseBoundaryVNode(host)) {
+        host = createSuspenseHost(diffContext, jsxKey);
+      }
+    }
+  }
```
Varixo, line 1624: isnt this just expect inline component or expect virtual?

packages/qwik/src/core/shared/cursor/cursor-props.ts
```diff
@@ -101,3 +114,19 @@ export function getCursorData(vNode: VNode): CursorData | null {
...
+export function findEnclosingSuspense(vnode: VNode | null): VNode | null {
+  let cur: VNode | null = vnode;
+  while (cur) {
+    if (cur.flags & VNodeFlags.SuspenseBoundary) {
+      return cur;
+    }
+    cur = cur.slotParent || cur.parent;
+  }
+  return null;
+}
```
Varixo, line 132: this function should goes to vnode-utils.ts file, its not related to cursors itself?

PS remove WeakMap reference from comment

packages/qwik/src/core/client/suspense-diff.ts
Varixo: you even added markSuspenseDirty

packages/qwik/src/core/ssr/ssr-render-jsx.ts
```diff
@@ -237,6 +263,48 @@ function processJSXNode(
...
+          enqueue(
+            setParentOptions(
+              options,
+              options.currentStyleScoped,
+              options.parentComponentFrame,
+              options.currentSuspensePriority
+            )
+          );
+          enqueue(
+            new JSXNodeImpl(
```
Varixo, line 291: _jsxSorted?

packages/qwik/src/core/client/suspense-diff.ts
wmertens: (AI): not as-is.

markVNodeDirty() in vnode-dirty.ts (line 108) is only a scheduler primitive: set dirty bits, propagate to an existing cursor, or create a normal cursor. It does not know about JSX payloads.
NODE_DIFF also needs a payload. executeNodeDiff() reads :nodeDiff and clears it in chore-execution.ts (line 185), so callers must still do setNodeDiffPayload(...) first. That is the normal pattern in reactive-primitives/utils.ts (line 106).
Suspense content is special: in suspense-diff.ts (line 174) the q-sus content root must become its own cursor via addCursor() (line 24). If you call markVNodeDirty(contentRoot, NODE_DIFF) while the boundary is already being processed, it will usually merge that work into the current blocking cursor instead of forcing a dedicated Suspense content cursor.
That dedicated cursor also carries Suspense-specific metadata: $suspense$ and $suspenseBootstrap$ in cursor.ts (line 31), and finishWalk() uses that in cursor-walker.ts (line 242) to resume the boundary correctly when the deferred content cursor completes.
So the boundary itself can use markVNodeDirty() for UI-state resync, and it already does in suspense-internal.ts (line 56). But the deferred content-root path needs more than “mark dirty”: payload + forced dedicated cursor + Suspense bootstrap metadata.

If this should be cleaner, I’d extract a higher-level helper like scheduleSuspenseContentRender(...), not teach markVNodeDirty() about NODE_DIFF and Suspense.

packages/qwik/src/core/client/suspense-diff.ts
wmertens: AI: It is useful. It is a thin adapter, but not semantic duplication.

syncSuspenseBoundary() in suspense-diff.ts (line 152) is the shared primitive. It assumes the caller already has:

a DiffContext
resolved props
the decision whether deferred content should be queued
diffSuspenseBoundaryNode() in suspense-diff.ts (line 194) is the scheduler entry point used from chore-execution.ts (line 175). When a Suspense boundary gets NODE_DIFF because its state changed, that does not mean “there is a JSX payload on this host”. It usually just means “re-evaluate fallback/visibility state” from code like suspense-internal.ts (line 56).

So this wrapper is where the runtime:

reads ELEMENT_PROPS off the host
creates and drains a temporary DiffContext
calls syncSuspenseBoundary(..., false, 0) so it does boundary-only resync and does not queue deferred content again
Without it, the generic NODE_DIFF executor would treat the boundary like a normal node diff, which is the wrong model for Suspense. So yes, it is small, but it is the right place for the Suspense-specific NODE_DIFF behavior.

If you want to simplify this area, I’d rename it rather than remove it. Something like executeSuspenseBoundarySync() would make its role clearer.

packages/qwik/src/core/client/vnode-diff.ts
wmertens: AI: expectSuspense() is not a semantic duplicate of either path.

What it shares with expectVirtual() (line 1537) is mostly the front half: “find/reuse/create the keyed virtual host.” In expectSuspense() (line 1584), that is basically lines 1589-1616.

After that, the behavior diverges completely:

expectVirtual() (line 1537) just ensures the virtual wrapper exists. It has no props/state/scheduling behavior.
The inline component branch in expectComponent() (line 1715) ensures a host, then immediately runs executeComponent() and queues its output.
expectSuspense() (line 1584) attaches Suspense boundary state, compares old/new children, sets timeout/props, may reset boundary state, and then calls syncSuspenseBoundary() (line 152), which can both diff fallback UI now and schedule deferred content on the q-sus content root cursor.
So: not duplicate overall. There is some duplicated “keyed virtual host resolution” scaffolding, but the actual work each branch performs after host selection is different enough that merging them directly would make the code worse.

packages/qwik/src/core/client/suspense-diff.ts
Varixo: still I think this is pointless, children are always different. Rerun = always new jsx = children array is different instance. Am I wrong?

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,215 @@
...
+  if (isPromise(result)) {
+    diffContext.$asyncAttributePromises$.push(
+      result.finally(() => {
+        cleanupDiffContext(fallbackDiffContext);
+      }) as Promise<void>
+    );
+    return;
+  }
+  cleanupDiffContext(fallbackDiffContext);
+}
```
Varixo, line 131: we should instead mark it as dirty and do diff via markers instead?

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1552,6 +1581,64 @@ function expectVirtual(diffContext: DiffContext, type: VirtualType, jsxKey: stri
...
+
+  const wasExistingSuspense = isSuspenseBoundaryVNode(host);
+  host = ensureSuspenseBoundaryAttached(diffContext.$container$, host!);
+
+  const oldProps = wasExistingSuspense
+    ? vnode_getProp<PropsProxy | null>(host, ELEMENT_PROPS, diffContext.$container$.$getObjectById$)
+    : null;
+  const suspensePriority = (getCursorData(diffContext.$cursor$)?.priority ?? 0) - 1;
+  const shouldRenderContent = !wasExistingSuspense || suspenseContentChanged(oldProps, jsxProps);
```
Varixo: instead maybe just do the same condition as component to rerender? not sure if it applies

packages/qwik/src/core/client/suspense-diff.ts
```diff
@@ -0,0 +1,215 @@
...
+    return;
+  }
+  syncSuspenseBoundary(diffContext, host, props, false, 0);
+  const result = drainAsyncQueue(diffContext);
+  if (isPromise(result)) {
+    return result.finally(() => {
+      cleanupDiffContext(diffContext);
+    });
+  }
+  cleanupDiffContext(diffContext);
```
Varixo: why we need to handle this here? syncSuspenseBoundary will mark it as dirty and run vnode_diff which is already running that

also dont now why new diffcontext is needed here. vnode_diff already creates new one

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1552,6 +1581,64 @@ function expectVirtual(diffContext: DiffContext, type: VirtualType, jsxKey: stri
...
+    host,
+    QSuspenseTimeout,
+    normalizeSuspenseTimeout(directGetPropsProxyProp(jsxNode, 'timeout')) ?? 200
+  );
+
+  if (shouldRenderContent) {
+    resetSuspenseState(host, !wasExistingSuspense);
+  }
+
+  syncSuspenseBoundary(diffContext, host, jsxProps, shouldRenderContent, suspensePriority);
```
Varixo: its weird that we dont call some diffing here like diffSuspenseBoundaryNode. For me it looks like diffSuspenseBoundaryNode is not needed and sync... can be renamed to diff...

wmertens:
closing in favor of #8588

PR #8569, opened 2026-04-22
fix(CI): benchmarks update still fail on mac M chips

maiieul:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

Problem: The old baseline was a tight integer loop with a serial dependency chain over a ~20KB array that fits entirely in L1. That kind of workload is limited by sequential scalar latency, which Intel handles almost as well as M4 — so M4 was only ~1.3× faster on the baseline while being ~2× faster on the actual Qwik benchmarks. The factors (benchmark/baseline) were therefore ~1.5–2× too small on M4.

Fix: Rewrote [baseline.ts](vscode-webview://1ae97gug94p9glogkd2nv2fe77v6r841lmp1evgquq8k0u1lmfv8/packages/qwik/src/core/bench/baseline.ts) to mirror the real scenarios' operation mix:

Object allocation with nested shape transitions (JSX/VNode-like)
String concatenation of an HTML-like table (SSR-like)
Map set+get round-trip (VNode/serdes bookkeeping)
Tree-walk serialization to a JSON-like string
These are the same kinds of operations the M4's wider core + better memory subsystem accelerates — so the baseline now scales proportionally.

Result: Factors went from 33–50% off build/v2 to within 5–17%, stable across runs (±2–3%). The validator's skew tolerance (~30% for stable benchmarks, ~200% for noisy ones) easily covers this, so CI should pass.

# Checklist


wmertens (approved):
LGTM

PR #8572, opened 2026-04-22
feat(worker): worker$ support for running QRL functions in Web Workers

Varixo:
This PR:
- adds worker$ for browser and server supporting also Node/Bun/Deno
- optimizer should be shipped together with core package for local packages testing
- adds Deno/Bun dev tests for worker$
- adds standard e2e tests for worker$


.github/workflows/ci.yml
```diff
@@ -40,6 +40,7 @@ on:
 env:
   # Disable incremental build, speeds up CI
   CARGO_INCREMENTAL: 0
+  CI_NODE_VERSION: 22.18.0
```
wmertens, line 43: not 24.x?

Maybe we should use latest node for building and 22 for testing?

packages/qwik-router/src/middleware/request-handler/async-request-store.ts
```diff
@@ -1,24 +1,22 @@
 import { isServer } from '@qwik.dev/core/build';
+import { _getAsyncLocalStorage } from '@qwik.dev/core/internal';
```
wmertens, line 2: hmm one of my prs also changes this

wmertens (commented):
not fully reviewed yet

.github/workflows/ci.yml
Varixo: we say qwik is compatible with node 22, so we should run CI on node 22, no?

PR #8573, opened 2026-04-22
fix(vite): resolve relative rollup input paths against vite root

maiieul:
# What is it?
- Bug

# Description

Fixes

> ✗ Build failed in 5ms
[vite-plugin-qwik] Qwik input "src/entry.preview.tsx" not found.

When `build.ssr` is a string (or `qwikVite({ ssr.input, client.input })` is relative), the raw path was forwarded to `rollupOptions.input` unchanged. Rollup resolves from CWD, so in monorepo/nx setups where CWD ≠ Vite root this fails with `Could not resolve entry module`. Input is now normalized against the Vite root in `normalizeOptions`, like `rootDir`/`srcDir`/`outDir`.

Also fixes issues with rolldown with better input paths handling.

### DIFF
--- a/packages/qwik-vite/src/plugins/plugin.unit.ts
+++ b/packages/qwik-vite/src/plugins/plugin.unit.ts
@@ -181,6 +181,16 @@ test('input array', async () => {
   ]);
 });
 
+test.runIf(process.platform === 'win32')('input array, win32', async () => {
+  const plugin = await mockPlugin();
+  const opts = await plugin.normalizeOptions({
+    rootDir: 'C:\\proj',
+    input: ['src\\cmps\\a.tsx', 'C:\\abs\\b.tsx'],
+  });
+  // relative paths are resolved against rootDir and normalized; absolute paths pass through
+  assert.deepEqual(opts.input, ['C:/proj/src/cmps/a.tsx', 'C:\\abs\\b.tsx']);
+});
+
 test('outDir', async () => {
   const plugin = await mockPlugin();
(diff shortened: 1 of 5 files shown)


packages/qwik-vite/src/plugins/vite.ts
```diff
@@ -227,7 +227,7 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
       };
 
       const opts = await qwikPlugin.normalizeOptions(pluginOpts);
-      input ||= opts.input;
+      input = opts.input;
```
maiieul, line 230: Only change that I didn't have in my rolldown PR. All the rest is also needed for rolldown.

packages/qwik-vite/src/plugins/plugin.unit.ts
```diff
@@ -165,17 +165,20 @@ test('tsconfigFileNames, empty array fallback to default', async () => {
...
     input: ['src/cmps/a.tsx', 'src/cmps/b.tsx'],
   });
-  // we don't provide input so that we don't override the vite input
-  assert.deepEqual(opts.input, undefined);
+  // relative inputs are resolved to absolute so rollup can find them
+  assert.deepEqual(opts.input, [
+    normalizePath(resolve(cwd, 'src/cmps/a.tsx')),
+    normalizePath(resolve(cwd, 'src/cmps/b.tsx')),
+  ]);
 });
```
Varixo, line 182: can we add a test for Windows-like paths?

Varixo (approved):
LGTM, added small comment

PR #8578, opened 2026-04-25
fix: allow Temporal in valid-lexical-scope-lint

FloezeTv:
# What is it?

- Bug

# Description

Serialization-support for the types in `Temporal` was added in #8495.
However, the `valid-lexical-scope`-lint would still warn that they would not be allowed to be used in a different scope because they would not be serializable.
This PR adds the serializable types from `Temporal` to the allowlist of the lint.

# Checklist

- [X] I added a changeset with `pnpm change`

### DIFF
--- a/.changeset/bright-showers-taste.md
+++ b/.changeset/bright-showers-taste.md
@@ -1,0 +1,5 @@
+---
+'eslint-plugin-qwik': patch
+---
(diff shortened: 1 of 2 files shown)


wmertens (approved):
LGTM

Issue #8293, opened 2026-02-03
[🐞] v2: qwikloader should wait until a container has state


wmertens:
### Which component is affected?

Qwik Runtime

### Describe the bug

when you run the e2e test "streaming" and click during the stream, the container is resumed before the vnodes/state has arrived.

Qwikloader should keep track of when a container is ready to run non-sync handlers, perhaps by expecting e.g. "ready" in the _qwikEv array for that container.

### Reproduction

see e2e tests

### Steps to reproduce

`pnpm serve`

http://localhost:3300/e2e/streaming

click while it loads

### System Info

```shell
any
```

### Additional Information

note that before, you could click but nothing happened (it also didn't remember you clicked), which is slightly better than now, where you click and the container never works any more.

PR #8579, opened 2026-04-25
perf(core): yield state processing during resume

Varixo:
Yield state processing during resume to increase performance


wmertens:
does this also fix https://github.com/QwikDev/qwik/issues/8293 ?

Varixo:
> does this also fix #8293 ?

unfortunatelly not, but I did fix it here #8581

PR #8581, opened 2026-04-27
fix(qwikloader): interactivity during streaming

### LINKED ISSUE #8293: [🐞] v2: qwikloader should wait until a container has state

Varixo:


.changeset/legal-trains-show.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: correctly handle interactivty during html streaming
```
maiieul, line 5: So before this change it only captured user events after the end of the stream?

packages/qwik/src/qwikloader.ts
```diff
@@ -108,6 +110,17 @@ const resolveContainer = (containerEl: QContainerElement) => {
   }
 };
 
+const waitForContainerReady = (container: QContainerElement) =>
+  container.getAttribute('q:container') === 'paused' && doc.readyState === 'loading'
```
wmertens, line 114: I think it would be better to add code that calls container.ready() after the state. That way, a container can resume while the HTML is still streaming, like in multi containers

.changeset/legal-trains-show.md
wmertens: no, but it could resume qwik before the state streamed

wmertens (approved):
LGTM 🚀

PR #8593, opened 2026-04-30
FEAT: move devtools into qwik #

JerryWu1234:
# What is it?

- Feature / enhancement
- Bug
- Infra

# Description

# Checklist


packages/browser-extension/public/qwikloader.js
wmertens, line 1: I'm confused as to what this is doing here. The browser extension doesn't look to be a Qwik application, and even if it was, the application would ship its own qwikloader.

wmertens (approved):
LGTM

packages/devtools/ui/src/utils/helperState.ts
wmertens, line 1: can you rename this to helperState?

wmertens (changes requested):

packages/devtools/ui/src/entry.dev.tsx
wmertens, line 1: you can delete this file

packages/browser-extension/public/qwikloader.js
Aejkatappaja: I added that one. Panel is a Qwik app, panel.html does import('@devtools/ui/entry.extension.qwik.mjs') and mounts it client-side. But @devtools/ui ships as a pre-built lib, so there's no qwikVite() SSR step injecting the qwikloader for us.

Vendored copy was the quick path to ship. Agreed it's fragile, drifts from @qwik.dev/core. Now that everything lives in the monorepo, easiest fix is probably copying @qwik.dev/core/dist/qwikloader.js at extension build time so it stays version-locked.

  Can do that in a follow-up?

packages/devtools/ui/src/entry.dev.tsx
Aejkatappaja: [0938441](https://github.com/QwikDev/qwik/pull/8593/commits/0938441f9ca7caba1127060f121d68242b195012)

packages/devtools/ui/src/utils/helperState.ts
Aejkatappaja: [0938441](https://github.com/QwikDev/qwik/pull/8593/commits/0938441f9ca7caba1127060f121d68242b195012)

packages/browser-extension/public/qwikloader.js
wmertens: yes ok

wmertens (approved):

PR #8605, opened 2026-05-02
feat: server-only folder and suffix

Varixo:
Implementation for https://github.com/QwikDev/qwik-evolution/discussions/202

### DIFF
--- a/.changeset/bright-servers-protect.md
+++ b/.changeset/bright-servers-protect.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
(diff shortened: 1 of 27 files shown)


.changeset/bright-servers-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/core': patch
```
maiieul: Should be major as it can break when people use .server in their project

maiieul (commented):
Double check that it doesn't throw when there is an `isServer`

wmertens (approved):
LGTM

PR #8608, opened 2026-05-02
fix(docs): enable iframe CORS headers

thejackshelton:
## Summary
- Ports only the iframe docs fix from #8596.
- Applies REPL CORS headers in the docs root layout so embedded docs iframes can load correctly.

## Testing
- pnpm exec prettier --check packages/docs/src/routes/layout.tsx


Varixo (approved):
LGTM

PR #8609, opened 2026-05-02
fix(core): allow duplicate qwik via shared singletons

wmertens:
Another step towards https://github.com/QwikDev/qwik-evolution/discussions/318


wmertens:
hmm _captures is also a singleton, we need to change it to a holder object

PR #8611, opened 2026-05-05
feat: out of order streaming for Suspense

Varixo:


packages/docs/src/routes/docs/labs/suspense/index.mdx
```diff
@@ -134,6 +141,245 @@ Suspense can show its fallback when content inside it is still loading, includin
 
 Suspense does not do the loading work. It only controls the fallback while that work finishes.
 
+## Out-of-order Streaming
+
+Out-of-order streaming is an optional server rendering mode for `<Suspense>`.
```
wmertens, line 146: I think it should be the default when the experimental is enabled.

packages/docs/src/routes/docs/labs/suspense/index.mdx
```diff
@@ -134,6 +141,245 @@ Suspense can show its fallback when content inside it is still loading, includin
...
+
+Server rendering means Qwik creates HTML on the server and sends it to the browser. Streaming means
+the browser can receive that HTML in chunks instead of waiting for one complete HTML string.
```
wmertens: ...HTML string (but still within the same HTTP connection).

packages/docs/src/routes/docs/labs/suspense/index.mdx
```diff
@@ -134,6 +141,245 @@ Suspense can show its fallback when content inside it is still loading, includin
...
+import Root from './root';
+
+export default createRenderer((opts) => {
+  return {
+    jsx: <Root />,
+    options: {
+      ...opts,
+      streaming: {
+        ...opts.streaming,
+        outOfOrder: { strategy: 'suspense' },
```
wmertens: do we plan other strategies? Why not a simple boolean and if we want to tweak things later we can optionally accept an object?

packages/docs/src/routes/docs/labs/suspense/index.mdx
```diff
@@ -134,6 +141,245 @@ Suspense can show its fallback when content inside it is still loading, includin
...
+
+Then, when the profile request resolves, Qwik sends the finished profile HTML and swaps it into the
+Suspense boundary.
+
+If the profile request finishes before the server needs to stream the fallback, Qwik can send the
+finished profile directly and the fallback may never appear.
+
+### Fallbacks
+
+The fallback is real UI. Choose markup that is useful while the user waits:
```
wmertens, line 318: Is the fallback always added to the stream? Worth explaining either way.

packages/docs/src/routes/docs/labs/suspense/index.mdx
```diff
@@ -134,6 +141,245 @@ Suspense can show its fallback when content inside it is still loading, includin
...
+
+### Interactivity
+
+Qwik keeps the streamed HTML resumable.
+
+That means the shell of the page can become interactive even while a Suspense section is still
+waiting. If your fallback has buttons or other event handlers, they can also resume while the
+fallback is visible, after the root page state has loaded in the browser.
+
+When the real content arrives, Qwik removes the fallback and inserts the resolved content. The
```
wmertens: Qwik hides the fallback, right? Not removes?

packages/docs/src/routes/docs/labs/suspense/index.mdx
```diff
@@ -134,6 +141,245 @@ Suspense can show its fallback when content inside it is still loading, includin
...
+It is less useful when the slow content is required before the rest of the page makes sense. In that
+case, showing a complete page a little later may be clearer than streaming a partial page early.
+
+### Static Builds
+
+Out-of-order streaming is an SSR feature. It helps when the browser is receiving an HTTP response
+while the server is still rendering.
+
+For static site generation, the HTML file is created ahead of time. The user receives the finished
+file later, so there is no live server stream to reveal a fallback first.
```
wmertens, line 375: Ah and during SSG, OoOS should be disabled of course.

wmertens (commented):
Initial comments

packages/qwik/src/core/shared/serdes/serialize.ts
```diff
@@ -698,24 +821,68 @@ export class Serializer {
...
+      }
+      this.$rootIdx$ = this.$serializationContext$.$roots$.length;
+    }
+  }
 
-    if (this.$forwardRefs$.length) {
-      let lastIdx = this.$forwardRefs$.length - 1;
-      while (lastIdx >= 0 && this.$forwardRefs$[lastIdx] === -1) {
-        lastIdx--;
+  private getForwardRefsPayload(): Array<number | string | number[]> | null {
```
wmertens, line 851: If I'm reading this right, if there is a slow promise first and a faster one later, ooos could send only the faster one and the slow will never be sent?

packages/qwik/src/core/shared/vnode-data-types.ts
```diff
@@ -80,3 +80,9 @@ export const VNodeDataChar = {
   SLOT: /* ************** */ 126, // `~` - `q:slot' - Slot name
   SLOT_CHAR: /* ******* */ '~',
 };
+
+export const getSegmentVNodeId = (segmentId: string, localIndex: number): number => {
+  const segmentIndex = parseInt(segmentId, 10) - 1;
+  const diagonal = segmentIndex + localIndex;
+  return -((diagonal * (diagonal + 1)) / 2 + localIndex + 1);
```
wmertens, line 101: I don't understand how this works, can you add a comment?

wmertens (changes requested):
I'm worried about forwardRefs, can you write a test with Suspense and a Promise that only resolves after the first chunk, followed by a Promise that resolves before the first chunk?

wmertens (commented):
Looks awesome so far, 114/128 files reviewed

I wish you didn't call them Segments, now we have SSR segments and JS segments and they're different things.

Issues for later PRs:
- REPL state display is broken
- would probably be better to push a special item in _qwikEv instead of using an event to know when a container has its state
- the OoOS patcher is quite big, perhaps put it in a module?

PR #8613, opened 2026-05-06
fix: node crash after not handled aborted connections

Varixo:
There were three problems:
- EPIPE / ECONNRESET were not treated as normal client disconnects.
The old code only handled messages like write after end and The stream has been destroyed.
- ServerResponse / request socket error events were not listened to.
In Node, an emitted error event without a listener can crash the process.
- The Node adapter dropped the getWritableStream(..., resolve) callback.
Other adapters resolve the response promise when the response stream is created. Node was not doing that, so handled.response could hang.

### DIFF
--- a/packages/qwik-router/src/middleware/node/http.ts
+++ b/packages/qwik-router/src/middleware/node/http.ts
@@ -31,10 +31,23 @@ export function getUrl(req: IncomingMessage | Http2ServerRequest, origin: string
   return normalizeUrl((req as any).originalUrl || req.url || '/', origin);
 }
 
-// when the user refreshes or cancels the stream there will be an error
-function isIgnoredError(message = '') {
-  const ignoredErrors = ['The stream has been destroyed', 'write after end'];
-  return ignoredErrors.some((ignored) => message.includes(ignored));
+// Client disconnects are expected during streaming SSR: the socket is gone,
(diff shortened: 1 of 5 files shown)


packages/qwik-router/src/middleware/node/http.ts
```diff
@@ -46,20 +59,39 @@ export function normalizeUrl(url: string, base: string) {
...
-  const closedPromise = closed
-    ? Promise.resolve()
-    : new Promise<void>((resolve) => {
-        res.once('close', () => {
-          closed = true;
-          resolve();
-        });
-      });
+  let responseError: unknown;
+  let resolveClosed = () => {};
```
wmertens: you can just do `let resolveClosed: () => void`, it will always be initialized

packages/qwik-router/src/middleware/node/http.ts
```diff
@@ -46,20 +59,39 @@ export function normalizeUrl(url: string, base: string) {
...
+  const closedPromise = new Promise<void>((resolve) => {
```
wmertens: you can make this `= closed ? Promise.resolve() : ...`

wmertens (approved):
LGTM

Some nitpicks, see if you want to change those or not, LGTM

PR #8620, opened 2026-05-10
fix(router): honor SSR backpressure via drain event for compression compatibility

maiieul:
# What is it?
- Bug

# Description
Since #8557 the Node SSR adapter resolves each chunk's write promise through `res.write`'s per-chunk callback. `compression` (and similar wrappers) replace `res.write` with a 2-arg signature `(chunk, encoding)` and drop that callback, hanging streaming SSR mid-response.

Switch to Node's standard backpressure protocol — return value + `'drain'` event — and resolve `close()` on `'finish'`, calling `res.end()` without a callback so `compression@1.x`'s 2-arg `res.end` doesn't blow up on `Buffer.from(<callback>)`.

|                                              | Pre-#8557                          | #8557                          | This PR                                                |
| -------------------------------------------- | ---------------------------------- | ------------------------------ | ------------------------------------------------------ |
| **Pause when buffer full?**                  | ❌ never                            | ✅ via per-write callback       | ✅ via `'drain'` event                                  |
| **Robust to wrappers like `compression`?**   | (didn't matter — never paused)     | ❌ hangs if callback dropped    | ✅ — `'drain'` / return value is the universal contract |

Reproduced and verified end-to-end against a real Express + `compression` app.

### DIFF
--- a/.changeset/compressed-streams-drain.md
+++ b/.changeset/compressed-streams-drain.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix(router): Node SSR no longer hangs when using `compression` (or other middleware that wraps `res.write` / `res.end`).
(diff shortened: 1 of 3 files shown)


wmertens (commented):
LGTM! 🧑‍🌾

PR #8635, opened 2026-05-15
fix(devtools/ui): Preloads overflow, CodeBreak button placement, inline Qwik logo

Aejkatappaja:
## Summary

Three small UI fixes in `packages/devtools/ui` spotted while testing the in-app overlay (via the docs site) and the browser extension panel.

### Preloads — outer container clipped content

The Preloads panel root used `overflow-hidden` on an `h-full` flex container. When the stat-cards grid and filter header took most of the available vertical space (common in narrower devtools panels), the table below was pushed out of view with no way to scroll. On narrower widths, the stat cards were also clipped horizontally for the same reason.

Switched the outer container to `overflow-auto` and replaced `h-full min-h-0` on the column stack with `min-h-full` so the content can grow past the viewport and scroll naturally. The inner table keeps its own horizontal scroll wrapper for the `min-w-[1080px]` table, so the sticky header behaviour is preserved.

**File:** `packages/devtools/ui/src/features/Preloads/Preloads.tsx`

### CodeBreak (StateParser) — Parse State button overlapped the next card

The content container inside the Input State card had the classes `min-h-0 flex-1 flex-col space-y-3 p-3` — note the missing `flex`. Without `flex`, `flex-col` is a no-op and the textarea's `h-full` expanded to the full card height, pushing the `Parse State` button outside the card. On the mobile-stacked layout (single column), that button then landed directly on top of the `Parsed State` card's header, making it visually overlap and hard to click.

Rather than reintroducing the missing `flex` and hoping the stack stays balanced at all widths, moved the `Parse State` button into the Input State card header alongside the title. More robust across panel widths, easier to reach, and consistent with the rest of the UI where per-card actions live in the header.

**File:** `packages/devtools/ui/src/features/CodeBreak/StateParser.tsx`

### Inline Qwik logo (remove remote `` dependency)

Three places referenced `https://qwik.dev/logos/qwik-logo.svg` via `` (DevtoolsButton, DevtoolsPanel title bar, DevtoolsContent overview header). In some contexts the remote fetch never resolves and the alt text is shown instead, leaving the panel/button looking broken.

Inlined the SVG as a new `QwikLogo` icon component and replaced all three usages. No more network dependency, renders identically on first paint.

**Files:**
- `packages/devtools/ui/src/components/Icons/Icons.tsx` (new `QwikLogo` export)
- `packages/devtools/ui/src/components/DevtoolsButton/DevtoolsButton.tsx`
- `packages/devtools/ui/src/components/DevtoolsPanel/DevtoolsPanel.tsx`
- `packages/devtools/ui/src/devtools/DevtoolsContent.tsx`

## Testing

- Built the devtools UI lib (`node --require ./scripts/runBefore.ts scripts/index.ts --devtools`) and ran the docs site with `qwikDevtools()` enabled.
- Visually verified the Preloads tab scrolls correctly at various devtools panel widths.
- Verified the Code Break Parse State button is rendered in the Input State card header, no overlap with the Parsed State card.
- Verified the Qwik logo renders inline in the button, title bar, and overview header without any network request.

## Notes

- Pure UI/CSS change in `@qwik.dev/devtools`, no public API surface affected.
- No changeset added — happy to add one if the maintainers prefer this trigger a patch bump.

### DIFF
--- a/.changeset/fix-devtools-ui-layout-polish.md
+++ b/.changeset/fix-devtools-ui-layout-polish.md
@@ -1,0 +1,9 @@
+---
+'@qwik.dev/devtools': patch
+---
+
+fix(devtools/ui): Preloads overflow, CodeBreak button placement, inline Qwik logo
+
+- Preloads panel: outer `overflow-hidden` clipped stat cards and prevented scroll when content exceeded the panel height. Switched to `overflow-auto` with `min-h-full` on the column stack so content scrolls on narrow or short panels.
+- CodeBreak StateParser: content container was missing the `flex` class, so the textarea pushed the Parse State button outside the card into the next grid row, overlapping the Parsed State header. Moved the button into the card header next to the title.
+- Inlined the Qwik logo as a local SVG component (`QwikLogo`) instead of fetching `https://qwik.dev/logos/qwik-logo.svg` via `<img src>`. Fixes the broken-image fallback in contexts where the remote fetch is blocked or unreachable.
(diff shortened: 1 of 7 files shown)


gioboa (approved):
LGTM 👏

PR #8636, opened 2026-05-15
perf: reduce state size

Varixo:
- global visible observer for links data & bundle prefetching
- write rootrefs deltas instead of raw rootref number
- cleanup unused task field
- common constants as serialization TypeIds
- omit cold static component props


wmertens (approved):
LGTM

PR #8651, opened 2026-05-20
chore(scripts): support branch refs in update-action-pins

TupiC:
# What is it?
- Infra

# Description
extends scripts/update-action-pins.ts so it can pin and refresh GitHub Actions that use branch refs (e.g. @main), not only version tags (e.g. @v4)

# Checklist


wmertens (approved):
LGTM

PR #8660, opened 2026-05-22
[codex] Improve missing Suspense flag diagnostic

PatrickJS:
## Summary

Adds an early Qwik Vite transform diagnostic when a module imports `Suspense` from `@qwik.dev/core` but the app has not enabled `qwikVite({ experimental: ['suspense'] })`.

The new error tells the developer exactly which Vite config flag is missing and includes the file path that triggered the check.

## Problem

Without the `suspense` experimental flag, async UI that depends on Suspense coordination can fail in a confusing way during SSR. In practice this can look like a stalled SSR request or an unrelated optimizer/runtime symptom, which sends developers looking in the wrong place.

The existing runtime error is useful when the `Suspense` component function actually runs, but it is too late for some dev-server paths. Catching the import during transform gives a direct configuration error before the app gets into an opaque async render state.

## Changes

- Adds a focused missing-Suspense-flag diagnostic in the Qwik Vite plugin transform.
- Adds a unit test covering `Suspense` import without the experimental flag.

## Validation

- `pnpm build.core`
- `pnpm vitest --run packages/qwik-vite/src/plugins/plugin.unit.ts`

### DIFF
--- a/packages/qwik-vite/src/plugins/plugin.ts
+++ b/packages/qwik-vite/src/plugins/plugin.ts
@@ -80,6 +80,14 @@ export enum ExperimentalFeatures {
   insights = 'insights',
 }
 
+const MISSING_SUSPENSE_EXPERIMENTAL_FLAG =
+  `Suspense requires qwikVite({ experimental: ['suspense'] }). ` +
+  `Add the 'suspense' experimental flag in your Vite config before using ` +
+  `Suspense, Reveal, or async UI that depends on Suspense coordination. ` +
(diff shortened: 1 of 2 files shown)


wmertens (commented):
This adds another full code scan on every transform, while I think it could instead be done in the Suspense component when trying to use it?

PatrickJS:
I think @Varixo said there was a better error in his branch

maiieul:
We have a better diagnostic now. Thanks.

PR #8666, opened 2026-05-25
fix(router): recover SPA navigation after animation timeout

Varixo:
Fixes SPA navigation recovery when browser view transitions fail or time out. The router now keeps fallback navigation behavior without running the same navigation twice, preventing inconsistent page updates.

### DIFF
--- a/.changeset/curly-experts-dig.md
+++ b/.changeset/curly-experts-dig.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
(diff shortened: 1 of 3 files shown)


wmertens (commented):
I'd like to hold off on merging this, I believe the router split branch also handles this so I'd like to merge that first and then check if it's still an issue.

Varixo:
Closing it is already fixed by the route loaders branch

PR #8667, opened 2026-05-27
fix(core): don't HTML-escape SSR tag nesting errors

maiieul:
# What is it?
- Bug

# Description
SSR tag-nesting validation errors were run through `escapeHTML` before being thrown, so `<div>` showed up as `&lt;div&gt;` and `'` as `&#39;` in the logs. The error is logged as a plain JS `Error` and never written into HTML, so the escaping only made it unreadable. Dropped it and updated the two specs.


Varixo (approved):
LGTM

PR #8670, opened 2026-05-27
fix: release paused cursors that are abandoned before they resume

Varixo:
Fixes a SPA navigation timeout caused by abandoned paused cursors keeping the container pending count alive


wmertens (approved):
LGTM

PR #8673, opened 2026-05-28
fix(router): avoid top-level await in generated SSG entry

maiieul:
# What is it?
- Bug

# Description
SSG builds with the edge/worker adapters (vercel-edge, netlify-edge, cloudflare-pages) crash on Rollup-vite. The generated SSG entry used a top-level `await`; on the worker/edge SSR target Vite inlines dynamic imports, and Rollup then reads the inlined `./system` namespace before it's initialized across that top-level await ([rollup/rollup#4166](https://github.com/rollup/rollup/issues/4166)). The entry now starts the SSG/worker promise without awaiting at the top level (it still keeps the process alive via worker handles and calls `process.exit`). rolldown-vite and node-target adapters were unaffected; this also keeps them correct.

### DIFF
--- a/.changeset/edge-ssg-top-level-await.md
+++ b/.changeset/edge-ssg-top-level-await.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix(router): SSG builds with edge/worker adapters no longer crash on Rollup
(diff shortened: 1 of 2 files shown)


wmertens (approved):
LGTM

PR #8684, opened 2026-05-29
fix(core): inline components context resolving

maiieul:
### Root cause
Commit 37cb49e2 ("fix: finding inline component's host subscription") changed inline components in a projection to use their parent/lexical-owner component frame instead of the nearest enclosing one. That correctly fixed signal subscription (re-render targeting), but executeComponent was using that single host for two different things:

`$effectSubscriber$` — which component re-executes when a signal changes (wants the parent)
`$hostElement$` — where useContext starts its tree walk (wants the nearest enclosing, i.e. the projection target)
The compiled MDX content (WrappedMdxContent/_createMdxContent) is an inline component that calls useMDXComponents() = useContext(MDXContext). After the commit, its context resolved from the lexical owner instead of the MDXProvider it's projected into, so it got the default empty map → no styled components.

### The fix

packages/qwik/src/core/shared/component-execution.ts — decouple the two:

$hostElement$ now uses renderHost (the inline node's actual position, whose parent is the enclosing provider) → context resolves correctly.
$effectSubscriber$ still uses subscriptionHost (the parent component) → the commit's subscription fix is preserved.
For component$ (where renderHost === subscriptionHost) nothing changes. The commit's own changes to vnode-diff.ts and ssr-render-jsx.ts are kept untouched.

### DIFF
--- a/packages/qwik/src/core/tests/use-context.spec.tsx
+++ b/packages/qwik/src/core/tests/use-context.spec.tsx
@@ -51,14 +51,6 @@ const useFooFn = () => {
   });
 };
 
-// Module-scoped helpers for the MDX-provider regression test. They must live at module scope
-// (like compiled MDX content) so the inline component is serializable when projected.
-const inlineProjectionContext = createContextId<{ value: string }>('inline-projection');
-const InlineProjectionContent = () => {
-  const ctx = useContext(inlineProjectionContext, { value: 'default' });
-  return <div>{ctx.value}</div>;
-};
-
 describe.each([
   { render: ssrRenderToDom }, //
   { render: domRender }, //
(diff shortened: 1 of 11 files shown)


packages/qwik/src/core/tests/use-context.spec.tsx
```diff
@@ -404,6 +412,36 @@ describe.each([
         </Component>
       );
     });
+    it('inline component projected into a context provider resolves the context', async () => {
```
Varixo: I would put it in inline-component.spec.ts, because its for inline component

packages/qwik/src/core/tests/use-context.spec.tsx
```diff
@@ -404,6 +412,36 @@ describe.each([
...
+      const App = component$(() => {
+        return (
+          <Layout>
+            <InlineProjectionContent />
+          </Layout>
+        );
+      });
+
+      const { document } = await render(<App />, { debug });
+      if (render === ssrRenderToDom) {
```
Varixo: we shouldnt add condition like this, remove it. Just check for both renders

packages/qwik/src/core/tests/use-context.spec.tsx
```diff
@@ -404,6 +412,36 @@ describe.each([
...
+          <Provider>
+            <Slot />
+          </Provider>
+        );
+      });
+      // Mirrors the MDX provider pattern: the compiled MDX content is an inline component
+      // that reads context (`useMDXComponents`) and is projected, through a layout, into a
+      // `component$` that provides it. Context must resolve from the projection target
+      // (the provider), not the inline component's lexical owner.
+      const App = component$(() => {
```
Varixo: the comment is not useful for me, you can see what is going on and comment make this file just larger + doesnt matter that its for mdx components

packages/qwik/src/core/tests/use-context.spec.tsx
```diff
@@ -51,6 +51,14 @@ const useFooFn = () => {
   });
 };
 
+// Module-scoped helpers for the MDX-provider regression test. They must live at module scope
+// (like compiled MDX content) so the inline component is serializable when projected.
+const inlineProjectionContext = createContextId<{ value: string }>('inline-projection');
```
Varixo: I would remove it, its standard thing for inline components in inline-component.spec.ts

Varixo (changes requested):

wmertens (approved):
LGTM 🚞

PR #8685, opened 2026-05-29
fix(core): repair OOOS Suspense state patch root serialization

Varixo:
Fixes out-of-order Suspense state patch serialization so segment-local roots no longer get treated as already-streamed root-container roots. This keeps resumed event handlers working correctly for streamed Suspense content, including hoisted QRL params inside loops.


wmertens (approved):
LGTM 🐁

PR #8691, opened 2026-06-01
fix(core): remove rollup workaround for stable hashes

wmertens:
We had a workaround in place that configured Rollup to only do 1 parallel operation at a time, that is no longer needed thanks to https://github.com/rollup/rollup/pull/6362.


maiieul (dismissed):
LGTM 👍

PR #8695, opened 2026-06-04
fix(core): stabilize repeated HMR updates

thejackshelton-kunaico:
## Summary
- match Qwik HMR updates using captured optimizer dev paths instead of relying on current data-qwik-inspector attributes
- skip acknowledging deleted/stale captured hosts
- invalidate lazy QRL refs before component HMR handlers run, without treating lazy invalidation as a handled update
- add regression coverage for repeated remove/restore of a stateful component

## Verification
- pnpm build.core.dev
- pnpm vitest run packages/qwik/src/core/tests/hmr.spec.tsx
- pnpm tsc.check
- Kunai browser QA: remove/add/remove/add layout shell repro passes with local Qwik + qwik-bundler

### DIFF
--- a/.changeset/stale-hmr-hosts.md
+++ b/.changeset/stale-hmr-hosts.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: make repeated HMR JSX remove and restore updates reload QRLs before rerendering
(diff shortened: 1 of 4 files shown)


wmertens (approved):
LGTM 🍭

PR #8698, opened 2026-06-05
feat(router): ev.internalRequest information

wmertens:


Varixo (approved):
LGTM

packages/docs/src/routes/docs/(qwikrouter)/guides/rewrites/index.mdx
```diff
@@ -24,9 +24,13 @@ src/routes/articles/
 
 
 ```tsx title="src/routes/plugin@article-rewrite.tsx"
-import type { RequestHandler } from "@qwik.dev/router";
+import type { RequestHandler } from '@qwik.dev/router';
+
+export const onRequest: RequestHandler = async ({ url, rewrite, internalRequest }) => {
+  if (internalRequest) {
```
maiieul, line 30: Are there really no better alternatives? This is not so great DX wise since devs must know about it... Ideally redirects should just work without having to remember to check for internal requests.

wmertens: that's impossible:

Either middleware doesn't run for loader requests, and it will miss logged-in state etc,
or the dev always redirects even for loaders, which breaks things.

This is because now each loader gets its own request, which allows sane loader caching.

In any case, no matter which API we would use, we still need to educate the dev. Having a prop directly visible in the type on requestEv is the most direct thing we can do.

PR #8699, opened 2026-06-05
feat(router): SSR and routeLoader eTag and cacheKey handling

wmertens:
Adds caching and eTag support to routeloaders, and adds auto eTag generation for SSR.

Breaking change if using previous beta: the `cacheKey` callback arguments changed, now you get the requestEv.

Changes:

- Split caching into separate SSR and loader caches (LRU) and add clearLoaderCache() alongside clearSsrCache().
- Introduce shared ETag utilities (normalizeETag, hash, performETagMatch) and apply normalized ETags to cacheKey callbacks while quoting only for HTTP headers.


packages/qwik-router/src/middleware/request-handler/handlers/loader-request-event.ts
```diff
@@ -0,0 +1,108 @@
...
+            return loaderValues[loaderOrAction.__id];
+          }
+          return loadRouteLoader(loaderOrAction, getLoaderRequestEvent(loaderOrAction));
+        }
+        return undefined;
+      }) as RequestEventInternal['resolveValue'],
+      enumerable: true,
+    },
+  }) as RequestEventInternal;
+}
```
Varixo: we had something like this somewhere, no?

Varixo (commented):
LGTM

PR #8705, opened 2026-06-08
refactor(devtools): single canonical devtools hook runtime

Aejkatappaja:
## Problem

The browser extension shipped its own copy of the devtools hook runtime in `packages/browser-extension/public/devtools-hook.js`, duplicating the logic the Vite plugin already keeps in the plugin runtime. Both had to be kept in sync by hand.

## Solution

Make the plugin installer the single source of truth and generate the extension script from it.

- `__qwik_install_hook_runtime__` (`packages/devtools/plugin/src/runtime/installers.ts`) is the one canonical implementation. It covers signal serialization, component snapshots, component detail lookup, render listeners, and signal editing.
- Added `createExtensionHookRuntime()` which wraps the same installer in a plain-script IIFE for content-script injection.
- `public/devtools-hook.js` is now generated by `scripts/gen-devtools-hook.mjs`, wired into the extension `prebuild` and `predev` scripts. It carries a generated-file banner and is prettier-ignored.
- Removed the component-key matching that was copy-pasted across `getComponentDetail` and `setSignalValue`; both now use a shared `findComponentKey` helper.
- The string builders live behind an internal `codegen` build entry that is deliberately not part of the published package exports, so it stays out of the public API and is consumed only by a relative path inside the monorepo.

## Tests

- Behavioral unit tests for the canonical installer: install guard, signal snapshot filtering, component tree name extraction, detail lookup by chunk and by name, signal editing, render listeners (including throwing-listener isolation).
- A drift guard test that fails if the committed `devtools-hook.js` stops matching the generated output.

## Follow-ups

The other injected hooks (vnode bridge, perf, inspect, nav) still duplicate their respective runtimes and can be migrated the same way in separate PRs.

## References

QwikDev project board, card "p1 Runtime Hook Duplication": https://github.com/orgs/QwikDev/projects/6/views/1#


packages/browser-extension/public/devtools-hook.js
JerryWu1234, line 1: Can't this file be submitted ?
we can run [gen-devtools-hook.mjs‎](https://github.com/QwikDev/qwik/pull/8705/changes#diff-128f90efc6ff41e874ca7be6f0dceac182fd9e7e1ae0c2789daea206b95a6e30) this script before dev and build

Aejkatappaja: Good call, done ✅. 

Dropped the committed devtools-hook.js, it's now gitignored and generated by `gen-devtools-hook.mjs` on `predev/prebuild`. 

The parity unit test generates the script in-memory from the canonical installer, so the two injection paths can't drift.

packages/browser-extension/package.json
```diff
@@ -14,6 +14,9 @@
   },
   "private": true,
   "scripts": {
+    "generate": "node scripts/gen-devtools-hook.mjs",
+    "prebuild": "node scripts/gen-devtools-hook.mjs",
```
JerryWu1234: we don't need a individual commands. 
we can merge these commands with build and dev so on

Aejkatappaja: [c27a0a2](https://github.com/QwikDev/qwik/pull/8705/commits/c27a0a247a6e2b1c6a0cad2f4900c2747993d589)

JerryWu1234:
@wmertens done
please take a look to check if ther  are any other suggestions

wmertens (approved):
LGTM

packages/devtools/plugin/src/runtime/installers.ts
```diff
@@ -132,6 +132,30 @@ export function __qwik_install_hook_runtime__(options: HookRuntimeOptions) {
     return getOrCreateRoot()?.[options.componentStateKey];
   };
 
+  const findComponentKey = (componentName: string, qrlChunk: string | null): string | null => {
+    const state = getState();
```
JerryWu1234, line 136: I found this function is very common, so we can extract this function as a utility function late

PR #8706, opened 2026-06-08
fix(router): fetch non-serialized route loader data after resume

Varixo:
Fixes serializationStrategy: "never" route loaders so resumed clients fetch missing SSR loader data instead of reading undefined. It preserves loader identity with an _UNINITIALIZED marker


wmertens (approved):
LGTM 💈

PR #7518, opened 2025-04-15
feat: Add Standard Schema validator to qwik-city

adrian-budau:
# What is it?

- Feature / enhancement

# Description

Based on #7281 

Adds a `schema$` validator accepting any [Standard Schema](https://github.com/standard-schema/standard-schema) confirming validation library

# Checklist

- [x] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [x] I performed a self-review of my own code
- [ ] I added a changeset with `pnpm change`
- [ ] I made corresponding changes to the Qwik docs
- [x] I added new tests to cover the fix / functionality

PR #7281, opened 2025-01-25
feat: Add Standard Schema validator to Qwik Router

fabian-hiller:
# What is it?

- Feature / enhancement

# Description

I implemented a `schema$` validator function based on [Standard Schema](https://github.com/standard-schema/standard-schema) that should replace `zod$` and `valibot$` in the long run.

> I added [`type-fest`](https://github.com/sindresorhus/type-fest) to simplify the type implementation, but I could write the types mysefle instead. Please give me feedback.

New features:
- Add new `schema$` function that supports any Standard Schema copatiple schema library including Zod, ArkType and Valibot

Breaking changes:
- Remove export of Zod's `z`
  - Users should import that from Zod directly
- Change return type of `.fieldErrors`
  - Arrays paths are changed from `array[].key` to `array.${number}.key`
  - This allows to directly access specific array index error messages
  - Furthermore, it simplifies the implementation

Other changes:
- Deprecate `zod$` and `valibot$` as well as any directly related functions and types
  - We could also remove `valibot$` as it was in `@beta` stage and marked as experimental

# Checklist

- [ ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [x] I performed a self-review of my own code
- [ ] I added a changeset with `pnpm change`
- [ ] I made corresponding changes to the Qwik docs
- [ ] I added new tests to cover the fix / functionality

PR #8707, opened 2026-06-08
Standard schema validator

gabrielgrant:
# What is it?

Feature / enhancement

# Description

Moves forward the work started previously in https://github.com/QwikDev/qwik/pull/7518 and https://github.com/QwikDev/qwik/pull/7281 to add Standard Schema support

# Checklist

### DIFF
--- a/packages/docs/src/routes/docs/(qwikrouter)/validator/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikrouter)/validator/index.mdx
@@ -166,11 +166,11 @@ export const useAction = routeAction$(
 
 ## Standard Schema validators
 
-Use `schema$()` to validate actions with any schema library that implements [Standard Schema](https://standardschema.dev/), such as Zod, Valibot, or ArkType. The validator infers the submitted input type, passes the validated output type to the action callback, and returns failed field errors as arrays of messages.
+Use `schema$()` to validate actions with any schema library that implements [Standard Schema](https://standardschema.dev/), such as Zod, Valibot, or ArkType. The validator infers the submitted input type, passes the validated output type to the action callback, and returns failed field errors in the same shape as other Qwik validators.
 
 ```tsx
-import { component$ } from '@builder.io/qwik';
-import { Form, routeAction$, schema$ } from '@builder.io/qwik-city';
+import { component$ } from '@qwik.dev/core';
+import { Form, routeAction$, schema$ } from '@qwik.dev/router';
 import { z } from 'zod';
 
 const userSchema = z.object({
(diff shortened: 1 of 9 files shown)


gabrielgrant:
AFAICT these CI failures seem to be timeouts setting up playwright, not actual failures, but I don't believe I'm able to re-run the jobs without pushing dummy commits to the branch

gabrielgrant:
Would be great if @wmertens and @fabian-hiller could take a look at this when you have a moment. Thank you!

A few specific design questions:

1. This adds types etc into the repo to avoid an additional dependency as suggested in https://github.com/QwikDev/qwik/pull/7518#issuecomment-2816904285 -- this is a question of Qwik policy/preferences (personally I'd generally go the other way and rely on upstream deps whenever possible, but I'm generally building projects, not frameworks)
2. You'll probably notice `StandardSchemaValidatorErrorKeyDotNotation` is pretty verbose/complicated. Seems to work for nicer typing feedback than just `Record<string, string[]>`, but would certainly understand if the implementation complexity is deemed too high
3. this implementation only gives access to the simplified dot notation -- it does not preserve the original Standard Schema issues array. May be worth changing that to also expose the original array too

gabrielgrant:
This is now rebased onto/PRed against build/v2

Since this was originally against v1, I left the existing explicit integrations with Valibot & Zod intact. If this is going to be added as part of the v2 transition, though, it's possible we may want to remove/deprecate those lib-specific integrations, and promote `schema$` as the primary API. Thoughts?

packages/docs/src/routes/docs/(qwikrouter)/validator/index.mdx
```diff
@@ -164,6 +164,76 @@ export const useAction = routeAction$(
...
 ```
 
+## Standard Schema validators
+
+Use `schema$()` to validate actions with any schema library that implements [Standard Schema](https://standardschema.dev/), such as Zod, Valibot, or ArkType. The validator infers the submitted input type, passes the validated output type to the action callback, and returns failed field errors in the same shape as other Qwik validators.
+
+```tsx
+import { component$ } from '@qwik.dev/core';
+import { Form, routeAction$, schema$ } from '@qwik.dev/router';
+import { z } from 'zod';
```
wmertens, line 174: you have to `import * as z from 'zod'` so that tree shaking works

packages/qwik-router/src/runtime/src/server-functions.unit.ts
```diff
@@ -188,4 +206,157 @@ describe('types', () => {
...
+    const action = routeAction$(
+      (data) => {
+        expectTypeOf(data).toEqualTypeOf<{ id: number }>();
+        return { ok: true };
+      },
+      schema$(() =>
+        createStandardSchema<{ id: string }, { id: number }>((value) => ({
+          value: { id: Number((value as { id: string }).id) },
+        }))
+      )
```
wmertens, line 258: QRLs can hold objects with functions in them, so maybe that would be a more ergonomic definition? 
```tsx
schema$({ validate: v => ({ value: { id:  Number((value as { id: string }).id) } })})
```
(no idea what I'm doing)

PR #8709, opened 2026-06-09
fix(router): const assignment in route loaders

Varixo:
Fixes routeLoader$ filtered search caching by storing the last filtered search and route path in a captured object, avoiding "assignment to constant variable" errors after QRL transform


wmertens (approved):
LGTM

PR #8712, opened 2026-06-09
refactor(devtools): single source of truth for shared protocol types

Aejkatappaja:
## Problem

The VNode tree node, component detail entry, and render event shapes were declared three times with identical fields:

- `packages/browser-extension/src/shared/types.ts` (`VNodeTreeNode`, `ComponentDetailEntry`, `RenderEvent`)
- `packages/devtools/ui/src/devtools/page-data-source.ts` (same names)
- `packages/devtools/kit/src/client-bridge.ts` (`DevtoolsVNodeTreeNode`, `DevtoolsComponentDetailEntry`, `DevtoolsRenderEvent`)

Three copies of the same data contract had to be kept in sync by hand.

## Solution

Make `@qwik.dev/devtools/kit` the single source of truth.

- The three interfaces now live once in the kit protocol module: `DevtoolsVNodeTreeNode` in `protocol/vnode.ts`, `DevtoolsRenderEvent` in `protocol/perf.ts`, `DevtoolsComponentDetailEntry` in `protocol/hooks.ts`.
- `kit/client-bridge.ts` imports them from the protocol module instead of declaring its own copies, so it becomes a consumer of the shared types too.
- The browser extension and the devtools UI drop their local declarations and re-export the kit types under their existing local names, so every consumer keeps a single import source and cannot drift.
- The `Devtools` prefix is kept on purpose: a bare `RenderEvent` would collide with the DOM lib global.

No public API change: the same type names are still exported from `@qwik.dev/devtools/kit`, just from the protocol module rather than the client bridge.

## Verification

- `@qwik.dev/devtools/kit` builds and its generated `.d.mts` exports all three types.
- `tsc` on the devtools UI passes with zero errors.
- `tsc` on the browser extension reports only two preexisting wxt `TS2742` errors, unrelated to this change.
- eslint passes on every touched file.

## References

QwikDev project board, card "p1 Shared Type Duplication": https://github.com/orgs/QwikDev/projects/6/views/1

### DIFF
--- a/.changeset/shared-protocol-types.md
+++ b/.changeset/shared-protocol-types.md
@@ -1,0 +1,11 @@
+---
+'@qwik.dev/devtools': patch
+---
+
+refactor(devtools): single source of truth for shared protocol types
+
+The VNode tree node, component detail entry, and render event shapes were declared
+three times: in the browser extension, in the devtools UI, and in the kit client
+bridge. They now live once in @qwik.dev/devtools/kit (protocol module) as
+DevtoolsVNodeTreeNode, DevtoolsComponentDetailEntry, and DevtoolsRenderEvent, and
+every consumer imports them from there.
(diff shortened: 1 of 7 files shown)


JerryWu1234:
@wmertens i done please take a look

wmertens (approved):

PR #8713, opened 2026-06-09
refactor(devtools): generate the extension VNode bridge from one shared source

Aejkatappaja:
## Problem

`packages/browser-extension/public/vnode-bridge.js` duplicated the VNode bridge logic the Vite plugin already owns in `packages/devtools/plugin/src/virtualmodules/vnodeBridge.ts` (via `createVNodeRuntime()` / `__qwik_install_vnode_runtime__`): tree building, prop serialization, component name normalization, DOM element resolution, highlighting, and component tree update posting. Both had to be kept in sync by hand.

## Solution

Generate the extension bridge from the single canonical source, the same way `devtools-hook.js` is handled.

- Expose `createVNodeRuntime` through the internal `codegen` entry of `@qwik.dev/devtools` (build-time only, not a published subpath).
- Generalize the extension generate step into `scripts/gen-injected-scripts.mjs`, which now emits both `devtools-hook.js` and `vnode-bridge.js` from their canonical builders. It replaces `gen-devtools-hook.mjs` and stays wired as the first step of `build` and `dev` (and the firefox variants).
- Stop tracking `public/vnode-bridge.js`: it is gitignored and generated on build/dev, matching the `devtools-hook.js` approach.

The extension bridge is now byte-for-byte the same module the plugin injects (same `createVNodeRuntime()` output), so the two cannot drift.

## Verification

- `@qwik.dev/devtools` builds and `codegen.mjs` exports `createVNodeRuntime`.
- The generate step writes both injected scripts; `vnode-bridge.js` is valid ESM and exposes the full bridge surface (`getVNodeTree`, `resolveElementToComponent`, `getElementRect`, `highlightNode`, `unhighlightNode`, `getNodeProps`, component tree update posting), installed with the bound `@qwik.dev/core/internal` helpers.
- eslint passes on the touched source.

## References

QwikDev project board, card "p1 VNode Bridge Duplication": https://github.com/orgs/QwikDev/projects/6/views/1


JerryWu1234:
done, @wmertens please take a look

wmertens (approved):
LGTM

Issue #8724, opened 2026-06-11
[🐞] V2 Updating route-level CSS files isn't reflected immediately by vite dev's HMR


briancarbone:
### Which component is affected?

Qwik Rollup / Vite plugin

### Describe the bug

Updating component-level CSS is reflected immediately in the browser, but updating route-level CSS is not. I've tested this both on the latest build/v2 and rolldown (Vite 8) branches; hot reload of route-level CSS modules works on neither.

### Reproduction

n/a

### Steps to reproduce

Import, for example, `routes/index.module.css` into `routes.tsx`. Change to the tsx file are reflected immediately in `vite run dev`, but changes to the CSS module are not.

### System Info

```shell
System:
    OS: Linux 7.0 cpe:/o:nixos:nixos:26.11 26.11 (Zokor)
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 30.67 GB / 62.71 GB
    Container: Yes
    Shell: 5.3.9 - /nix/store/cgjr3kj3hs7ngznyws5qfg16c8scpys0-bash-interactive-5.3p9/bin/bash
  Binaries:
    Node: 26.3.0 - /nix/store/nk72hijagb0yc9rax5rr4gw8mpvmq7vw-nodejs-26.3.0/bin/node
    npm: 11.16.0 - /nix/store/nk72hijagb0yc9rax5rr4gw8mpvmq7vw-nodejs-26.3.0/bin/npm
  npmPackages:
    typescript: 6.0.3 => 6.0.3
    vite: ^8.0.16 => 8.0.16
```

PR #8725, opened 2026-06-12
fix(qwik-router): hot-reload CSS imported by route files

briancarbone:
What is it?

- Bug

Description

In dev, editing CSS imported by a route file (index.tsx / layout.tsx) had no effect. The dev server had to be restarted to see the change. CSS imported by a component updated instantly, as expected.

Qwik resolves these route CSS imports itself and injects them as <link> tags, so the module only exists in the SSR module graph. Vite watches its own client module graph plus the route source-file globs, so it never watched these files, and a change fired no HMR event at all. Component CSS worked because Vite tracks and handles it natively.

The fix:

- Register each injected CSS file with Vite's watcher so edits are seen.
- On change, send a css-update to the client so Vite swaps the <link> in place, matching native component CSS behavior.
- CSS already in the client graph is left to Vite's native HMR, so component styling is untouched.

Scoped to route-injected CSS in dev mode. Production builds and component CSS are unaffected.

Checklist
  - [x] I added a changeset with pnpm change
  
Fixes https://github.com/QwikDev/qwik/issues/8724

### DIFF
--- a/.changeset/router-css-hmr.md
+++ b/.changeset/router-css-hmr.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: hot-reload route-imported CSS in dev without a server restart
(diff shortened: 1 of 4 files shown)


wmertens (approved):
LGTM

wmertens (approved):

PR #8730, opened 2026-06-12
fix(router): rerun catchall route loaders on SPA navigation

Varixo:
Ensure route loader signals track route dependencies even when initialized from SSR-injected data, so navigating between URLs that match the same catchall route refreshes loader data with the new params and URL.


wmertens (approved):
LGTM 🏝️

PR #8735, opened 2026-06-12
perf: reduce component rerender when props are not changed

Varixo:
dont rerender component when they are the same signal or signal with the same value or the same class or style


wmertens (approved):
LGTM

PR #8736, opened 2026-06-13
fix(core): rewind cursor when attaching dirty descendants

Varixo:
Ensure signal-bound DOM updates are processed when a signal changes while an existing render cursor is already in progress.


packages/qwik/src/core/shared/vnode/vnode-dirty.ts
```diff
@@ -109,6 +109,10 @@ function findAndPropagateToBlockingCursor(container: Container, vNode: VNode): b
       // remember that cursor's nearest boundary for async/suspense bookkeeping.
       setNearestCursorBoundary(vNode, cursorBoundary);
       propagatePath(current);
+      const cursorData: CursorData = getCursorData(current)!;
+      if (cursorData.position !== current) {
+        cursorData.position = vNode;
```
wmertens: Can't this cause children to be processed before dirty parents?

Varixo:
Closing for now, probably fixed by another PR https://github.com/QwikDev/qwik/pull/8743

Issue #8741, opened 2026-06-16
V2: No supported way to exclude prerendered routes from the server bundle


briancarbone:
I prerender a large content section (a few hundred MDX pages) and deploy to a single-artifact edge runtime. The prerendered HTML is served from KV and those routes never run on the server, but their compiled components are still bundled into the server build, which blows the binary size limit.

`ssg.include` only controls what gets prerendered, but it does not remove those routes from the generated route table (@qwik-router-config), which statically imports every route in `routesDir`.

Is there a supported way to keep a static-only route out of the server bundle? `createRoutes` and `generateQwikRouterConfig` in the router vite plugin emit import() for every route with no exclusion. The SSG step prerenders HTML and injects the static paths array but never prunes the table.

Using
@qwik.dev/router f535f59
@qwik.dev/core f535f59

PR #8742, opened 2026-06-17
feat(qwik-router): exclude prerendered routes from the server route plan

briancarbone:
# What is it?

  - Feature / enhancement
  - Fixes https://github.com/QwikDev/qwik/issues/8741

  # Description

  Add an opt-in `serverExcludeRoutes` option (glob patterns, same syntax as the SSG adapter's `ssg.include`). Matching static page routes are omitted from the runtime server route plan, so their components tree-shake out of the deployed server bundle. This keeps size-capped edge runtimes (e.g. Cloudflare Workers) from growing with prerendered page count.

  SSG still prerenders the routes and the client/SPA build is unaffected. The SSG build entry imports a new `@qwik-router-config?ssg` variant that always carries the full route plan; dev and the client build keep every route as well. Exclusion applies to the production server build only.

  Restricted to static (non-dynamic) routes with no server `action$` or per-request SSR fallback; the user certifies a listed route is fully static.

  ## How it works

  - Matched with the existing `createRouteTester` (the same matcher SSG uses), so the glob syntax is identical to `ssg.include`.
  - In `createRoutes`, a matched static route's loader is skipped, so its trie node prunes and the route's chunk tree-shakes
  out. `routeIdMap` membership stays the single source of truth (the `_R` loader-hash emission now mirrors it).
  - A new `@qwik-router-config?ssg` virtual-module variant carries the full plan for the SSG build entry; the default config
  (runtime server) gets the filtered plan. Dev and the client build also use the full plan, so local dev and SPA navigation are unaffected.

  ## Why opt-in and static-only

  At codegen time the build has no route-handler info (no `action$`/`onRequest` detection, loaders aren't optimized yet), so it cannot prove a prerendered route is safe to drop: a route can be prerendered for GET yet still need runtime SSR (e.g. a POST `action$`, which `isStaticPath` does not intercept). The option is therefore a user certification, and v1 is limited to static (non-dynamic) routes to avoid partial-prerender edge cases. Default behavior is byte-identical (empty list).

  # Checklist

AI-assisted (Opus 4.8), could use a more thorough review.

### DIFF
--- a/.changeset/exclude-prerendered-routes-from-server.md
+++ b/.changeset/exclude-prerendered-routes-from-server.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': minor
+---
+
+feat(router): automatically omit fully-prerendered, server-free routes from the production SSR route plan so their chunks tree-shake out of size-capped server bundles.
--- a/packages/qwik-router/modules.d.ts
+++ b/packages/qwik-router/modules.d.ts
@@ -17,6 +17,11 @@ declare module '@qwik-router-config' {
   export default defaultExport;
 }
 
+declare module '@qwik-router-config?ssr' {
+  export * from '@qwik-router-config';
+  export { default } from '@qwik-router-config';
+}
+
 declare module '@qwik-city-plan' {
   export * from '@qwik-router-config';
   export { default } from '@qwik-router-config';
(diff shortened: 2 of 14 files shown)


briancarbone:
This would be better implemented by detecting routes with non-GET server entry points; then you could have just `ssg` in the config while still generating the necessary server code. In any case, stripping prerendered routes from the bundle is actually required for me to stay within edge platform limits.

wmertens (changes requested):
I think the concept is great, but I think it's SSR that should be special-cased (`?ssr`) for the routerconfig because both the client and SSG need the full trie, no?

Furthermore, a filtered trie should also remove all intermediate nodes that no longer have a leaf, I'm not sure that's happening, can you add a test?

wmertens (approved):
LGTM

PR #8743, opened 2026-06-17
fix: preserve pending UI updates across frame yields

Varixo:
This fixes an issue where large UI updates spanning multiple frames could restart incorrectly and lose pending work. The update preserves in-progress render state when a cursor advances back to itself, preventing missed UI changes after yielding to the browser.


wmertens (approved):
LGTM

PR #8748, opened 2026-06-18
fix(router): preserve route loader subscriptions during seq cleanup

Varixo:
Fixed SPA navigation in cases where route loader data could stop updating after navigating between catch-all routes. Pages now correctly re-render their content when returning through client-side navigation, including layouts that share loader data through context.

### DIFF
--- a/packages/qwik/src/core/reactive-primitives/types.ts
+++ b/packages/qwik/src/core/reactive-primitives/types.ts
@@ -123,14 +123,10 @@ export interface AsyncSignalOptions<T> extends ComputedOptions {
   timeout?: number;
 }
 
-export const enum SignalFlags {
+export const enum ComputedSignalFlags {
   INVALID = 1,
   RUN_EFFECTS = 2,
-}
-
-export const enum WrappedSignalFlags {
-  // should subscribe to value and be unwrapped for PropsProxy
-  UNWRAP = 4,
+  PRESERVE_ON_SEQ_CLEANUP = 4,
 }
 
 export const enum SerializationSignalFlags {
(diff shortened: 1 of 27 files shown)


packages/qwik/src/core/reactive-primitives/types.ts
```diff
@@ -50,6 +50,16 @@ export type AsyncQRL<T> = QRLInternal<AsyncFn<T>>;
...
+  /**
+   * When true, the signal is not cleaned up when it is found in a component's sequential scope
+   * during component cleanup.
+   *
+   * This is intended for shared signals whose lifecycle is owned outside of the component that read
+   * them.
+   *
+   * Defaults to `false`.
+   */
+  preserveOnSeqCleanup?: boolean;
```
wmertens: this should be an internal-only option, no?

packages/qwik-router/src/runtime/src/qwik-router-component.tsx
```diff
@@ -590,6 +590,12 @@ export const useQwikRouter = (props?: QwikRouterProps) => {
       }
       updateRouteLoaderPaths(routeLoaderCtx, loadedRoute.$loaderPaths$, trackUrl);
       const routeLoaders = ensureRouteLoaderSignals(contentModules, loaderState, routeLoaderCtx);
+      // if (!isServer) {
```
wmertens: this should be removed

wmertens (changes requested):

wmertens (approved):
LGTM 🦋

PR #8751, opened 2026-06-20
feat(qwik-router): add `exclude` option to `rewriteRoutes`

### LINKED ISSUE #8741: V2: No supported way to exclude prerendered routes from the server bundle

briancarbone:


Varixo:
Hello! Thank you for your work, looking good. One more thing - you need to update docs here https://next.qwik.dev/docs/guides/rewrites/#rewrites

briancarbone:
Should be good to go now. 🙂

PR #8752, opened 2026-06-20
fix(qwik): honor stopPropagation() from a deferred event handler

maiieul:
# What is it?
- Bug

# Description
Split out of #8745 (it is unrelated to ErrorBoundary).

qwikloader walks the event path synchronously and collects one task list per element, then flushes them. A handler whose QRL is still importing can only call `stopPropagation()` after that walk has already moved on, so ancestor handlers on the bubbling path ran anyway. `runEventTasks` now re-checks `cancelBubble` between per-element task groups, so a deferred stop still skips later elements; a synchronous stop truncates the walk as before. Covered by new `qwikloader.behavior.unit.ts` cases (bubble + capture phase); the brotli budget bump reflects the added logic.


Varixo:
I dont understand it, thats why we have stoppropagation:*, no?

wmertens (approved):
LGTM

PR #8760, opened 2026-06-22
feat(router): render SSG in a dedicated Vite build environment

briancarbone:


packages/qwik-vite/src/plugins/vite.ts
```diff
@@ -398,8 +397,9 @@ export function qwikVite(qwikViteOpts: QwikVitePluginOptions = {}): any {
 
     configEnvironment(name: string, _config: EnvironmentOptions, _env: ConfigEnv) {
       // Use environment name to distinguish server vs client — config.consumer is not yet set
-      // at the time this hook is called.
-      const isServer = name === 'ssr';
+      // at the time this hook is called. Adapters may add their own server environment (e.g. `ssg`
+      // for static generation), which needs the same server treatment as `ssr`.
+      const isServer = name === 'ssr' || name === 'ssg';
```
wmertens, line 399: Hmm shouldn't this read the consumer from the environment options?

packages/qwik-router/src/adapters/shared/vite/index.ts
```diff
@@ -176,88 +198,91 @@ export function viteAdapter(opts: ViteAdapterPluginOptions) {
...
-            try {
-              const content = await fs.promises.readFile(staticPathsFile, 'utf-8');
-              staticPaths.push(...JSON.parse(content));
-              // Clean up the temporary file
-              await fs.promises.unlink(staticPathsFile);
-            } catch {
-              // No static paths file — SSG may have produced no pages
-            }
-          }
+        // Read (and remove) the static paths written by run-ssg.js, then drop the throwaway dir.
```
wmertens, line 269: 🤔 ah - I thought that notFound is a no-op nowadays, and therefore staticPaths is vestigial?

wmertens (commented):
Looking great, still wondering about a few things like: 

- maybe add an option to keep the SSG build, default no?
- how well does this work with a v1 project that got migrated?
- staticPaths generation/handling should be removed I think

packages/qwik-router/src/adapters/shared/vite/index.ts
briancarbone: No, actually, `staticPaths` is useful for adapter implementations. `isStaticPath` calls it and determines whether you hit the CDN or the router.

packages/qwik-vite/src/plugins/vite.ts
briancarbone: The consumer environment is undefined for 'ssr' because it's the implicit environment (same with the client environment).

wmertens (approved):
LGTM

wmertens (changes requested):
@briancarbone Hmm, if you visit https://briancarbone-feat-ssg-build.qwik-8nx.pages.dev/docs/advanced/sitemaps/ via the sidebar (broken link, to fix apparently but not due to this PR) then you get the custom 404. But if you visit directly you get the generic 404 error.

It does work in current v2 https://qwikdev-build-v2.qwik-8nx.pages.dev/docs/advanced/sitemaps/

PR #8771, opened 2026-06-25
fix(router): render error.tsx inside its layouts on a thrown error

briancarbone:


wmertens (commented):
On my phone, can't do a proper review yet, but could you add the route config as one of the SSG snapshot tests? Before this change. Then after the change we can easily see the effect

e2e/qwik-e2e/apps/qwikrouter-ssg-snapshot/expected.config.txt
```diff
@@ -0,0 +1,26 @@
...
+const SubLayout = ()=>import("<app>/src/routes/sub/layout");
+
+/** Qwik Router Routes (2) */
+export const routes = {
+  _I: ()=>import("<app>/src/routes/index"),
+  _R: "__LOADERS:<app>/src/routes/index.tsx__",
+  "sub": {
+    _L: SubLayout,
+    _R: "__LOADERS:<app>/src/routes/sub/404.tsx|<app>/src/routes/sub/layout.tsx__",
+    _4: [ SubLayout, ()=>import("<app>/src/routes/sub/404") ],
```
wmertens: it seems to me that during _4 discovery, SubLayout could be added in-memory instead of on-disk. Why is it better to do this on-disk?

e2e/qwik-e2e/tests/qwikrouter/ssg-snapshot.e2e.ts
```diff
@@ -135,9 +158,24 @@ async function checkBrotliBudget(label: string, content: string, budget: number)
   ).toBeLessThanOrEqual(budget);
 }
 
+// enforce:'pre' so we capture qwik-router's load() output before the bundler rewrites its imports.
```
wmertens, line 163: Note that this makes the config easy to read but does not capture the routeloader injection, which deserves a comment.

wmertens (changes requested):
I'm not sure about the utility here? Why is it better to bake?

wmertens (approved):
LGTM

PR #8772, opened 2026-06-26
fix(qwik): surface container resume failures instead of hanging

maiieul:
# What is it?
- Bug

# Description
The container state-resume pipeline (`process-state-data.ts`) swallowed a thrown resume: `scheduleProcessContainerStateData`'s `onError` marked the container done but never called `markContainerDataReady`, so every `whenContainerDataReady` waiter hung forever — and it logged nothing, so the failure was invisible (a silent hang rather than a surfaced error).

Now it reports the error via `logError` and unblocks waiters so the rest of the page can still resume (graceful degradation). Includes a regression test that drives the pipeline with a throwing iterator and asserts it resolves instead of timing out (it times out without the fix).

Split out of the ErrorBoundary PR (#8745), where this was discovered — it's an independent core-resume robustness fix.

### DIFF
--- a/.changeset/surface-resume-failures.md
+++ b/.changeset/surface-resume-failures.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix: surface a failed container resume — report it and unblock `whenContainerDataReady` waiters — instead of swallowing the error into a silent hang.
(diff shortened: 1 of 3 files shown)


wmertens (approved):
LGTM

PR #8774, opened 2026-06-26
feat(build): keep Qwik and Router .d.ts fresh in build.core.dev

maiieul:
# What is it?
- Infra

# Description

`build.core.dev` did zero declaration work, so editing a public signature left stale `.d.ts` until a full `build.core`, meaning slower iterations cycles.

image

> For context, the alternative to get the same fresh types is build.core at ~77s — so the dev loop is ~15× faster and now correct.

It now re-emits fresh Qwik **and Router** types incrementally — a single `emitDeclarationOnly` tsc over both packages (with core/router resolved to source) — and writes one-line re-export shims pointing at the per-file `dts-out`, instead of the slow, non-incremental api-extractor rollup.

A few seconds warm (~5s, ~6.6s after a router edit) vs ~75s for a full `build.core`. `build.watch` is untouched (stays esbuild-only), and prod `build.core` is unchanged — it overwrites the shims with the real rollups.

### DIFF
--- a/.ruler/AGENTS.md
+++ b/.ruler/AGENTS.md
@@ -69,6 +69,8 @@ pnpm vitest run packages/qwik/src/core/tests/use-task.spec.tsx
 pnpm playwright test e2e/qwik-e2e/tests/events.e2e.ts --browser=chromium --config e2e/qwik-e2e/playwright.config.ts
 ```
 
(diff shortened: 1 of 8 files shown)


wmertens (approved):
LGTM

PR #8727, opened 2026-06-12
test: fix non-asserting e2e checks (missing awaits, no-op assertions)

voidmatcha:
# What is it?

- Docs / tests / types / typos

# Description

This PR strengthens existing Playwright checks in the Qwik starter and docs E2E suites. It does not add product behavior or rename tests.

The fixes are focused on assertions that currently do not enforce the behavior they describe:

- Prefix `await` on 7 web-first assertions whose promises were previously discarded (`toHaveURL` and `toHaveClass`).
- Replace `expect(await input.getAttribute("prevented")).toBeDefined()` with the repo's existing custom matcher idiom, `await expect(input).hasAttribute("prevented")`. The old form passes for `null`, so it did not verify that the attribute exists.
- Replace a discarded `isVisible()` boolean and a bare `getByText()` locator statement with `await expect(...).toBeVisible()`.

Blame check:

- The docs mobile menu class checks trace to 1230712613 from 2025-07-15.
- The docs resource locator check traces to dd80d5e6ca from 2025-01-17.
- The affected starter checks trace at least as far back as 7e742eb3a from 2024-06-25.

Verification:

- Re-fetched `origin/main` on 2026-06-13.
- Applied commit 4d25770799accf43eb409dc4c845b46acf7eeaed to a temp worktree based on current `origin/main` and ran:
  `bash scripts/pr-preflight.sh tmp/preflight-qwik e2e/docs-e2e/tests/Docs/navBarOnMobile.spec.ts starters/e2e/e2e.sync-qrl.spec.ts starters/e2e/e2e.use-id.spec.ts starters/e2e/qwikcity/nav.spec.ts starters/e2e/qwikcity/resource.spec.ts`
- Result: `PREFLIGHT 0 fail(s)` with scanner delta `total from 16 to 9, p0 from 9 to 2, ast from 5 to 0`; AST artifacts clean; diff hygiene clean; authoring clean with 0 added comments and no test renames.
- Not run locally: repo TypeScript, lint, or Playwright E2E. This clone has no `node_modules`, and the Qwik E2E path requires the repo dependency and app setup. The preflight therefore skipped tsc, eslint, and Playwright because local repo binaries were absent.

Checklist:

- [x] My code follows the developer guidelines of this project
- [x] I performed a self-review of my own code
- [x] Changeset not needed: test-only assertion fixes, no package release note
- [x] Docs not needed: no public docs behavior changes
- [x] Existing tests updated by strengthening their assertions

---

[...]

PR #8777, opened 2026-06-26
test(qwik-e2e): assert non-asserting e2e checks

voidmatcha:
# What is it?

# Description

Two e2e checks in the qwik-e2e suite silently always pass, so they cannot catch a regression:

- sync-qrl.e2e.ts asserted expect(await input.getAttribute('prevented')).toBeDefined().
  getAttribute returns string | null and both pass toBeDefined(), so the check never fails.
  The sync-qrl app sets the attribute to String(e.defaultPrevented), and the test intent is to
  confirm the default was prevented, so this now asserts toHaveAttribute('prevented', 'true').
- use-id.e2e.ts called await totalIdsLocator.isVisible() and discarded the boolean, asserting
  nothing. Replaced with the web-first await expect(totalIdsLocator).toBeVisible().

Both edits strengthen existing checks without changing test names or behavior. Test-only change,
no published package touched, so no changeset is needed.

This re-targets QwikDev/qwik#8727 onto build/v2 at wmertens' request: that PR was approved on main
but closed because development is focused on build/v2 and merging would conflict. The same
anti-patterns are present in the restructured v2 suite, so the fix is reproduced here.

# Testing

Verified statically: the new assertions use Playwright matchers valid for this suite, and the
sync-qrl change matches the app component that sets the attribute to String(e.defaultPrevented).
The e2e specs were not run locally, since the run needs pnpm install plus a core build and a
browser download. This is disclosed rather than claimed.

---

Found while reviewing the test suite with [`e2e-skills/e2e-reviewer`](https://github.com/voidmatcha/e2e-skills#skill-2-e2e-reviewer--quality-review).

### DIFF
--- a/e2e/qwik-e2e/tests/sync-qrl.e2e.ts
+++ b/e2e/qwik-e2e/tests/sync-qrl.e2e.ts
@@ -24,6 +24,6 @@ test.describe('resuming', () => {
     // clicking checkbox does not toggles the checked state, because default is prevented.
     await input.click();
     await expect(input).toBeChecked();
-    await expect(await input.getAttribute('prevented')).toBeDefined();
+    await expect(input).toHaveAttribute('prevented', 'true');
   });
 });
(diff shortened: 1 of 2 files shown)


wmertens (approved):
LGTM

PR #8778, opened 2026-06-26
feat(router): make view transitions opt-in via the viewTransition prop

maiieul:
# What is it?
- Feature / enhancement

# Description
View transitions on SPA navigation are now opt-in. `viewTransition` defaults to `false` — set `<QwikRouterProvider viewTransition={true}>` to enable. Default-on caused slowness and de-synced transitions with SPA navigation + Suspense.


.changeset/opt-in-view-transition.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/router': minor
```
wmertens: can you make this major? It's a breaking change

maiieul: Ah didn't see your review. Major should only be for v1 breaking changes no?

PR #8784, opened 2026-06-30
fix: preserve props during eager SSR resume

Varixo:
This fixes an eager SSR resume issue where statePrewarm could restore component props proxies with missing constant props. During deserialization, referenced prop objects may be allocated before they are fully inflated, so the props proxy owner now preserves those references instead of normalizing them away.


wmertens (approved):
LGTM 🐕

PR #8785, opened 2026-06-30
feat: support Vite 8 with the Rolldown bundler

briancarbone:


wmertens:
looks great but at this point I don't think we want to require Vite 8 yet. Can't we already get some of this with Vite 6 or 7?

briancarbone:
> looks great but at this point I don't think we want to require Vite 8 yet. Can't we already get some of this with Vite 6 or 7?

It seems you _could_ feasibly use Rolldown with Vite 7 via the `rolldown-vite` package... I hasten to wonder whether the version restriction could be relaxed without any other changes.

wmertens:
I see two changes here: the graph fix and the vite8.

Ideally, the graph fix is a separate PR.

For vite8, I see that the incompatibility is mostly the config options? Can't we detect the vite version and use the correct key based on that?

Vite 8 has some breaking changes around CJS compatibility, and if we can prolong vite 7 compatibility, we should.

briancarbone:
> I see two changes here: the graph fix and the vite8.
> 
> Ideally, the graph fix is a separate PR.
> 
> For vite8, I see that the incompatibility is mostly the config options? Can't we detect the vite version and use the correct key based on that?
> 
> Vite 8 has some breaking changes around CJS compatibility, and if we can prolong vite 7 compatibility, we should.

It's not just the config options key, unfortunately. `manualChunks` was removed in Rolldown, first replaced by `advancedChunks` and then `codeSplitting`, so the chunk backend (rollup vs. rolldown) is sufficiently different.

AliMahmoudDev:
This looks solid! I like the approach of keeping changes minimal and focused. Small, targeted PRs like this are much easier to review and less likely to introduce regressions.

AliMahmoudDev:
Vite 8 + Rolldown support is exciting! This is a significant bundler upgrade. One thing to test: making sure tree-shaking still works correctly with Rolldown, as some bundlers handle side effects differently.

Varixo (commented):
thanks for the fix here. It should be a different PR to merge it sooner

packages/qwik/src/core/client/process-vnode-data.ts
```diff
@@ -591,9 +591,14 @@ function* processVNodeDataImpl(
             node as Element,
             contentBoundaryId
           );
-          nextNode = nextSibling(node);
+          let cursor: Node | null = node;
+          while (cursor && !(nextNode = nextSibling(cursor))) {
+            cursor = cursor.parentNode;
+          }
```
Varixo, line 597: good finding, but we need test for it, and also probably should be separate PR

briancarbone: added tests

briancarbone:
> thanks for the fix here. It should be a different PR to merge it sooner

No rush

maiieul:
Closing as superseded by https://github.com/QwikDev/qwik/pull/8909. We can implement the HMR fixes as follow ups now.

PR #8786, opened 2026-07-01
feat(cli): handle nub package manager commands

colinhacks:
The Qwik CLI already detects nub — `getPackageManager()` passes through the `which-pm-runs` name — but two command builders assume every non-npm manager accepts a bare `<pm> <script>`. That doesn't hold for [nub](https://nubjs.com): it has no implicit script shortcut, so `nub dev` is not valid; scripts run via `nub run dev`. Its executor is `nubx`, not a bare `nub <bin>`.

This routes nub through the `run`-prefixed form in `pmRunCmd()` (alongside npm) and maps it to `nubx` in `runInPkg()`. Install is unaffected — `installDeps` already spawns `<pm> install`.

### DIFF
--- a/packages/qwik/src/cli/utils/install-deps.ts
+++ b/packages/qwik/src/cli/utils/install-deps.ts
@@ -10,7 +10,7 @@ export function installDeps(pkgManager: string, dir: string) {
 }
 
 export function runInPkg(pkgManager: string, args: string[], cwd: string) {
-  const cmd = pkgManager === 'npm' ? 'npx' : pkgManager;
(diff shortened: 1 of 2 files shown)


wmertens (dismissed):
LGTM

colinhacks:
Thanks for the LGTM. The red checks (Release, All requirements are met) look like unrelated fork-CI — a cancelled macOS/webkit E2E job upstream of the release gate, not this diff. Let me know if anything's needed to get it over the line.

PR #8787, opened 2026-07-01
fix: prevent cold dev route loader requests from failing during SPA

Varixo:
In dev, the first SPA navigation to a cold route could request route loader data before the route-local loader metadata was fully available. This made q-loader requests fail with "Loader not found" / empty responses, especially for catchall routes or loaders not visible as direct route module exports. This PR adds a dev-only route loader registry so loader objects created during module evaluation can still be resolved by id for cold q-loader requests. It also keeps dev prefetch disabled to avoid speculative q-loader requests before dev metadata has settled.

### DIFF
--- a/.changeset/floppy-pandas-count.md
+++ b/.changeset/floppy-pandas-count.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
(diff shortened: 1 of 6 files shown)


wmertens (approved):
LGTM 🕺

PR #8789, opened 2026-07-02
fix(qwik-vite): keep reachable deps when reducing cyclic bundle graphs

briancarbone:


maiieul (commented):
Nice! I removed this in my Vite 8 PR because that led to the same bug you faced I think. Your solution is better because it's more performant. Can you trim/remove the unnecessary comments and only keep what's critical to understand the code?

briancarbone:
Should be better now

maiieul:
Sorry I forgot you had done a PR for this. I merged a similar PR already with your code 🙈. Thanks @briancarbone!

PR #8793, opened 2026-07-02
feat(router): blockSSR

wmertens:
Routeloaders block SSR by default and can be opted-out.


maiieul:
Comparing the two shapes for deferred/streamed loader data — a `defer`-marked loader (its own request) vs a **bare `Promise` field** returned inside a loader's value (rides the parent's request). They're not substitutes; each has real exclusives that follow from which side of the request boundary the streamed unit sits on.

### Capability comparison

| Capability | **`defer` loaders** | **bare-promise fields** |
|---|---|---|
| Per-unit HTTP caching (cacheKey / eTag / CDN, cross-route cache entry) | ✓ per loader | ✓ per loader (just split it out) · ✗ per *field* |
| Independent revalidation after an action | ✓ per loader | ✓ per loader · ✗ per *field* |
| Typed `.error` (`ServerError` envelope, prod redaction, closest-`ErrorBoundary` routing) per **streamed** unit | ✓ per loader | ✓ per loader · ✗ per *field* |
| Static knowability (build-time envelope enforcement, pre-head await set known without executing, typed `RequestEvent` with no `status`/`redirect`) | **✓ exclusive** | ✗ discovered only after the loader body runs; runtime guard only |
| Ships on existing machinery (whole-signal OOOS streaming) | ✓ | ✗ (needs a pending-promise wire protocol — the serializer currently await-flattens nested promises) |
| Co-location (streamed work closes over critical data, no `resolveValue` / embedded re-run) | ✗ | **✓ exclusive** |
| Declarations needed for critical + streamed | ≥ 2 loaders | 1 loader |
| Streamed units share one request / one middleware run (SPA nav) | ✗ one request per deferred loader | **✓** fields ride the parent's request |
| Dynamic deferral (await the fast path / stream the slow path per request) | ✗ (bot/SEO case is better a framework knob) | ✓ (niche) |
| **Dynamic cardinality** (N streamed units, N known only at runtime) | **✗ inexpressible** | **✓ exclusive** |

### The one thing only bare promises can express: dynamic cardinality

A search page: fast ID query, slow per-item hydration, and you want each result to **pop in individually** as its own fetch settles.

```tsx
// bare-promise fields — one streamed unit PER ITEM, count known only at runtime
export const useSearch = routeLoader$(async (ev) => {
  const ids = await searchIds(ev.query.get('q'));         // fast, critical, envelope-capable
  return { ids, items: ids.map((id) => fetchItem(id)) };  // N bare promises, N = runtime
});

// render: each card reveals as ITS OWN promise settles
{data.value.ids.map((id, i) => (
  <Await key={id} promise={data.value.items[i]} fallback={<SkeletonCard />}>
    {(item) => <ProductCard item={item} />}
  </Await>
))}
```

Why `defer` loaders can't match it:

```tsx
// Attempt 1 — one deferred loader with Promise.all: loses per-item streaming
export const useItems = routeLoader$(async (ev) => {
  const ids = await ev.resolveValue(useSearchIds);
  return Promise.all(ids.map(fetchItem));  // signal resolves when the SLOWEST item does
}, { defer: true });                        // → the whole grid pops at once, no per-item pop-in

// Attempt 2 — one loader per item: impossible
// loaders are static module-level exports; you can't declare useItem_0 … useItem_N
// when N depends on the request. One loader = one signal; there is no per-instance
// parameterization of a routeLoader$.
```

The unit of deferral for a `defer` loader is *the loader*, and loaders are statically declared — so a runtime-determined number of independently-streamed units is inexpressible. A bare promise's unit is *a value*, and values are dynamic (the serializer already walks values recursively, so promises inside arrays are natural for it).

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
```diff
@@ -310,24 +310,34 @@ function createResolveRequestHandlers() {
...
+      let allBlockSSRLoaders: Promise<void> | undefined;
+      for (let i = 0; i < routeLoaders.length; i++) {
+        const loader = routeLoaders[i];
+        const promise = loadRouteLoader(loader, requestEv);
+        // Handle every rejection so a background loader can't crash the request.
+        promise.catch(() => {});
+        if (loader.__blockSSR) {
+          allBlockSSRLoaders = (
+            allBlockSSRLoaders ? allBlockSSRLoaders.then(() => promise) : promise
+          ) as Promise<void>;
```
Varixo, line 343: why no Promise.all?

.changeset/lazy-route-loaders.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/router': minor
+---
+
+route loaders gain a `blockSSR` option (default `true`); set `blockSSR: false` to run a loader in the background without blocking SSR
```
Varixo: there should be "feat:" prefix

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
wmertens: this way, the first error in route order will throw even if there are multiple errors.
Also, the loads all start at the same time so the .then() chaining doesn't cause a waterfall

jordanw66:
In the PR it mentions the recommended setting for loaders is `blockSSR: false`

Should that be the default instead? Or less radically, could there be a global setting to change the default to `false`?

wmertens:
we discussed this and decided to make blockSSR itself experimental, but the functionality  is required for better v1 compatibility

PR #8796, opened 2026-07-03
feat(core): async auto tracking and async computed 

wmertens:
Converts async functions into generator functions and adds a harness so that reactive tracking can happen across await.

This also means that there is no more reason for AsyncSignal, so the async machinery moves to ComputedSignal, we will drop Asyncsignal before V2.0


wmertens:
Parking the autoTrack aspect for now, but moving useAsync into useComputed in another PR

PR #8797, opened 2026-07-03
fix: preserve scroll after spa action submits

Varixo:
The old prevUrl came from routeLocation.prevUrl, which can still point to the route before the current page. After SPA navigation, a plain routeAction$ submit was therefore mistaken for a form navigation from a different page, causing scroll reset. Now we use the actual URL at the start of the current navigation.


wmertens (approved):
LGTM

Issue #5929, opened 2024-03-03
[🐞] Error on response attempting to close WritableStream with Bun adapter


octet-stream:
### Which component is affected?

Qwik City (routing)

### Describe the bug

When using Qwik with Bun adapter (Bun.serve) with disabled SSG, the server throws an error attempting to close locked `WritableStream`. From what I can tell, the stream becomes locked after the `pipeTo` call [here](https://github.com/BuilderIO/qwik/blob/420f382a87a9f28829c28a46fdefab0b9430f56e/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts#L443) and it stays on that stage when `writebleStream.close()` is called [here](https://github.com/BuilderIO/qwik/blob/420f382a87a9f28829c28a46fdefab0b9430f56e/packages/qwik-city/middleware/request-handler/resolve-request-handlers.ts#L476). The server itself however responds without any issue, this error for some reason just shows up in logs, but I get this during every server response response.

### Reproduction

https://github.com/octet-stream/qwik-bun-stream-issue

### Steps to reproduce

1. Clone this repository;
2. Install dependencies via `bun i`;
3. Build the project for production `bun run build`;
4. Start the app via `bun run serve`;
5. Open [http://localhost:3000](http://localhost:3000) in your browser;

After the main page shows up, look at the terminal and you'll see following error:

```
$ bun server/entry.bun.js
Server started: http://localhost:3000/
636 |     } finally {
637 |       await stream.ready;
638 |       await stream.close();
639 |       await pipe;
640 |     }
641 |     await writableStream.close();
                          ^
TypeError: WritableStream.close method can only be used on non locked WritableStream
      at writableStreamCloseForBindings (:1:21)
      at /Users/octetstream/projects/qwik-bun-stream-issue/server/entry.bun.js:641:11
```

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 62.06 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.2 - ~/Library/Caches/fnm_multishells/2006_1708258753953/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - ~/Library/Caches/fnm_multishells/2006_1708258753953/bin/npm
    pnpm: 8.15.4 - /opt/homebrew/bin/pnpm
[...]

Issue #4381, opened 2023-05-31
[🐞] `server$` writes to a closed connection and crashes server with `ERR_STREAM_DESTROYED`


DustinJSilk:
### Which component is affected?

Qwik City (routing)

### Describe the bug

Clicking the button and refreshing the browser with the following code causes the server to crash:

```tsx
const search = server$(async () => {
  // Simulate a long running request
  await new Promise((resolve) => setTimeout(resolve, 5000));

  return "Some data";
});

export default component$(() => {
  return (
    <>
      <button
        onClick$={async () => {
          await search();
        }}
      >
        Click me then refresh
      </button>
    </>
  );
});
```

It looks like qwik is writing to a closed connection without checking if it is still available.
https://github.com/BuilderIO/qwik/pull/4290 might fix it, im hoping this issue expedites its release as this is crashing my production server regularly

### Reproduction

https://github.com/DustinJSilk/qwik-issue-server-crash

### Steps to reproduce

Run the app
Click the button
Refresh the browser
Watch the server crash

### System Info

```shell
N/A
```

DustinJSilk:
Fixed by #4290 !!

PR #8798, opened 2026-07-04
fix(router): await server$ stream close so a client disconnect can't crash Bun

aggyomfg:
## What is it?

- Bug

## Description
A streaming `server$` (an `async function*` consumed as a stream) crashes the
**entire Bun process** when the client disconnects mid-stream (SPA navigation
away, tab close, refresh, network drop).

When the client goes away, the response writable is errored, so `write`/`close`
reject. The async-iterator branch of `runServerFunction` called `stream.close()`
**without awaiting or catching it**, so the rejection went unhandled. On Bun the
default `unhandledRejection` behavior is to abort the process (Node only logs a
warning, which is why this stayed hidden there). After the crash every following
request fails with `ECONNREFUSED` — the server is gone.

The fix extracts the streaming branch into `streamServerFunctionResult` and
`await`s + swallows both the per-item `write` and the final `close`, so a
disconnected writable tears down quietly instead of leaking a rejection.

**Reproduction:** a streaming `server$` on the Bun adapter — open the stream,
read one chunk, then abort the request while it is still producing; the process
exits.

**Related:** #5929 (fixed the `TextEncoderStream` polyfill, not this close),
#4381 (Node `ERR_STREAM_DESTROYED` on the same class of teardown),
oven-sh/bun#18228 (Bun's stance: the framework must catch the close).

Test drives `streamServerFunctionResult` with a writable whose `close()` rejects (the
post-disconnect state) and asserts no unhandled rejection escapes. It is red
without the fix (leaks `Cannot close a writable stream that is closed or
errored`) and green with it. The process-abort half is Bun-specific and can't be
reproduced under Node, but the leaked-rejection root cause is covered here.

## Checklist

### DIFF
--- a/packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
+++ b/packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
@@ -554,20 +554,7 @@ function createResolveRequestHandlers() {
       }
       throwIfControlFlowSignal(result);
       if (isAsyncIterator(result)) {
-        ev.headers.set('Content-Type', 'text/qwik-json-stream');
-        const writable = ev.getWritableStream();
-        const stream = writable.getWriter();
-        for await (const item of result) {
-          if (isDev) {
-            verifySerializable(item, qrl);
-          }
-          const message = await _serialize(item);
-          if (ev.signal.aborted) {
-            break;
-          }
-          await stream.write(encoder.encode(`${message}\n`));
-        }
-        stream.close();
+        await streamServerFunctionResult(ev, result, qrl);
       } else {
         verifySerializable(result, qrl);
         ev.headers.set('Content-Type', 'application/qwik-json');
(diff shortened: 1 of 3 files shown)


packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
```diff
@@ -578,6 +565,31 @@ function createResolveRequestHandlers() {
...
+    for await (const item of result) {
+      if (isDev) {
+        verifySerializable(item, qrl);
+      }
+      const message = await _serialize(item);
+      if (ev.signal.aborted) {
+        break;
+      }
+      // Swallow rejection: writable may be errored by a client disconnect.
+      await stream.write(encoder.encode(`${message}\n`)).catch(() => {});
```
maiieul: Could we log the error message in dev mode at least?

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
```diff
@@ -578,6 +565,31 @@ function createResolveRequestHandlers() {
     }
   }
 
+  // Streams an async-iterator server function result to the response writable. A mid-stream
```
maiieul: Please remove this comment

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
aggyomfg: Added dev-logs.

packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
aggyomfg: removed.

wmertens (approved):
LGTM

wmertens:
@aggyomfg the unit test is failing on an undefined qrl

aggyomfg:
> @aggyomfg the unit test is failing on an undefined qrl

fixed in https://github.com/QwikDev/qwik/commit/55638e4f1981ae6e2b0baa2952f3d626eb442f49

PR #8803, opened 2026-07-06
refactor(devtools): extract findComponentKey as a shared runtime utility

Aejkatappaja:
## Problem

`findComponentKey` was inlined inside the hook runtime installer, and the "name after the last underscore" derivation was duplicated across `getComponentDetail`, `setSignalValue`, and `getComponentTreeSnapshot`.

## Solution

Promote both to top-level functions in `installers.ts`:

- `__qwik_find_component_key__(state, componentName, qrlChunk)`: pure (it takes the state record instead of closing over it), prefers an exact QRL chunk suffix, falls back to a case-insensitive name match.
- `__qwik_derive_component_name__(path)`: the shared name derivation, reused by the key lookup and by `getComponentTreeSnapshot`.

Both are declared as top-level `function` declarations, so `Function.prototype.toString()` keeps their names, and the runtime emitter includes them in the injected bundle alongside the installer. The Vite plugin virtual module and the generated extension script both still resolve them. Behaviour is unchanged.

Because the utilities are now pure, they are unit tested directly.

## Verification

- kit and plugin build; both `createHookRuntime()` (plugin) and `createExtensionHookRuntime()` (extension) emit the two utilities and reference them from the installer.
- Executed the generated extension runtime against a stubbed window: component tree naming, component-detail lookup by name and by QRL chunk, and signal editing all behave as before.
- `hook-runtime.unit.ts` passes, including new direct tests for both utilities. eslint passes on the touched files.

## References

QwikDev project board, card "p1 extract findComponentKey function as a utility function": https://github.com/orgs/QwikDev/projects/6/views/1


JerryWu1234:
approved

maiieul (approved):

PR #8804, opened 2026-07-07
refactor(devtools): share the page data source contract between ui and extension

Aejkatappaja:
## Problem

The browser extension's `createRemotePageDataSource()` in `extension-data-provider.ts` structurally mirrored the `PageDataSource` interface declared in `packages/devtools/ui/src/devtools/page-data-source.ts`, without depending on it. A change to the UI data-source contract would not be caught against the extension implementation.

## Solution

Make `PageDataSource` a single shared contract.

- Move the `PageDataSource` interface into `@qwik.dev/devtools/kit` (next to the related `InPageBridge`). All of its member types already live in kit, so no new imports are needed.
- The UI overlay source imports `PageDataSource` from kit (and re-exports it under the same name), and `InPageDataSource` keeps implementing it.
- The extension annotates `createRemotePageDataSource(): PageDataSource`, so TypeScript now checks the remote implementation against the shared contract.

## Verification

- kit builds and exports `PageDataSource`.
- `tsc` on the browser extension passes (only two pre-existing wxt `TS2742` errors remain, unrelated to this change), confirming the remote data source satisfies the contract.
- `tsc` on the devtools UI passes with zero errors.
- eslint passes on the touched files.

## References

QwikDev project board, card "p2 Page Data Source Shape Duplication": https://github.com/orgs/QwikDev/projects/6/views/1

### DIFF
--- a/.changeset/shared-page-data-source.md
+++ b/.changeset/shared-page-data-source.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/devtools': patch
+---
+
+refactor: share the page data source contract between the ui and extension
--- a/packages/browser-extension/src/shared/extension-data-provider.ts
+++ b/packages/browser-extension/src/shared/extension-data-provider.ts
@@ -1,5 +1,6 @@
 import type {
   Component,
+  PageDataSource,
   QwikDevtoolsComponentSnapshot,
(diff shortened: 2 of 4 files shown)


JerryWu1234:
approved

maiieul (approved):

PR #8805, opened 2026-07-07
fix: loading route loaders for dev server

Varixo:
This PR removes the dev-only global route loader registry and replaces it with route-scoped preparation for q-loader requests.

The registry allowed loaders to be resolved from global state, even when they were not available through the currently matched route. That made dev behavior less deterministic and different from production. Instead, the dev middleware now prepares the matched route before handing the request to the router: it refreshes routing state, loads server plugins and route modules, and makes the requested loader available only for that request.

### DIFF
--- a/.changeset/tidy-candles-guess.md
+++ b/.changeset/tidy-candles-guess.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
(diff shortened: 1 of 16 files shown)


wmertens (approved):
LGTM

PR #8807, opened 2026-07-07
feat(core): async useComputed$

wmertens:
This PR deprecates `useAsync$` and instead moves all async the machinery to ComputedSignal.

This means that the synchronous part of the compute function gets reactive-tracked automatically, and after the first await, `track()` needs to be used.


packages/qwik/src/core/reactive-primitives/impl/computed-signal-impl.ts
```diff
@@ -36,35 +108,99 @@ export class ComputedSignalImpl<T, S extends QRLInternal = ComputeQRL<T>>
...
    * The compute function is stored here.
    *
-   * The computed functions must be executed synchronously (because of this we need to eagerly
-   * resolve the QRL during the mark dirty phase so that any call to it will be synchronous). )
+   * Sync computed functions must be executed synchronously (because of this we need to eagerly
+   * resolve the QRL during the mark dirty phase so that any call to it will be synchronous).
    */
   $computeQrl$: S;
-  $flags$: ComputedSignalFlags | SerializationSignalFlags;
+  $flags$: number;
```
Varixo, line 116: I think it should be an enum

packages/qwik/src/core/reactive-primitives/impl/computed-signal-impl.ts
Varixo, line 1: I dont quite like that. The implementation is really loooong, even if someone is not using features, they need to download all the code, can we do it more tree shakeable? 🤔

wmertens: It's 200 brotli bytes more for the router e2e build hmmm.

wmertens: Well, we can't determine if the result is a Promise at build time, and making the async machinery lazy loaded dependent on that requires overhauling the core build and that's definitely a different PR.

wmertens:
@copilot resolve the merge conflicts in this pull request

PR #8808, opened 2026-07-08
fix(qwik): dedupe already-streamed-chore SSR warning to once per host

maiieul:
# What is it?
- Bug

# Description
The SSR dev warning "A chore was scheduled on a host element that has already been streamed" fired once per chore, so a re-dirtied streamed host (e.g. a server-side `setInterval` bumping a signal) spammed N identical copies. Now deduped to once per host, so distinct problematic hosts still each warn once. Pre-existing dev-only warning, unrelated to any feature work.


wmertens (approved):
LGTM

PR #8809, opened 2026-07-08
fix(core): allow invalid harmless tag nesting

wmertens:
We should only refuse DOM structures that change when the browser parses the HTML. `<div>` inside `<button>` is retained by the browser.


Varixo (changes requested):
Even if all the browser around behave here the same the implementation is wrong imo. You just need to remove these tags from config

wmertens:
@Varixo you mean, don't even warn for invalid-spec but valid-parsing HTML?

PR #8810, opened 2026-07-08
build(repo): raise TypeScript lib floor to es2022

maiieul:
# What is it?
- Infra

# Description
Bumps the TypeScript `lib` floor from es2021 to es2022 in the root tsconfig (core, router, optimizer, vite, react all inherit it). Safe against our support matrix — the lowest declared runtime is Node 18.17, and every es2022 API (Error.cause, Object.hasOwn, `.at()`, class static blocks) lands by ~Node 16.11. No public API change.

Also drops one now-redundant `as Error` cast the bump makes unnecessary, and points qwik-vite's built-in dev linter at es2022 (patch).

### DIFF
--- a/.changeset/dev-eslint-es2022.md
+++ b/.changeset/dev-eslint-es2022.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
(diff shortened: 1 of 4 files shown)


wmertens (approved):
LGTM 🥏

PR #8726, opened 2026-06-12
fix(qwik-router): prevent duplicated assetsDir in static paths

intellix:
# What is it?

- Bug (with tests)

# Description

Issue (from v1) but still applies for us: https://github.com/QwikDev/qwik/issues/8151

# Checklist

- [ ] My code follows the [developer guidelines of this project](https://github.com/QwikDev/qwik/blob/main/CONTRIBUTING.md)
- [ ] I performed a self-review of my own code
- [x] I added a changeset with `pnpm change`
- [ ] I made corresponding changes to the Qwik docs
- [x] I added new tests to cover the fix / functionality

PR #8811, opened 2026-07-08
fix(router): respect vite build.assetsDir in static paths

### LINKED ISSUE #8151: [🐞] V1 build.assetsDir causes public files to not work

maiieul:
# What is it?

- Bug

# Description

Fixes #8151. Supersedes #8726 — thanks @intellix for the analysis and the test fixtures reused here.

`postBuild` walked the client output labeled with `join(basePathname, assetsDir)` while public files live at the dist root. With `build.assetsDir` set this doubled the assets segment in static paths, mis-prefixed root public files like robots.txt, and let `cleanStaticGenerated` delete prerendered pages. The walk now starts at `basePathname` with the ignore prefixes derived from `assetsDir`, and the computed prefixes are injected into `isStaticPath` so staticFile-based adapters serve bundles under a custom assetsDir too.

The deno e2e variant now builds with `assetsDir: 'q'` and prerenders a few routes (including a static loader sidecar) to cover this end-to-end.

### DIFF
--- a/.changeset/assets-dir-static-paths.md
+++ b/.changeset/assets-dir-static-paths.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: respect vite build.assetsDir in static paths and public files
--- a/.changeset/user-static-paths-raw.md
+++ b/.changeset/user-static-paths-raw.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: user-provided static file paths match without a trailing slash
--- a/e2e/adapters-e2e/adapters/deno/vite.config.ts
+++ b/e2e/adapters-e2e/adapters/deno/vite.config.ts
@@ -10,6 +10,13 @@ export default extendConfig(baseConfig, () => {
         input: ['src/entry.deno.tsx'],
       },
     },
-    plugins: [denoServerAdapter({ name: 'deno' })],
+    plugins: [
+      denoServerAdapter({
+        name: 'deno',
+        // Prerender routes with and without loaders to cover static serving of
+        // pages and loader sidecars.
+        ssg: { include: ['/profile/', '/loaders/*'] },
+      }),
+    ],
   };
 });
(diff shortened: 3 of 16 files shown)


maiieul:
Superseded by #8817, which removes Qwik's special `build.assetsDir` handling entirely (v2 breaking change) in favor of native Rollup `output.assetFileNames` — that relocates assets consistently on client + SSR, so the assetsDir-specific static-path machinery here is no longer needed.

The two assetsDir-independent correctness fixes from this PR — base-pathname-aware static prefixes and the user-static raw/unslashed match — are intentionally dropped for now; they can be revisited via #8812 / #8813.

PR #8816, opened 2026-07-09
fix(qwik-vite): bootstrap QRL segment parents missing from the dev graph

maiieul:
# What is it?

- Bug

# Description

Needed for Vitest browser mode in the monorepo.

In dev, `load()` serves a QRL segment by lazily transforming its parent module — but only when the parent is already in the requesting server's module graph. If it isn't, it logs `module … does not exist in the build graph!` and the segment 404s.

That happens whenever SSR runs in a _different_ Vite server than the one resolving QRLs at runtime — e.g. Vitest browser mode, or any split server/browser SSR setup. The browser's Vite server never imported the component modules, so their segments can't be served.

`resolveId` already records the segment → parent link (it derives the parent from the segment URL). This PR uses it: when the parent isn't in the graph, transform it by id to bootstrap it instead of erroring. If it genuinely can't be transformed, it falls through to the existing graceful "not found" rather than throwing.

Dev-only — the branch is gated behind `devServer`, so production builds are unaffected.

Surfaced by [vitest-browser-qwik](https://github.com/QwikDev/vitest-browser-qwik), which currently works around this by pre-crawling the browser module graph before every SSR render; with this fix that workaround can go away.

# Checklist

### DIFF
--- a/.changeset/dev-qrl-parent-bootstrap.md
+++ b/.changeset/dev-qrl-parent-bootstrap.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+Dev-mode QRL segments now resolve even when their parent module was never loaded in the requesting Vite server (e.g. Vitest browser mode or SSR rendered in a separate server). The dev server transforms the parent on demand instead of erroring with "module does not exist in the build graph".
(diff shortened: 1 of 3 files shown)


wmertens (approved):
LGTM

PR #8818, opened 2026-07-09
fix: preserve transitive server$ modules in SSR builds

Varixo:
Fixes SSR server function registration when server$ is only reachable through stripped client-side code or transitive imports. The router now preserves discovered server function modules through tree-shaking, including packages marked with sideEffects: false.

Examples:

1. Transitive server$ import
```ts
// route.tsx
import { ProductPage } from './product-page';
// product-page.tsx
import { emitViewContent } from './emit-view-content';

export const ProductPage = component$(() => {
  // ...
});
// emit-view-content.ts
const emitViewContentServer = server$(async function () {
  return 'ok';
});

export async function emitViewContent() {
  return emitViewContentServer();
}
```
Fixed path:

route -> product-page -> emit-view-content -> server$
2. server$ only used inside stripped client code
```ts
import { emitViewContent } from './emit-view-content';

useVisibleTask$(async () => {
  await emitViewContent();
});
```
Before, SSR stripping removed the visible task usage, so the import could disappear from the SSR graph.

3. Dynamic import inside stripped client code
```ts
useVisibleTask$(async () => {
  const { emitViewContent } = await import('./emit-view-content');
  await emitViewContent();
});
```
Before, the literal dynamic import was lost when useVisibleTask$ was stripped from SSR.

4. Modules inside packages with sideEffects: false
```ts
{
  "sideEffects": false
}
// emit-view-content.ts
const emitViewContentServer = server$(async function () {
  return 'ok';
});

export async function emitViewContent() {
  return emitViewContentServer();
}
```
Before, the router generated only:

import './emit-view-content';
Rollup could tree-shake that away. Now it keeps the module exports live.

### DIFF
--- a/.changeset/silver-frogs-flash.md
+++ b/.changeset/silver-frogs-flash.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: register aliased server functions used only in client code like visible tasks
--- a/e2e/qwik-e2e/apps/qwikrouter-test.prod/src/root.tsx
+++ b/e2e/qwik-e2e/apps/qwikrouter-test.prod/src/root.tsx
@@ -1,8 +1,9 @@
 import { RouterOutlet, useQwikRouter } from '@qwik.dev/router';
 import { RouterHead } from './components/router-head/router-head';
+import { component$ } from '@qwik.dev/core';
 import './global.css';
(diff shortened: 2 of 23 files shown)


wmertens (approved):
LGTM 🏌️

Issue #8794, opened 2026-07-03
[v2] in-order streaming: no flush before awaiting a component-body promise (already-rendered UI stays buffered)


maiieul:
## Which component is affected?

Qwik Runtime (SSR streaming, v2 / `build/v2`)

## Describe the bug

Under the default in-order buffering strategy (`auto`), the SSR writer does not flush the stream buffer before awaiting a qwik component body that returns a promise.

In `packages/qwik/src/core/ssr/ssr-render-jsx.ts`, the `isQwikComponent` branch enqueues `await jsxOutput` with **no** `streamHandler.flush()` (build/v2 ~line 314), while the plain `isPromise(value)` branch right above it enqueues `ssr.streamHandler.flush()` before its await (~line 148).

Consequence: everything already rendered before the pending component — including interactive UI the user could be seeing — sits in the server buffer until the async component resolves. With a slow data dependency the user stares at a partial/stale page even though the HTML for the content above was fully produced. With `streaming.inOrder.strategy: 'direct'` the prefix flushes correctly, which confirms the buffering site.

Found while writing the mid-stream ErrorBoundary e2e for #8745 (commit `a9fc79292` works around it by wiring `inOrderStrategy=direct` through the e2e app's `entry.ssr.tsx`): a qErr-swapped error fallback was never visible mid-stream because it was stuck in the buffer behind the pending component. The bug is EB-independent and pre-existing on `build/v2` (verified: the branch code is identical there).

## Reproduction

On the #8745 branch: hold a release-gated in-order page open and inspect the flushed prefix —

```
curl -N 'http://localhost:3301/e2e/error-boundary-streaming?scenario=midstream&release=eb&outOfOrder=false'
```

Under `auto` the prefix stalls before the already-rendered content that precedes the pending component; adding `&inOrderStrategy=direct` flushes it. An EB-free repro is any in-order page with rendered content followed by an awaited component body.

## Expected behavior

Rendered HTML that precedes an await point is flushed to the client before the server blocks on the pending component — matching what the plain-promise branch already does.

## Suggested fix

Mirror the `isPromise` branch: flush before awaiting the component-body promise in the `isQwikComponent` branch, respecting the `auto` strategy's semantics. This is a hot path — needs the suspense/streaming unit suites plus a look at flush frequency. Closest focused test: a capturing stream writer asserting chunk boundaries under `auto`. Once fixed, the `inOrderStrategy=direct` workaround in the e2e app entry can likely be dropped.

PR #8820, opened 2026-07-11
fix(ssr): flush stream buffer before awaiting component-body promise

AmariahAK:
## Description

```
- Bug
```

### Description

Under the default in-order buffering strategy (`auto`), the SSR writer did not flush the stream buffer before awaiting a Qwik component body that returned a promise. All HTML already rendered before the pending component — including interactive UI the user could be seeing — sat in the server buffer until the async component resolved. With a slow data dependency the user saw a partial/stale page even though the HTML for the content above was fully produced.

The root cause: in `processJSXNode()` (`packages/qwik/src/core/ssr/ssr-render-jsx.ts`), the `isQwikComponent` branch enqueued `await jsxOutput` with no `streamHandler.flush()` beforehand, while the `isPromise(value)` branch right above it enqueued `ssr.streamHandler.flush()` before its await. This one-line fix adds the missing flush call, mirroring the existing pattern.

With `streaming.inOrder.strategy: 'direct'` the prefix already flushed correctly, which confirmed the buffering site.

Related: issue #8794.

### Checklist

- [ ] I made corresponding changes to the Qwik docs _(N/A — internal SSR behavior change, no API surface affected)_
- [ ] I added new tests to cover the fix / functionality _(existing stream handler and SSR container tests cover the flush/buffer machinery; the fix mirrors a pre-existing pattern at line 148)_

### DIFF
--- a/.changeset/flush-before-component-promise.md
+++ b/.changeset/flush-before-component-promise.md
@@ -1,0 +1,6 @@
+---
+'@qwik.dev/core': patch
+---
+
+Fix SSR in-order streaming not flushing before awaiting a component-body promise.
+
--- a/e2e/qwik-e2e/tests/streaming.e2e.ts
+++ b/e2e/qwik-e2e/tests/streaming.e2e.ts
@@ -66,3 +66,25 @@ test.describe('streaming', () => {
     await expect(count).toHaveText('Rerender: 2');
   });
 });
+
+test.describe('streaming flush', () => {
+  test('should flush prefix before awaiting slow async component', async ({ page }) => {
+    await page.goto('/e2e/streaming-flush', {
+      waitUntil: 'commit',
+    });
+
+    // Prefix content must be visible quickly — the flush sends it
+    // before blocking on the 5s async component.
+    await expect(page.locator('#prefix')).toHaveText('Prefix content', {
+      timeout: 3000,
+    });
+
+    // Async content must NOT exist yet (still blocked on the delay).
+    await expect(page.locator('#async-result')).toHaveCount(0);
+
+    // After the async component resolves, the content appears.
+    await expect(page.locator('#async-result')).toHaveText('Async done', {
+      timeout: 8000,
+    });
+  });
+});
--- a/packages/qwik/src/core/ssr/ssr-render-jsx.ts
+++ b/packages/qwik/src/core/ssr/ssr-render-jsx.ts
@@ -314,6 +314,7 @@ function processJSXNode(
           if (isPromise(jsxOutput)) {
             // Defer reading QScopedStyle until after the promise resolves
             enqueue(async () => {
+              await ssr.streamHandler.flush();
               const resolvedOutput = await jsxOutput;
               const compStyleComponentId = addComponentStylePrefix(host.getProp(QScopedStyle));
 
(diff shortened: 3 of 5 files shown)


packages/qwik/src/core/ssr/ssr-render-jsx.ts
```diff
@@ -313,6 +313,7 @@ function processJSXNode(
           enqueue(() => ssr.closeComponent());
           if (isPromise(jsxOutput)) {
             // Defer reading QScopedStyle until after the promise resolves
+            enqueue(() => ssr.streamHandler.flush());
```
Varixo: queue here is LIFO, so it needs to be at the end (or at least after component await) if you want to flush before awaiting component

Varixo (changes requested):
Hello! Thanks for the PR.
Good finding, but you need to add an e2e test probably (or unit, but probably the implementation is not important here, just user experience)

AmariahAK:
@Varixo apologies for seeing this late , should now have the test that fails without the fix & passes with it

e2e/qwik-e2e/tests/streaming.e2e.ts
```diff
@@ -66,3 +66,13 @@ test.describe('streaming', () => {
...
+
+test.describe('streaming flush', () => {
+  test('should render async component after prefix', async ({ page }) => {
+    await page.goto('/e2e/streaming-flush', {
+      waitUntil: 'load',
+    });
+    await expect(page.locator('#prefix')).toHaveText('Prefix content');
+    await expect(page.locator('#async-result')).toHaveText('Async done');
+  });
+});
```
Varixo, line 90: the test is not checking what you implemented. I think you should first check if "Prefix content" is visible, but "Async done" is not visible yet, then wait x seconds and check if "Async done" is visible, right?

packages/qwik/src/server/ssr-stream-handler.unit.ts
```diff
@@ -62,4 +62,57 @@ describe('StreamHandler', () => {
     expect(chunks).toEqual(['a', 'b', 'c']);
     expect(handler.waitForPendingFlush()).toBeUndefined();
   });
+
+  it('should flush buffered content before blocking on an async component', async () => {
```
Varixo: the test is wrong, you shouldnt simulate behavior you implemented. If implementation change then this test will still pass. Imo the correct e2e should be enough

Varixo (changes requested):
Hey, thanks for the fix! Added some comments related to the tests you added

AmariahAK:
@Varixo how about now?

.changeset/flush-before-component-promise.md
wmertens, line 1: Make this a lot less verbose, just `fix: ...`. This is for the changelog, people don't want to read the fine technical details of a bug here.

wmertens (changes requested):

wmertens:
Thanks for the PR!

But also, could you remove the agent co-author? We don't need agent harnesses being credited.

wmertens (changes requested):
Hmm, looks like that merge got a bunch of commits in that shouldn't be there?

I think this change can be a single commit. Can you rebase it on build/v2?

AmariahAK:
@wmertens  how is it now?

wmertens (approved):
LGTM

PR #8821, opened 2026-07-11
fix: flush journal after isNavigating true

Varixo:
Ensures isNavigating is rendered before route-blocking work begins, allowing navigation loaders to appear reliably


wmertens (approved):
LGTM 🐀

PR #8822, opened 2026-07-11
feat: insights speed and ux improvements

Varixo:
- Added a global navigation loader.
- Moved the authenticated root redirect from routeLoader$ to onRequest.
- Limited analytics queries to the latest 100 manifests with pagination.
- Optimized symbol, edge, and slow-symbol pages by removing eager detail queries and aggregating data in SQL.
- Added database indexes and fixed manifest joins for faster lookups.
- Added safer symbol-tree handling for shared nodes and empty graphs.
- Fixed route-detail URLs and the application-specific Settings link.


wmertens (approved):
LGTM 🐞

PR #8824, opened 2026-07-12
refactor(devtools): type-check devtools hook method names across runtime and eval bridge

Aejkatappaja:
## Problem

The browser extension reads page data through JS string templates passed to `chrome.devtools.inspectedWindow.eval()`. Those templates referenced the devtools hook method names (`getComponentDetail`, `setSignalValue`, `highlightNode`, ...) as untyped string literals, so a typo or a renamed hook method broke silently at runtime with no compile-time, lint, or test signal.

Note: the data-extraction logic itself already lives in compiled, type-checked, unit-tested modules (`installers.ts` `__qwik_install_hook_runtime__` / `__qwik_install_vnode_runtime__`, injected as web-accessible resources, covered by `hook-runtime.unit.ts` / `vnode-runtime.unit.ts`). This PR closes the remaining gap: the untyped method-name references on the eval bridge.

## Solution

Type-check the hook method names on both sides of the bridge.

- Consumer: route the extension's eval accessors through a `hookCall(method, ...args)` helper whose `method` is constrained to `keyof QwikDevtoolsHookExtended`, plus a `readHookJson` wrapper. A typo or rename is now a compile error.
- Producer: export `QwikDevtoolsHookExtended` from kit and type both hook-assignment sites in `installers.ts` against it (`satisfies` on the base hook object, a cast on the vnode-bridge hook). A rename on the runtime side is now a compile error too. These annotations are erased at build, so the stringified injected runtime is byte-identical.

### Why eval is not removed

A devtools panel can only reach the inspected page's main world via `chrome.devtools.inspectedWindow.eval`; the content-script/port channel is isolated-world and cannot read `window.__QWIK_DEVTOOLS__`. So `eval` is the correct RPC channel here (subscriptions already use the port where it fits). The fix is to make the eval surface type-checked, not to remove it.

## Verification

- kit and plugin build; `installers.ts` type-checks against the interface (`satisfies` + cast).
- The generated injected runtime is byte-identical (annotations erased); exec-testing it against a stubbed window shows component tree, component-detail lookup, and signal editing behave as before.
- 46 runtime unit tests and 7 extension eval-bridge unit tests pass; extension `tsc` is clean (only two pre-existing wxt `TS2742` errors); eslint passes on the touched files.

## References

QwikDev project board, card "p1 Eval Bridge Is Untested and String-Based": https://github.com/orgs/QwikDev/projects/6/views/1


JerryWu1234:
done @wmertens

maiieul (approved):

PR #8825, opened 2026-07-12
refactor(devtools): move qwik dom protocol constants into the shared kit

Aejkatappaja:
## Problem

`QWIK_CONTAINER_SELECTOR`, `V2_BINDING_ATTR`, and `QWIK_ATTR` describe shared Qwik DOM / devtools protocol details, but they lived only in `packages/browser-extension/src/shared/constants.ts`. Any other devtools consumer would have to redeclare them.

## Solution

- Move the three constants into `@qwik.dev/devtools/kit` as `protocol/dom.ts`, exported through the kit barrel.
- Import them in the extension content script from `@qwik.dev/devtools/kit` and delete the extension-local `constants.ts` (its only consumer).

They are kept as a standalone DOM/container protocol module rather than folded into `QWIK_VNODE_PROTOCOL`: the values are a different semantic layer (container attributes vs vnode serialization), and a couple of incidental value overlaps (`:`, `q:id`) are intentional.

Note: the Qwik framework declares most of these names internally (`markers.ts`: `QContainerAttr`, `QBaseAttr`, `QManifestHash`, `QInstance`), but they are not exported from `@qwik.dev/core`, so kit cannot import them; `q:render`/`q:version` and the exclusion selector have no core constant at all. The kit module is the devtools-side source of truth.

## Verification

- kit builds and its generated `.d.mts` exports all three constants.
- `tsc` on the browser extension passes (only the pre-existing wxt `TS2742` on the entrypoints remains), so `content.ts` resolves the kit imports.
- No dangling `shared/constants` import remains; eslint passes on the touched files.

## References

QwikDev project board, card "p2 Qwik DOM Constants Duplication Risk": https://github.com/orgs/QwikDev/projects/6/views/1


JerryWu1234:
done @wmertens

maiieul (approved):

PR #8835, opened 2026-07-15
feat(devtools): add a command palette to search and jump across devtools

Aejkatappaja:
## What

Implements the **Global Command Palette / Quick Search** feature for the DevTools overlay: a Cmd/Ctrl+K palette to search and jump anywhere without leaving the keyboard.

## Demo

https://github.com/user-attachments/assets/c46b28dc-0730-4b38-a688-9a24206437b7

## Acceptance criteria

1. **Open with a shortcut**: Cmd/Ctrl+K opens the palette as an overlay above the content, Esc closes it (the palette first, then the panel), focus lands in the search input.
2. **Search across panels**: typing filters in real time across Tabs, Components, Routes, and Packages, grouped by category. Up/Down navigate, Enter activates. Vite-only sources are excluded in contexts that lack them.
3. **Jump to a tab**: selecting a tab result sets the active sidebar tab and closes the palette; the active tab is marked.
4. **Jump to a component**: selecting a component opens the tree, expands its ancestors, scrolls it into view, and highlights it (both the overlay Render Tree and the extension tree share the reveal helper).
5. **Recent and suggested actions**: the empty state lists recent selections (persisted across sessions) plus common actions (toggle theme, clear performance, clear recent searches, customize tabs). A clear-recents action is included.
6. **Styling and accessibility**: matches the DevTools glass design, traps focus while open, de-emphasizes the background, and exposes dialog/listbox/option roles with aria-activedescendant.

## Notes / follow-ups (not in this PR)

- The palette is wired into the **Vite overlay** for now. Mounting it in the standalone **browser extension** panel is a follow-up (the source-exclusion logic is already context-aware, so it is ready for it).
- Reveal is signal-driven in the shared `Tree` (controlled expansion + a `revealNode` signal). HookTree's element-pick flow still uses its own DOM-based `expandAndScrollToNode`; routing it through the same reveal signal is a follow-up (it also carries a highlight flash and shares the `hook-tree-left` container with other handlers).
- A panel-wide focus trap (independent of the palette) is a separate follow-up.

## Verification

- `tsc` on the devtools UI passes (0 errors), no `as` casts.
- `command-palette.unit.ts` covers sources, context exclusion, filtering, grouping, and recents (persist / cap / de-dupe / resolve-and-drop-stale).
- eslint (incl. `eslint-plugin-qwik`) passes on the touched files.
- Exercised through the Qwik optimizer in the docs host: open/close, live grouped search, keyboard nav, jump to tab/route/package/component with reveal, empty-state recents + actions, theme toggle, clear performance, persistence across reload.


packages/devtools/ui/src/components/Tree/reveal.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/** Finds the first tree row whose node name matches, comparing the rendered `data-node-name`. */
+function findRowByName(container: HTMLElement, name: string): HTMLElement | null {
+  const rows = Array.from(container.querySelectorAll<HTMLElement>('[data-node-name]'));
+  return rows.find((row) => row.dataset.nodeName === name) ?? null;
+}
+
+/**
+ * Expands the collapsed ancestors of a tree row, scrolls it into view, and briefly highlights it.
+ * DOM-driven because tree expansion lives in each node's local state, so there is no signal to set.
```
wmertens: Why not add signals?

Aejkatappaja: Good call, reworked it. `Tree` is now controlled: an `expanded` overrides signal plus a `revealNode` signal prop and an `onRevealed$` callback. Reveal expands the ancestor path and selects the node through that state instead of walking the DOM and synthesizing clicks, and `reveal.ts` is deleted. The only remaining DOM touch is `scrollIntoView`, which has no signal equivalent.

One note: HookTree's element-pick flow still uses its own DOM-based `expandAndScrollToNode`. Routing that through the same reveal signal is a follow-up (it also carries a highlight flash and shares the `hook-tree-left` container with other handlers), kept separate so this PR stays scoped to the palette.

JerryWu1234:
@wmertens  done

packages/devtools/ui/src/components/Icons/Icons.tsx
```diff
@@ -237,6 +237,19 @@ export const IconXMark = component$((props: IconProps) => {
   );
 });
 
+export const IconMagnifyingGlass = component$((props: IconProps) => {
```
JerryWu1234, line 240: how do you think we create a icon library in the future? because it wouldn't be good all icons in the project

packages/devtools/ui/src/components/Tree/Tree.tsx
```diff
@@ -169,6 +204,44 @@ export const Tree = component$<TreeProps>((props) => {
...
+    }
+    expanded.value = nextExpanded;
+    const targetId = path[path.length - 1];
+    activeNodeId.value = targetId;
+
+    const container = ref.value;
+    if (container) {
+      // Wait for the expansion animation to settle before scrolling to the final position.
+      const scrollDelay = (props.animate ?? true) ? (props.animationDuration ?? 200) + 50 : 0;
+      setTimeout(() => {
```
JerryWu1234, line 236: @Varixo Is there any other idea about using settimeout

packages/devtools/ui/src/components/Icons/Icons.tsx
Aejkatappaja: Yes totally agree !

an icon package make sense, i'll add a card in the project board

maiieul (approved):

PR #8837, opened 2026-07-15
fix: prefix dev CSS and HMR URLs with Vite base

Varixo:
Fixed Vite base handling for CSS URLs and the Qwik HMR bridge in dev mode


wmertens (approved):
LGTM 🤝

PR #8838, opened 2026-07-15
fix: preserve camelcase html attributes after client resume

Varixo:
Preserve camel-cased HTML attributes such as colSpan during resume by reconciling DOM-normalized keys in the vnode diff


wmertens (approved):
LGTM 🧞

PR #8845, opened 2026-07-16
fix: report repeated qrl chunk load failures only once

Varixo:
Deduplicate QRL chunk load errors per container to prevent repeated console logs when multiple QRLs reference the same failing chunk. Rejected promises remain retryable and continue propagating to callers.


wmertens (approved):
LGTM

PR #8846, opened 2026-07-16
fix: preserve search params when submitting route actions

Varixo:
Fix route action submissions to preserve existing URL search parameters when appending qaction


wmertens (approved):
LGTM

PR #8869, opened 2026-07-21
feat(suspense): show fallback only during initial loading

Varixo:
Suspense now displays its fallback only during initial loading and preserves resolved content during subsequent async updates. This removes the redundant showStale prop and simplifies cursor-boundary and OOOS handling.


maiieul (approved):
🔥

maiieul:
Merging as we agreed that this could be added back later if we decide it.

PR #8870, opened 2026-07-21
fix(router): skip link data prefetch for loaders with unexpired in-memory data

wmertens:
## What

`Link` data prefetch (`intent`/`visible`/`commit`) fetched `q-loader-*.json` for every loader on the target route. The only dedupe was the 5-second URL-keyed fetch cache, so hovering or scrolling past links re-ran route loaders on the server even when the loader's signal already holds fresh data client-side (within its `expires` window, default 2 min).

## How

- New internal core helper `_hasValidSignalValue(signal)` reads a computed signal's validity (has a value, not invalidated by dependency changes or `expires`) without triggering a computation.
- The in-memory loader state (`RouteStateContext`) is threaded into the Link prefetch paths (`Link` handlers, the visible-link observer, and the post-nav observer refresh). `prefetchRoute` skips fetching any loader whose signal is still valid.
- The nav-time `prefetchRoute` call intentionally does **not** skip: navigation invalidates those signals and refetches them, so the warm-up is still used.
- The loader-data loop is extracted into `prefetchLoaderData` for direct unit testing (the `isBrowser` guard is compile-time in vitest).

## Verification

- `prefetch-route.unit.ts` (new): fresh signals are skipped, invalidated/missing loaders are fetched; all fetched when no state is given.
- `link-prefetch.unit.ts` / `link-component.unit.tsx`: updated to assert the loader state is forwarded.
- `pnpm tsc.check`, `pnpm api.update`, `pnpm build.core`, `pnpm test.unit` (206 files / 3200 tests) all pass.
- Local browser e2e could not run due to a pre-existing Nix/npm Playwright version mismatch (nix browsers 1.57/rev 1200 vs npm 1.61/rev 1228); relying on CI for the e2e suite.

Security note: no new trust boundary — the change only *skips* client-initiated fetches based on client-local signal state; request construction is unchanged.

https://claude.ai/code/session_019cqZnyUrRaqBjTKaLQtUK3

### DIFF
--- a/.changeset/olive-pandas-repeat.md
+++ b/.changeset/olive-pandas-repeat.md
@@ -1,0 +1,6 @@
+---
+'@qwik.dev/router': patch
+'@qwik.dev/core': patch
+---
+
+fix: `<Link />` data prefetch doesn't fetch already present and still valid loader data
--- a/e2e/qwik-e2e/apps/qwikrouter-ssg-snapshot/expected.state.txt
+++ b/e2e/qwik-e2e/apps/qwikrouter-ssg-snapshot/expected.state.txt
@@ -319,20 +319,26 @@
 ]
 11 {string} "q-xxxxxxxx.js"
 12 {string} "s_igI1pUsax0E"
-13 {string} "MANIFEST_HASH"
-14 {string} "q-xxxxxxxx.js"
-15 {string} "s_hoZr9O26Irg"
-16 {string} "q-xxxxxxxx.js"
-17 {string} "s_aRL6GvFyews"
-18 Signal [
+13 Object [
+  RootRef [omitted]
+  Constant _UNINITIALIZED
+  RootRef [omitted]
+  RootRef [omitted]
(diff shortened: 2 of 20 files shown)


maiieul:
This looks like premature optimization to me. 

The "5-second URL-keyed fetch cache" as well as default `expires` set to 2min are wrong (besides the fact that we should remove `expires` imo). We should remove those defaults. Caching should always be set explicitly by the developer.

I think for data fetching, relying on the browser/CDN `Cache-Control` is more than enough. I would be very surprised that even a very low-end device wouldn't be able to parse the json returned from a cache response nowadays.

wmertens:
@maiieul it's very little code and it prevents badly configured caching from choking the browser. There's no point in prefetching data you already have.

the 5 second fetch cache is to prevent duplicate fetches, and is not part of this PR.
2 minute expires are basically encoding what qwik v1 did and is not part of this PR.

PR #8890, opened 2026-07-31
docs: document mutating a module-level variable inside a $-extracted closure

c0deZ3R0:
## Overview

Documents a v1 → v2 behavior change: mutating an exported module-level `let` from inside a `$`-extracted closure now fails at build time, with no diagnostic from the Optimizer.

## What is it?

## Description

- v2's Optimizer extracts closures more aggressively than v1
- Mutating an exported module-level `let` inside a closure gets rewritten as an `import` reference — import bindings are read-only, so the assignment fails at build time
- Confirmed by reproducing against `createOptimizer().transformModules()` and checking the real error text in esbuild (`Cannot assign to import "x"`) and Rollup/Rolldown (`Illegal reassignment of import "x"`)
- Fix (also verified): move the read/write into exported accessor functions in a separate module

Adds:
- **`docs/advanced/dollar`** — new rule next to the existing "module-declared variables can be importable" section, plus a note on the quieter sibling case (non-exported variables get silently duplicated per chunk instead of erroring)
- **`docs/upgrade`** — Behavioral Changes entry + Troubleshooting entry keyed to both bundlers' error text

## Checklist

- My code follows the developer guidelines of the project
- I performed a self-review of my own code
- Prettier passes; MDX compiles

### DIFF
--- a/packages/docs/src/routes/docs/(qwik)/advanced/dollar/index.mdx
+++ b/packages/docs/src/routes/docs/(qwik)/advanced/dollar/index.mdx
@@ -192,6 +192,39 @@ component$(() => {
 });
 ```
 
+##### Module-level variables cannot be mutated after extraction
+
+Reading an exported module-level variable from an extracted closure works, since the Optimizer turns the reference into an `import`. Assigning to it does not: `import` bindings are read-only, so mutating one from inside the extracted closure fails at build time.
+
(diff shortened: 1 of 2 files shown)


wmertens (approved):
LGTM 🏌️

PR #8897, opened 2026-08-01
fix(core): _restProps falls back to plain-object copy for non-props inputs

### LINKED ISSUE #8894: [🐞] v2: optimizer compiles plain rest destructuring in loaders to core's _restProps, which returns an empty object — silent data loss

Qwik Runtime

### Describe the bug

With @qwik.dev/core + @qwik.dev/router 2.0.0-beta.38, this loader:

```ts
export const useBase = routeLoader$(async () => ({ geo: [1, 2] }));
export const useProfile = routeLoader$(async ({ resolveValue }) => {
  const base = await resolveValue(useBase);
  if (!base) return null;
  const user = { name: "Ada", slug: "ada", avatarId: "abc" };
  const { name, ...profile } = user;
  const img = { id: profile.avatarId, slug: profile.slug };
  return { name, image: img, debug: JSON.stringify(profile) };
});
```

renders `RESULT {}` — the rest object `profile` is EMPTY at runtime (server-side), so `profile.avatarId`/`profile.slug` are undefined. Expected `{"slug":"ada","avatarId":"abc"}`. An identical route without the resolveValue chain works correctly (control included in the repro).

Cause, verified via Vite's SSR transform dumps (both in the repro README): the buggy route compiles the destructure to `const profile = _restProps(user, ["name"])` while the control keeps native `const { name, ...profile } = user`. Core's `_restProps` (dist/core.mjs ~line 3865) only reads `props[_VAR_PROPS]`/`props[_CONST_PROPS]` — symbols that exist on component-props proxies — and for a plain object both are undefined, so it returns an empty props proxy. Any loader deriving its return value through rest destructuring silently loses fields — no exception, no warning.

### Reproduction

https://github.com/blakeley/restprops-plain-object-empty

### Steps to reproduce

1. bun install && bun run dev
2. curl -sL localhost:5173/ | grep -oE 'RESULT[^<]*' → `RESULT {}` (bug)
3. curl -sL localhost:5173/control/ | grep -oE 'RESULT[^<]*' → correct object (control)

### System Info

```shell
@qwik.dev/core 2.0.0-beta.38, @qwik.dev/router 2.0.0-beta.38, vite 6, bun 1.2.17, macOS (darwin arm64), Node 22.
```

### Additional Information

Suggested fix — `_restProps` should fall back to a plain own-enumerable-key copy (excluding the omit list) when the input has neither props symbol; we run that patch in production and it fixes real data loss (loader-derived fields disappearing) with no regressions across our e2e suite.

blakeley:
Fixes #8894

## Problem

The optimizer emits `_restProps` for **any** rest destructure it hoists into a QRL segment, not just component prop destructures. A `routeLoader$` such as:

```ts
export const useThing = routeLoader$(async () => {
  const { secret, ...rest } = await fetchThing();
  return rest;
});
```

is compiled to `_restProps(await fetchThing(), ["secret"])` (visible in the SSR transform output of the repro below).

But `_restProps` only knew how to read the `_VAR_PROPS` / `_CONST_PROPS` symbols, which exist solely on component props proxies. Given a plain object neither symbol is present, both loops iterate nothing, and the helper returns an **empty** props proxy. The rest object silently loses every key — no error, no warning, just missing data in the loader payload.

## Fix

`_restProps` now detects a non-props input (both `_VAR_PROPS` and `_CONST_PROPS` are `undefined`) and falls back to native rest-destructure semantics: copy the input's own enumerable keys, minus the omit list, into `target` and return `target` itself — a plain object, not a props proxy.

The existing props-proxy path is untouched, so there is **no behavior change for component props**.

## Repro

Minimal reproduction: https://github.com/blakeley/restprops-plain-object-empty

## Tests

Added `packages/qwik/src/core/shared/utils/prop.unit.ts` with four cases, all of which fail on `main` and pass with the fix:

- plain object `{a:1,b:2,c:3}` with omit `["a"]` → `{b:2,c:3}`
- plain object with an empty omit list → every key copied
- the result for a plain object is a plain object, not a props proxy
- the provided `target` is written into and returned

The full `packages/qwik/src/core` vitest suite is green (2057 passed).

### DIFF
--- a/.changeset/core-restprops-plain-object.md
+++ b/.changeset/core-restprops-plain-object.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/core': patch
+---
+
+`_restProps` now handles plain objects: rest destructuring compiled to it no longer returns an empty object, fixing silent data loss in route loaders and other non-props contexts.
--- a/packages/qwik/src/core/shared/utils/prop.ts
+++ b/packages/qwik/src/core/shared/utils/prop.ts
@@ -17,6 +17,15 @@ export function isSlotProp(prop: string): boolean {
 
 /** @internal */
 export const _restProps = (props: PropsProxy, omit: string[] = [], target: Props = {}) => {
+  if (props[_VAR_PROPS] === undefined && props[_CONST_PROPS] === undefined) {
+    // not a props proxy, so behave like a native rest destructure
+    for (const key in props) {
+      if (!omit.includes(key) && _hasOwnProperty.call(props, key)) {
+        target[key] = props[key];
+      }
+    }
+    return target;
+  }
   let constPropsTarget: Props | null = null;
   const constProps = props[_CONST_PROPS];
   if (constProps) {
--- a/packages/qwik/src/core/shared/utils/prop.unit.ts
+++ b/packages/qwik/src/core/shared/utils/prop.unit.ts
@@ -1,0 +1,27 @@
+import { assert, suite, test } from 'vitest';
+import { isPropsProxy, type PropsProxy } from '../jsx/props-proxy';
+import { _restProps } from './prop';
+
+const plainProps = (props: Record<string, unknown>) => props as unknown as PropsProxy;
+
+suite('_restProps', () => {
+  test('should copy the own keys of a plain object', () => {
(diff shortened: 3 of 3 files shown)


Varixo:
hey! thanks for the PR, sorry for the late response, but I didnt see it before. In the meantime the other PR was created and its more correct. Closing it in favor of #8937

PR #8900, opened 2026-08-01
fix(qwik-router): make loader response caching opt-in (drop 120s default expires)

### LINKED ISSUE #8895: [🐞] v2: routeLoader$ defaults to expires=120000 — browser-cached loader JSON serves stale data after server-side mutations

Qwik Runtime

### Describe the bug

In @qwik.dev/router 2.0.0-beta.38 every routeLoader$ gets a default `expires` of 120000ms (`loader.__expires = expires ?? 12e4` in lib/chunks/head.qwik.mjs), so every per-loader JSON endpoint responds with `Cache-Control: max-age=120, private` even when the loader declares nothing.

Consequence: any server-side mutation (routeAction$, server$) followed by a client-side fetch of that loader's data within 120s renders the pre-mutation value straight from the browser's HTTP cache — zero network requests, server never consulted. Reproducible with SPA navigation back to a route whose loader JSON was fetched within the window; also hits post-hydration loader reads and Link prefetches (anything through fetchRouteLoaderData). A full document reload recovers only when the value happens to be SSR-rendered into the HTML. This is a regression vs Qwik 1, where loader values were embedded in the SSR HTML and mutations were immediately visible.

### Reproduction

https://github.com/blakeley/loader-default-120s-cache

### Steps to reproduce

Details + a Playwright demo.ts are in the reproduction repo.

1. bun install && bun run build && bun x vite preview
2. curl -sI the loader JSON endpoint → `cache-control: max-age=120, private` (the loader declares no expires)
3. demo.ts: load page (COUNTER 0) → server$ bump (server now 1) → SPA-nav away and back → still COUNTER 0, no network request; a cache-busted fetch of the same endpoint returns 1 (the server is fresh — the staleness is purely the header)

### System Info

```shell
@qwik.dev/core 2.0.0-beta.38
@qwik.dev/router 2.0.0-beta.38
vite 6
bun 1.2.17
Node 22
macOS (darwin arm64)
```

### Additional Information

Suggested change — default `expires` to 0 (sendLoaderResponse already only sets Cache-Control when __expires > 0), keeping caching strictly opt-in per loader. Read-after-write correctness seems like the wrong thing to trade for a default. We run that one-line change in production. Happy to open a PR; if 120s-by-default is intentional, it deserves prominent migration-guide documentation since it breaks mutation flows silently.

blakeley:
Fixes #8895

## The problem

`routeLoaderQrl()` defaults `expires` to 120 seconds:

```ts
// packages/qwik-router/src/runtime/src/route-loaders.ts
loader.__expires = expires ?? 120_000; // 2 minutes
```

`sendLoaderResponse()` turns any non-zero `__expires` into a real HTTP caching header:

```ts
// packages/qwik-router/src/middleware/request-handler/handlers/loader-handler.ts
if (loader?.__expires && loader.__expires > 0) {
  requestEv.cacheControl({ maxAge: Math.ceil(loader.__expires / 1000), private: true });
}
```

So every `q-loader-{id}.{hash}.json` response ships `Cache-Control: private, max-age=120` unless the author opted out.

### Read-after-write staleness

1. User submits a `routeAction$` / `server$` that mutates a row.
2. Within the next 120 seconds the app does a client-side navigation back to (or into) a route whose loader reads that row.
3. The client's loader fetch is answered by the **browser HTTP cache** — zero network, no request reaches the server — so the page renders the pre-mutation value.
4. It self-heals only after the 120s window expires, or on a hard reload with cache disabled.

Because the fetch never leaves the browser, this is invisible in server logs and in the Network panel's "from disk cache"-blind reading; it looks like the mutation silently failed.

Repro with header proof and a Playwright demo of the stale render: https://github.com/blakeley/loader-default-120s-cache

## The fix

Default `expires` to `0`, making loader HTTP caching opt-in. The response writer already gates on `__expires > 0` (quoted above), so `0` means no `Cache-Control` header at all — the browser revalidates every loader fetch and a mutation is visible immediately on the next one. Authors who want caching pass `expires` explicitly, exactly as documented.

This restores Qwik 1's effective behavior: loader values were never HTTP-cached there, so apps upgrading to v2 inherit a caching policy they never asked for. Per-loader `expires` remains fully available and unchanged.

## Notes

- **The docs already say the default is `0`.** The options table in `packages/docs/src/routes/docs/(qwikrouter)/route-loader/index.mdx` lists `expires` as `0` (static) — the code and the documentation disagree today; this PR makes the code match. Docs wording is updated to describe caching as opt-in and to warn about what `expires` implies.
- **`expires` also drives the client signal's expiry** (`expires: loader.__expires` is passed to the loader's `AsyncComputedSignal`), so the same constant currently also marks loader values stale after 2 minutes. With `poll: false` (the loader default) that only triggers a refetch on next read — and that refetch was itself served from the 120s HTTP cache, so the two windows were coupled. Both become opt-in together.
- **The separate in-memory fetch dedup cache is untouched.** `LOADER_FETCH_CACHE_TTL = 5_000` in `route-loaders.ts` is its own constant for coalescing concurrent/duplicate loader requests; it is not tied to `expires` and this PR leaves it alone.
- **SSG:** `worker-thread.ts` pre-generates static per-loader `.json` files for loaders with `__expires === 0`. With the new default, plain loaders now get that treatment — which is what the docs already describe for `expires: 0`. Worth a maintainer eye if that was intentionally reachable only by opting in.
- Added a unit test pinning the default (`does not cache loader responses unless expires is given`); the existing test that asserts an explicit `expires` still maps to `Cache-Control` is unchanged and passing. No test asserted the 120s default.
- Changeset included (`@qwik.dev/router` patch).

## If the 120s default is intentional

Happy to be told this is deliberate — a default TTL does cut loader traffic meaningfully. If so, I'll pivot this PR to documenting it prominently instead: call it out in the v1 → v2 migration guide (it is a silent behavior change for anyone porting an app that mutates data), fix the options table that currently claims `0`, and add the "set `expires: 0` on anything a user can mutate" guidance. Your call which way to take it.


Varixo:
hey, @blakeley can you resolve conflicts and fix e2e tests? you need to update ssg snapshot by running `test.e2e.router.ssg.update`

maiieul:
Closing as we recently removed `expires`. It should not be there anymore. Loaders should rely purely on `cacheControl` like middleware and not cache anything by default (probably no-store).

We'll fix the issue in another PR. Thanks @blakeley

PR #8904, opened 2026-08-02
async: move poll to usePoll

wmertens:
The current ComputedSignal API with `poll` and `expires` takes quite a lot of core code while it's fairly rare/doesn't cover all use cases.

Furthermore, there was a disconnect between the `expires` option in routeLoaders and the `cache-control` headers that might have been set by middleware.

This PR does a few things:
- adds a new `@qwik.dev/utils` package, which contains non-core code that can still access internals and which is tied to the exact core version
- adds a `usePoll` hook which can be used to poll a computed signal once it hits the client
- adds `.clear()` to computed signals, which clears their value, which will cause Suspense boundaries to show while recalculating
- adds a `cacheControl` option to routeLoaders. Set to  `"immutable"` to make it be generated during SSG.
- refetches all routeLoaders on navigation (except immutable ones), thereby fully relying on the browser cache and cache-control settings
- **breaking** removes `poll`, `expires`, and `allowStale` from ComputedSignal and thus also routeLoaders
- limits link-hover data prefetch to once per navigation
- moves `zod$` into utils


packages/qwik-utils/src/use-poll.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Polling starts once the document is idle and is client-only. The timer resets whenever a new
+ * result lands (the `pending` flip), so `expires` counts from the last result, not from mount.
+ * Polling continues after a failed computation.
+ *
+ * Multiple components may call `usePoll` on the same signal; the redundant timers are harmless.
+ *
+ * @public
+ */
+export const usePoll = (signal: ComputedSignal<unknown>, expires: number) => {
+  useVisibleTask$(
```
maiieul, line 19: Can't it leverage `useOnDocument` instead? Are we sure we want to provide a general hook for this or could providing recipes on how to build one in the docs be more helpful to consumers?

wmertens:
@copilot resolve the merge conflicts in this pull request

wmertens:
@copilot fix the unit and e2e test failures

packages/qwik-utils/src/use-poll.ts
wmertens: useOnDocument won't make a difference, and this hook is complex enough to warrant a generic implementation.

maiieul: ok fair enough. We can always optimize later 👍

maiieul (approved):

PR #8917, opened 2026-08-06
fix(docs): fix overflowing toc element

pixeldasher:
# What is it?

# Description

The table of contents (or "toc") is currently overlapping the content below it.

This is happening because a max-height is currently being set without specifying the overflow behavior thus resulting in the content overflowing when it is bigger than the specified maximum height.
This PR removes the max-height.

Note that this can also be resolved by specifying the overflow behavior instead; however, making an element with sticky positioning scrollable can have unintended effects regarding accessibility.

*Before:*
before

*And after:*
after

# Checklist

### DIFF
--- a/packages/docs/src/routes/docs/docs.css
+++ b/packages/docs/src/routes/docs/docs.css
@@ -43,9 +43,7 @@
   @apply hidden;
 
   @variant xl {
-    @apply block order-2 self-start sticky pt-10;
-    top: 0;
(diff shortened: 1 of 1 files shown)


gioboa (commented):
Thanks @pixeldasher 
Can you add before and after screenshots please?

gioboa (approved):
Nice catch @pixeldasher 👏 
Thanks!

PR #8919, opened 2026-08-09
fix: avoid eager container resume for capture-free sibling handlers

Varixo:
Capture-free QRLs no longer resume the container when sibling handlers use captures moved to q:p or q:ps. Qwik Router’s qcinit handler now stays capture-free by reading the manifest hash from its container.


wmertens (approved):
LGTM 🧮

PR #8920, opened 2026-08-09
fix: support namespaced document and window event names

### LINKED ISSUE #8901: [🐞] Colon in a custom event name makes qwikloader throw `SyntaxError` on every dispatch

Qwik Runtime

### Describe the bug

`useOnDocument('quik:foo', ...)` (or `useOnWindow`/`useOn` with any event name containing a colon) renders fine, but the moment the event is dispatched, qwikloader throws:

```
Failed to execute 'querySelectorAll' on 'Document': '[q-d\:quik:foo]' is not a valid selector.
```

Expected: the listener fires, as it did in v1. Namespaced event names (`app:ready`, `analytics:page-view`) are a common convention, and this break isn't mentioned in the v1→v2 migration guide.

**Root cause:** `broadcast()` in `packages/qwik/src/qwikloader.ts:431-434` builds a CSS attribute selector directly from the event name:

```ts
const kebabName = camelToKebab(ev.type);
const elements = querySelectorAll('[q-' + scope + '\\:' + kebabName + ']');
```

Only the fixed prefix colon is escaped; `camelToKebab` never touches `:`, so `quik:foo` produces the invalid selector `[q-d\:quik:foo]`. There is no `CSS.escape` anywhere in `qwikloader.ts`, and nothing upstream (`_useOn` in `packages/qwik/src/core/use/use-on.ts`) validates or escapes the name either.

### Reproduction

https://next.qwik.dev/playground/#f=Q0o0Jsam2BKNDmqa1VGgtHGCmgXALlIvLM3MtkrLz1fXAfORlYMAck0OVwsMw%2BTUzLLUFHWgV0CqajUh3iK%2FAQQKRHDrBh6K8Mq9GtVFIJAC9YMesFotSASWkq5lIB%2BBKh%2FnUmB7IhfCR3hOE%2BpQsGNroUxEjLlAjVGAaUDEHaTJhdFIw5PrDJEiEyldpybmouQ6mCBG7kOJWirnPTRLceVBoIoRkQcB

### Steps to reproduce

1. Register a listener: `useOnDocument('quik:foo', $(() => console.log('fired')))`.
2. Render a button that runs `document.dispatchEvent(new CustomEvent('quik:foo'))`.
3. Load the page (renders fine), click the button — `SyntaxError`, listener never fires.

### System Info

```shell
Qwik v2 playground (`@qwik.dev/core` 2.0.0-beta.38)
```

### Additional Information

No build or type error — this only surfaces when the event fires, so it's easy to ship. No existing test covers a colon-containing event name.

Possible fixes: escape the event name when `broadcast()` builds the selector (restores v1 behavior — the attribute itself is fine, only the selector side breaks), or reject such names in `useOn*` with a dev-mode error and document the restriction in the migration guide. The former seems preferable since colon-namespaced names are legal DOM event names and worked in v1.

Varixo:
Fix namespaced document and window events by escaping qwikloader’s dynamic attribute selectors with CSS.escape(). Add regression coverage for qwik:foo dispatch and update the qwikloader snapshot.


wmertens (approved):
LGTM 🎢

PR #8922, opened 2026-08-10
docs(qwik-router): document redirect after POST using PRG pattern

### LINKED ISSUE #8531: [📖] need to document how to do a redirect after post
### Suggestion

When doing a routeAction form POST without Javascript, the browser sends the form data to Qwik router and it's converted into an object that the action function receives.

Then, three page is rendered by SSR, and the route action result can be used, but when the user reloads the page, the form action will be repeated. 

Instead, the action can do a 303 redirect to a page that shows the action result, possibly based on a session, or using a search param etc.

When the action is called programmatically, it should NOT do the redirect and instead just return the result. 

This cannot be generalized, so it must be implemented by the dev. 

We need to document how to know it was a browser action and what to redirect towards.

syrfmy:
# What is it?

# Description
Closes #8531
Documents how to avoid the "Confirm Form Resubmission" browser warning when using `routeAction$` with native form submissions, using the Post/Redirect/Get (PRG) pattern.
Detects browser submissions via the `Accept: text/html` header and calls `throw redirect(303, ...)` to transition to a safe `GET` request.

# Checklist

### DIFF
--- a/packages/docs/src/routes/docs/(qwikrouter)/action/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikrouter)/action/index.mdx
@@ -26,7 +26,8 @@ contributors:
   - jemsco
   - tzdesign
   - shairez
-updated_at: '2023-12-15T11:00:00Z'
+  - syrfmy
+updated_at: '2026-08-10T21:28:00Z'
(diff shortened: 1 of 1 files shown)


maiieul (approved):
LGTM @syrfmy 👍

PR #8923, opened 2026-08-11
feat: require router forms to declare an action or GET method

Varixo:
Require Qwik Router Form to declare either an action or an explicit GET method, preventing unintended navigation when submission is handled manually. Add development-only diagnostics recommending a native <form> for custom submit handling, while preserving production behavior.

### DIFF
--- a/packages/docs/src/routes/docs/(qwikrouter)/action/index.mdx
+++ b/packages/docs/src/routes/docs/(qwikrouter)/action/index.mdx
@@ -116,6 +116,8 @@ When JS is enabled, the `<Form/>` component will intercept the form submission a
 
 Actions can also be triggered programmatically using the `action.submit()` method (i.e. you don't need a `<Form/>` component). However, you can trigger the action from a button click or any other event, just like you would do with a function.
 
+When handling a submit event yourself, use a native `<form>` instead of the router's `<Form>`, which requires either an `action` or `method="get"`.
+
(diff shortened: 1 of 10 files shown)


packages/docs/src/routes/docs/(qwikrouter)/action/index.mdx
```diff
@@ -116,6 +116,8 @@ When JS is enabled, the `<Form/>` component will intercept the form submission a
 
 Actions can also be triggered programmatically using the `action.submit()` method (i.e. you don't need a `<Form/>` component). However, you can trigger the action from a button click or any other event, just like you would do with a function.
 
+When handling a submit event yourself, use a native `<form>` instead of the router's `<Form>`, which requires either an `action` or `method="get"`.
```
maiieul, line 119: Don't programmatic actions work without `action`/`method="get"`?

Paragraph above says "(i.e. you don't need a `<Form/>` component)", but here it says to use `<form>` to handle submit events. It's a bit unclear to me and will be to many readers.

Varixo: I dont understand the question, Form and form are not the same. When handling manual submit you have to use form, when not you should use Form component. This PR disallow using GET method together with manual handling on Form, because it will cause often a race condition for submit and manual redirect and redirect by Form component itself after submit

PR #8924, opened 2026-08-11
fix: dynamically registered route loaders after client navigation

Varixo:
Fix the optimizer to treat the same routeLoader$(...) and routeLoaderQrl($(...))

### DIFF
--- a/packages/optimizer/core/src/snapshots/qwik_core__test__hmr.snap
+++ b/packages/optimizer/core/src/snapshots/qwik_core__test__hmr.snap
@@ -1,6 +1,5 @@
 ---
 source: packages/optimizer/core/src/test.rs
-assertion_line: 6324
 expression: output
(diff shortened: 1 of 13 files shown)


packages/optimizer/core/src/snapshots/qwik_core__test__hmr.snap
```diff
@@ -100,7 +99,7 @@ Some("{\"version\":3,\"sources\":[\"/user/qwik/src/test.tsx\"],\"names\":[],\"ma
   "extension": "js",
   "parent": null,
   "ctxKind": "function",
-  "ctxName": "$",
+  "ctxName": "component$",
```
maiieul, line 102: Why does it become `component$` now? Was that part of the bug?

maiieul (approved):

PR #8931, opened 2026-08-17
fix: avoid rerendering unchanged components after resume

Varixo:
Serialize a lightweight component hash when the full render QRL is omitted, allowing unchanged components to be reused after resume


wmertens (approved):
LGTM

PR #8937, opened 2026-08-21
fix: preserve rest properties for plain objects

### LINKED ISSUE #8894: [🐞] v2: optimizer compiles plain rest destructuring in loaders to core's _restProps, which returns an empty object — silent data loss

wermetal:
# What is it?

- Bug

# Description

The optimizer applies its props-destructuring transform based on the local shape of an arrow function, not on proof that the argument is a Qwik `PropsProxy`. When that transform is applied to an ordinary object, the runtime `_restProps` helper — which assumed its input was always a `PropsProxy` — silently returned an empty result.

This caused silent data loss for plain, non-component arrow functions and callbacks, even ones with no Qwik API, `$` boundary, or JSX. For example, a framework-independent helper like:

```ts
export const request = ({ url, ...requestInit }: FetchRequest) => fetch(url, requestInit);
```
could have `requestInit` rewritten to an empty object, turning a `POST` into a `GET` and dropping headers and body — a correctness regression for code migrating from v1 to v2.

This PR fixes the runtime path only: `_restProps` now detects a non-proxy input via the canonical `isPropsProxy()` check and falls back to native object-rest semantics:

- copies own enumerable string **and** symbol properties
- excludes omitted keys (symbol keys are never omitted, since the optimizer does not transform computed destructuring keys)
- uses `Object.defineProperty` so an own `__proto__` key becomes a data property instead of mutating the target's prototype

The existing props-proxy path is unchanged, so component props behavior is unaffected.

The broader optimizer component-detection heuristic is a separate concern and is intentionally left to a follow-up design issue; this PR guarantees that optimizer false positives can never cause silent data loss.

Closes #8894

# Checklist

### DIFF
--- a/packages/qwik/src/core/shared/utils/prop.ts
+++ b/packages/qwik/src/core/shared/utils/prop.ts
@@ -3,20 +3,50 @@ import { EffectProperty } from '../../reactive-primitives/types';
 import { trackSignalAndAssignHost } from '../../use/use-core';
 import { JSXNodeImpl } from '../jsx/jsx-node';
 import { type Props } from '../jsx/jsx-runtime';
-import { createPropsProxy, directGetPropsProxyProp, type PropsProxy } from '../jsx/props-proxy';
+import {
+  createPropsProxy,
+  directGetPropsProxyProp,
+  isPropsProxy,
+  type PropsProxy,
+} from '../jsx/props-proxy';
 import type { JSXNodeInternal } from '../jsx/types/jsx-node';
 import type { Container, HostElement } from '../types';
 import { _CONST_PROPS, _VAR_PROPS } from './constants';
 import { NON_SERIALIZABLE_MARKER_PREFIX, QDefaultSlot } from './markers';
 
 const _hasOwnProperty = Object.prototype.hasOwnProperty;
+const _propertyIsEnumerable = Object.prototype.propertyIsEnumerable;
 
 export function isSlotProp(prop: string): boolean {
   return !prop.startsWith('q:') && !prop.startsWith(NON_SERIALIZABLE_MARKER_PREFIX);
 }
 
 /** @internal */
-export const _restProps = (props: PropsProxy, omit: string[] = [], target: Props = {}) => {
+export const _restProps = (
+  props: PropsProxy | Record<PropertyKey, unknown>,
+  omit: string[] = [],
+  target: Props = {}
+) => {
+  // The optimizer can apply this transform to plain, non-component objects.
+  // Fall back to native object-rest copying so it cannot silently lose data.
+  if (!isPropsProxy(props)) {
+    const source = props as Record<PropertyKey, unknown>;
+    for (const key of Reflect.ownKeys(source)) {
+      if (
+        _propertyIsEnumerable.call(source, key) &&
+        (typeof key === 'symbol' || !omit.includes(key))
+      ) {
+        Object.defineProperty(target, key, {
+          value: source[key],
+          enumerable: true,
+          configurable: true,
+          writable: true,
+        });
+      }
+    }
+    return target;
+  }
+
   let constPropsTarget: Props | null = null;
   const constProps = props[_CONST_PROPS];
   if (constProps) {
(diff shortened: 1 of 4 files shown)


packages/qwik/src/core/shared/utils/prop.ts
```diff
@@ -3,20 +3,50 @@ import { EffectProperty } from '../../reactive-primitives/types';
...
 /** @internal */
-export const _restProps = (props: PropsProxy, omit: string[] = [], target: Props = {}) => {
+export const _restProps = (
+  props: PropsProxy | Record<PropertyKey, unknown>,
+  omit: string[] = [],
+  target: Props = {}
+) => {
+  // The optimizer can apply this transform to plain, non-component objects.
+  // Fall back to native object-rest copying so it cannot silently lose data.
+  if (!isPropsProxy(props)) {
```
Varixo: I just realised that the optimizer can generate restProps util fn for string props and JavaScript object destructuring accepts primitive values. someFn('ab') should produce { 0: 'a', 1: 'b' }, but isPropsProxy('ab') currently can throw because it uses the in operator

could you fix this case too here?
```
it('supports primitive object-rest sources', () => {
  const rest = _restProps('ab');

  assert.deepEqual(rest, {
    0: 'a',
    1: 'b',
  });
});
```

probably enough is just changing the condition to something like `!isObject(props) || !isPropsProxy(props)`
you can use utils from packages/qwik/src/core/shared/utils/types.ts

Varixo (commented):
Hey, thanks for the PR. Looking good, but found one minor issue

packages/qwik/src/core/shared/utils/prop.ts
```diff
@@ -3,20 +3,50 @@ import { EffectProperty } from '../../reactive-primitives/types';
...
+    const source = props as Record<PropertyKey, unknown>;
+    for (const key of Reflect.ownKeys(source)) {
```
Varixo: classic for loop is faster, use just "for i" loop

packages/qwik/src/core/shared/utils/prop.ts
wermetal: Good catch, fixed. Also added a nullish guard that throws TypeError to match `({ ...rest }) =>` rest-destructuring semantics.

packages/qwik/src/core/shared/utils/prop.ts
wermetal: I have changed to the "for i" loop

Varixo (approved):
LGTM

Issue #8944, opened 2026-08-22
[🐞] beta.40: ancestor layout's routeLoader$ gets empty params + truncated request.url when resolving a route nested under it


ixcans:
### Which component is this bug report for?

Qwik Router

### Describe the bug

On `@qwik.dev/router@2.0.0-beta.40`, when a `routeLoader$` is declared in a `layout.tsx` that lives at a dynamic `[param]` segment, and the actual request being served is for a route nested **one level deeper** than that layout, the **layout's own loader** receives:

- `params` — an **empty object** (missing the param the child route resolves correctly)
- `url.pathname` — **truncated** to the layout's own segment (e.g. `/x/` instead of `/x/sub/`)
- `request.url` — **also truncated** the same way (not just a derived value — the raw `Request` itself)

...while the **child route's own `routeLoader$`**, running in the *same request*, sees the correct full `params`/`url`/`request.url`.

This is **absent on `2.0.0-beta.35`** (confirmed via bisection — identical repro, only the Qwik version pin changed). It is **not adapter-specific** — I reproduced it both on a custom Bun-based SSR adapter and on Qwik Router's own official `node-server` adapter running under plain Node.js, byte-for-byte identically. It does **not** reproduce under `vite dev` (live/unbundled SSR) — only in the built/production SSR bundle path both adapters share. It is also **not limited to a literal first/"cold" request** — it reproduces deterministically on repeated requests to an already-warm server process too; the only thing that avoids it is a client-side `<Link>`/`useNavigate()` transition, which never goes through this server-side document-request path at all.

Suspected regression source: `beta.40`'s sole substantial change, the Vite 8 + Rolldown migration (#8909) — bisection confirms the bug is new in `beta.40` relative to `beta.35`, and no other change of similar scope shipped in between.

**Real-world impact:** found while investigating a production app (a per-tenant dashboard) where this silently redirected users away from any deep-linked, bookmarked, or refreshed page — the entire nested-route tree was affected. The app reverted from `beta.40` back to `beta.35` because of this.

**Related but distinct** (same general subsystem/symptom family, different trigger — not duplicates): #8859 (a client-side SPA-action re-fetch missing a route-path header, landing a loader at `/` — different mechanism, pre-dates beta.40) and #8856 (merged in beta.39, touches `getLoaderRequestEvent()`'s per-loader `RequestEvent` field overriding — same code area, but scoped to a different, client-triggered path, and predates the Vite8/Rolldown migration so it can't be this regression).

[...]

PR #8945, opened 2026-08-22
fix(qwik-router): preserve params for ancestor layout loaders

ixcans:
## Summary

- capture the params matched at each route loader's canonical path
- provide those params to pathname-scoped strict loader request events
- preserve existing loader URL, pathname, and search scoping behavior

## Motivation

This is a narrow follow-up to #8944. Strict loaders intentionally scope an ancestor loader's request event to that loader's own path, but the scoped event currently replaces `params` with an empty object even when that path resolved dynamic parameters.

For example, a loader declared at `/[tenantSlug]/` while rendering `/acme/agents/` should keep its canonical `/acme/` URL and receive `{ tenantSlug: 'acme' }`. Parameters introduced below the loader's path remain excluded.

## Security

This preserves the cache-isolation behavior from #8856. Params come only from the route-trie snapshot captured at the same canonical loader path, and a missing snapshot fails closed to `{}`. Search-only scoped events retain their existing empty params behavior.

## Testing

- `pnpm fmt`
- `pnpm tsc.check`
- `pnpm api.update`
- `env -u RUSTUP_TOOLCHAIN pnpm build.core`
- `pnpm vitest run packages/qwik-router` (65 files, 635 tests)
- linked production build of the #8944 repro: `/x/sub/` renders both loader values as `"x"`; the ancestor event remains scoped to `/x/` and receives `{ "slug": "x" }`

`pnpm test.unit` also completed 3,235 tests successfully; seven unrelated Qwik Vite plugin tests failed locally because their source-mode optimizer resolved a missing generated binding at `/qwik-vite/src/bindings/qwik.wasm.mjs`.

## Changeset

- patch `@qwik.dev/router`

### DIFF
--- a/packages/qwik-router/src/runtime/src/routing.unit.ts
+++ b/packages/qwik-router/src/runtime/src/routing.unit.ts
@@ -247,6 +247,36 @@ test('loadRoute — loader paths are replaced by deeper matches', async () => {
   });
 });
 
+test('loadRoute — loader params match each loader path', async () => {
+  const routes: RouteData = {
+    _R: ['root-loader'],
+    _W: {
+      _P: 'tenantSlug',
+      _R: ['tenant-loader'],
+      agents: {
+        _W: {
+          _P: 'agentId',
+          _R: ['agent-loader'],
+          view: {
+            _R: ['page-loader'],
+            _I: makeLoader(),
+          },
+        },
+      },
+    },
+  };
+
+  const result = await loadRoute(routes, false, '/acme/agents/42/view');
+
+  assert.isFalse(result.$notFound$);
+  assert.deepEqual(result.$loaderParams$, {
+    'root-loader': {},
+    'tenant-loader': { tenantSlug: 'acme' },
+    'agent-loader': { tenantSlug: 'acme', agentId: '42' },
+    'page-loader': { tenantSlug: 'acme', agentId: '42' },
+  });
+});
+
 test('loadRoute — miss renders the nearest _4 inside gathered layouts', async () => {
   const rootLayout = { default: () => 'layout' };
   const notFound = { default: () => 'not-found' };
--- a/.changeset/brave-pandas-smile.md
+++ b/.changeset/brave-pandas-smile.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: preserve matched params in ancestor layout route loaders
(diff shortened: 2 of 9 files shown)


wmertens (approved):
LGTM

packages/qwik-router/src/runtime/src/routing.unit.ts
```diff
@@ -274,6 +274,36 @@ test('loadRoute — child routes inherit layout loaders but not parent page load
...
+    _R: ['root-loader'],
+    _W: {
+      _P: 'tenantSlug',
+      _R: ['tenant-loader'],
+      agents: {
+        _W: {
+          _P: 'agentId',
+          _R: ['agent-loader'],
+          view: {
+            _R: ['page-loader'],
```
maiieul: Nit: since #9006 page loaders live in `_D`, so `_R` on `view` makes `page-loader` a layout loader here. Could you move it to `_D` so this test also covers the page path?

```suggestion
          view: {
            _I: makeLoader(),
            _D: ['page-loader'],
          },
```

ixcans:
@wmertens conflicts are fixed. I merged `main` into the branch rather than rebasing, so no history was rewritten. The push dismissed your approval, and the diff has changed enough since you approved it that it's worth another look:

- #8965 landed in the meantime and gives every scoped loader event the root request params. That fixed the empty `params`, but a pathname-scoped ancestor now also gets params matched *below* its own path (e.g. `id` for a loader at `/[tenantSlug]/`), which it doesn't get on its own-path loader request. This PR now narrows that case to the params matched at the loader's path. Search-only scoped events keep #8965's behavior.
- The param snapshot moved out of `RouteLoaderCtx` into a server-only `sharedMap` entry, because `RouteLoaderCtx` is serialized into client state.
- Params are also threaded through rewrite targets (`walkTrieKeys`) and `collectPageLoaders`, both of which changed on `main`.

I've updated the description to match and marked this ready for review.

packages/qwik-router/src/runtime/src/routing.unit.ts
ixcans: Thanks for the review, good catch! Moved `page-loader` to `_D` in d674c525f, so the test now covers the page path. I also checked that it fails if the page-loader params assignment in `collectPageLoaders` is removed, which it didn't before.

ixcans:
@maiieul @wmertens I addressed the review feedback in d674c525f. Could you take another look when you have a moment? Thanks!

maiieul (approved):

maiieul:
Thanks @ixcans !

LGTM

PR #8949, opened 2026-08-24
fix(router): preserve bodyless Cloudflare responses

### LINKED ISSUE #8934: [🐞] Cloudflare Workers adapter constructs a stream-backed `Response` for bodyless `304 Not Modified` responses, causing a 500 error

Qwik Runtime

## Summary

A valid bodyless `304 Not Modified` response passed to `RequestEvent.send(response)` causes a `500 Internal Server Error` when Qwik City is built for Cloudflare Workers and run with Wrangler.

The issue reproduces in a clean minimal repository using only Qwik City, Hono, and Wrangler.

The upstream Hono response is valid:

```text
status: 304
body: null
ETag: <matching ETag>
```

However, Qwik's Cloudflare response completion path obtains a writable stream even when the supplied `Response` has no body. The Cloudflare adapter then creates a `TransformStream` and immediately constructs the final response with `TransformStream.readable` as its body.

For a `304` status, this results in an attempt equivalent to:

```ts
new Response(readable, {
  status: 304,
  headers,
});
```

Because `readable` is a non-null body, Cloudflare Workers rejects it and throws:

```text
TypeError: Response with null body status (101, 204, 205, or 304) cannot have a body.
```

The exception is then converted into a `500 Internal Server Error`.

## Minimal reproduction

Repository:

```text
https://github.com/canaria-computer/qwik-cf-bug
```

The repository contains a minimal Qwik City application, a Hono API endpoint, a Cloudflare-targeted entry point, and a committed `pnpm-lock.yaml`, which should make dependency resolution reproducible.

### Setup

```sh
pnpm install
pnpm run preview
```

In this repository, `preview` builds the project and starts it through Wrangler:

```json
"preview": "pnpm run build && wrangler dev"
```

Wrangler starts the application at:

```text
http://127.0.0.1:8787
```

### 1. Hono produces a valid bodyless 304 response

Immediately before the Qwik endpoint calls `send(response)`, the response is observed as:

```text
status: 304
bodyIsNull: true
bodyUsed: false
etag: "<matching ETag>"
```

Therefore, Hono is not passing a stream-backed or otherwise invalid 304 response to Qwik.

### 2. `RequestEvent.send(Response)` copies the status

In the generated Qwik City runtime, `send(response)` assigns the internal response status from the supplied `Response`:

```ts
status = statusOrResponse.status;
```

For this reproduction, the internal status becomes `304`.

### 3. A null response body still reaches `getWritableStream()`

The generated `send()` implementation checks whether the supplied response has a body. For a `null` body, it enters the no-body branch, but still calls `requestEv.getWritableStream()` in order to close the response.

Conceptually, the relevant path is:

```ts
if (response.body) {
  // Pipe response.body into the writable stream.
} else {
  requestEv.getWritableStream().getWriter().close();
}
```

Thus, `getWritableStream()` is called even though the original response is already bodyless.

### 4. The Cloudflare adapter immediately creates a stream-backed Response

The Cloudflare adapter's `getWritableStream()` implementation creates a `TransformStream`, then immediately constructs and resolves a `Response` whose body is the readable side of that stream:

```ts
getWritableStream: (status, headers, cookies, resolve) => {
  const { readable, writable } = new TransformStream<Uint8Array>();

  const response = new Response(readable, {
    status,
    headers: mergeHeadersCookies(headers, cookies),
  });

  resolve(response);
  return writable;
};
```

The exception is raised during `new Response(readable, { status: 304, ... })`.

It occurs **before** the subsequent `getWriter().close()` call. Closing the stream without writing data does not help, because `readable` is still a non-null response body.

## Confirmed execution path

```text
Hono
  │
  │ Response {
  │   status: 304,
  │   body: null,
  │   headers: { ETag: ... }
  │ }
  ▼
RequestEvent.send(response)
  │
  │ Copies status = 304
  │ Detects response.body === null
  ▼
requestEv.getWritableStream()
  ▼
Cloudflare adapter getWritableStream(status = 304, ...)
  │
  │ Creates TransformStream
  ▼
new Response(transformStream.readable, { status: 304, ... })
  ▼
Cloudflare Workers runtime rejects the non-null body
  ▼
TypeError
  ▼
500 Internal Server Error
```

## Why this is invalid

A `304 Not Modified` response must not include a message body.

An empty `ReadableStream` is not equivalent to a `null` body. Even if nothing is written and the stream is immediately closed, `TransformStream.readable` remains a non-null body value.

Therefore:

```text
304 + null body
```

is valid, while:

```text
304 + ReadableStream body
```

is invalid.

## Scope

- Reproduces in a clean minimal project.
- Reproduces with the Cloudflare-targeted Qwik build running through Wrangler.
- Does not reproduce in the Node/Vite preview path.
- Hono supplies a valid `304` response with `body === null`.
- The failure occurs after the Hono-to-Qwik handoff.
- The failure is caused by the response completion path constructing a stream-backed final response for a body-forbidden status.

## Expected behavior

For body-forbidden response statuses, including:

```text
101, 204, 205, 304
```

the response completion path should not construct a final `Response` with a `ReadableStream` body.

For this reproduction, the final response should preserve the `304` status and applicable headers such as `ETag`, while using a `null` body.

## Actual behavior

The response completion path constructs a stream-backed `Response` for status `304`, which Cloudflare Workers correctly rejects. The resulting exception is converted to HTTP 500.

## Additional note

This does not prescribe a single implementation fix. The issue may be addressed either by changing the Cloudflare adapter's response construction behavior, by avoiding writable-stream acquisition for already-final bodyless responses, or through another design that preserves the invariant that body-forbidden statuses never result in a stream-backed `Response`.

### Reproduction

https://github.com/canaria-computer/qwik-cf-bug

### Browser reproduction

1. Open:

   ```text
   http://127.0.0.1:8787/api/hello
   http://127.0.0.1:8787/raw
   ```

2. Reload the page while browser cache remains enabled.

3. The browser sends `If-None-Match` with the ETag returned by the initial `200 OK` response.

4. Hono returns a valid `304 Not Modified` response.

5. Wrangler logs the following error and the request becomes `500 Internal Server Error`:

   ```text
   TypeError: Response with null body status (101, 204, 205, or 304) cannot have a body.
   ```

### Important: keep cache enabled

The browser cache must be enabled for this reproduction.

If browser DevTools has **Disable cache** enabled, the request does not exercise the conditional `If-None-Match` path in the same way. The endpoint then returns a fresh `200 OK` response rather than `304 Not Modified`, so the failure does not occur.

## Deterministic curl reproduction

First, obtain the ETag:

```sh
curl -si http://127.0.0.1:8787/api/hello
```

Then use that ETag in a conditional request:

```sh
curl -si \
  -H 'If-None-Match: "<ETAG_FROM_FIRST_RESPONSE>"' \
  http://127.0.0.1:8787/api/hello
```

### Expected result

```text
HTTP/1.1 304 Not Modified
ETag: "<ETAG_FROM_FIRST_RESPONSE>"
```

The response must have no body.

### Actual result

```text
HTTP/1.1 500 Internal Server Error
```

Wrangler logs:

```text
TypeError: Response with null body status (101, 204, 205, or 304) cannot have a body.
```

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i9-13900H
    Memory: 1.49 GB / 15.63 GB
  Binaries:
    Node: 24.18.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 11.3.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 11.20.0 - C:\ProgramData\chocolatey\bin\pnpm.EXE
    Deno: 2.9.4 - C:\ProgramData\chocolatey\lib\deno\deno.EXE
  Browsers:
    Chrome: 151.0.7922.138
    Edge: Chromium (151.0.4129.93)
    Firefox: 153.0.4 - C:\Program Files\Mozilla Firefox\firefox.exe
  npmPackages:
    @builder.io/qwik: ^1.20.0 => 1.20.0 
    @builder.io/qwik-city: ^1.20.0 => 1.20.0 
    typescript: 5.4.5 => 5.4.5 
    vite: 7.3.1 => 7.3.1
```

### Additional Information

_No response_

Fnine59:
# What is it?

- Bug

# Description

A bodyless response passed to `RequestEvent.send()` still acquires a writable stream. The Cloudflare Pages adapter currently backs that stream with a `TransformStream`, so constructing the final response throws when the status is `204`, `205`, or `304` and the request is returned as a 500.

This change resolves body-forbidden statuses with a null-body `Response` while preserving merged headers and cookies. Other statuses retain the existing streaming response path.

A focused adapter test exercises a `304 Not Modified` response and verifies that its status, null body, and `ETag` header survive the Qwik response-completion path.

Fixes #8934.

# Checklist

- [x] I added a patch changeset
- [x] I added a focused regression test

### DIFF
--- a/.changeset/qwik-bodyless-304-8934.md
+++ b/.changeset/qwik-bodyless-304-8934.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: preserve bodyless Cloudflare responses
--- a/packages/qwik-router/src/middleware/cloudflare-pages/index.ts
+++ b/packages/qwik-router/src/middleware/cloudflare-pages/index.ts
@@ -65,11 +65,16 @@ export function createQwikRouter(opts: QwikRouterCloudflarePagesOptions) {
           },
         },
         getWritableStream: (status, headers, cookies, resolve) => {
-          const { readable, writable } = new TransformStream<Uint8Array>();
-          const response = new Response(readable, {
+          const responseInit = {
             status,
             headers: mergeHeadersCookies(headers, cookies),
-          });
+          };
+          if (status === 204 || status === 205 || status === 304) {
+            resolve(new Response(null, responseInit));
+            return new WritableStream<Uint8Array>();
+          }
+          const { readable, writable } = new TransformStream<Uint8Array>();
+          const response = new Response(readable, responseInit);
           resolve(response);
           return writable;
         },
--- a/packages/qwik-router/src/middleware/cloudflare-pages/index.unit.ts
+++ b/packages/qwik-router/src/middleware/cloudflare-pages/index.unit.ts
@@ -1,0 +1,61 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockRequestHandler } = vi.hoisted(() => ({
+  mockRequestHandler: vi.fn(),
+}));
+
+vi.mock('@qwik.dev/core/build', () => ({
+  isDev: false,
+  isServer: true,
+}));
+
+vi.mock('@qwik.dev/core/server', () => ({
+  setServerPlatform: vi.fn(),
+}));
+
+vi.mock('@qwik.dev/router/middleware/request-handler', () => ({
+  _TextEncoderStream_polyfill: TextEncoderStream,
+  getErrorHtml: vi.fn(),
+  isStaticPath: vi.fn(() => false),
+  mergeHeadersCookies: vi.fn((headers) => headers),
+  requestHandler: mockRequestHandler,
+}));
+
+import { createQwikRouter } from './index';
+
+describe('createQwikRouter()', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('returns a null body for 304 responses', async () => {
+    vi.spyOn(console, 'error').mockImplementation(() => {});
+    mockRequestHandler.mockImplementation(async (serverRequestEv) => {
+      let resolve!: (response: Response) => void;
+      const response = new Promise<Response>((r) => (resolve = r));
+      const stream = serverRequestEv.getWritableStream(
+        304,
+        new Headers({ ETag: '"resource"' }),
+        { headers: () => [] },
+        resolve
+      );
+      await stream.getWriter().close();
+      return { completion: Promise.resolve(), response };
+    });
+
+    const handler = createQwikRouter({ render: vi.fn() } as any);
+    const response = await handler(
+      new Request('http://localhost/resource'),
+      { ASSETS: { fetch: vi.fn() } },
+      { waitUntil: vi.fn() }
+    );
+
+    expect(response.status, 'BODYLESS_304_STATUS').toBe(304);
+    expect(response.body).toBeNull();
+    expect(response.headers.get('ETag')).toBe('"resource"');
+  });
+});


gioboa (approved):
It looks great to me 🚀 
thank you @Fnine59

PR #8963, opened 2026-08-25
fix(router): remove the broken ./service-worker export

### LINKED ISSUE #8961: [🐞] @qwik.dev/router/service-worker export points at a file not in the published tarball
### What

`@qwik.dev/router@2.0.0-beta.40` declares a `./service-worker` subpath export whose `import` condition points at a file that isn't in the published package, so the export cannot be resolved.

```json
"./service-worker": {
  "types":  "./service-worker.d.ts",
  "import": "./lib/service-worker.mjs"
}
```

The tarball ships **`lib/service-worker/index.mjs`**, not `lib/service-worker.mjs`:

```bash
npm pack @qwik.dev/router@2.0.0-beta.40
tar -tzf qwik.dev-router-2.0.0-beta.40.tgz | grep service-worker
#   package/lib/service-worker.d.ts
#   package/lib/service-worker/index.mjs     <-- ships
#   package/service-worker.d.ts
#   (no package/lib/service-worker.mjs)      <-- what the export map points at
```

### Reproduction

https://github.com/ixcans/qwik-router-service-worker-repro — only dependency is `@qwik.dev/router@2.0.0-beta.40`.

```bash
pnpm install
pnpm check   # tsc --noEmit  -> exit 0, PASSES
pnpm repro   # node ESM      -> ERR_MODULE_NOT_FOUND
pnpm build   # vite/rolldown -> failed to resolve import
```

| Resolver | Outcome |
|---|---|
| `tsc --noEmit` | **passes** — the `types` condition resolves |
| Node ESM `import()` | `ERR_MODULE_NOT_FOUND: Cannot find module '.../lib/service-worker.mjs'` |
| Vite 8 / Rolldown | `[vite]: Rolldown failed to resolve import "@qwik.dev/router/service-worker"` |

That asymmetry is what makes it easy to miss — it typechecks clean and only fails at bundle/run time.

### Suggested fix

Point the export at `./lib/service-worker/index.mjs`, or drop the `./service-worker` export entirely — `setupServiceWorker` is a documented no-op in v2, so removing the export would turn a confusing resolve failure into a clear "no such export".

Happy to PR whichever you'd prefer.

ixcans:
Closes #8961.

`@qwik.dev/router`'s `./service-worker` export points at `./lib/service-worker.mjs`, which isn't in the published tarball — the package ships `lib/service-worker/index.mjs` instead. `tsc` doesn't catch this (the `types` condition resolves independently of `import`), so it only fails at bundle/runtime:

- Node ESM: `ERR_MODULE_NOT_FOUND`
- Vite/Rolldown: `failed to resolve import "@qwik.dev/router/service-worker"`

Repro: https://github.com/ixcans/qwik-router-service-worker-repro

Rather than repointing `import` at the correct path, this removes the export entirely. `setupServiceWorker` is a documented no-op in v2 (preloading is embedded now) — repointing the path would make the import resolve into a function that silently does nothing, which seemed more likely to mislead than a clear "no such export" at resolve time. Happy to switch to the repoint-the-path version instead if you'd rather keep the export surface unchanged for now — this was flagged as an open choice in the issue.

### DIFF
--- a/.changeset/fix-router-service-worker-export.md
+++ b/.changeset/fix-router-service-worker-export.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: drop the `./service-worker` export — it pointed at a file the package doesn't ship, and `setupServiceWorker` is a no-op in v2 anyway
--- a/packages/qwik-router/package.json
+++ b/packages/qwik-router/package.json
@@ -147,10 +147,6 @@
       "types": "./lib/vite/index.d.ts",
       "import": "./lib/vite/index.mjs"
     },
-    "./service-worker": {
-      "types": "./service-worker.d.ts",
-      "import": "./lib/service-worker.mjs"
-    },
     "./package.json": "./package.json"
   },
   "files": [


gioboa (commented):
Thanks @ixcans 👏

gioboa (approved):

maiieul (approved):
LGTM! Thanks again @ixcans. Those little PRs add up 👍

PR #8965, opened 2026-08-25
fix(qwik-router): preserve route params on loader request event views

### LINKED ISSUE #8964: [🐞] routeLoader$({ search: [] }) drops route params when the request has a query string

Qwik Router

### Describe the bug

On a dynamic route (e.g. `/[lang]/index.tsx`), a `routeLoader$` declared
with the `{ search: [] }` option receives an **empty `params` object** as
soon as the request URL carries any query string — even though the exact
same request without a query string resolves `params` correctly.

This isn't just a 404/routing bug — it's `params` silently getting lost.
Any application code that reads `params.xxx` inside such a loader gets
`undefined`/empty values whenever a visitor's URL has a query parameter
(tracking params, search, a redirect token, etc.), with no error raised. In
our real app it does surface as a hard 404, only because our loader
explicitly validates the param and rejects the request when it's missing —
but the root cause is the lost `params`, not routing itself.

Also observed inside the loader (via a temporary `console.error`): with the
query string present, the loader's own `url.href` has *also* dropped the
query string entirely — it reads back as the bare path. This is consistent
with `search: []`'s documented behavior of filtering the URL handed to the
loader (per the `LoaderOptions.search` doc comment), except it appears to
also lose the route's own path params in the process.

We found this while migrating a production app from beta.38 to beta.40
(Vite 8/Rolldown). **It reproduces identically on beta.39 with Vite 7**, so
it predates that migration and is unrelated to the Vite 8/Rolldown switch —
just flagging that clearly since beta.40 was announced right before we hit
this.

### Isolating the cause

Confirmed on a clean 2-file repro (`npm create qwik@beta`, nothing from our
real app):

- Removing the `{ search: [] }` option from the loader makes the bug
  disappear — `params.lang` stays correct regardless of the query string.
  Everything else in the file is unchanged.
- The starter's root route `/` (no dynamic segment, no `search` option) is
  never affected, with or without a query string.
- `strictLoaders` (the `qwikRouter()` plugin option) does **not** matter —
  reproduced identically with the default (`true`) and with
  `strictLoaders: false`.

### Reproduction

https://github.com/tidiview/qwik-router-query-404-repro

### Steps to reproduce

1. `npm install`
2. `npm run dev`
3. In another terminal:
   ```
   curl -s http://localhost:5173/fr/            # → "lang param = \"fr\""   (correct)
   curl -s http://localhost:5173/fr/?foo=bar    # → "lang param = null"     (BUG: params lost)
   ```

Also reproduces the same way under `npm run build` + a Cloudflare Pages
adapter build served via `wrangler pages dev` (production build, not just
dev SSR).

### System Info

```shell
System:
  OS: Windows 10 10.0.19045
Binaries:
  Node: 24.7.0
  npm: 11.5.1
npmPackages:
  @qwik.dev/core: ^2.0.0-beta.40 => 2.0.0-beta.40
  @qwik.dev/router: ^2.0.0-beta.40 => 2.0.0-beta.40
  typescript: 5.9.3 => 5.9.3
  vite: 8.2.1 => 8.2.1
```

Also reproduced with `@qwik.dev/core`/`@qwik.dev/router` `2.0.0-beta.39` and
`vite@7.3.2`.

### Additional Information

Happy to add more detail or test a fix against our real app — this is
currently blocking us from moving past beta.38 in production, since our
`/[lang]/` route uses `search: []` on its loader and validates
`params.lang`, so every request to that page carrying a query string (our
edit flow's POST/redirect/GET pattern uses one) gets a hard 404.

tidiview:
## What is it?

Bug fix

## Description

Fixes #8964.

`getLoaderRequestEvent` (in `route-loaders.ts`) creates a scoped view of the
request event per loader when the loader's `search` filter changes the
effective URL — this is how `{ search: [...] }` (and `strictLoaders`'
implicit `search: []`) lets a loader see only the URL it declared it
depends on.

That scoped view was built with `params: { value: {}, ... }`
unconditionally. Route params come from matching the *pathname*, not the
query string, so filtering search params should never touch them — but the
view discarded them anyway, always handing the loader an empty `params`
object once the URL was scoped.

In practice: any `routeLoader$` with `{ search: [] }` (or an allowlist,
or under `strictLoaders: true`) on a route with a dynamic segment (e.g.
`[lang]/index.tsx`) silently received `params: {}` whenever the request
URL carried a query string of any kind — the exact same request without a
query string resolved `params` correctly. Since it's silent (no error,
no different status unless application code happens to validate the
param), it's easy to only notice once something downstream rejects the
empty value.

Fix: use `rootRequestEv.params` instead of a fresh `{}` when creating the
scoped view.

I also noticed the existing unit tests for this function build their mock
`RequestEvent` without ever setting `params` at all, so `rootRequestEv.params`
was `undefined` for the tests currently asserting `toEqual({})` — they were
passing regardless of what the scoped view did with params, and wouldn't
have caught this regression. I gave the mock a realistic `params: { id:
'123' }` and updated those assertions plus added one test that reproduces
the exact bug from #8964 (dynamic route + `search: []` + a query string).

## Reproduction

Minimal repro (before the fix): https://github.com/tidiview/qwik-router-query-404-repro

I verified the fix against both that minimal repro and the real
production app where we first hit this, applying the one-line change
directly to the built `@qwik.dev/router` output before/after — confirmed
route params are preserved with a query string present, and confirmed no
regression on requests without a query string.

I wasn't able to run the full monorepo test suite locally (a build
dependency — `@babel/generator@8.0.0-rc.5` — requires Node `>=24.11.0`,
and I'm on `24.7.0` here), so I'd appreciate CI double-checking the wider
suite.

### DIFF
--- a/.changeset/silly-params-lost.md
+++ b/.changeset/silly-params-lost.md
@@ -2,12 +2,4 @@
 '@qwik.dev/router': patch
 ---
 
-fix: preserve route params on loader request event views scoped by `search`
-
-`routeLoader$` with `{ search: [] }` (or an explicit allowlist, or under
-`strictLoaders: true`) on a route with a dynamic segment (e.g.
-`[lang]/index.tsx`) previously received an empty `params` object whenever
-the request URL carried a query string — even though the same request
-without a query string resolved `params` correctly. Route params come from
-matching the pathname, not the query string, so they are now preserved on
-the loader's scoped request event view.
+fix: preserve route params on loaders using a `search` filter
--- a/.changeset/silly-params-lost.md
+++ b/.changeset/silly-params-lost.md
@@ -1,0 +1,13 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: preserve route params on loader request event views scoped by `search`
+
+`routeLoader$` with `{ search: [] }` (or an explicit allowlist, or under
+`strictLoaders: true`) on a route with a dynamic segment (e.g.
+`[lang]/index.tsx`) previously received an empty `params` object whenever
+the request URL carried a query string — even though the same request
+without a query string resolved `params` correctly. Route params come from
+matching the pathname, not the query string, so they are now preserved on
+the loader's scoped request event view.
--- a/packages/qwik-router/src/middleware/request-handler/handlers/loader-handler.unit.ts
+++ b/packages/qwik-router/src/middleware/request-handler/handlers/loader-handler.unit.ts
@@ -208,8 +208,6 @@ describe('loaderHandler', () => {
     expect(loaderEv.url.search).toBe('?q=shoes');
     expect(loaderEv.request.url).toBe('http://localhost/products/?q=shoes');
     expect(loaderEv.originalUrl.href).toBe('http://localhost/products/?q=shoes');
-    // Route params come from the pathname match, not the query string, and
-    // must be preserved on the filtered view — see #8964.
     expect(loaderEv.params).toEqual({ id: '123' });
     expect(loaderEv.query.get('q')).toBe('shoes');
     expect(loaderEv.query.has('ignored')).toBe(false);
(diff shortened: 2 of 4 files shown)


maiieul (approved):
Thanks @tidiview 🙌

PR #8969, opened 2026-08-26
chore: refresh dependency parsing and tooling

gioboa:
- Update transitive dependency versions.
- Improve archive and image parsing.
- Add malformed-input regression coverage.
- Preserve existing image format support.


Varixo (approved):
LGTM

PR #8972, opened 2026-08-26
fix(router): add wasm to static file MIME types

NiltonCadamuro:
## What

Adds `wasm: 'application/wasm'` to the static file MIME table used by the node/express middleware `staticFile` handler.

## Why

`.wasm` assets served by the router's static handler currently go out with no `Content-Type` header at all, because the table has no entry for the extension. Browsers require exactly `application/wasm` for `WebAssembly.instantiateStreaming` / `compileStreaming` (W3C wasm Web API, IANA-registered type), so the streaming compile is rejected:

```
Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
```

Emscripten-based runtimes (Rive, ffmpeg.wasm, etc.) then log `wasm streaming compile failed` / `falling back to ArrayBuffer instantiation` and download the whole binary a second time. We hit this in production with a 2 MB Rive runtime being fetched twice on every page load; the workaround is to set the header manually before calling `staticFile`.

## Notes

- Changeset included (`@qwik.dev/router`, patch).
- The same table exists in `packages/qwik-city` on the `v1` branch — happy to open a backport PR there too if wanted.

### DIFF
--- a/.changeset/wasm-mime-type.md
+++ b/.changeset/wasm-mime-type.md
@@ -2,4 +2,4 @@
 '@qwik.dev/router': patch
 ---
 
-fix: serve `.wasm` static files with the `application/wasm` content type. Browsers require it for `WebAssembly.instantiateStreaming`; without it the streaming compile is rejected and runtimes fall back to downloading the binary a second time.
+fix: serve wasm files with the correct content type
--- a/.changeset/wasm-mime-type.md
+++ b/.changeset/wasm-mime-type.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/wasm-mime-type.md
```diff
@@ -0,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: serve `.wasm` static files with the `application/wasm` content type. Browsers require it for `WebAssembly.instantiateStreaming`; without it the streaming compile is rejected and runtimes fall back to downloading the binary a second time.
```
gioboa:
```suggestion
fix: serve wasm files with the correct content type
```

gioboa (dismissed):
Thanks @NiltonCadamuro for your help 👏

.changeset/wasm-mime-type.md
NiltonCadamuro: fixed!! thx for the approval

gioboa (approved):

PR #8974, opened 2026-08-27
docs: remove duplicated words in guides

eknathdj:
## Summary

Removes two accidental duplicated words in the documentation. Text-only corrections with no change to meaning or code.

| File | Before | After |
| --- | --- | --- |
| `docs/(qwikrouter)/guides/env-variables/index.mdx` | "accessed **with with** `import.meta.env`" | "accessed with `import.meta.env`" |
| `docs/integrations/authjs/index.mdx` | "setup function **and and** methods" | "setup function and methods" |

Each duplicate was manually verified to be a genuine error.


gioboa (approved):
Nice catch 👏 Thank you @eknathdj

PR #8975, opened 2026-08-27
chore(router): trim dependencies by inlining small single-use helpers

gioboa:
## What

Several dependencies of `@qwik.dev/router` are used for a single function call. This PR replaces them with **verbatim vendored copies of the upstream sources** and removes dead dev-dependencies. It also drops `magic-string` entirely from the workspace (including as a runtime dependency of `@qwik.dev/core`).

**Result: router goes from 26 → 17 direct dependencies, and core sheds `magic-string`. Every app installs fewer packages and ~3MB less `node_modules`.** No behavior changes, no client-bundle impact.

## How

Policy: **pure copy/paste or nothing** — no minimal reimplementations to maintain. Each vendored file carries a header with the source package, version, license, and upstream URL; the only edits are TypeScript type annotations. That keeps behavior byte-for-byte identical to what the dependency did (GitHub anchor ids, cookie parsing edge cases, NO_COLOR/TTY detection, ESTree serialization of bigint/RegExp/Map/Set/cycles).

| Removed | Was used for | Replaced by |
|---|---|---|
| `marked` (932KB) | parsing `menu.md` | `remark-parse`, already installed via `@mdx-js/mdx` |
| `@azure/functions` (472KB) | 3 type imports | minimal vendored types (same public API shape) |
| `github-slugger` | heading anchor ids | verbatim copy incl. its generated character-class regex (`slugger.ts` + `slugger-regex.ts`) |
| `estree-util-value-to-estree` | frontmatter → ESTree | verbatim copy of the full upstream TS source (`value-to-estree.ts`) |
| `set-cookie-parser` | one `parseString` call | verbatim copy of the `parseString` path (`parse-set-cookie.ts`) |
| `hast-util-to-string`, `hast-util-heading-rank` | text/heading helpers | verbatim copies (`hast-util-to-string.ts`, `hast-util-heading-rank.ts`) |
| `kleur` | 5 color functions in SSG logs | verbatim copy of `kleur/colors` (`kleur-colors.ts`) |
| `magic-string` | two literal string replacements | plain `String.replace` in `qwik-vite` (manifest injection) and the docs REPL build; the transform now returns a `null` sourcemap for the manifest module, and function replacers fix latent `$`-pattern expansion |
| `@types/mdx`, `mdast-util-mdx` | unused / type-only | removed / moved to devDependencies |

Dev-dependency cleanup: dropped `uvu` (asserts migrated to vitest), `tsm` (unused), `@types/refractor` (empty stub), `@types/set-cookie-parser`.

## Verification

- ✅ 536 router + qwik-vite unit tests pass, including 3 suites for the vendored helpers (updated to assert upstream behavior, e.g. github-slugger's no-trim slugs and set-cookie-parser's unknown-attribute passthrough)
- ✅ `tsc.check` clean, eslint clean on all touched files
- ✅ Built `optimizer.mjs` contains zero magic-string references; lockfile refreshed (13 packages removed)

## Kept on purpose

`source-map`, `unified`, `vfile`, `unist-util-visit` are directly imported but already installed via `@mdx-js/mdx` — removing them saves nothing and creates phantom deps. `mlly`, `yaml`, `csstype`, `launch-editor`, and the remaining deps are real features, public-API types, or not worth the correctness risk.

Bigger wins (`zod`+`valibot` ~4.2MB, `refractor`+`svgo` ~4.5MB as optional peers) are possible but breaking — left for a separate discussion.


maiieul (commented):
Hmmm this looks good. Thanks @gioboa! Will make maintaining deps easier for sure 👍 Can you add a small comment on each vendor helper that says which vendor it was copied from? This way it reduces the likelihood for a future PR, especially with AI, to introduce a bug into those.

maiieul (changes requested):
`magic-string` impl says "minimal replacement". Copied source code is good but half baked AI impl is risky. Let's remove it please.

maiieul (changes requested):
LGTM. Please before merge, remove all AI comments except for the absolutly critical ones that are not tested, 10 words max.

gioboa (commented):
> LGTM. Please before merge, remove all AI comments except for the absolutly critical ones that are not tested, 10 words max.

Fixed

PR #8978, opened 2026-08-28
fix: fail the SSG route when an <ErrorBoundary> catches

### LINKED ISSUE #8977: ErrorBoundary: SSG bakes a caught error's fallback at 200 and reports success
## Describe

During a static build (`mode: 'static'`), an `<ErrorBoundary>` catch does not fail the route. The render succeeds, the fallback streams, and the status stays 200.

- The SSG worker judges success by status only: `result.ok = (status >= 200 && status < 300) || …` (`packages/qwik-router/src/ssg/worker-thread.ts`).
- No SSG code reads `RenderToStreamResult.errorBoundaryCaught` (added in #8745).

Result: a build-time blip bakes the fallback into the static HTML. The build reports success. Crawlers index the fallback. A retry button has no live endpoint to hit.

## Expected

- The static build fails the route when a boundary catches, the same as a rejected render. `errorBoundaryCaught` is the hook.
- Optional: a generator retry knob — re-render the route N times before the build fails.

## Notes

Follow-up to #8745. Applies to every error kind, not to one error class.

maiieul:
# What is it?
- Bug

# Description
Fixes #8977. SSG baked a caught `<ErrorBoundary>` fallback into static HTML at 200 and reported success. The render middleware now records the catch on the request, and the SSG worker fails the route: the build exits 1 and the path stays out of `staticPaths`, so a deployed app would still SSR it.

### DIFF
--- a/packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
+++ b/packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
@@ -26,6 +26,7 @@ import { ensureSlash } from '../../utils/pathname';
 import { performETagMatch, hash, normalizeETag, setETagHeader } from './etag-hash';
 import {
   getRequestMode,
+  RequestEvErrorBoundaryCaught,
   RequestEvETagCacheKey,
   RequestEvHttpStatusMessage,
   RequestEvShareServerTiming,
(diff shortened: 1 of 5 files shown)


packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts
```diff
@@ -816,6 +817,9 @@ The request origin "${inputOrigin}" does not match the server origin "${origin}"
           await stream.write((result as any as RenderToStringResult).html);
         }
         boundaryErrored ||= (result as RenderToStreamResult).errorBoundaryCaught === true;
+        if (boundaryErrored) {
+          requestEv.sharedMap.set(RequestEvErrorBoundaryCaught, true);
+        }
```
Varixo, line 822: should it be cleared somewhere?

maiieul: - createRequestEvent builds a fresh sharedMap per request (request-event-core.ts:57). The flag dies with the request.
- The render middleware is the last handler in the chain. Two paths re-enter the chain on the same sharedMap, and neither can see a stale flag.

Varixo (approved):
LGTM

PR #8981, opened 2026-08-28
fix(core): stop rebuilding a headless component's output on every re-render

### LINKED ISSUE #8980: [🐞] v2: a component which renders no DOM element of its own and uses `useVisibleTask$` is destroyed and rebuilt on every re-render

Qwik Runtime

### Describe the bug

A component which renders no DOM element of its own gets a placeholder `<script>` so that a document or window `useOn` event - `useVisibleTask$` among them - has something to sit on. The function which puts that placeholder into the output, [`injectPlaceholderElement`](https://github.com/QwikDev/qwik/blob/main/packages/qwik/src/core/shared/component-execution.ts), has five separate problems. All five are measured below, on `2.0.0-beta.42` and on `main` (`ae74a2ca`), in a client only render and after an SSR resume.

### 1. The component is destroyed and rebuilt on every re-render

```tsx
const Child = component$<{ text: string }>((props) => {
  useTask$(() => count('mount'));           // no track(), so this runs once per instance
  return <span>{props.text}</span>;
});

export const Parent = component$(() => {
  const isLoaded = useSignal(false);
  useVisibleTask$(() => {
    isLoaded.value = true;                  // one ordinary state change
  });
  return <Child text={eager(isLoaded.value)} />;   // a component as the root, no element around it
});
```

```
mount     2      <- destroyed and rebuilt
cleanup   1
```

Wrapping the child in a `<div>` or in `<></>` gives `mount 1`.

The child loses its state and its DOM. After SSR the whole subtree is rebuilt on resume - new fragment, new component vnode, new `<span>`:

```
without the fix, after resume            with a fixed injectPlaceholderElement
<Component q:id="4">                     <Component q:id="4">
  <Fragment[1048575]>       new            <Fragment[0]>
    <Component[1048575]>    new              <Component[0] q:id="20">   same instance
      <span>                new                <span :="bu_0">          same element
```

**Why.** The output is wrapped together with the placeholder in a fragment **without a key**, and a new wrapper is built on every render:

```ts
if (jsx.type !== Fragment) {
  return _jsxSorted(Fragment, null, null, [jsx, placeholder], 0, null);
  //                                                             ^^^^
}
```

`expectVirtual` in `vnode-diff.ts` never matches a fragment without a key against the existing tree:

```ts
const checkKey = type === VirtualType.Fragment;
const isSameNode = currentIsVirtual && currentKey === jsxKey && (checkKey ? !!jsxKey : true);
...
// For fragments without a key, always create a new virtual node (ensures rerender semantics)
if (jsxKey === null || diffContext.$isCreationMode$) { /* insert a brand new vnode */ }
```

A new vnode is inserted, `descend()` then turns on creation mode, and everything below is built from zero. Nothing written by a user reaches this rule, because the optimizer puts a key on every JSX node in dev and in production - this wrapper is the only fragment in core that has none.

That also explains why a `key` on the child changes nothing: the replacement happens one level above it.

### 2. A component returning a list of nodes loses the placeholder

```tsx
export const Banner = component$(() => {
  useVisibleTask$(() => console.log('visible'));
  return [<Left key="l" />, <Right key="r" />];
});
```

Rendered HTML: `<span>left</span><span>right</span>` - no `q-d:qinit` anchor at all, so after a resume the task can never run.

```ts
// For an array of nodes, we inject the placeholder into the first element.
if (isArray(jsx) && jsx.length > 0) {
    injectPlaceholderElement(jsx[0], placeholder);   // <- the returned wrapper is discarded
    return jsx;
}
```

The recursive call returns a new fragment and the caller ignores it. It only works when `jsx[0]` happens to be a fragment, because that branch modifies the node in place.

### 3. A component returning a signal or a promise loses the placeholder

Same cause - both fall through to `// For anything else we do nothing`. (Neither shape is in the `JSXOutput` type, but the runtime handles both: `findFirstElementNode` has an `isSignal` and an `isPromise` branch.)

### 4. The placeholder is pushed into the node the component returned

```ts
jsx.children.push(placeholder);   // the node the component returned, not a copy
```

Fine while a component rebuilds its JSX every render. Not fine when it returns a node it does not rebuild - a hoisted or a shared piece of JSX - because that node then keeps the `<script>` for good.

### 5. So the node carries that placeholder into every other component which returns it

```tsx
const VIEW = <>shared</>;

const First = component$(() => {
  useVisibleTask$(() => log.push('first-task'));
  return VIEW;                        // leaves a <script> inside VIEW
});

const Innocent = component$(() => VIEW);   // no hooks at all
```

```
scripts=2   attrs=["mock-chunk#_task#0","mock-chunk#_task#0"]
<div>shared<script q-d:qinit="mock-chunk#_task#0"></script>shared<script q-d:qinit="mock-chunk#_task#0"></script></div>

firing only the script rendered by `Innocent`  ->  ["first-task","first-task"]
```

`Innocent` registers nothing, yet it renders a `<script>` pointing at `First`'s visible task, and firing it runs that task - twice, because the handler was added to the same shared node on both renders.

## Expected behaviour

1. A re-render of such a component re-renders what it returned, it does not rebuild it.
2. The placeholder is rendered whatever the component returns.
3. The node a component returns is not modified.

### Reproduction

https://github.com/wermetal/qwik-v2-remount-issue-reproduction

### Steps to reproduce

```bash
git clone https://github.com/wermetal/qwik-v2-remount-issue-reproduction
cd qwik-v2-remount-issue-reproduction/ssr
pnpm install
pnpm build.all
```

`build.all` builds the client and ssr bundles, renders the HTML with `@qwik.dev/core/server` the way a server would, serves it, opens it in the Chrome already installed on the machine and prints what each case measured. Output on `2.0.0-beta.42`:

```
1. a component with a component as its root is rebuilt
    FAIL - client side mount=1 (expected 0, the server already mounted it)
2. a component returning a list renders no placeholder, so its task never runs
    FAIL - task ran=0 (expected 1), <script q-d:qinit> in the section=0 (expected 1)
3. the same for a component returning a signal
    FAIL - task ran=0 (expected 1), <script q-d:qinit> in the section=0 (expected 1)
4. the placeholder is pushed into the node the component returned
    FAIL - children of the shared node=2 (expected 1, it was written to)
5. that node then renders a placeholder in a component with no hooks
    FAIL - <script q-d:qinit> in the section=2 (expected 1, only the component which registered the task)
control: <></> between the component and its child     PASS - client side mount=0
control: <div> between the component and its child     PASS - client side mount=0
```

The anchors are counted in the HTML the server produced, not in the live DOM, because qwikloader removes those attributes while it resumes the page. `pnpm verify --headed` to watch it in a browser, `pnpm serve` to look at the page yourself.

The repository root is the same five cases in a client only render (`pnpm install && pnpm build && pnpm serve`). Problems 2 and 3 show there as a missing anchor, but their task still runs: a client only render starts a visible task itself, without the `<script>`. It also has `?case=loop`, where the component being rebuilt owns a subscription - rebuilding it drops andremakes that subscription, resubscribing replays the newest value, which is one more emit, and it runs to the 2000 cap in about two seconds:

```
subscribe      2001
unsubscribe    2000
Leaf.mount     1
```
 

### System Info

```shell
System:
    OS: macOS 26.5.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 96.03 MB / 24.00 GB
    Shell: 5.9.2 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 25.2.1
    npm: 11.7.0
    pnpm: 10.26.0
  Browsers:
    Chrome: 151.0.7922.170
    Firefox: 154.0
    Safari: 26.5.2
  npmPackages:
    @qwik.dev/core: 2.0.0-beta.42 => 2.0.0-beta.42
    typescript: ^5.6.0 => 5.9.3
    vite: ^8.2.2 => 8.2.2
```

### Additional Information

I have a fix and tests ready and I intend to submit a PR.

wermetal:
# What is it?

- Bug

# Description

Fixes #8980

A component which renders no DOM element of its own gets a placeholder `<script>`, so that a document or window `useOn` event - `useVisibleTask$` among them - has something to sit on. `injectPlaceholderElement` put that placeholder into the output in four ways which were wrong:

1. **Everything below the component was destroyed and rebuilt on every re-render.** The output was wrapped together with the placeholder in a fragment **without a key**, and a new wrapper was built on every render. `expectVirtual` never matches an unkeyed fragment against the existing tree, so a new vnode was inserted, `descend()` turned on creation mode, and every component below lost its state and its DOM. After SSR the whole subtree was rebuilt on resume.

   Nothing a user writes reaches that rule, because the optimizer keys every JSX node in dev and in
   production - this wrapper was the only fragment in core without a key.

2. **A component returning a list of nodes lost the placeholder**, so its `useVisibleTask$` never
   ran after a resume:

   ```ts
   if (isArray(jsx) && jsx.length > 0) {
       injectPlaceholderElement(jsx[0], placeholder);   // the returned wrapper was discarded
       return jsx;
   }
   ```

   The recursive call returns a new fragment and the caller ignored it. It only worked when `jsx[0]`
   happened to be a fragment, because that branch modified the node in place.

3. **A component returning a signal or a promise lost it too** - both fell through to
   `// For anything else we do nothing`.

4. **The placeholder was pushed into the node the component returned.** For a component returning a
   node it does not rebuild - hoisted or shared JSX - that node kept the `<script>` for good, and
   every other component returning the same node then rendered a `<script>` carrying the first
   component's task.

## The change

Whatever the component returns is now placed, together with the placeholder, in a single fragment
carrying a stable key, and the returned node is never written to:

- an **array** becomes the children of that fragment,
- a **fragment** is rebuilt with one more child, keeping its own key when it has one,
- **everything else** - a component, an element, text, a signal, a promise, nothing at all - is
  wrapped in it.

Because the shape of the output no longer decides the shape of the tree, a component which returns a
single node on one render and a list on the next keeps its children instead of rebuilding them.

## Behaviour changes

- A component returning a list is rendered inside that fragment, one level deeper than before, with
  the placeholder as its last child. `qinit` handlers run in document order, so a list whose first
  item was a fragment used to run the parent's task *before* the tasks of the components inside the
  list; it now runs after them, which is what a component with a single root has always done. The
  VDOM assertion in `should trigger in empty components array` is updated for this.
- A fragment returned without a key by such a component is given the wrapper key, so it is matched
  across re-renders like every other output. The same component *without* a document or window
  `useOn` event still gets a new fragment on every render, because unkeyed fragments are
  deliberately never matched by `expectVirtual`.

## Not changed

The rule in `expectVirtual` itself. Relaxing it would let a fragment reuse the vnode of a
**component** standing in the same position, with that component's QRL, props and tasks still
attached; doing it safely needs the diff to know which kind of virtual node it is looking at, and a
vnode does not carry that in a production build. Worth noting that before `eb493deb3` ("fix: signal
wrapper should not rerender") every unkeyed virtual node was re-created, and that commit made an
exception for the other kinds and left fragments behind - so the rule looks like leftover behaviour
rather than a decision. It still affects JSX built at runtime, or shipped by a package compiled with
another toolchain, since that JSX has no keys.

`error-boundary.ts` also builds its client side fallback in an unkeyed fragment, so a boundary
showing its fallback rebuilds that subtree on every re-render. Left alone here, because `reset` may
depend on it.

## Tests

Added to `use-visible-task.spec.tsx`, each running under both `domRender` and `ssrRenderToDom`, all
of them failing on `main`:

- `should not re-create the root child component on re-render` - the child records the prop it was
  born with, so `vtrue|vfalse` proves the instance survived; it also asserts the `<span>` is the same
  DOM node after a resume
- `should not re-create the root child when output changes from child to array`
- `should trigger in empty components returning a list of components` - the anchor is rendered, and
  under SSR it carries the QRL rather than an empty attribute
- `should not modify a jsx node which the component returns` - the returned node is untouched, and
  the component which registered nothing renders no `<script>`
- `should keep every instance of a headless component when there are many of them` - twenty of them
  on one page, keyed and unkeyed, none mixed up by the shared wrapper key
- `should trigger in empty components array` - updated for the new placeholder position

# Checklist


.changeset/lucky-moons-remount.md
```diff
@@ -0,0 +1,40 @@
...
+
+- A component returning a list is now rendered inside that fragment, one level deeper than before,
+  with the placeholder as the last child of it. `qinit` handlers run in document order, so a list
+  whose first item was a fragment used to run the parent's task before the tasks of the components
+  inside the list; it now runs after them, which is what a component with a single root has always
+  done.
+- A fragment returned without a key by such a component is given the wrapper key, so it is matched
+  across re-renders like every other output. The same component without a document or window `useOn`
+  event still gets a new fragment on every render, because unkeyed fragments are deliberately never
+  matched by `expectVirtual`.
```
Varixo: this is way to long, please make it short like 1-2 sentences max, no implementation details

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -334,47 +334,56 @@ function findFirstElementNode(jsx: JSXOutput): ValueOrPromise<JSXNodeInternal<st
...
 
+/**
+ * Key of the fragment which wraps a headless component's output together with its `useOn`
+ * placeholder `<script>`.
+ *
+ * The wrapper is re-created on every render of the component, so it needs a stable identity.
+ * Unkeyed fragments are intentionally never matched by `expectVirtual`, which would tear down and
+ * re-create the whole subtree - and with it every component instance below it - on every
+ * re-render.
+ */
```
Varixo: the comment is not necessary - if you think its needed please shorten it to max two lines, dont give implementation details here

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -334,47 +334,56 @@ function findFirstElementNode(jsx: JSXOutput): ValueOrPromise<JSXNodeInternal<st
...
+const PLACEHOLDER_FRAGMENT_KEY = ':useOn';
```
Varixo: this is in wrong place, look at markers.ts file

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -32,10 +32,16 @@ import { ErrorProvider } from '../../testing/rendering.unit-util';
...
 import { Task, TaskFlags } from '../use/use-task';
+import type { JSXNodeInternal, JSXOutput } from '../shared/jsx/types/jsx-node';
 
 const debug = false; //true;
 Error.stackTraceLimit = 100;
 
+/** A function call is left eager by the optimizer, so reading a signal through it re-renders. */
+function stringify(value: unknown) {
+  return `v${value}`;
+}
```
Varixo: this is assumption that may change later. We can change the optimizer and then test will pass even if it shouldnt or vice versa

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+          </span>
+        );
+      });
+
+      /**
+       * The component renders no host element of its own, so `useVisibleTask$` makes it render a
+       * placeholder `<script>` next to the child, wrapped in a fragment. That wrapper is re-created
+       * on every render, so it has to keep its identity - otherwise the child below it is destroyed
+       * and rebuilt, losing its state and its DOM.
+       */
```
Varixo: this is unnecessary, test name already describes the problem

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+        });
+        return <Child text={stringify(isLoaded.value)} />;
+      });
+
+      const { document } = await render(<Cmp />, { debug });
+      const spanBeforeRerender = document.querySelector('span');
+      if (render === ssrRenderToDom) {
+        await trigger(document.body, 'script', 'd:qinit');
+      }
+      await getTestPlatform().flush();
```
Varixo: its deprecated, remove it

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+       * placeholder `<script>` next to the child, wrapped in a fragment. That wrapper is re-created
+       * on every render, so it has to keep its identity - otherwise the child below it is destroyed
+       * and rebuilt, losing its state and its DOM.
+       */
+      const Cmp = component$(() => {
+        const isLoaded = useSignal(false);
+        useVisibleTask$(() => {
+          isLoaded.value = true;
+        });
+        return <Child text={stringify(isLoaded.value)} />;
```
Varixo: instead of function just add isLoaded.value to the component body - it will create component subscription and will rerender when isLoaded changes. Or if you want something better:
```
const text = `v${isLoaded.value}`;
return <Child text={text} />;
```

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+          </span>
+        );
+      });
+
+      const Cmp = component$(() => {
+        const isLoaded = useSignal(false);
+        useVisibleTask$(() => {
+          isLoaded.value = true;
+        });
+        const child = <Child text={stringify(isLoaded.value)} />;
```
Varixo: I think the same here

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+      if (render === ssrRenderToDom) {
+        await trigger(document.body, 'script', 'd:qinit');
+      }
+      await getTestPlatform().flush();
+
+      // the child sees the new prop, but it is still the instance created with the old one
+      expect(document.querySelector('span')?.textContent).toBe('vtrue|vfalse');
+      // and it is the same dom node, not a rebuilt one (only observable after a resume, because a
+      // client only render has already re-rendered by the time it returns)
+      expect(document.querySelector('span')).toBe(spanBeforeRerender);
```
Varixo: if you want to look at rerenders better to do a pattern (globalThis as any).counter++; in component body. its easier to understand and cleaner

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+
+      const { document } = await render(<Cmp />, { debug });
+      const spanBeforeRerender = document.querySelector('span');
+      if (render === ssrRenderToDom) {
+        await trigger(document.body, 'script', 'd:qinit');
+      }
+      await getTestPlatform().flush();
+
+      expect(document.querySelector('span')?.textContent).toBe('vtrue|vfalse');
+      expect(document.querySelector('span')).toBe(spanBeforeRerender);
```
Varixo: the same pattern here as above ((globalThis as any).counter++)

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+
+      const { vNode, document } = await render(<Cmp />, { debug });
+      if (render === ssrRenderToDom) {
+        await trigger(document.body, 'script', 'd:qinit');
+      }
+      await getTestPlatform().flush();
+
+      // the placeholder goes into the rendered output, the returned node stays untouched
+      const children = shared.children;
+      expect(Array.isArray(children) ? children.length : 1).toBe(1);
```
Varixo: we should know if children are array or not, right?

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+          </span>
+        );
+      });
+
+      const Headless = component$<{ n: number }>((props) => {
+        const isLoaded = useSignal(false);
+        useVisibleTask$(() => {
+          isLoaded.value = true;
+        });
+        return <Child n={props.n} text={stringify(isLoaded.value)} />;
```
Varixo: (globalThis as any).counter++ pattern again

packages/qwik/src/core/tests/use-visible-task.spec.tsx
```diff
@@ -904,6 +936,182 @@ describe.each([
...
+
+      const texts = Array.from(document.querySelectorAll('span')).map((s) => s.textContent);
+      // every instance kept the value it was born with, and kept its own number
+      expect(texts).toHaveLength(COUNT * 2);
+      expect(new Set(texts).size).toBe(COUNT * 2);
+      expect(texts.filter((t) => !t!.endsWith('|vfalse]'))).toEqual([]);
+      expect(texts[0]).toBe('[0:vtrue|vfalse]');
+      expect(texts[COUNT]).toBe('[100:vtrue|vfalse]');
+      // one placeholder per instance, they do not share or steal each other's
+      expect(document.querySelectorAll('script[q-d\\:qinit]').length).toBe(COUNT * 2);
```
Varixo: these expectations are not doing what the test title says. To check it you should do a test with some item reordering

Varixo (changes requested):
Hello, thank you for the PR. I added some comments.
However additional wrapper is a bigger problem than it looks. This will have big serialization impact.

how about something easier. We could create no fragment wrapper at all.
1. lets do injectPlaceholderElement returning only [jsx, placeholder] for any case
2. createPlaceholderScriptNode will get :useOn as a key instead of null

didnt check it but it should work (of course the use-on and yours tests will need changes and removing fragment wrapper from expectations)

Varixo (approved):
LGTM, thanks for your work!

PR #8984, opened 2026-08-30
fix(preloader): don't over-preload large lazy-import maps (CMS/component registries)

intellix:
## Problem

A component that references a **large map of lazy imports** — a CMS component registry, or a router/switch that can render any one of *N* components — causes the preloader to preload **every one of the *N* bundles** the first time that component renders, even though only one branch is ever taken.

Root cause: when a bundle is certain to load (probability `1`), the queue elevates **all** of its imports to certain — including its *dynamic* imports. For a selection map with 100+ lazy children that means 100+ bundles get force-preloaded. The coarse `1/10` probability encoding then rounds small probabilities up and compounds them across propagation chains, so even indirect children cross the preload threshold.

Real-world impact: on a production Qwik app (large website) whose Storyblok registry maps ~120 bloks to lazy imports, the homepage preloaded **~3,600** JS bundles. With this change it preloads **~1,697** — only the bloks the page actually renders, plus their genuine dependencies.

## Why this is a fresh PR (re: #8983)

#8983 tried just the cascade half and was correctly closed — as @maiieul noted, on its own it **under-preloads**: a lazy modal referenced by an on-page component would no longer have its bundle ready before the click. This PR keeps modal-style deps preloaded and only damps genuine large fan-outs. That distinction is enforced by a regression test (below).

## The fix — four coordinated changes

1. **Cascade** (`queue.ts`) — only **static** imports (`$importProbability$ === 1`) of a certain bundle are elevated to certain. Dynamic imports keep propagating **multiplicatively** instead of being forced to `1`.
2. **Fan-out-aware edge probability** (`convertManifestToBundleGraph`) — a bundle's dynamic-import edges are damped by `min(1, FANOUT_FREE / fanOut)`. A 2-child component (a modal + its handler) is essentially undamped; a 120-child selection map scores each edge far lower. **Low fan-out is unchanged** — this is what preserves modal preloading.
3. **Finer probability encoding** — the bundle graph is serialised at **1/100** instead of 1/10. A damped registry edge keeps a distinct low value (e.g. `0.06`) instead of rounding up to `0.1` and compounding across a chain (which otherwise pushed unused bundles over the floor on SPA navigation).
4. **Preload floor** (`$minPreloadProbability$`, default `0.2`) — bundles scoring below the floor aren't speculatively preloaded.

## Regression coverage

`queue.unit.ts` adds a guard that a **certain bundle's lazy imports (a modal + its handler) are still queued for preload** — i.e. the change does not reintroduce the under-preloading of #8983 — alongside a fan-out test proving a large selection map is damped. `bundle-graph.unit.ts` covers the encoding/fan-out math.

- `packages/qwik/src/core/preloader/queue.unit.ts` — 8 passed
- `packages/qwik-vite/src/plugins/bundle-graph.unit.ts` — 8 passed

## Files

- `packages/qwik/src/core/preloader/queue.ts` — cascade + floor threshold
- `packages/qwik/src/core/preloader/constants.ts` — `$minPreloadProbability$`
- `packages/qwik/src/core/preloader/bundle-graph.ts` — 1/100 decode
- `packages/qwik-vite/src/plugins/bundle-graph.ts` — fan-out damping + 1/100 encode
- unit tests + changeset

Happy to split this into smaller PRs (e.g. cascade+fan-out first, encoding second) if that's easier to review.

### DIFF
--- a/packages/qwik/src/core/preloader/constants.ts
+++ b/packages/qwik/src/core/preloader/constants.ts
@@ -12,6 +12,11 @@ export const doc = isBrowser ? document : undefined!;
 export const config = {
   $DEBUG$: false,
   $maxIdlePreloads$: 25,
+  // Bundles below this probability are not speculatively preloaded. Combined with fan-out-aware
+  // edge probabilities (see convertManifestToBundleGraph), this keeps large lazy maps/registries
+  // (many mutually-exclusive dynamic imports, each scored low) from preloading every entry, while
+  // still preloading likely code (a modal, a route's components) that scores well above it.
+  $minPreloadProbability$: 0.2,
 };
 
 // Determine which rel attribute to use based on browser support
--- a/.changeset/fix-preloader-dynamic-import-cascade.md
+++ b/.changeset/fix-preloader-dynamic-import-cascade.md
@@ -1,0 +1,16 @@
+---
+'@qwik.dev/core': patch
+---
+
+fix(preloader): don't force-preload every dynamic import of a certain bundle
+
+When a bundle reaches ~100% probability, the preloader elevated **all** of its
+dependencies — static *and* dynamic — to ~99% and preloaded them, bypassing the
+`maxIdlePreloads` throttle. Static imports are genuinely certain to load with
+their importer, but dynamic imports are a runtime choice. As a result, a certain
+bundle that merely references a large map of lazy imports (e.g. a CMS/router
+component that can render any of N components) would preload all N bundles even
+though only one branch is ever taken.
+
+Dynamic imports now keep propagating multiplicatively (parentProbability ×
+edgeProbability); only static imports are elevated to certain.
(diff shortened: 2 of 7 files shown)


packages/qwik/src/core/preloader/constants.ts
```diff
@@ -12,6 +12,11 @@ export const doc = isBrowser ? document : undefined!;
 export const config = {
   $DEBUG$: false,
   $maxIdlePreloads$: 25,
+  // Bundles below this probability are not speculatively preloaded. Combined with fan-out-aware
+  // edge probabilities (see convertManifestToBundleGraph), this keeps large lazy maps/registries
+  // (many mutually-exclusive dynamic imports, each scored low) from preloading every entry, while
+  // still preloading likely code (a modal, a route's components) that scores well above it.
+  $minPreloadProbability$: 0.2,
```
maiieul, line 19: So when are those bundles preloaded? Should they never be preloaded? What happens if they're not preloaded and the user needs them?

PR #8988, opened 2026-09-01
fix(core): fire onQVisible$ for client-rendered elements

maiieul:
# What is it?
- Bug

# Description
After an SPA nav, on the new page:
- onQvisible$ does not fire at all
- useVisibleTask$ with intersection-observer strategy fires eagerly right after the navigation even when the element is far below the fold

Claude:
`onQVisible$` (and `useOn('qvisible')`) never fired for elements rendered on the client — e.g. after a SPA navigation — because the qwik loader's IntersectionObserver only finds elements carrying the `q-e:qvisible` attribute, which only SSR wrote. CSR now writes the attribute, vnode-diff queues a loader notification that the flush phase sends once the journal lands in the DOM, and the loader re-scans on repeated registrations.

With the loader path working, client-created `useVisibleTask$` with the default `intersection-observer` strategy now genuinely waits for the element to become visible instead of running eagerly at render. `document-ready`/`document-idle` strategies keep running right away (the document is already ready/idle). Headless components fall back to `d:qinit` as before, which the loader now re-processes after the flush — this also means client-rendered `useOnDocument('qinit'/'qidle')` handlers fire after the flush instead of never. The loader trigger only ever performs a task's initial run.

In Vitest there is no loader, so component tests for client-rendered visible tasks must simulate it with `trigger(..., 'qvisible')` (or `'d:qinit'` for headless), same as the SSR arms always did.


packages/qwik/src/core/shared/cursor/cursor-flush.ts
```diff
@@ -28,10 +38,24 @@ export function executeFlushPhase(cursor: Cursor, container: Container): void {
...
+function journalAddsQVisibleListener(journal: VNodeJournal): boolean {
+  for (let i = 0; i < journal.length; i++) {
+    const operation = journal[i];
+    if (operation instanceof SetAttributeOperation && operation.attrName === QVisibleAttr) {
+      return true;
+    }
+  }
+  return false;
+}
+
```
Varixo: imo its wrong. Scheduler shouldnt know about any diff logic. This fixes only an effect of this. This should be fixed somewhere in vnode-diff

packages/qwik/src/core/shared/component-execution.ts
```diff
@@ -213,8 +214,13 @@ function addUseOnEvents(
...
         }
         if (targetElement) {
-          if (targetElement.type === 'script' && key === qVisibleEvent) {
+          if (targetElement.type === 'script' && key === QVisibleAttr) {
             eventKey = 'q-d:qinit';
+            const isSsr = qTest ? isServerPlatform() : isServer;
+            if (!isSsr) {
+              // qinit already fired, so run the visible tasks now
+              scheduleVisibleTasksNow(container, useOnEvents[key].qrls);
+            }
```
Varixo: not sure what it does, but for sure its in very wrong place. It is creating unwanted side effect. The function should only add use on effects to jsx. maybe vnode diff should be responsible for that, hard to tell without context

packages/qwik/src/core/tests/component.spec.tsx
```diff
@@ -2451,9 +2451,7 @@ describe.each([
     });
 
     const { document } = await render(<Cmp />, { debug });
-    if (render === ssrRenderToDom) {
-      await trigger(document.body, 'div', 'qvisible');
-    }
+    await trigger(document.body, 'div', 'qvisible');
```
Varixo, line 2454: the changes here and in the other tests look wrong. why we need to trigger qvisible on pure csr render?

packages/qwik/src/core/shared/cursor/cursor-flush.ts
maiieul: Claude:
> Agreed, fixed in eb1e4c3ae. The flush timing is needed — the loader can only observe the element after the journal lands in the DOM — but the event knowledge now lives in vnode-diff: registerEventHandlers queues the loader notification on CursorData and executeFlushPhase just drains the queue after the flush, without knowing which events or why.

packages/qwik/src/core/tests/component.spec.tsx
maiieul: Claude:
> This PR changes client-rendered visible tasks to wait for visibility instead of running eagerly at render. Eager was never a real choice before: the qvisible listener was dead on CSR, so running at render was the only way the task ran at all. Vitest has no qwikloader or IntersectionObserver, so tests simulate the observer with trigger(..., 'qvisible') — the same convention the SSR arms always used. Same story for the d:qinit triggers on headless components.

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1101,12 +1103,33 @@ function registerEventHandlers(
...
   // TODO only do these when not already present
-  if (key.charAt(2) !== 'e') {
+  if (key.charAt(2) !== 'e' || scopedKebabName === QVisibleScopedEvent) {
     vnode_setAttr(diffContext.$journal$, vnode, key, '');
   }
-  registerQwikLoaderEvent(diffContext, scopedKebabName);
+  if (
+    scopedKebabName === QVisibleScopedEvent ||
+    scopedKebabName === 'd:qinit' ||
+    scopedKebabName === 'd:qidle'
```
Varixo: maybe it should be a const enum for those events

also the condition `scopedKebabName === QVisibleScopedEvent` is doubled (here and condition above). We can check it once and use the result in both places

packages/qwik/src/core/client/vnode-diff.ts
```diff
@@ -1101,12 +1103,33 @@ function registerEventHandlers(
...
+  }
+}
+
+function queueQwikLoaderEventAfterFlush(diffContext: DiffContext, eventName: string) {
+  const cursorData = getCursorData(diffContext.$cursor$);
+  if (cursorData) {
+    const loaderEvents = (cursorData.notifyQwikLoaderEvents ||= []);
+    if (!loaderEvents.includes(eventName)) {
+      loaderEvents.push(eventName);
+    }
```
Varixo: its not really clear to me why we queue or not based on existence of cursor data on cursor

packages/qwik/src/core/shared/cursor/cursor-flush.ts
```diff
@@ -29,6 +34,13 @@ export function executeFlushPhase(cursor: Cursor, container: Container): void {
     _flushJournal(journal);
     cursorData.journal = null;
   }
+  const loaderEvents = cursorData.notifyQwikLoaderEvents;
+  if (loaderEvents) {
+    cursorData.notifyQwikLoaderEvents = null;
+    for (let i = 0; i < loaderEvents.length; i++) {
+      registerQwikLoaderEvent(container as ClientContainer, loaderEvents[i]);
+    }
+  }
```
Varixo: I see the only problem with this - its very specific to one edge case, but all cursors need to pay for it.
keeping null on every cursor is additional 8b for every cursor. The solution is also not very general to use it in other place in the future.

What if we can make global WeakMap that will contain cursor -> notifyQwikLoaderEvents. This way we create entires only when needed and we dont add more memory to every cursor

The other idea is to make generic solution like onFlush or something, but for now maybe its overthinking

packages/qwik/src/core/tests/component.spec.tsx
Varixo: I am not really conviced, the qvisible worked before for csr, and I think it should work this way, because this is how it works in real app. Unit test should simulate it and trigger event when needed for csr

packages/qwik/src/core/client/vnode-diff.ts
maiieul: Done in 7b2479636 — QwikLoaderScanEvent const enum for e:qvisible/d:qinit/d:qidle, and the qvisible check is computed once and reused.

packages/qwik/src/core/client/vnode-diff.ts
maiieul: No good reason — vnode_diff always walks a real cursor, so the fallback was dead defensive code, and if it ever ran it would push before the flush and reintroduce the original bug. Removed in 7b2479636; it now asserts the cursor data like executeFlushPhase does.

packages/qwik/src/core/shared/cursor/cursor-flush.ts
maiieul: Went with the WeakMap in 7b2479636: pendingQwikLoaderEvents lives in cursor-props keyed by CursorData, with queueQwikLoaderEvent / takeQwikLoaderEvents helpers. Nothing is allocated unless a scan event is registered, mergeCursorData moves entries between the two data objects, and the CursorData field is gone. Agree a generic onFlush is overthinking it for now.

packages/qwik/src/core/tests/component.spec.tsx
maiieul: Hmmm so the issues are, after an SPA nav, on the new page:
`onQvisible$` does not fire at all
`useVisibleTask$` with `"intersection-observer"` strategy fires eagerly right after the navigation even when the element is far below the fold.

I added this at the top of the PR description.


Claude:
> on main, onQVisible$ on a client-rendered element never fired in a real browser — that's the bug this PR started from (SPA-navigate to the preloader-test counters page, scroll to the bottom, no log). The element had no q-e:qvisible attribute and no q:observed, so the loader never observed it. useVisibleTask$ only appeared to work on CSR because it ran eagerly at render and never went through qvisible at all.

packages/qwik/src/core/shared/cursor/cursor-flush.ts
```diff
@@ -29,6 +34,12 @@ export function executeFlushPhase(cursor: Cursor, container: Container): void {
     _flushJournal(journal);
     cursorData.journal = null;
   }
+  const loaderEvents = takeQwikLoaderEvents(cursorData);
```
Varixo: getPendingQwikLoaderEvents?

we have only get*/retrieve* and we should be consistent

maiieul: Renamed in 0c6cf8e15. Note it still removes the entry on read — the flush must not re-notify next time round, and mergeCursorData relies on that to move entries between cursors — so it's a get-and-clear despite the get prefix. Happy to split it into get + clear if you'd rather keep get* pure.

Varixo (dismissed):
LGTM

PR #8989, opened 2026-09-01
refactor: better docs score

Varixo:
- use lightweight sidemenu
- delay preloading until first paint and load


packages/qwik/src/server/types.ts
```diff
@@ -39,7 +39,7 @@ export interface PreloaderOptions {
    * Bundles that reach 100% probability (static imports of other bundles) will always be preloaded
    * immediately, no limit.
    *
-   * Defaults to `25`
+   * Defaults to `20`
```
maiieul: It's still 25. See my other comment.

packages/qwik/src/core/preloader/queue.unit.ts
```diff
@@ -88,6 +88,28 @@ test('appends preloads directly to head within a trigger slice', async () => {
   expect(document.head.querySelectorAll('link').length).toBe(2);
 });
 
+test('limits speculative preloads to ten by default', async () => {
```
maiieul: This test is a false postive. When I run preloader-test, the `ssrPreloads` are still at 5 and `maxIdlePreloads` are at 25.

Imo we can remove the `ssrPreloads` if that affects lighthouse, and the `maxIdlePreloads` need to remain at 25 because it speeds up total preloading time without impacting reprioritization times too much. So we should not touch this and instead ensure that the client side idle modulepreload appends start after critical downloads and CPU work so that lightouse is happy.

maiieul (changes requested):

packages/qwik/src/core/preloader/queue.unit.ts
```diff
@@ -88,7 +88,7 @@ test('appends preloads directly to head within a trigger slice', async () => {
   expect(document.head.querySelectorAll('link').length).toBe(2);
 });
 
-test('limits speculative preloads to ten by default', async () => {
+test('limits concurrent speculative preloads to maxIdlePreloads', async () => {
```
maiieul, line 91: Are you sure this test tests anything at all? If not let's remove it.

packages/qwik/src/core/preloader/queue.ts
```diff
@@ -284,6 +285,8 @@ export const preload = (item: string | string[], probability?: number) => {
 };
 
 if (isBrowser) {
+  setPreloader(preload);
```
maiieul, line 288: Why do we need to do this now?

Varixo: because I removed static import from qrl-class, so the preloader has to register itself on evaluation, otherwise requestPreload has nothing to call and every QRL and router hint is dropped

PR #8995, opened 2026-09-04
fix(router): run non-blocking loaders without async local storage

### LINKED ISSUE #8994: `blockSSR: false` route loader throws and truncates the SSR document when AsyncLocalStorage is unavailable
Repro repository: **https://github.com/thelfroes/qwik-blockssr-als-repro**

**One line:** a `routeLoader$` declared `{ blockSSR: false }` reaches
`ServerRouteLoaderCapture.load()` in a continuation that has no invoke context, so
`getRequestEvent()` returns `undefined`, `load()` throws mid-render, and the response ships as an
HTTP **200** containing two `<html>` elements and no resume payload.

Versions: `@qwik.dev/core` / `@qwik.dev/router` `2.0.0-beta.40` (stock, unpatched),
`q:version 2.0.0-beta.40-dev+a459d55`, vite 8.2.2, wrangler 4.128.0 / workerd, compat date
`2025-07-18`.

---

## Minimal reproduction

Seven source files, no network, no env. Full app and captured HTML: **https://github.com/thelfroes/qwik-blockssr-als-repro**

```bash
npm install
npm run build
npm run serve:als-off     # wrangler pages dev dist --compatibility-flags=nodejs_als
curl -s http://localhost:8791/ | wc -c        # => 962   (broken)
```

The whole app is one layout and one page:

```tsx
// src/routes/layout.tsx
const slow = <T,>(value: T, ms: number): Promise<T> =>
  new Promise((resolve) => setTimeout(() => resolve(value), ms));

export const useLateLoader = routeLoader$(
  async () => slow('late-loader-value', 60),
  { blockSSR: false }, // <-- delete this line; everything else identical
);

export const useBlockingLoader = routeLoader$(async () =>
  slow('blocking-loader-value', 60),
);
```

`--compatibility-flags=nodejs_als` exposes `AsyncLocalStorage` as a global but **not** as
`process.getBuiltinModule('node:async_hooks')`, which is the only path
`packages/qwik/src/core/shared/platform/async-local-storage.ts` resolves. Qwik prints its own
`AsyncLocalStorage is not available` warning at startup, which is how you know the harness is armed.
Any runtime where that resolution fails should behave the same way.

`blockSSR` is experimental in beta.40, so the config carries
`qwikVite({ experimental: ['blockSSR'] })`.

### Three arms, 10 requests each

| arm        | `blockSSR: false` | ALS     | bytes   | `<script type="qwik/state">` | complete | throws          |
| ---------- | ----------------- | ------- | ------- | ---------------------------- | -------- | --------------- |
| **broken** | yes               | **off** | **962** | **absent**                   | **no**   | **1 / request** |
| working    | no                | off     | 4803    | present                      | yes      | 0               |
| control    | yes               | on      | 4802    | present                      | yes      | 0               |

10 / 10 in every arm. The broken arm is deterministic — no concurrency or load required. The
ALS-off server logged one `Unable to determine the current RequestEvent.` per request, 12 / 12
across its whole lifetime.

---

## What and where exactly is it truncated

`captures/broken.html`, 962 bytes:

| byte range | content                                                                                |
| ---------- | -------------------------------------------------------------------------------------- |
| 0 – 14     | `<!DOCTYPE html>`                                                                      |
| 15 – 196   | the Qwik container's opening tag, `<html q:render="ssr" … q:container="paused" … :>`   |
| **197**    | **cut** — a complete second `<html>…</html>` 500 error document is welded in from here |

```
<!DOCTYPE html><html q:render="ssr" q:route="/" q:container="paused" … q:instance="o1cyg5" :><html>
                                                                                              ^ byte 197
<head>
  <meta charset="utf-8">
  <meta http-equiv="Status" content="500">
  <title>500 Internal Server Error</title>
  …
</head>
<body><p><strong>500</strong> <span>Internal Server Error</span></p></body>
</html>
```

- **First thing missing:** the real `<head>` content — every `modulepreload`, the `<title>`, and
  the entire `<body>`.
- **`<script type="qwik/state">`: absent.** Nothing resumes; the page is completely inert.
  (The marker is `qwik/state`. `qwik/json` appears in a healthy document only as the _attribute
  value_ of `q:func="qwik/json"`.)
- **Closing `</html>`: present**, but it belongs to the welded error page, not to the real render.
  A "does it end with `</html>`" check therefore passes on a broken document.
- `q:container="paused"` is still on the first tag, so the client sees a container it can never
  resume.
- Status line is **HTTP 200**. Behind a CDN this is a cacheable success.

The healthy `captures/working.html` (4803 bytes) has the full head, both loader values, a single
`</html>`, and the `qwik/state` payload.

The thrown error, from the wrangler log of the broken arm:

```
Error: Unable to determine the current RequestEvent.
  at gd.load (…/server/build/q-4usrIM51.js:8:3855)
```

---

## The code

All line numbers below are from `QwikDev/qwik` at `e478dfbd9` (origin/main, 2026-09-01). I checked
HEAD directly rather than relying on a released tag — **the naked `getRequestEvent()` is still
present at HEAD.** It is also present in `d7ab00689`.

**1. The throw** — `packages/qwik-router/src/runtime/src/route-loaders.ts:148`

```ts
  load() {
    const requestEv = getRequestEvent();
    if (!requestEv) {
      throw new Error('Unable to determine the current RequestEvent.');   // :151
    }
    // Use pre-computed value from loadersMiddleware if available,
    // to avoid re-running the loader after the response stream is open.
    const values = getRouteLoaderValues(requestEv);
    if (this.hash in values) {
      return values[this.hash];
    }
    // A background (blockSSR:false) loader must not touch the page response.
    const ev = this.blockSSR ? requestEv : detachResponseFromEvent(requestEv);
    return loadRouteLoaderByQrl(this.hash, this.qrl, this.validators, ev);
  }
```

Note the ordering: the pre-computed value at `:155` would satisfy the call, but `:149` throws
before it is ever reached.

**2. The two resolvers** — `route-loaders.ts:491`

```ts
export const getRequestEvent = (
  thisArg?: unknown,
): RequestEvent | undefined => {
  if (!isServer) {
    throw new Error('getRequestEvent() can only be used on the server.');
  }
  return (
    _asyncRequestStore?.getStore() ||
    [thisArg, _getContextEvent()].find(isRequestEvent)
  );
};
```

`_getContextEvent()` (`packages/qwik/src/core/use/use-core.ts:256`) returns
`tryGetInvokeContext()?.$event$` — it only works while an invoke context is on the stack.
`_asyncRequestStore` is `undefined` whenever ALS cannot be resolved
(`packages/qwik-router/src/middleware/request-handler/async-request-store.ts:7`), and it is the only
thing wrapping the request (`user-response.ts:49`).

**3. The actual discriminator** — `packages/qwik-router/src/middleware/request-handler/resolve-request-handlers-core.ts:334`

```ts
let allBlockSSRLoaders: Promise<void> | undefined;
for (let i = 0; i < routeLoaders.length; i++) {
  const loader = routeLoaders[i];
  const promise = loadRouteLoader(loader, requestEv);
  promise.catch(() => {});
  if (loader.__blockSSR) {
    allBlockSSRLoaders = (
      allBlockSSRLoaders ? allBlockSSRLoaders.then(() => promise) : promise
    ) as Promise<void>;
  }
}
return allBlockSSRLoaders;
```

Every loader is started with an explicit `requestEv`, but only `__blockSSR` ones are awaited.
That is the whole difference:

- **Blocking loader** — resolved before render begins. `capture.load()` runs synchronously inside
  the render's invoke context, so `_getContextEvent()` supplies the event. Works with ALS off.
- **Non-blocking loader** — still pending when render reaches its signal. The compute suspends and
  resumes in a continuation with no invoke context, so `_getContextEvent()` yields nothing and the
  ambient ALS store is the only resolver left. Lose that and `load()` throws.

---

## Two measurements that contradict the obvious guesses

These both cost us real time, so they are worth stating up front.

**It is not "ALS off breaks everything."** With every loader blocking, ALS-off renders every route
healthy — **0 / 78** across two independent builds of a much larger production app, plus 10 / 10 in
the minimal repro. The failure requires `blockSSR: false`. Any fix validated against "ALS off ⇒
everything fails" is being validated against a baseline that does not exist.

**It is not concurrency-dependent.** In the larger app the broken arm fails **2 / 18 sequentially**
(and 5 / 60 under 6-wide concurrency). The trigger is upstream latency racing the render — "was
this loader still pending when render reached it" — and concurrency only raises the odds by slowing
the loader down. In the minimal repro a fixed 60 ms delay makes it 10 / 10 with no concurrency at
all.

---

## What we did not establish

We have not shown that anything _drops_ an ALS frame.

A standalone workerd probe — `als.run(store, cb)` where `cb` returns a `Response` wrapping an open
`TransformStream`, with the store read from a continuation that outlives `cb`, after the first
flush, across `await writer.ready` backpressure, a real dynamic `import()`, a subrequest `fetch()`,
and micro/macrotask hops — **lost 0 frames in 339 requests at 24-wide concurrency**.

So the mechanism is better described as _the continuation never had a context to begin with_ than as
_the runtime lost one_. In the reproduction here ALS is simply absent, and the non-blocking
continuation has no invoke context either — two independent resolvers, neither available.

We also see this failure in production on a runtime where ALS **is** available, intermittently and
under load. We have not been able to reproduce that half, and we cannot currently explain it: if
ALS is live and `_asyncRequestStore.run()` wraps the request, `getStore()` should hold. That is the
part where an upstream opinion would help most.

---

## Suggested direction

`load()` throwing before it consults `getRouteLoaderValues()` looks like the load-bearing detail:
the value it needs is often already computed, and the middleware holds a `requestEv` it could pass
through. Making the compute path independent of an ambient lookup — rather than making the ambient
lookup more reliable — would close every route into `load()` at once, and would make `blockSSR:
false` safe on runtimes without `node:async_hooks`.

Failing that, `load()` throwing _mid-stream_ is what turns a loader error into a cacheable HTTP 200
with two `<html>` elements. Even with the root cause unfixed, not welding a complete error document
into an already-flushed stream would stop a single failure from being cacheable.

Varixo:
Fix non-blocking route loaders failing during SSR when AsyncLocalStorage is disabled by explicitly passing the RequestEvent. Also remove the router-specific request event type and fallback handling from Qwik Core.

### DIFF
--- a/.changeset/fuzzy-pandas-wait.md
+++ b/.changeset/fuzzy-pandas-wait.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/router': patch
+---
+
+fix: run non-blocking route loaders without async local storage
--- a/packages/qwik-router/src/runtime/src/qwik-router-component.tsx
+++ b/packages/qwik-router/src/runtime/src/qwik-router-component.tsx
@@ -226,7 +226,12 @@ export const useQwikRouter = (props?: QwikRouterProps) => {
   // Then set .value from middleware-computed loader values (inert, non-reactive data).
   const loaderState = {} as Record<string, ComputedSignal<unknown>>;
   const contentModulesForInit = env.loadedRoute.$mods$ as ContentModule[];
-  const loaders = ensureRouteLoaderSignals(contentModulesForInit, loaderState, routeLoaderCtx);
+  const loaders = ensureRouteLoaderSignals(
+    contentModulesForInit,
+    loaderState,
+    routeLoaderCtx,
+    env.ev
+  );
   for (const loader of loaders) {
     if (loader.__id in env.loaderValues) {
       const value = env.loaderValues[loader.__id];
@@ -613,7 +618,12 @@ export const useQwikRouter = (props?: QwikRouterProps) => {
       if (!isServer) {
         invalidateNavRouteLoaders(loaderState);
       }
-      const routeLoaders = ensureRouteLoaderSignals(contentModules, loaderState, routeLoaderCtx);
+      const routeLoaders = ensureRouteLoaderSignals(
+        contentModules,
+        loaderState,
+        routeLoaderCtx,
+        isServer ? env.ev : undefined
+      );
       if (shouldInvalidateActionLoaders) {
         // Actions force revalidation (fetch cache: 'reload') for their loaders
         if (actionLoaderHashes !== undefined) {
--- a/packages/qwik-router/src/runtime/src/route-loaders.spec.tsx
+++ b/packages/qwik-router/src/runtime/src/route-loaders.spec.tsx
@@ -27,6 +27,7 @@ import {
 } from '@qwik.dev/core/internal';
 import { createDocument } from '@qwik.dev/core/testing';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { RouteLoaderCtxContext, RouteStateContext } from './contexts';
 import {
   ensureRouteLoaderSignal,
   getModuleRouteLoaders,
@@ -285,6 +286,36 @@ describe('route loader store + computed signal tracking', () => {
 
     expect(signal.error).toBeUndefined();
     expect(signal.value).toBe('late-loader-value');
+  });
+
+  it('passes the RequestEvent to a loader created during render', async () => {
+    let resolve!: (value: string) => void;
+    const pendingValue = new Promise<string>((r) => (resolve = r));
+    const loader = routeLoaderQrl(
+      createQRL(null, 'render-loader', () => pendingValue)
+    ) as LoaderInternal;
+    const requestEv = {
+      sharedMap: new Map(),
+      cookie: {},
+      url: new URL('http://localhost/'),
+    } as unknown as RequestEvent;
+    const loaderPromise = loadRouteLoader(loader, requestEv);
+    const host = vnode_newVirtual();
+    container.setContext(host, RouteStateContext, {});
+    container.setContext(host, RouteLoaderCtxContext, { loaderPaths: {} });
+    container.$serverData$.qwikrouter = { ev: requestEv };
+    const ctx = newInvokeContext();
+    ctx.$container$ = container;
+    ctx.$hostElement$ = host;
+
+    const signal = invoke(ctx, loader);
+
+    resolve('render-loader-value');
+    await loaderPromise;
+    await signal.promise();
+
+    expect(signal.error).toBeUndefined();
+    expect(signal.value).toBe('render-loader-value');
   });
 
   it('should verify store is reactive', async () => {
--- a/packages/qwik-router/src/runtime/src/route-loaders.spec.tsx
+++ b/packages/qwik-router/src/runtime/src/route-loaders.spec.tsx
@@ -27,8 +27,14 @@ import {
 } from '@qwik.dev/core/internal';
 import { createDocument } from '@qwik.dev/core/testing';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { getModuleRouteLoaders, routeLoaderQrl, setLoaderSignalValue } from './route-loaders';
-import type { LoaderInternal, RouteModule } from './types';
+import {
+  ensureRouteLoaderSignal,
+  getModuleRouteLoaders,
+  loadRouteLoader,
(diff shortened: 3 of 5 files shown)


wmertens (commented):
I don't understand why this is needed?
Shouldn't routeLoaders always have access to the request that caused them to run?

wmertens (approved):
LGTM

PR #8996, opened 2026-09-04
fix: preserve nested QRLs captured by library components

### LINKED ISSUE #8990: [🐞] A page with a @qds.dev/ui component does not resume on a full page load (build only)

Qwik Runtime

### Describe the bug

A route containing a @qds.dev/ui component (a select in my case) does not resume when loaded directly — typed URL, refresh, plain anchor. Reaching the same route through a Link works. It only happens in a production build; pnpm dev is fine.
Nothing on the page responds — not just the qds component: a plain Qwik button placed next to it does not work either. On beta.43 the console shows TypeError: e is not a function and TypeError: Cannot read properties of undefined (reading 'apply'). On beta.40–beta.42 the page is equally dead but throws nothing, so this is not a regression in beta.43 — beta.43 only makes an older failure visible.

### Reproduction

https://github.com/gimonaa/qwik-beta-43

### Steps to reproduce

run pnpm install && pnpm build && pnpm preview

### System Info

```shell
System:
    OS: macOS 26.6.2
    CPU: (12) arm64 Apple M4 Pro
    Memory: 507.64 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.0 - /Users/alessandrogimona/.nvm/versions/node/v22.13.0/bin/node
    npm: 10.9.2 - /Users/alessandrogimona/.nvm/versions/node/v22.13.0/bin/npm
    pnpm: 11.25.0 - /Users/alessandrogimona/Library/pnpm/bin/pnpm
  Browsers:
    Chrome: 152.0.7977.75
    Firefox: 154.0.1
    Safari: 26.6.2
```

### Additional Information

- /select/ reached via Link from / → works
- /select/ loaded directly → dead
- /plain/ loaded directly → works. Same page, same navigation, without the qds component: that is what points at the component rather than at the setup.

Varixo:
Fix nested inlinedQrl calls captured by another QRL not being extracted or included in the manifest. This prevented affected pages from resuming after a production full-page load.

### DIFF
--- a/.changeset/nested-qrl-captures.md
+++ b/.changeset/nested-qrl-captures.md
@@ -1,0 +1,5 @@
+---
+'@qwik.dev/optimizer': patch
+---
+
+fix: preserve nested QRLs captured by library components
--- a/packages/optimizer/core/src/test.rs
+++ b/packages/optimizer/core/src/test.rs
@@ -7317,6 +7317,38 @@ export function qwikifyQrl(reactCmp$, opts) {
 }
 
 #[test]
+fn inlined_qrl_in_capture_is_extracted() {
+	let output = test_input!(TestInput {
+		code: r#"
+import { inlinedQrl } from '@qwik.dev/core';
+
+const context = {};
+export const handler = inlinedQrl(() => {}, "outer_abc", [
+	inlinedQrl(() => {}, "inner_def", [context]),
+]);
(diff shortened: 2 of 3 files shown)


wmertens (approved):
LGTM

PR #8998, opened 2026-09-05
fix: prevent render loops and stalls after async failures

Varixo:
- prevent infinite retries when component chunk imports fail
- handle rejected task suspension promises so rendering can finish
- await asynchronous computations on streamed SSR nodes and propagate failures


wmertens (approved):
LGTM

PR #9005, opened 2026-09-09
fix(preloader): preload only qrls, ignore dynamic imports

Varixo:
Now preloader will preload only QRLs and will ignore dynamic imports. 
There is a rule - dynamic import should be handled by user and qrl is handled by Qwik


packages/qwik-vite/src/plugins/bundle-graph.ts
```diff
@@ -65,25 +68,21 @@ export function convertManifestToBundleGraph(
...
-        // we only want to include dynamic imports that belong to the app
-        // e.g. not all languages supported by shiki
-        (dep) =>
-          graph[dep] &&
-          // either there are qrls
-          (graph[dep].symbols ||
-            // or it's a dynamic import from the app source
-            graph[dep].origins?.some((o) => !o.includes('node_modules')))
-      ) || [];
+    const followedImports = trustedNodes.has(bundleName)
```
maiieul: Wasn't checking `graph[dep].symbols` like in the past version enough to verify it's a qrl and not just a dynamic import?

packages/qwik-vite/src/plugins/bundle-graph.ts
```diff
@@ -44,6 +44,8 @@ export function convertManifestToBundleGraph(
   }
   // All known chunks and symbols
   const graph = { ...manifest.bundles };
+  // Symbol and adder nodes carry their own edges; only chunk edges get the qrl rule
```
maiieul: what's an "edge"? Which rule? This is not clear. Either remove or make it clear pls.

packages/qwik-vite/src/plugins/bundle-graph.ts
Varixo: - it was skipping all dynamic imports directly in src/
- it kept a user import('./widget.tsx') whenever the target chunk held any QRL segment
- at chunk level import("./widget") and qrl(() => import("./segment")) are the same dynamicImports entry, so no target-side condition can separate them

packages/qwik-vite/src/plugins/bundle-graph.ts
maiieul: Ah I see what you meant here. Are you sure bundle graph adders shouldn't count like QRLs as preloaded stuff? I'm thinking they should.

packages/qwik-vite/src/plugins/bundle-graph.ts
```diff
@@ -64,26 +64,19 @@ export function convertManifestToBundleGraph(
...
-        // we only want to include dynamic imports that belong to the app
-        // e.g. not all languages supported by shiki
-        (dep) =>
-          graph[dep] &&
-          // either there are qrls
-          (graph[dep].symbols ||
-            // or it's a dynamic import from the app source
-            graph[dep].origins?.some((o) => !o.includes('node_modules')))
-      ) || [];
+    const dynamicImports = bundle.qrlImports?.filter((dep) => graph[dep]) || [];
```
maiieul: Can't we name this `qrlImports`?

maiieul (changes requested):

maiieul (approved):

PR #5542, opened 2023-12-08
chore: AWS Lambda Adapter: CloudFront, S3 & Function URL

ptz0n:
# Overview

# What is it?

- [x] Feature / enhancement
- [ ] Bug
- [x] Docs / tests / types / typos

# Description

A more performant and robust AWS Lambda adapter, using S3 and Lambda Function URL origins behind CloudFront.

# Use cases and why

- Simple and flexible stack for deploying Qwik to AWS.
- Less overhead, removing API Gateway.
- Serving static files from S3.

# Checklist:

- [ ] My code follows the [developer guidelines of this project](https://github.com/BuilderIO/qwik/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] Added new tests to cover the fix / functionality

PR #9007, opened 2026-09-10
chore: AWS Lambda Adapter: CloudFront, S3 & Function URL

ptz0n:
# What is it?

- [x] Feature / enhancement
- [ ] Bug

# Description

Replaces the Serverless Framework-based deploy (`serverless.yml`, `serverless-offline`, `serverless deploy`) in the AWS Lambda adapter with a plain CloudFormation SAM template + `Makefile`, deployed via `aws cloudformation package`/`deploy` directly.

- S3 origin for static assets, Lambda Function URL origin for SSR, one CloudFront distribution in front of both. No API Gateway.
- One fewer third-party framework dependency for a template this minimal — `entry_aws-lambda.tsx` and the node-server adapter's `vite.config.ts` are unchanged, this only replaces the deployment layer.
- Adds a `README.md` fragment (the adapter didn't have one), matching the format used by `express`/`node-server`/etc.

This picks up where #5542 left off. That PR was closed for going stale in 2024, not on technical grounds — no review was ever left on it (see the discussion there). This is a fresh implementation against current `main`, informed by that draft plus a standalone reference repo I've been maintaining at [ptz0n/qwik-cloudfront-s3-lambda-url](https://github.com/ptz0n/qwik-cloudfront-s3-lambda-url).

# Use cases and why

- A simple, flexible stack for deploying Qwik to AWS without adopting a separate deploy framework on top of CloudFormation.
- Anyone already on CloudFormation/SAM elsewhere in their infra gets a consistent tool instead of a second one just for this adapter.

# Checklist:

- [ ] I have made corresponding changes to the documentation (the [aws-lambda docs page](https://qwik.dev/docs/deployments/aws-lambda/) still describes the Serverless Framework flow — separate follow-up)
- [ ] Added new tests to cover the fix / functionality (infra/config change — verified with a full build: scaffolded a fresh app from `starters/apps/base` + this adapter fragment against current `main`, confirmed `npm run build` passes typecheck, lint, and produces the expected `server/entry_aws-lambda.js` + `dist/build/*` output the CloudFormation template targets)


maiieul (changes requested):
Thanks @ptz0n. I don't know enough about cloudformation and AWS serverless tbh so I had Claude review it.

Here's Claude's review below. Let me know if you want to tackle it. Otherwise I can attempt improving it when I find some time for it and will ping you for review. Ofc I'd prefer that you tackle it since you probably know more about AWS than me :). Feel free to ask questions if anything's unclear!

**Verdict: request changes.** Dropping Serverless makes sense, since the old template pins Serverless v3 and the `nodejs18.x` runtime, both out of support. But this template can't deploy. Even after a fix for that, actions, cookies and public files would still break. The PR only checked `npm run build`; a deploy would have failed at the first step.

**Blockers**

1. **Circular dependency, so the stack can't be created.** `ORIGIN: !GetAtt CDN.DomainName` (`serverless.yml:27`) makes the function depend on the CDN, the CDN depends on `SSRFunctionUrl`, and the URL depends on the function. cfn-lint reports E3004 on all three. The author's reference repo has the same template.
   - Fix: drop `ORIGIN` and add a viewer-request CloudFront Function that copies `host` into `x-forwarded-host`.
   - Then set `HOST_HEADER=x-forwarded-host`, which the node middleware already reads ([[http.ts:24](https://claude.ai/epitaxy/packages/qwik-router/src/middleware/node/http.ts:24)](packages/qwik-router/src/middleware/node/http.ts:24)). The protocol is already `https`, because serverless-http's socket sets `encrypted = true`.
2. **The SSR behavior breaks Qwik Router** (`serverless.yml:110-119`).
   - `AllowedMethods` has no POST, so CloudFront rejects every `routeAction$`, `<Form>` and `server$` call.
   - `QueryString: false` and `Cookies: none` strip `?qaction=`, `url.searchParams` and sessions.
   - The legacy `ForwardedValues` setup caches SSR HTML at the edge for 24h when the response has no `Cache-Control`.
   - Fix: allow all 7 methods and use the managed `CachingDisabled` and `AllViewerExceptHostHeader` policies. AWS says the second one is meant for Function URL origins.
3. **The Function URL returns 403.** Since October 2025, new URLs need `lambda:InvokeFunction` as well as `lambda:InvokeFunctionUrl`, and the template grants only the second (`:32`). Fix: use SAM's `FunctionUrlConfig: { AuthType: NONE }`. I checked the SAM translator source (v1.109) and it creates both grants, so `SSRPermission` and `SSRFunctionUrl` can go.
4. **Public files 404.** Only `build/*`, `*manifest.json`, `service-worker.js` and `favicon.ico` go to S3. Everything else hits a Lambda that no longer ships `dist/` (`CodeUri: ../server`; the old config packaged `dist/**`).
   - That breaks `/favicon.svg` (the file the starters actually link), `/robots.txt`, `/assets/*` (images, and CSS over the 20 KB inline limit) and the playground's `public/fonts`.
   - Fix: route `assets/*` to S3 and ship `dist/` next to `server/`.
5. **Hardcoded names clash.** S3 bucket names are global, so the README defaults fail for anyone after the first user. `FunctionName: qwik-ssr` (`:19`) and the OAC name `PublicBucketAccess` (`:69`) clash between two stacks in one account. Fix: let CloudFormation name the bucket and output the name, and use `${AWS::StackName}` in the OAC name.

**Should fix**
- No `Compress: true` on the S3 behaviors, so JS ships uncompressed. `aws s3 sync` sets no `Cache-Control`, while the old config marked `build/**` as `immutable`.
- The Makefile updates the Lambda before it syncs assets, so for a short time new HTML can point at chunks that aren't uploaded yet.
- There's no `Outputs` section, so users never see their URL.
- [[The docs page](https://claude.ai/epitaxy/packages/docs/src/routes/docs/deployments/aws-lambda/index.mdx)](packages/docs/src/routes/docs/deployments/aws-lambda/index.mdx) still says `serverless deploy`. It's in the same repo, so I'd update it in this PR rather than later.
- No changeset. The Workers adapter used a `create-qwik` patch changeset (a748c3005).
- CI is red: Lint Package fails prettier on the adapter's `package.json`, probably the one-line `docs` array.

**Nits:** the TODO and StackOverflow comments get copied into every user's app. `cloudformation/serverless.yml` is a confusing name without Serverless; `template.yaml` is the SAM default. The Makefile hardcodes `npm run build` and needs `make`, which Windows users won't have.

**Checked and fine:** the `handler` export exists, and the serverless-http 4 bump looks safe. Node 24 loads the ESM bundle without a `package.json`; I tested `require()` of an ESM file on Node 24.

The template I sent covers blockers 1–5 and the S3 compression and caching fixes. Want me to post this as a review on the PR?

Sources:
- [[Control access to Lambda function URLs](https://docs.aws.amazon.com/lambda/latest/dg/urls-auth.html)](https://docs.aws.amazon.com/lambda/latest/dg/urls-auth.html)
- [[Managed origin request policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-origin-request-policies.html)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-origin-request-policies.html)
- [[Managed cache policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html)
- [[CloudFront expiration (24h default TTL)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html)](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html)

PR #9008, opened 2026-09-11
fix(core): keep locale across async withLocale without ALS

xia-chao:
# What is it?

- Bug

# Description

`withLocale('zh-CN', async () => { ... })` is supposed to keep `getLocale()` as `zh-CN` for the whole function, including after `await`.

That already works when AsyncLocalStorage is present. When it is not (browser build, some edge runtimes), the old code used `try/finally`. The `finally` runs the moment the function returns a Promise, so after `await` the locale is gone and `getLocale()` throws `Reading locale outside of context.`

This change waits until that Promise finishes (success or error), then puts the previous locale back. Sync `withLocale` is unchanged.

Related: this is the leftover hole after #8999, which taught Qwik how to *find* ALS on more runtimes. This PR is about the path that still has no ALS.

# Checklist

### DIFF
--- a/packages/qwik/src/core/use/use-locale.ts
+++ b/packages/qwik/src/core/use/use-locale.ts
@@ -1,7 +1,8 @@
-import { tryGetInvokeContext } from './use-core';
 import { getAsyncLocalStorage } from '@qwik.dev/core/async-local-storage';
 import { isServer } from '@qwik.dev/core/build';
 import type { AsyncLocalStorage } from 'node:async_hooks';
+import { isPromise } from '../shared/utils/promises';
+import { tryGetInvokeContext } from './use-core';
 
 let _locale: string | undefined = undefined;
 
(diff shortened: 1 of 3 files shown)


packages/qwik/src/core/use/use-locale.ts
```diff
@@ -55,11 +59,29 @@ export function withLocale<T>(locale: string, fn: () => T): T {
...
+          throw reason;
+        }
+      ) as T;
+    }
+    restore();
+    return result;
+  } catch (err) {
+    restore();
+    throw err;
   }
```
Varixo: instead of this code we already have similar util function called safeCall. You can use it like this:
```ts
  return safeCall(
    fn,
    (value) => {
      restoreLocale();
      return value;
    },
    (error) => {
      restoreLocale();
      throw error;
    }
```

Varixo (commented):
hey! thanks for the PR, added small comment

Varixo (approved):
LGTM, thanks for the PR!

PR #9010, opened 2026-09-14
fix(router): serve static files under the configured base path

### LINKED ISSUE #8812: [🐞] staticFile middlewares don't strip basePathname before file resolution
With a vite `base` (e.g. `/base/`), `isStaticPath` can match a request, but the node/deno/bun staticFile middlewares resolve files as `join(staticFolder, url.pathname)` — the base segment is not stripped, and dist has no `base/` directory, so resolution fails on all three (`packages/qwik-router/src/middleware/node/index.ts`, `middleware/deno/index.ts`, `middleware/bun/index.ts`).

Split out of #8811 (assetsDir static paths fix), which made `isStaticPath` base-aware but did not touch file resolution.

KayanoLiam:
# What is it?

- Bug

# Description

Fixes #8812.

With a base such as `/docs/`, a request for `/docs/hello.txt` can be recognized as static but is read from `dist/docs/hello.txt`, while the file is actually at `dist/hello.txt`. Build and asset directory prefixes also omit the base.

Use the configured base for build/asset classification, then strip it at the filesystem boundary in the Node, Deno, and Bun middleware. Preserve full URL paths in the prerendered-path lookup and keep the existing trailing-slash behavior. Deno and Bun derive the filename from the stripped pathname too.

Filesystem access still follows static-path eligibility checks, and the base must match at a slash boundary. Encoded path segments remain encoded.

The focused suite covers public files, nested bases, root/index pages, both trailing-slash modes, build/assets without manifest entries, and paths outside the base. All 50 tests pass; restoring the original production implementations produces 23 failures.

Validation:
- `pnpm build.core` and the Router TypeScript check pass.
- Changed-file ESLint, Prettier, and diff checks pass.
- The package unit suite passes with one existing loopback-socket test excluded: 3673 passed, 20 skipped (19 existing skips plus that exclusion), and 19 todo. The excluded test is `should make Node response progress before render completes`; this environment does not permit its localhost listener.
- Deno/Bun adapter tests use real filesystem-backed API shims under Node. Native runtime and browser E2E tests were not run.

# Checklist

- [x] I added a changeset


maiieul (approved):

maiieul:
Thanks @KayanoLiam 

LGTM

PR #9015, opened 2026-09-16
feat(migration): handle the rest migration cases

Varixo:


maiieul (approved):
LGTM except a bit weird that `qwikRouter({strictLoaders: false})` is in router-config.ts. We can make a follow up PR if need be.

PR #9017, opened 2026-09-18
fix: TypeScript optimizer worker and stripped loader handling

wmertens:
## What is it?

Bug fixes

## Description

This PR fixes several issues in the TypeScript optimizer related to worker threads, stripped loaders, and constant folding:

### Key Fixes

1. **Worker thread memory management**: Worker threads now disable raw-transfer parsing to prevent reserving ~6 GB of address space per thread, which breaks child process spawning in builds. Added `QWIK_TS_OPTIMIZER_RAW_TRANSFER` environment variable to control this behavior.

2. **Stripped loader sentinel numbering**: Fixed collision where multiple stripped route loaders were wired to the same client QRL. Each stripped loader now gets its own sentinel binding from a unified numbering sequence shared with top-level `worker$` declarations.

3. **Worker and stripped loader interaction**: Top-level `worker$` declarations now participate in the same sentinel numbering sequence as stripped loaders, preventing collisions while respecting `regCtxName` configuration.

4. **Constant folding in nested ternaries**: Fixed dead code elimination to properly handle constant ternaries whose kept branch contains further constant folds. The optimizer now correctly parenthesizes object and function branches left by nested folds.

5. **Library mode event handlers**: Fixed library-mode event handlers to preserve the `.m()` moved-captures marker when collapsing to inlined QRLs, and properly include dev metadata in `inlinedQrlDEV` calls.

6. **Component prop captures**: Fixed bare `$()` component prop references outside a component boundary to keep their captures in the `.w([...])` wrapper.

### Documentation Updates

- Clarified build commands: `pnpm build.full` builds everything with TypeScript optimizer; `pnpm build.rust` is only needed for Rust optimizer work
- Updated setup instructions to clarify that Rust toolchain is optional for day-to-day development
- Added guidance on baselining TypeScript optimizer failures and comparing optimizer outputs

### Library Package Updates

Updated library packages (`@qwik.dev/router`, `@qwik.dev/react`, `@qwik.dev/utils`, `@qwik.dev/devtools`, `supabase-auth-helpers-qwik`) to build with the TypeScript optimizer via `tsOptimizer: true` in their Vite configs.

## Testing

- Added 3 new test cases for stripped loader and worker sentinel numbering
- Added 2 new test cases for library-mode event handler and dev metadata handling
- Added 2 new test cases for nested constant folding edge cases
- Added test case for bare component prop captures
- Added unit tests for raw transfer policy decision logic
- Existing optimizer tests pass with the fixes applied

https://claude.ai/code/session_01XuiGUYMZEWufvNFQLpqeMG


packages/ts-optimizer/src/optimizer/rewrite/lib-mode-collapse.ts
```diff
@@ -35,10 +35,16 @@ import { createTransformSession } from '../edit/transform-session.js';
...
   readonly qrlSymbolName: string;
+  /** The decl carried the `.m()` moved-captures marker, which the literal must keep. */
+  readonly hasMovedCaptures: boolean;
+  /** Source text of the dev metadata argument of a `_noopQrlDEV` decl. */
+  readonly devMetaText?: string;
   readonly start: number;
   readonly end: number;
 }
 
+const MOVED_CAPTURES_MARKER = '.m()';
```
Varixo: is this correct place for marker like this?

packages/ts-optimizer/tests/optimizer/ast/parse-raw-transfer.test.ts
```diff
@@ -0,0 +1,23 @@
+import { readFileSync } from 'node:fs';
+import { describe, expect, it } from 'vitest';
+import { disableRawTransfer, parseWithRawTransfer } from '../../../src/optimizer/ast/parse.js';
+
+const vmDataMb = () =>
+  Number(/VmData:\s+(\d+)/.exec(readFileSync('/proc/self/status', 'utf8'))![1]) / 1024;
+
+describe('parseWithRawTransfer', () => {
+  // Raw transfer reserves ~6 GB per thread; pool workers must not, or fork() in the host fails.
+  it.skipIf(process.platform !== 'linux')(
```
Varixo: why only linux?

packages/ts-optimizer/src/optimizer/rewrite/lib-mode-collapse.ts
wmertens: Fair point, it was wedged between two interfaces. Moved it up with the module's other top-level declarations and gave it a doc comment saying what it is (folded into the `.m()` marker commit).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/ts-optimizer/tests/optimizer/ast/parse-raw-transfer.test.ts
wmertens: Because it read `VmData` from `/proc/self/status`. Switched to `process.memoryUsage().arrayBuffers`, which reports the raw-transfer buffer in full (6144 MB for one raw parse, 0 for a plain one) on every platform, so the test now runs everywhere and the Linux skip is gone.

---
_Generated by [Claude Code](https://claude.ai/code)_

Varixo (approved):

PR #9018, opened 2026-09-18
fix(qwikloader): don't queue user events behind qwik lifecycle handlers

maiieul:
# What is it?
- Bug

# Description
On slow networks the first click after load was ignored until the router's `qcinit` imports finished, because every event handler waited in one shared task queue. Qwik lifecycle events (`qinit`, `qidle`, `qvisible`, `qcinit`, ...) now run right away, and browser events keep their order.

Regression from https://github.com/QwikDev/qwik/commit/7ae409deb7432f800c1a33206108e059cc4af1dd (#8540), which added the shared queue.


Varixo (approved):
lgtm

PR #9024, opened 2026-09-18
fix(router): share the static file lookup between node, deno and bun

maiieul:
# What is it?
- Bug

# Description
Follow-up to #9010: Node, Deno and Bun now resolve static files through one shared helper instead of three copies. This fixes Node erroring on prerendered pages in dotted folders like `/blog/v1.2/`, and replaces the duplicated adapter tests with one table test.


yslpn:
Thanks, we’re waiting for this fix so we can remove the local patch for https://valibot.dev/

Varixo (approved):
LGTM

